Issue #18, opened 2013-08-20
drush_get_projects() does not properly load project status url in rare cases


iamEAP:
Copied from https://drupal.org/node/2044487

PR #20, opened 2013-08-20
Ensuring that "project status url" is set regardless of project nesting.

iamEAP:
In particular, this fixes cases wherein sub-extensions' names happen to come prior to the primary extension's name alphabetically (e.g. the "bulk_export" submodule of ctools).

See #18


jonhattan:
I don't see the point in supporting that a module has a different status url than other modules part of the same project.

Btw an improvement to status url support was done recently in #1460

PR #64, opened 2013-09-05
User-specific temp directories

arnested:
The "drush use @site-alias" feature does not work on a system with more than one user account using the feature.

From a real life experience today:

```bash
$ whoami
arne
$ ls -ld /tmp/drush-env
drwxr-xr-x 2 root root 4096 2013-09-04 18:25 /tmp/drush-env
```
Since another user has already created the `drush-env` folder I'm not able to set my environment.

This pull request incorporates the user ID in the folder name.

### DIFF
--- a/drush.complete.sh
+++ b/drush.complete.sh
@@ -9,7 +9,7 @@
 which drush > /dev/null || alias drush &> /dev/null || return
 
 __drush_ps1() {
-  f="${TMPDIR:-/tmp/}/drush-env/drush-drupal-site-$$"
+  f="${TMPDIR:-/tmp/}/drush-env-${USER}/drush-drupal-site-$$"
   if [ -f $f ]
   then
     DRUPAL_SITE=$(cat "$f")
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
Okay, I finally circled back to this issue, ignored the PR that was here, went to fix it, decided that the proposed solution was too heavyweight to be placed into __drush_ps1.  I don't think it would be a good idea to add a loop-and-find-first-writable  folder to this function that runs every time the bash prompt is rendered.

Once I decided that, I ended up re-implementing the PR that was already here.  Except the PR here has a unit test.

So, my suggestion is to ignore my recent PR, and commit the PR here.  The other functions that need cache directories have already gotten per-user components, as near as I can tell (c.f. drush_directory_cache()), so the only odd-man-out is site-set, which needs to be independent of the other library functions in order to stay simple to parallel __drush_ps1.

arnested:
:+1:

arnested:
I rerolled the pull request so it will apply on the `6.x` branch again (the pull request was originally opened against the `6.x` branch).

The patch applies to the `master` branch as well (with minor offsets but nothing `patch(1)` or `git cherry-pick fa163da840f3bd4d877f33c58e21c24ee342ddf7` can't handle automatically).

PR #97, opened 2013-09-12
Allow use of destination parameter for modules and themes

mradcliffe:
Original patch by acbramley. I made the fork, applied the patch using attribution, and then worked on a test.

Original issue: http://drupal.org/node/2011284


ergonlogic:
I'm closing this issue due to inactivity. Post here if anyone wants it reopened.

As a possible work-around, couldn't you download a theme or module as a 'library', and use the provided 'destination' parameter?

PR #130, opened 2013-09-19
No field labels

greg-1-anderson:
A user may want to see just the command output data, without any extra field labels; it should therefore be possible to tell Drush to omit the field labels.

Example:

```
$ drush status 'Drush configuration'
 Drush configuration   :  /home/ga/local/config/drushrc.php /home/ga/.drush/drushrc.php 

$ drush status 'Drush configuration' --no-field-labels
 /home/ga/local/config/drushrc.php /home/ga/.drush/drushrc.php 
```

The --no-field-labels version could more easily be processed in a bash for loop, for example.


weitzman:
This is now possible in master branch. See #2460

Issue #80, opened 2013-09-11
drush pm-enable doesn't work when running against an Oracle db


joe-udwin-lrn:
Drush pm-enable does not currently work against an oracle db.

This issue was raised before drush moved to github:
https://drupal.org/node/1180240

Work has already gone into a patch on the above thread:
https://drupal.org/files/oracle-1180240-3_0.patch

Any chance we can get this integrated into the main branch?

Thanks!

PR #133, opened 2013-09-19
Modifications required to make drush pe-enable work on oracle

Joe-U-Questionmark:
This is the fix to issue 80
(https://github.com/drush-ops/drush/issues/80) which is essentially a
copy of this patch (https://drupal.org/files/oracle-1180240-3_0.patch)
applied to main with an additional bug fix


greg-1-anderson:
Really old - closing.

PR #241, opened 2013-11-04
Specify multiple types in watchdog commands

rkarajgi:
Specify multiple types in WS command such as --type=user,cron


greg-1-anderson:
This can now be done with --filter and regular expressions

PR #353, opened 2014-01-03
Fix #134. Added db-drop-existing-tables switch to site-install

### LINKED ISSUE #134: site-install deletes all the existing tables in a database
drush site-install deletes all the existing tables in a database (including ones that are not related to Drupal!).

This makes housing multi-site Drupal installations in the same database or sharing the database with another application impossible.

I can see this would be useful for development and testing but I can't think of any situation where this would make sense in a production environment.

I think site-install should be changed to disable this functionality by default and optionally allow someone to pass an additional parameter to delete tables if required.

dustinmoris:
Added 'db-drop-existing-tables' switch to site-install command to optionally not delete existing tables in the target database, so that Drupal can co-habit a database with other applications. This switch defaults to TRUE if not specified, so that normal functionality is unaffected.

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -273,6 +273,7 @@ function core_drush_command() {
         'description' => 'Password for the "db-su" account. Optional.',
         'example-value' => 'pass',
       ),
+      'db-drop-existing-tables' => 'Defaults to TRUE. Remove all tables in the target database before creating Drupal tables.',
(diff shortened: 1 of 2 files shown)


rgpublic:
Would be very useful if it existed for Drush9. Currently even if I use --db-prefix, it seems the whole database is just wiped which is quite... unexpected :-(

PR #453, opened 2014-02-11
Issue-254: db prefix can be a string or an array

douggreen:
sql-sync fails because the prefix can be an array. This PR handles whether the prefix is a string or an array. See https://api.drupal.org/api/drupal/includes%21database%21database.inc/function/DatabaseConnection%3A%3AsetPrefix/7


weitzman:
Very old PR. The underlying code has changed. Not sure if this is still an open issue.

PR #462, opened 2014-02-13
Provide hook to alter order of hook invocations.

ergonlogic:
Eventually I'll get this right...

This is a follow-up to [Provide weight-like mechanism to ensure order of hook calls](https://drupal.org/node/2031383). It provides a similar hook for native drush hooks in _drush_invoke_hooks().

It replaces https://github.com/drush-ops/drush/pull/461, which failed testing due to commit conflicts. This time I squashed everything into one commit. Fingers crossed.


weitzman:
See AnnotatedCommand hooks

Issue #545, opened 2014-03-21
Drush backend invoke does not consider port number when deciding local vs remote


greg-1-anderson:
Drush backend_invoke uses ssh to execute commands on remote systems, but will directly execute (e.g. with proc_open) commands on the local system.

There is a problem with the test Drush uses to determine local vs. remote in that it only considers the name/address of the target host, and ignores the ssh port number that might be set in ssh-options.  This interferes with communicating with remote hosts via an ssh tunnel, e.g. as Vagrant will set up.

c.f. http://drupal.stackexchange.com/questions/107339/how-do-you-write-a-drush-alias-for-a-local-vagrant-box/107349?noredirect=1#comment127408_107349

PR #546, opened 2014-03-21
Ensure that Drush will use backend_invoke() for vagrant & c.

greg-1-anderson:
This PR will cal through to backend_invoke() instead of executing a command locally if ssh-options are defined.  c.f.: #545

This is just a quick implementation for testing.  We still need to decide if we should inspect the contents of ssh-options before deciding whether the command should be local or remote.

The conflicting use case is that some folks may have Drush aliases that they share between multiple machines; in this case, Drush currently takes some pains to insure that it does not use ssh to call a "remote" site if the alias to said site is being used on the server where it is running.  In other words, if 'remote-host' is set to 'foo.com', then Drush will pretend it is not set if the local machine is 'foo.com'.  This is the root cause for why Drush aliases to sites running on virtualbox / vagrant do not work right.

This PR should probably be enhanced to require that ssh-options does contain -p, but the -p parameter is not -p 22.  Is this the right test.

### DIFF
--- a/examples/example.aliases.drushrc.php
+++ b/examples/example.aliases.drushrc.php
@@ -154,6 +154,9 @@
  *   hosting the Drupal instance. **Important Note: The remote-host option
  *   must be omitted for local sites, as this option controls whether or not
  *   rsync parameters are for local or remote machines.
+ * - '#check-local': Test to see if 'remote-host' is the local machine; if
+ *   it is, then 'remote-host' will be ignore. Useful when sharing aliases
+ *   among multiple machines.
  * - 'remote-user': The username to log in as when using ssh or rsync.
  * - 'os': The operating system of the remote server.  Valid values
  *   are 'Windows' and 'Linux'. Be sure to set this value for all remote
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -1106,7 +1106,7 @@ function _drush_backend_generate_command($site_record, $command, $args = array()
   $ssh_options = $site_record['ssh-options'];
   $os = drush_os($site_record);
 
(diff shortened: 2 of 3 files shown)


ArtuGit:
I have this problem now.

```
$aliases['nxlog.dev'] = array(
    '#check-local' => TRUE,    
    'uri' => 'dev.nxlog.co',
    'root' => '/var/www/drupalvm/nxlog/www',
    'remote-user' => 'vagrant',
    'remote-host' => '127.0.0.1',
    'ssh-options' => "-p 2222 -i $home/.vagrant.d/insecure_private_key",    
  );

``` 
I have Drush which was installed using Composer.

 Drush script           :  /opt/drush-8.x/vendor/drush/drush/drush.php              
 Drush version        :  8.1.7

I cannot apply this patch because there are not such lines in such files.
I guess the patch is obsolete, but 
`'#check-local' => TRUE,    `
does not work too.

greg-1-anderson:
`#check-local` has been removed from Drush.  For backend invoke, try using something other than `localhost` or `127.0.0.1` to address your vargrant vm.  e.g. add an entry to /etc/hosts mapping `vagrant` to `localhost` and then use that.

Perhaps `_drush_backend_generate_command()` should replace:
```
  if (drush_is_local_host($hostname)) {
    $hostname = null;
  }
```
with:
```
  if (drush_is_local_host($hostname) && empty($ssh_options)) {
    $hostname = null;
  }
```

ArtuGit:
Thank you, Greg.
It turns out that 
`  'ssh-options' => '-o PasswordAuthentication=no -i ' . drush_server_home() . '/.vagrant.d/insecure_private_key',`
fixes this issue.

PR #627, opened 2014-05-09
Sanitize password to a randomly generated password

arnested:
Sanitizing the password to the fixed string 'password' is an insecure
default.

Since most people probably won't be aware of the sanitizing when doing an sql-sync they could end up with publicly available sites with insecure passwords.

### DIFF
--- a/commands/sql/sql.drush.inc
+++ b/commands/sql/sql.drush.inc
@@ -233,7 +233,7 @@ function sql_drush_command() {
     'description' => "Run sanitization operations on the current database.",
     'bootstrap' => DRUSH_BOOTSTRAP_DRUSH,
     'options' => array(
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This can be solved by setting the password you desired in a command-specific record in your sitealiases.  I think it would be sufficient to use a single, well-known password for all users (well-known == known internally to the development team), or, if you prefer, you can generate a random password in your drushrc.php file (it's php code -- you can call drush_generate_password() there.

Setting every user to a different random password does not seem like it is useful, as this would require you to use drush uli / set-password for any account you wanted to log in as.  If this is the effect you wanted, you could achieve it by setting a very-long constant password in your drushrc.php, and simply do not distribute it to anyone, and they will have to use uli to log in.

PR #638, opened 2014-05-22
Support PostgreSQL peer authentication

nathanweeks:
See https://drupal.org/node/1030732#comment-7578671 for details.


tstoeckler:
This seems to only be relevant to Drupal 6. Drush no longer does any of its own parsing of database information and since Drupal 7 the DB information is no longer stored as a single URL. I verified that I can run `drush sql-connect`, `drush sql-cli` and `drush sql-drop` with peer authentication just fine.

Closing this. Please re-open if you find any remaining issues in Drush with peer authentication in Postgres.

PR #639, opened 2014-05-22
Fix #631 : D8 include files might have .php in future

### LINKED ISSUE #631: add flexibility to accommodate adding .php extension to php filenames like foo.module foo.module.php to aid in D8 core issue https://drupal.org/node/7269
Motivation:
https://drupal.org/node/7269 Add .php extension to PHP files
is postponed on this issue.

Drupal.org continuous integration testing (testbot) depends on drush, and changing the filename extensions of several files will break testing.

Proposed resolution:
add flexibility to accommodate adding .php extension to php filenames like foo.module foo.module.php 

Old → New
foo.module  → foo.module.php
foo.install → foo.install.php
foo.inc → foo.inc.php
bar.theme   → bar.theme.php
baz.profile → baz.profile.php

ParisLiakos:
This fixes absolutely necessary to get the testbot up and running in https://drupal.org/node/7269
which is `si` and `en` commands.

Related issue https://github.com/drush-ops/drush/issues/631

### DIFF
--- a/commands/core/drupal/site_install.inc
+++ b/commands/core/drupal/site_install.inc
@@ -8,7 +8,11 @@ function drush_core_site_install_version($profile, array $additional_form_option
     $profile = 'standard';
   }
 
-  require_once DRUSH_DRUPAL_CORE . '/includes/install.core.inc';
+  $install_file = '/includes/install.core.inc';
+  if (file_exists(DRUSH_DRUPAL_CORE . $install_file . '.php')) {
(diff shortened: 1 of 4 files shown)


joelpittet:
https://www.drupal.org/node/2206501#comment-8714091 is fixed. 
+1 to get this patch in.

ParisLiakos:
thanks for assisting us here drush folks

Issue #53, opened 2013-09-01
Allow drush make to generate compiled and/or updated makefiles


ergonlogic:
Summary
-------

The patch below basically adds three options to the 'make' command:

**--generate**: Generate a makefile, based on the one passed in, with all versions resolved.
**--generate-update**: Generate an updated makefile, based on the one passed in, with all versions updated.
**--no-build**: Do not build a codebase. Makes the `build path` argument optional.

The first two default to printing to the terminal, but an output file may be specified. The last option allows these options to operate on makefiles without executing an actual build.

Examples
---------

**drush make example.make example --generate=compiled.make**

Create a 'snapshot' makefile (compiled.make) with the versions that were actually used in building the codebase.

**drush make example.make --generate-update --no-build**

Output an updated makefile to the terminal, without building a codebase, based on the source makefile (example.make).

Background & Patch
---------------------

See these two issues from drupal.org for the back-story:
 * [Issue #1212686: Create a versioned make file when running drush make with option --make-generate](https://drupal.org/node/1212686)
 * [Issue #1325818: Generate an upgrade makefile from an existing makefile?](https://drupal.org/node/1325818)

The latest patch that includes the changes from both issues is: https://drupal.org/files/drush-versioned_updated_makefiles-1325818-10.patch

weitzman:
dupe of #660 which has been merged.

PR #660, opened 2014-06-03
Allow drush make to generate compiled and/or updated makefiles

ergonlogic:
## Summary

The basic idea here allow drush make to output (or write) the equivalent to a composer.lock. This pull-request basically adds three options to the 'make' command:

**--lock**: Generate a makefile, based on the one passed in, with all versions _resolved_.
**--lock-update**: Generate an updated makefile, based on the one passed in, with all versions _updated_.
**--no-build**: Do not build a codebase. Makes the `build path` argument optional.

The first two default to printing to the terminal, but an output file may be specified. The last option allows these options to operate on makefiles without executing an actual build.
## Examples

**drush make example.make example --lock=example.lock**

Create a 'snapshot' makefile (example.lock) with the versions that were actually used in building the codebase.

**drush make example.make --lock-update --no-build**

Output an updated makefile to the terminal, without building a codebase, based on the source makefile (example.make).
## Background

See these two issues from drupal.org for the back-story:
- [Issue #1212686: Create a versioned make file when running drush make with option --make-generate](https://drupal.org/node/1212686)
- [Issue #1325818: Generate an upgrade makefile from an existing makefile?](https://drupal.org/node/1325818)

... and this other issue and pull request for previous discussion:
- #53
- https://github.com/drush-ops/drush/pull/68


weitzman:
FYI, the --lock-update option no longer exists. Use make-update. See #1142

PR #673, opened 2014-06-08
Convert most of exec.inc to use Symfony Process component.

weitzman:
Here is a start at using Process instead of exec() and friends. There isn't much to gain here yet. I'm actually most interested in @greg-1-anderson taking a look at backend.inc and seeing how much we can simplify _drush_backend_proc_open() and related code by using Process. I'm fine with losing some functionality there if it improves maintainability.

### DIFF
--- a/commands/core/archive.drush.inc
+++ b/commands/core/archive.drush.inc
@@ -267,13 +267,13 @@ function drush_archive_dump($sites_subdirs = '@self') {
   file_put_contents("{$tmp}/MANIFEST.ini", $contents);
 
   // Add manifest to archive.
-  drush_shell_cd_and_exec($tmp, "$tar --dereference -rf %s %s", $destination, 'MANIFEST.ini');
+  drush_run(array("$tar", "--dereference", "-rf", $destination, 'MANIFEST.ini'), $tmp);
 
   // Ensure that default/default.settings.php is in the archive. This is needed
   // by site-install when restoring a site without any DB.
   // NOTE: Windows tar file replace operation is broken so we have to check if file already exists.
   // Otherwise it will corrupt the archive.
-  $res = drush_shell_cd_and_exec($workdir, "$tar -tf %s %s", $destination, $docroot . '/sites/default/default.settings.php');
+  $res = drush_run(array("$tar", "-tf", $destination, $docroot . '/sites/default/default.settings.php'), $workdir);
   $output = drush_shell_exec_output();
   if (!$res || !isset($output[0]) || empty($output[0])) {
     drush_shell_cd_and_exec($workdir, "$tar --dereference -vrf %s %s", $destination, $docroot . '/sites/default/default.settings.php');
@@ -285,7 +285,7 @@ function drush_archive_dump($sites_subdirs = '@self') {
   }
 
   // Compress the archive
-  drush_shell_exec("gzip --no-name -f %s", $destination);
+  drush_run(array("gzip", "--no-name", "-f", $destination));
 
   // gzip appends .gz unless the name already ends in .gz, .tgz, or .taz.
   if ("{$destination}.gz" != $final_destination) {
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -383,7 +383,7 @@ function drush_config_export($destination = NULL) {
   $branch = FALSE;
   if ($remote) {
     // Get the branch that we're on at the moment
-    $result = drush_shell_cd_and_exec($destination_dir, 'git rev-parse --abbrev-ref HEAD');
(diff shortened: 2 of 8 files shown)


greg-1-anderson:
I think that this PR could be merged, but I'm on the fence about whether I want to tackle the backend invoke refactor & #1126 for Drush 7 stable or not.

weitzman:
I'm OK with delaying this until after Drush7.

greg-1-anderson:
Untagging for Drush 7.  I don't mind if progress commits are put in early, but I'm going to wait until we split off the 7.x branch before working on the core of backend invoke.

weitzman:
This PR was failing tests so I rerolled and did a force push to this `process` branch. Lets see how tests go. Also, `drush_shell_proc_open` is not yet working as evidenced by ssh command (see below). I'm not clear on what that function is supposed to return so help is welcome. 


    $ dr -v @buytaert.prod ssh
    proc_get_status() expects parameter 1 to be resource, object given in exec.inc:240
    proc_close() expects parameter 1 to be resource, object given in exec.inc:241`

greg-1-anderson:
I should be able to take a look at this in the next couple of days.

greg-1-anderson:
I just did a rebase with master and a force-push of this.  I backed out 8eee9be; I'll bring that back later, when I'm ready to make it work. First, I'm going to do a little refactoring, and see how the unit tests look.

greg-1-anderson:
Making some progress here.  The biggest reasons to use the Symfony Process component is a) rely on Symfony for shell argument escaping, and b) allow us to make separable classes that do not require all of Drush core to execute.  At the moment, both of these goals are more "best practices" than an immediate need, but it's good to move all of our code in this direction, especially w.r.t. low-level components (e.g. exec and logging).

One of the big differences between the Process component and our current code is that our current code defines sprintf strings such as "foo --flag=%s --other-option %s", whereas the Process component expects arguments to be presented as individual items in an array.  (While there is a single-string variant, this method presumes that none of the arguments need escaping, and is therefore often unsuitable for our purposes.)

To address the issue of using Symfony argument escaping, I added a new function `drush_run()`, that takes its arguments as an array of items that may need escaping (or a single string that does not need to be escaped).  Please take a look at the last commit, and let me know if you have any feedback.  (n.b. I split out the 'interactive' variant to deal with later.)

However, if we are going to go to the effort of changing our calls to drush_shell_exec everywhere they appear, it also seems like it would be wise to just forgo our wrappers, and call the Symfony process component directly.  This would allow us to take advantage of asynchronous commands, and so on.

There is, however, a fairly large impediment to dispensing with our wrappers, and that is --simulate mode.  We want to be able to transition to using the Process component without having to place references to the 'simulate' global everywhere, or worse, give up the feature.  The one idea I have here is that we could encapsulate this global into our own custom ProcessBuilder.  This might give us a good migration path later, when we are using more Symfony components and can use DI to manage the objects we use.

I'll include a custom ProcessBuilder in my next commit.

greg-1-anderson:
Maybe we should move forward here, also. Any thoughts on the proposed ProcessBuilder?  The main problem here is that code needs to explicitly use the ProcessBuilder in our namespace, rather than using the one provided by Symfony, in order to get a Process component that supports 'simulated' mode.

This is the same sort of conundrum I mentioned in #1799, where adopting standard components can hold us back when they do not support the functionality we need.

weitzman:
ProcessBuilder looks good to me. I'll review the rest of the PR soon. Thanks for returning to this one.

weitzman:
Did you say that we need some upstream changes before proceeding here?

greg-1-anderson:
Our custom ProcessBuilder is necessary because the Symfony process component does not support "Simpulated" mode. This means that if Drush started using third-party libraries that used the process component, those would not support simulated mode when it is set by Drush. For that reason, I thought it might make sense to contribute some improvements back upstream, so that this behaviour could remain consistent. However, I think that the right order to do it in would be to make a version of backend invoke that uses the Symfony Process component first; that way, when we do go to make an upstream contribution, it will be with reviewed, implemented, and tested code. Getting this issue merged is a pre-req for using the process component in backend invoke. (Or do both at the same time.)

weitzman:
Ok to merge when you are ready. We can branch if you think this is 9.x material.

greg-1-anderson:
I squash-rebased to see if the tests are still passing.  I think I might rename `drush_run` to `drush_exec` before committing.

greg-1-anderson:
The tests are green, but I don't think it's quite ready to merge yet. I'll do more work here on Monday.

weitzman:
Lets add a @deprecated to drush_shell_exec as well ... Looking good to me.

greg-1-anderson:
I think we can say that this was superseded by #3758

PR #675, opened 2014-06-10
Adds support for building a phar.

unn:
I had asked @weitzman about phar support at DrupalCon Austin during his session.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 tests/phpunit.xml
 vendor
+box.phar
+drush.phar
 
 /lib/Console_Table-1.1.3/
--- a/README.md
+++ b/README.md
@@ -421,6 +421,14 @@ You can put your Drush command file in a number of places:
 
 In any case, it is important that you end the filename with ".drush.inc", so that Drush can find it.
 
+PHAR
+--------
+
+Drush can be built into a phar for distribution as a single file. To build:
+
+1. Download and install box.phar from http://box-project.org.
+2. Run `box build` in the drush directory.
+
 
 FAQ
 ------
--- a/box.json.dist
+++ b/box.json.dist
@@ -1,0 +1,24 @@
+{
+    "chmod": "0755",
+    "directories": ["commands", "includes", "lib", "misc"],
+    "files": [
+        "drush.info"
+    ],
+    "finder": [
+        {
+            "name": "*.php",
+            "exclude": [
+                "phing",
+                "test",
+                "tests",
+                "Test",
+                "Tests",
+                "Tester"
+            ],
+            "in": "vendor"
+        }
+    ],
+    "main": "drush.php",
+    "output": "drush.phar",
+    "stub": true
+}
--- a/includes/command.inc
+++ b/includes/command.inc
@@ -1357,7 +1357,7 @@ function _drush_find_commandfiles($phase, $phase_max = FALSE) {
   switch ($phase) {
     case DRUSH_BOOTSTRAP_DRUSH:
       // Core commands shipping with drush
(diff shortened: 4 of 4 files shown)


weitzman:
I looked at wp-cli project and they recommend phar install method. When folks need to use an alternate PHP or php.ini, they document a wrapper script that calls into the phar with a fixed php and php.ini

weitzman:
Nice feature! [self-update phar](http://moquet.net/blog/distributing-php-cli/)

mathewpeterson:
This looks pretty good. Is there anything I can do to help get this merged in?

greg-1-anderson:
We need to figure out how to use the `drush` bash script as a stub here, or, failing that, come up with an alternate strategy that allows folks to use Drush without the wrapper script.  This also relates to #1343; with the bash script, we have a strategy to automatically select the correct `drush` to use, in the event that a site-local Drush is available.  There is a phpenv-like alternative proposal, but it requires that the user correctly select the right `drush` script, which has a greater margin for (user) error.

weitzman:
Might be unblocked by #1595 (to be merged soon)

weitzman:
I made a 'phar' branch in this repo and added a box.json.dist file. Unfortunately, we have the same problem as before. Using a phar, we happily enter into the drush finder but we can't pcntl_exec or passthru because those functions can't find the drush.launcher since its only accessible via PHP stream wrapper. The exact warning is `Warning: pcntl_exec(): Error has occurred: (errno 2) No such file or directory in phar:///Users/moshe.weitzman/tmp/drush-8.0.0-beta14-99-ga9f49a8.phar/includes/startup.inc on line 325`.

So, we are no closer than before. We have to give up on the niceties of the drush.launcher in order to get phar. Those niceties include having a single file that folks can download that has all dependencies downloaded. They no longer need Composer.

greg-1-anderson:
We could support a phar version of Drush if we gave up on the site-local Drush feature. In this scenario, though, the user would be in risk of "dependency hell" if the dependent libraries bundled in the phar were incompatible with the versions bundled with Drupal. The version of Drush would need ot match the version of Drupal exactly. Users would have problems if they had multiple versions of Drupal 8 on the same machine.

Perhaps we should discuss in a meeting, and summarize here.

weitzman:
Why do we have to give up on site local Drush? I'm imaging a scenario where starts with her global drush.phar and its finder hands off to a site local drush. That should work fine.

greg-1-anderson:
The question, though, is, if the phar enters at the drush finder instead of the drush launcher, what happens on invocations of drush that do -not- redispatch to a site-local Drush?

I suppose we could detect that we are in 'phar mode', and just call drush_main() directly from the finder in that instance.  It would also probably work if we just called drush_main() directly any time the finder selected a launcher that is right next to the finder.  I guess in phar mode, there would be no launcher next to the finder; we could just call drush_main() in that instance as well.

So, I think you're right, we should be able to build a phar and make all scenarios work.

greg-1-anderson:
Made a quick prototype. Maybe someone could try merging it in here, and see if we end up with all of the goodness of a phar, with the flexibility of the Drush finder.

weitzman:
Fixed up the prototype and pushed to 'phar' branch. Not quite done yet as including drush.php gives some ugly output due to shebang and also getting error `Fatal error: Call to undefined function drush_sql_get_class() in phar:///Users/moshe.weitzman/c/h/tmp/drush-8.0.0-rc1-6-gcf7578a.phar/includes/environment.inc on line 368`. Will work on this again soonish.

Then we have to figure out how to update documentation since phar users without a site-local drush don't get the niceties of the launcher (environment variables).

weitzman:
Lets continue in #1693

PR #726, opened 2014-07-18
Enabled configurable retry attempts for downloads

andrewlcg:
This is an attempt to fix a common bugbear of CI processes - if a git repo you need to clone is unavailable for a few seconds when you're running drush make the entire make process fails. This attempts to address this common problem by providing options to retry a download/clone multiple times.

make_download_factory has been modified to call the selected download function in a while loop until it's either successful or hits a limit defined by the drush option download-attempts. Additionally direct calls to the 'private' _make_download_file have been replaced by a new function make_download_simplefile which ensures the encapsulation of all downloads is carried out within make_download_factory.

A single failed download now provides a warning message, rather than firing a drush error, but failure to download after the number of attempts specified by download-attempts elicits an error.


kenorb:
+1

I had to run `drush make` 3-5x times until it downloaded all libraries at the same time.

greg-1-anderson:
pm-* is deprecated

PR #750, opened 2014-08-11
Do not restore owner/group of files when extracting module release tar

IlariExove:
This pull request is for 6.x branch. When extracting tarballs as root (or equivalent) file owner and group is restored. Drush should not restore file owner for drupal.org releases as it doesn't make sense.


weitzman:
Closing in favor of #751.

PR #751, opened 2014-08-11
Do not restore owner/group of files when extracting module release tar

IlariExove:
This pull request is for master branch. When extracting tarballs as root (or equivalent) file owner and group is restored. Drush should not restore file owner for drupal.org releases as it doesn't make sense.


greg-1-anderson:
This PR appears to have merit and may still be of value, but is really old. If anyone wanted to resurrect it, we could consider on the latest 8.x branch.

PR #766, opened 2014-08-20
Add a pm-projectinfo (pmpi) command

greg-1-anderson:
pm-projectinfo displays information on the projects available on the current Drupal site, including the list of extensions they contain (the reverse mapping of pm-info) and the Drush commands they define, if any.  If no projects are names, all are displayed.  With the --drush option specified, only projects that define Drush commands are listed.

Only projects that are part of a Drupal site are considered by this command.  Standalone Drush extensions are not visible.

This command came up as part of a discussion with Tim in Wellington.  He needed to be able to map from a project name (already downloaded) to the modules it contains.  While it seems that his actual use case might be better served by an improved pm-updatecode command (that respects module dependencies during partial upgrades -- maybe condtionally?), it does all the same seem to be really useful to have the inverse function for pm-info available.

### DIFF
--- a/commands/pm/pm.drush.inc
+++ b/commands/pm/pm.drush.inc
@@ -208,6 +208,24 @@ function pm_drush_command() {
     'drupal dependencies' => array($update),
     'aliases' => array('rf'),
   );
+  $items['pm-projectinfo'] = array(
+    'description' => 'Show a report of available  projects and their modules.',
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
I still think this is useful in core, but I clearly have not used it much.  Committed to master and 7.x-3.x branch of drush_extras.

PR #770, opened 2014-08-21
Fix 'array_merge(): Argument #1 is not an array context.inc:261'

bchavet:
I found that if you specify an 'include' option in a drush alias, the following warnings are returned during a drush rsync:

````
array_merge(): Argument #1 is not an array context.inc:261                                                   [warning]
array_unique() expects parameter 1 to be array, null given context.inc:261                                   [warning]
````

This is a result of $cache being set to bool(false) as a result of a call a little farther up to drush_get_context(DRUSH_INCLUDE). According to http://php.net/manual/en/function.array-merge.php, the recommended method of ensuring both arguments to array_merge() are arrays is to do a typecast, which is what this patch does.

### DIFF
--- a/includes/context.inc
+++ b/includes/context.inc
@@ -259,7 +259,7 @@ function drush_set_config_special_contexts(&$options) {
             }
           }
           else {
-            $cache = array_unique(array_merge($cache, $value));
+            $cache = array_unique(array_merge((array)$cache, (array)$value));
           }
           // Once we have moved the option to its special context, we
           // can remove it from its option context -- unless 'propagate-cli-value'


greg-1-anderson:
Are you trying to use 'include-paths' in your alias (to add a --include option to rsync), or did you actually mean to use 'include' (tell Drush where commandfiles are located)?  You cannot use 'include' in an alias record to pass --include options to rsync; use 'include-paths' instead.  Read `drush topic docs-strict-options` for an explanation.

In any event, I could not reproduce this issue using either 'include' or 'include-paths'.  Perhaps it has been fixed already; try again with Drush master, and open a new issue with more detailed information on how to reproduce if you are still having trouble.

PR #782, opened 2014-08-29
Add a hook_drush_make_post_download().

arnested:
Take action after a project in a make file has been downloaded.

Similar to hook_drush_pm_post_download() but invoked after each project
in a make file has been downloaded.

Notice that hook_drush_pm_post_download() is also invoked for make file
projects of type 'pm' (native Drupal.org projects).


greg-1-anderson:
Make is deprecated.

greg-1-anderson:
Make is deprecated.

PR #818, opened 2014-09-15
Process missing modules in drush pm-list command.

Chi-teck:
Sometimes modules have been removed without being uninstalled. That leads to [serious performance problem](https://www.drupal.org/node/1081266). I suppose it would be handy to track such modules with  `drush pm-list` command.

missing-modules

### DIFF
--- a/commands/pm/pm.drush.inc
+++ b/commands/pm/pm.drush.inc
@@ -765,19 +765,45 @@ function drush_pm_list() {
   }
 
   $result = array();
-  $extension_info = drush_get_extensions(FALSE);
-  uasort($extension_info, '_drush_pm_sort_extensions');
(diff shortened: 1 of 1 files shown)


jonhattan:
Which are the heuristics behind this? drush_get_extensions() won't include missing modules so you are comparing with entries in system table?

OTOH, this should be implemented in other place, probably within drush_get_extensions(), so other commands such as pm-info leverage it.

jonhattan:
This is fixed in D8, and there's a patch for D7 that has hit RTBC at least once https://www.drupal.org/node/1081266

OTOH it seems there's already a contrib tool - https://www.drupal.org/sandbox/joseph.olstad/2430429

PR #822, opened 2014-09-17
Don't include system.module for versions > 6.x

catch56:
This allows drush to work when running the core patch from https://www.drupal.org/node/1068932

Just avoids including system.module if we already got the version from bootstrap.inc


weitzman:
Core patch has not gone in

PR #842, opened 2014-09-24
Fix sql-sync not propagating command-specific options to sub-commands

juampynr:
Currently, the following example at `examples/example.aliases.drushrc.php` does not work:

``````
```
 #$aliases['live'] = array(
 #    'parent' => '@server,@dev',
 #    'uri' => 'mydrupalsite.com',
 #     'target-command-specific' => array (
 #       'sql-sync' => array (
 #         'skip-tables-list' => 'comments',
 #       ),
 #     ),
 #  );
```
``````

 The reason is that skip-tables-list is an option for sql-dump, not for sql-sync. The above would only work if the setting was set to the sql-dump command instead of the sql-sync command. However, this may be desired just for sql-sync so it is not a good solution. Besides, this must be frustrating developers which do not understand why tables are not ignored.

We should propagate sql-sync options to its sub-commands such as sql-dump or sql-sanitize. Perhaps on a follow up we could remove options which belong to sub-commads defined at sql-sync and instead use `allow-additional-options => ('sql-create', 'sql-dump', 'sql-sanitize')`.

### DIFF
--- a/commands/sql/sync.sql.inc
+++ b/commands/sql/sync.sql.inc
@@ -82,6 +82,15 @@ function drush_sql_sync($source, $destination) {
   $global_options = drush_redispatch_get_options() + array(
      'strict' => 0,
   );
+  // Merge command specific options for sql-sync with global options so they are applied
+  // on sub-commands such as sql-dump.
+  if (!empty($source_settings['command-specific']['sql-sync'])) {
+    $global_options = array_merge($global_options, $source_settings['command-specific']['sql-sync']);
+  }
+  // Same for source-command-specific options for sql-sync.
+  if (!empty($source_settings['source-command-specific']['sql-sync'])) {
+    $global_options = array_merge($global_options, $source_settings['source-command-specific']['sql-sync']);
+  }
 
   if (drush_get_context('DRUSH_SIMULATE')) {
     $backend_options['backend-simulate'] = TRUE;

### CODE REVIEW

**Comment by cafuego** [comment]
I'm on drush 6.5.0, which appears to include the code that addresses #923 but there is no difference in behaviour, the --(source|target)-db-url options are ignored.

**Comment by BarisW** [comment]
Same here, on drush 7.0-dev. 

drush sql-sync @client.vm @client.test --source-database=civicrm --target-database=civicrm

Doesn't work. It copies the default database to the correct destination (so --target-database works!).

`drush @client.vm sql-dump --database=civicrm` 
and
`drush @client.vm sql-drop --database=civicrm`

both work fine

**Comment by weitzman** [comment]
@BarisW - can you post a log with --verbose? Also post the content of your aliases and any relevant config. are you originating this call from @client.vm or @client.test or another host?

**Comment by weitzman** [comment]
I'm not quite sure why this issue was temporarily closed. Seems like i thought that this got merged. Anyway, the PR has handling for source-command-specific but not target. Is that deliberate? ping @greg-1-anderson

**Comment by weitzman** [comment]
Alternatively, what about adding more calls in drush_sql_sync_init like so:

```
drush_sitealias_command_default_options($source_settings, 'source-', 'sql-dump');
drush_sitealias_command_default_options($source_settings, 'source-', 'sql-create');
drush_sitealias_command_default_options($target_settings, 'target-', 'sql-query');
drush_sitealias_command_default_options($target_settings, 'target-', 'sql-sanitize');
```

weitzman:
Now that I think of it, Wouldn't the above calls be needed whenever a command calls out to another command. The command specific would be lost in that case as well. Seems like we could go crazy adding calls to drush_sitealias_command_default_options() from compound commands like updb (calls out to update-status), make, confirg-merge, etc.

weitzman:
I just experimented with drush_sitealias_command_default_options() and I can't figure out how it persists any changes that it makes to the site record. Maybe this is not the right function to use and I have been incorrect for these past few posts.

greg-1-anderson:
The current command-specific implementation is kind of tangled.  I'd like to rework it as part of the alias / context refactoring.  This would impact backend invoke as well, probably.

greg-1-anderson:
Drush 9 currently does not support command-specific options in site aliases. Might be nice to re-add this feature using the new APIs, but that would require a complete rework.

Issue #844, opened 2014-09-25
Add a command to destroy the ID map of a migration


tstoeckler:
When running a migration, the migrate system stores a map of already migrated IDs in the ID map plugin. (By default this is stored in the `{migrate_id_MIGRATION_ID}` table.)

Core does not provide any means to clear this through the UI, so that when developing a migration - and thus generally needing to run it over and over again until it works - one needs to manually truncate the appropriate table(s) each time.

This is very tedious, so a Drush command would be incredibly helpful.

See also https://www.drupal.org/node/2342695

PR #845, opened 2014-09-25
Add a migration-destroy-id-map command

tstoeckler:
See #844


weitzman:
migrate commands are in core drush anymore.

PR #880, opened 2014-10-14
Update Sqlpgsql.php for Windows

ChristopheMayet:
Under a windows environment with a Postgres database, the variable "PGPASSFILE=..." will be interpreted as a command (not found of course, as it's not a valid command).
The result is it asks for password (but silently with the -q parameter), so the drush command is blocked and the password can't be input by the prompt.

In order to set the variable properly, I added a "SET" command if it's under windows platform.
I don't know if it's the best way to handle environment issue here, but it seems to work.

### DIFF
--- a/lib/Drush/Sql/Sqlpgsql.php
+++ b/lib/Drush/Sql/Sqlpgsql.php
@@ -25,9 +25,14 @@ class Sqlpgsql extends SqlBase {
 
   public function command() {
     $environment = "";
+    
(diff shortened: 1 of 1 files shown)


kenorb:
Related: [drush can't find Drupal site after migrating to PostgreSQL](https://stackoverflow.com/q/36910614/55075).

weitzman:
Fixed in #2973

Issue #915, opened 2014-10-21
Drush shouldn't install completions for executables it doesn't install


orlitzky:
By default, drush installs bash completions for a bunch of commands, none of which it installs:

```
# Register our completion function. We include common short aliases for Drush.
complete -o bashdefault -o default -o nospace -F _drush_completion d dr drush drush5 drush6 drush7 drush.php
```

The only executable that drush actually installs is "drush", so it shouldn't lay claim to those other commands/aliases that it doesn't own.

PR #939, opened 2014-10-29
Only install completions for the "drush" command

orlitzky:
This is a potential fix for issue #915. I've tried to document the procedure everywhere that aliases/bash-completion are already mentioned in the docs.


weitzman:
Lots has changed with drush init. Dont think we want this anymore.

PR #1000, opened 2014-11-18
Fix drush logging to be compatible with latest Drupal 8.1.x code

Stormwind99:
Otherwise drush gives following error when using drupal-8.1.x 

PHP Fatal error:  Trait 'Drupal\Core\Logger\RfcLoggerTrait' not found in /root/.composer/vendor/drush/drush/lib/Drush/Log/DrushLog.php on line 12
Drush command terminated abnormally due to an unrecoverable error.                                                                      [error]
Error: Trait 'Drupal\Core\Logger\RfcLoggerTrait' not found in /root/.composer/vendor/drush/drush/lib/Drush/Log/DrushLog.php, line 12

### DIFF
--- a/commands/core/drupal/environment.inc
+++ b/commands/core/drupal/environment.inc
@@ -8,7 +8,6 @@
  */
 
 use Drupal\Core\Site\Settings;
-use Drupal\Core\Logger\RfcLogLevel;
 
 /**
  * Get complete information for all available modules.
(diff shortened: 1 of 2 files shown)


jonhattan:
I guess you mean drupal-8.0.x. In any case, this is a regression: `RfcLoggerTrait` was introduced a month before this PR was created, on 2014-10-14 in http://cgit.drupalcode.org/drupal/commit/?id=b82e28d5e467d4922068a8f59382ac9970e2e92c

The effect of this PR is that wd messages stop showing while a command is performing, because the loglevels doesn't match.

Issue #1006, opened 2014-11-21
The config-get command does not display overridden values


jhedstrom:
Set `$config['system.site']['name'] = 'foo'` in `settings.php`. Then

``` sh
$> drush config-get system.site
```

Note the original value is displayed, not the overridden one. Core [currently behaves the same](https://www.drupal.org/node/2379983), but figured here at least we could add an argument to retrieve overridden values instead.

PR #1007, opened 2014-11-21
Adds an option to view overridden config.

jhedstrom:
- Fixes issue #1006

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -40,7 +40,7 @@ function config_drush_command() {
         'value' => 'required',
       ),
       'include-overridden' => array(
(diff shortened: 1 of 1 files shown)


weitzman:
https://www.drupal.org/node/2392319 went in so more work is likely needed here.

jhedstrom:
Actually, the option still works as expected. The major change in the issue referenced above was with regards to how configs were pulled if they were going to be saved, eg, instead of

```php
$config = \Drupal::config('foo')->set('bar', 'baz')->save();
```
one now calls

```php
$config = \Drupal::configFactory()->getEditable('foo')->set('bar', 'baz')->save();
```

for reading config, the logic used herein is still applicable afaik.

PR #1034, opened 2014-12-04
Improve readability of help output

danshumaker:
This does a couple things I find very helpful for help output.  
1. It increases the column widths of the commands from 20 to 40.   I find 20 to be entirely unreadable when reading acquia and pantheon long commands.
2. It moves the aliases into their own column so they are easier to quickly read.  I feel readability is becoming a larger issue since now there are about 250 commands printed when you have both acquia and pantheon commands included.

Hopefully you also find this helpful.

Enjoy.

Dan

### DIFF
--- a/commands/core/help.drush.inc
+++ b/commands/core/help.drush.inc
@@ -375,8 +375,9 @@ function drush_help_listing_print($command_categories, $format) {
     // Build rows for drush_print_table().
     $rows = array();
     foreach($commands as $cmd => $command) {
-      $name = $command['aliases'] ? $cmd . ' (' . implode(', ', $command['aliases']) . ')': $cmd;
-      $rows[$cmd] = array('name' => $name, 'description' => $command['description']);
+      $name = $cmd;
+      $alias = $command['aliases'] ? ' (' . implode(', ', $command['aliases']) . ')': '';
+      $rows[$cmd] = array('name' => $name, 'aliases' => $alias, 'description' => $command['description']);
     }
 
     // Vary the output by mode: CLI or HTML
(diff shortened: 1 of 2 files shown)


grugnog:
drush_print_table() is already pretty responsive and auto-adjusts each column based on both the terminal width and the content of the column. Basically it determines the optimum width of each column to minimize the number of wrapped characters across all columns.

The issue here is that we are fixing the left hand column width, which I think was a workaround because we layout each table separately, and want the different tables in each command group to line-up. You can see the automatic wrapping behavior by dropping the fixed width - i.e.:
drush_print_table($rows, FALSE, array('label' => 40));

I think an underlying issue then is that the different command group tables do not line up. Whilst each individually is easier to read this way (for me, at least) as a whole they are a bit harder to scan for an individual command (although nowadays there are so many I normally need to scan for the group first).

There are a few potential ways we might proceed.

1) Adding a % based fixed width as proposed here might improve the issue for a given set of command names and terminal width, but I think is still going to be far from optimal in many combinations.

2) We could go with the "ragged" (but optimal) command group tables as above, but add a drush_print() below each section. For me at least this makes it quite a lot easier to scan, both for sections (which I find blend together in the current listing) and for commands within a section. i.e.
-      drush_print_table($rows, FALSE, array('name' => 20));
+      drush_print_table($rows, FALSE);
+      drush_print();

3) We could use the lower level function drush_table_column_autowidth() across all the group rows. Then, the column width stats/calculations would be run in advance, producing the optimal widths considering all the groups. The resulting widths passed into each table as fixed widths as it is printed.

4) More ambitious still would be to rework (or more likely rewrite) the table layout system to be more flexible, allowing "span" rows etc. This would allow us to output all the command groups as a single table.

weitzman:
I think 2) seems like a relatively quick experiment and might improve things. Thats where I would recommend starting.

greg-1-anderson:
Take off the Drush7 milestone if we don't want to fix this right now -- I just wanted to keep this from falling out of sight while we discuss, since there was an IRC mention.

grugnog:
I am taking a pass at improving the table interface to allow % and min/max width restrictions.

ccharlton:
I am available to test this. I am a heavy Drush user.

weitzman:
Table wrapping work is now in #2603

PR #1055, opened 2014-12-16
Fix #1054 by more selectively pruning empty project array items.

### LINKED ISSUE #1054: Support additional per-project parameters when using --lock option
With ```--lock``` we output a fully-realized makefile, but we lose important additional parameters. For example, ```projects[...][do_recursion] = 0``` in a source makefile won't carry over to the lockfile.

ergonlogic:
This fixes issue #1054. I basically just added a recursive array_filter() re-implementation that allows '0' values to remain.


greg-1-anderson:
+1

ergonlogic:
I re-factored make somewhat already in #1162. I'll have to re-check that this is still an issue, but then we also want to be able to output yaml in --lock and make-generate in #1110, and a couple others. So there's a number of related issues that'll tie into this.

PR #1057, opened 2014-12-17
Only prompt for en/dis confirmation when dependencies are impacted.

lkmorlan:
Port of: https://www.drupal.org/node/2037053#comment-7633493


weitzman:
Still makes sense IMO. Needs reviews.

weitzman:
This is in drush9

PR #1063, opened 2014-12-19
Add support for git repositories in the 'includes' directive.

pmatias:
# Git support for the includes directive
## Objective

Be able to retrieve makefiles from git repositories
## The reason

In our company we work with multiple makefiles for hundreds of different projects, and we are currently working on a better way to organise our several dozens of makefiles. One approach that we've found to be very practical for us is the following _example_ structure:

```
api = 2
core = 7.x

; platform.make contains core drupal and several main modules/themes/libraries
; that we need to have available for all of our projects
includes[platform] = "platform.make"

; sub_platform.make contains specific modules/themes/libraries for
; specific requirements on a different set of projects
includes[subplatform] = "sub_platform.make"

; [...]
```
## The problem

The only problem with this setup is that our makefiles are distributed in several _private_ git repositories (divide and conquer: this way we can have better control on what project/patches are inside on the different types of platforms) so getting the makefiles from raw urls is off the table since it will need some kind of authentication, and we're not going to write nasty stuff to make that happen.
## The solution

The 'includes' directive supports git repositories. This way, we get to assemble the main makefile the way we need it (also we get to define the core directive in the remote makefile that we want, and not the main one, giving us more control on Drupal versions).
## Code proposal

With this code, drush make will support the following sentences in the makefiles:

```
includes[remote][makefile] = 'drupal.make'
includes[remote][download][type] = "git"
includes[remote][download][url] = "git@github.com:organisation/repository.git"
; branch could also be tag or revision, code relays on make_download_git function
includes[remote][download][branch] = "7.x"
```

Also, despite our reasons of needing this feature, it seems like a natural feature that 'includes' could use, since it currently supports local files, relative path local files and remote urls.
## Final thoughts

The idea is to meet the requirement _without_ going against the community standards, so if you think that we can achieve the same with another approach, or you have a different approach for the submitted code in this PR, please let me know.

Thanks!

### DIFF
--- a/docs/make.md
+++ b/docs/make.md
@@ -327,7 +327,7 @@ makefiles to override the keys/values of former makefiles.
       download:
         type: "git"
         url: "git@github.com:organisation/repository.git"
-        # Branch could be tag or revision, it relays on the standard drush git download feature.
+        # Branch could be tag or revision, it relies on the standard drush git download feature.
         branch: "master"          
 
 ### Defaults


pmatias:
Thanks for the quick response on this PR, @jhedstrom (and everyone who provided feedback) ! 
One more question though, as pointed out by @dickolsson, should I create a separate PR for 6.x?

Thanks!

PR #1064, opened 2014-12-19
drush_is_local_host() doesn't seem to get run for shell alias commands that start with !

davereid:
I have the following in my sites/all/drush/drushrc.php:

``` php
$root = str_replace('/drush', '', dirname(__FILE__));
$options['shell-aliases']['build-theme'] = '!cd ' . $root . '/themes/custom/mytheme && npm install && bundle install && grunt --env=production';
```

And my Drush alias is defined as the following:

``` php
$aliases['not-really-remote'] = array(
  'root' => '/var/www/local.mysite.com/',
  'uri' => 'local.mysite.com',
  'remote-host' => 'Corsair250D', // This matches my current *local* computer's host name.
);
```

When I run drush @not-really-remote build-theme I get a SSH prompt, which is very unexpected:

```
davereid@Corsair250D$ drush @not-really-remote build-theme
The authenticity of host 'corsair250d (127.0.1.1)' can't be established.
ECDSA key fingerprint is 0a:e9:83:1c:59:0e:e1:b9:a1:2d:43:e8:83:77:e9:6f.
Are you sure you want to continue connecting (yes/no)?
```

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -1258,7 +1258,7 @@ function drush_core_execute() {
  * Helper function for drush_core_execute: run one shell command
  */
 function _drush_core_execute_cmd($site, $cmd) {
-  if (!empty($site['remote-host'])) {
+  if (!empty($site['remote-host']) && !drush_is_local_host($site['remote-host'])) {
     // Remote, so execute an ssh command with a bash fragment at the end.
     $exec = drush_shell_proc_build($site, $cmd, TRUE);
     return (drush_shell_proc_open($exec) == 0);


davereid:
Well we'll need to inform Acquia that they need to add #check-local on all the Drush aliases they are providing to subscriptions, and to have everyone re-download their alias files from Acquia once that's been completed. :/

greg-1-anderson:
The explicit #check-local was added to make it easier to talk to sites on virtual machines, e.g. on some other port on localhost.  It is not checked consistently.

An alternative to drush_is_local_host() would be to do this test yourself in your aliases.drushrc.inc file, and add in the 'remote-host' field if you are not local.

We need to decide what our final policy should be for Drush 7.

greg-1-anderson:
In Drush 8, we should remove #check-local completely, and have folks kill the 'remote-host' element in their alias alter hook if they want that behavior.

Example needed.

frederickjh:
I used the code after the `// TLDR; MAGIC STARTS HERE.` line in this gist, [derekderaps/drush-alias-remote-host-check.php](https://gist.github.com/derekderaps/6e0474b9f9a9012638d5), at the end of one of my alias files and it works OK as long as the host resolves to 127.0.0.1. 

I had one host (webenabled.net development environment) that they include a subdomain in the login details (**xyz**.dev6.webenabled.net). With the subdomain it did not resolve to 127.0.0.1, but I figured out that I do not need to included the subdomain for the alias to work and without the subdomain (dev6.webenabled.net) it resolves to 127.0.0.1.

frederickjh:
There is also has a blog post about that code: [Level up your Drush-Fu with aliases that work across all environments](http://www.mediacurrent.com/blog/make-your-drush-aliases-work-local-and-remote).

greg-1-anderson:
Shell aliases removed from Drush 9.

PR #1065, opened 2014-12-19
Added --force option to pm-download to skip MD5 validation.

davereid:
I hate having to wait for all the MD5 hashes to be updated properly in order to download new module updates. Can we provide an option to bypass it?

### DIFF
--- a/commands/pm/package_handler/wget.inc
+++ b/commands/pm/package_handler/wget.inc
@@ -70,12 +70,14 @@ function package_handler_download_project(&$request, $release) {
   }
 
   // Check Md5 hash.
-  if (drush_op('md5_file', $download_path) != $release['mdhash'] && !drush_get_context('DRUSH_SIMULATE')) {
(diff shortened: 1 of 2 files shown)


weitzman:
Seems like the DA is not going to fix this quickly. I'm going to merge this soon unless I hear differently.

weitzman:
I changed option name and moved option definition to wget engine. Thanks @davereid

davereid:
Can we please get this backported to 6.x too? Or I'm not too familiar what the policy is, but this affects all versions of Drush equally, not master.

drumm:
https://www.drupal.org/node/2357551 is what I mentioned in my last comment here. We're going to be able to deploy that, including purging caches on repackaging, in the next few days, maybe a week or two. Watch that issue and https://www.drupal.org/change-notifications for updates.

I don't remember if I got an answer to my ?nocache= question, I don't see one here.

PR #1066, opened 2014-12-21
Make drush use @alias work even if called from a script

jurgenhaas:
If `drush use @alias` is called from a script, then posix_getppid() will return a process id which is not the same as the process of the hosting shell and hence the stored alias won't be recognized by subsequent calls to drush which makes drush use @alias pretty useless. This only applies if drush is called through another script, which will be the case in a lot of scenarios.

To resolve this we should grab the session id of the current process which will go up the tree to the shell which is hosting any tree of scripts that are eventually calling drush.

### DIFF
--- a/includes/sitealias.inc
+++ b/includes/sitealias.inc
@@ -2020,7 +2020,7 @@ function drush_sitealias_get_envar_filename($filename_prefix = 'drush-drupal-sit
   }
 
   $tmp = getenv('TMPDIR') ? getenv('TMPDIR') : '/tmp/';
-  return "{$tmp}/drush-env/{$filename_prefix}" . posix_getppid();
+  return "{$tmp}/drush-env/{$filename_prefix}" . posix_getsid(posix_getppid());
 }
 
 /**


grugnog:
I agree that `use` should be avoided in scripts - currently you can either have the alias passed into the script as an argument or variable, or have the script operate on the "current" alias set by prior command (via either cd or `use`). Both of these avoid the potential for side effects.

@jurgenhaas can you describe what you are doing in more detail? There may be another solution, either by changing what you are doing slightly, or perhaps there is a more generally useful change to Drush that could support this.

jurgenhaas:
On my production servers I have to make sure, that drush is always executed in the context of the web server user account - www-data in my case. Therefore, the file ```/usr/local/bin/drush``` is a script that itself calls ```/usr/local/share/drush/drush/drush``` as the www-data user.

In this context, drush is always called via a script and therefore ```posix_getppid()``` returns a different id every time as it returns the pid of the script and not the hosting shell. That's what makes ```drush use``` useless unfortunately.

So, if there was any alternative way to force usage of drush by a specific user, then I'd be rather happy. Or, "re-engineering" the script ```/usr/local/bin/drush``` so that it handles ```drush use``` on its own without calling drush might work as well. But the simple change in the PR seems much more straight forward and I can't actually think of a negative side effect. Or is there any that I've missed?

grugnog:
I see - it sounds like the script is really an interactive wrapper, not an automated script. I wonder if you could replace this with an alias (which shouldn't affect the ppid). Alternatively, I wonder if you altered the script to finish with "exec sudo [...] drush [...]" that would help - it should do (as exec would terminates the wrapper) but I haven't tested it.

In terms of the patch, it seems like it may break other use cases, by making the "use" apply across multiple terminals (selecting a windowing or terminal manager app pid), rather than being terminal shell specific (as is the intent).

weitzman:
Seems like there is not enough support for this PR. Feel free to keep discussing alternatives here.

jurgenhaas:
I was getting back to this one and further tests showed that not the interactive script wrapper is the real problem, it's the fact that drush is called via ```sudo``` - that is changing the parent process ID every time it's been called. Where as the sid always remains the same but only within the current session not across terminals.

@grugnog I was trying if the proposed PR was having an effect on other terminals and it doesn't. Just run the following PHP script with ```sudo php script.php```and you will see that the PPID changes every time where the SID always remains the same. If you run the same script from multiple terminals on the same host, then you'll get different values. So there is no cross terminal impact.

Test-Script:

```
<?php

echo "PPID=" . posix_getppid();
echo "\n";
echo "SID=" . posix_getsid(posix_getppid());
```

However, all of this is only getting relevant because we are forcing drush to be executed as www-data so that we don't get permission issues down the road. If anyone had an alternative solution for this without using sudo, I'd be glad to hear any suggestions.

jurgenhaas:
I was getting back to this one and further tests showed that not the interactive script wrapper is the real problem, it's the fact that drush is called via ```sudo``` - that is changing the parent process ID every time it's been called. Where as the sid always remains the same but only within the current session not across terminals.

@grugnog I was trying if the proposed PR was having an effect on other terminals and it doesn't. Just run the following PHP script with ```sudo php script.php```and you will see that the PPID changes every time where the SID always remains the same. If you run the same script from multiple terminals on the same host, then you'll get different values. So there is no cross terminal impact.

Test-Script:

```
<?php

echo "PPID=" . posix_getppid();
echo "\n";
echo "SID=" . posix_getsid(posix_getppid());
```

However, all of this is only getting relevant because we are forcing drush to be executed as www-data so that we don't get permission issues down the road. If anyone had an alternative solution for this without using sudo, I'd be glad to hear any suggestions.

jurgenhaas:
I was getting back to this one and further tests showed that not the interactive script wrapper is the real problem, it's the fact that drush is called via ```sudo``` - that is changing the parent process ID every time it's been called. Where as the sid always remains the same but only within the current session not across terminals.

@grugnog I was trying if the proposed PR was having an effect on other terminals and it doesn't. Just run the following PHP script with ```sudo php script.php```and you will see that the PPID changes every time where the SID always remains the same. If you run the same script from multiple terminals on the same host, then you'll get different values. So there is no cross terminal impact.

Test-Script:

```
<?php

echo "PPID=" . posix_getppid();
echo "\n";
echo "SID=" . posix_getsid(posix_getppid());
```

However, all of this is only getting relevant because we are forcing drush to be executed as www-data so that we don't get permission issues down the road. If anyone had an alternative solution for this without using sudo, I'd be glad to hear any suggestions.

greg-1-anderson:
While I didn't like this PR in the context of scripts, I could potentially get behind it for the use case described above--especially since different terminals do not share sessions.

It is a slight detriment that, with this PR, a script that used `drush use` would also change the site being used in the terminal it was called from.  I have one remaining serious reservation about this, though, and that is, what happens if one dev violates best pracitce, and uses `use` in a script, and then passes that script on to another user, who is not accustomed to the `use` command?  In this scenario, if the bad script changes the used site, and the user has not set up their prompt to show which site is being used, they will find themselves mysteriously "stuck" in some Drupal context, with no good clues about why Drush is selecting some arbitrary site as the target for all commands.

If we were to commit something like this, I think perhaps we'd want to require that it be "enabled" via some option that folks who wanted it could set in their drushrc.php configuration.  Then, by default, you'd get a separate 'use' context for scripts / sudo, unless you enabled the 'persistent' flavor of 'use'.

jurgenhaas:
That sounds reasonable and maybe a command line option would be even better as it then won't be pregnant but for those specific SUDO scenarios.

greg-1-anderson:
cli options / drushrc.php settings are interchangeable.  `drush --foo=bar` == `$options['foo'] = 'bar';`

greg-1-anderson:
Also:

`alias sudrush='sudo drush --foo=bar'`

If folks are going to want to only *sometimes* enable this, then maybe in the default mode, `use` saves the site to both files (PPID-based, and SID-based)?  That might be a needless doubling of config files for an edge case, though.

greg-1-anderson:
Maybe the `sudrush` alias could copy the ppid file over the sid file, call sudo drush, then copy the sid file over the ppid file.  Then the complexity is only there for folks who want it.

jurgenhaas:
Copying ppid file would have to work 2-ways: when ```drush use @alias``` is used, the ppid file needs to be copied over to the sid file and for each subsequent ```drush``` call that is *not* ```drush use``` it needs to copy back from the sid to the new ppid file of the new sudo process. And all that needs to happen within that one sudo process that also calls drush because only inside there is a way to know its ppid, or am I missing anything?

This sounds really really hard and I'd prefer the cli option.

greg-1-anderson:
Closing old issues.

Issue #1086, opened 2015-01-07
No submodule mapping found.


mikeshiyan:
When downloading Drupal projects as submodules (`$ drush dl devel --package-handler=git_drupalorg --gitsubmodule`) drush (or git?) generates paths with double-backslashes in .gitmodules file:
```
[submodule "sites\\all\\modules\\contrib\\devel"]
	path = sites\\all\\modules\\contrib\\devel
	url = git://git.drupal.org/project/devel.git
```
which results in `No submodule mapping found in .gitmodules for path 'sites/all/modules/contrib/devel'` message on `$ git submodule update`.

But! When doing the same without drush (`$ git submodule add --branch 7.x-1.x http://git.drupal.org/project/devel.git sites/all/modules/contrib/devel`) paths contain forward slashes (`sites/all/modules/contrib/devel`) and work properly with other git submodule commands.

Drush Version: 7.0-alpha7;
OS: Win7.

mikeshiyan:
Also I noticed that .git directory is created inside the downloaded project root, and not in the superproject's one (drupal/.git/modules/...). Is it OK, because `$ git submodule update` doesn't work even with proper paths and without any error messages? I didn't found any info on this.

jonhattan:
in #566 we introduced the use of realpath() to prevent issues with symlinks. Now I've read that documentation for that function states that "On windows realpath() will change unix style paths to windows style."

@mikeshiyan Perhaps newer versions of git can work with windows paths?

PR #1087, opened 2015-01-07
Fixed Win backslashes in paths in .gitmodules.

mikeshiyan:
A solution to the https://github.com/drush-ops/drush/issues/1086.


jonhattan:
I wonder if this was an issue before the changes introduced in #566

PR #1088, opened 2015-01-08
Updated search-* commands to work with Drupal 8

imron:
Updated search-index and search-status to check for
drupal_major_version >= 8 and if so use drupal 8 specific
code to get the remaining search items and update the search
index.


weitzman:
Merged without testing, since the replaced code is out of date anyway. Thanks.

PR #1090, opened 2015-01-09
Keep configuration in confgis dirs during installation

webflo:
```drush site-install``` wipes all files in config dirs during installation. Please add an option to keep these config files. Otherwise it is not possible to install Drupal from CLI with [Config Installer](http://drupal.org/project/config_installer].

Example command

```
drush si config_installer --keep-config -y config_installer_staging_configure_form.staging_directory=../config/staging config_installer_site_configure_form.account.name=foobar config_installer_site_configure_form.account.pass=foobar config_installer_site_configure_form.account.mail=foobar@example.com
```

// cc @alexpott

### DIFF
--- a/commands/core/site_install.drush.inc
+++ b/commands/core/site_install.drush.inc
@@ -37,6 +37,7 @@ function site_install_drush_command() {
         'example-value' => 'directory_name',
       ),
       'writable' => 'Make CMI and other dirs writable by both web and CLI users. Suitable for non Prod environments.',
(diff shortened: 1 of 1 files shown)


weitzman:
Looks reasonable.

* The option should be named keep-config
* The option should only show for D8 sites. [An example](http://api.drush.org/api/drush/commands%21user%21user.drush.inc/function/user_drush_help_alter/master)

webflo:
Here we go.

PR #1094, opened 2015-01-12
Change REQUEST_URI to '/'.

### LINKED ISSUE #1093: Drush urls aren't properly detecting that clean urls are enabled
When drush is generating urls against a Drupal 8 site, either core or drush isn't properly detecting that clean urls are available, resulting in `index.php` being part of the url.

This is visible when running commands such as `uli`, where it isn't much of an issue, but also in search index commands, where it does become an issue as those urls are ultimately displayed to the end user.

jhedstrom:
- Fixes #1093 

This 'fixes' the issue in local testing, but I'm not sure if it is the proper change.

### DIFF
--- a/lib/Drush/Boot/bootstrap.inc
+++ b/lib/Drush/Boot/bootstrap.inc
@@ -564,7 +564,8 @@ function _drush_bootstrap_drupal_site_validate() {
     $_SERVER['PHP_SELF'] = '/index.php';
   }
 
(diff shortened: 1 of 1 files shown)


weitzman:
Would be good to test uli with prior versions of Drupal and make sure this doesn't break anything.

jhedstrom:
Manual testing of `drush uli` against Drupal 6, 7, and 8 works with this change.

weitzman:
Did some quick testing and the cherry picked into master. Thanks.

PR #1097, opened 2015-01-13
Explain how to set up ZSH autocompletion in the docs

colans:
It's very easy to set up shell auto-completion for ZSH as well as BASH, but this isn't documented anywhere officially.

I added the small amount of information to the documentation, which works quite well for me.  The details are originally from [Use Drush Autocomplete with Zsh](http://mark.shropshires.net/blog/use-drush-autocomplete-zsh).


weitzman:
Thanks @colans. Merged into master. Feel free to submit a new PR for 6.x branch if desired.

PR #1102, opened 2015-01-16
Use Symfony var-dumper component as Boris variable inspector

damiankloip:
The drush php command currently uses Boris, and the standard `ColoredInspector` to show variables.

The Symfony var-dumper component has much more going for it. It shows much more information about objects, like property visibility, array counts... The `ColoredInspector` does not show any protected properties (which is most properties in D8 now) or floats for example. See the screenshots below:

#### Array
drush_____sites_d8_core_ _php_ _264x50

#### Action config entity
drush_____sites_d8_core_ _php_ _264x50

#### Array
__forks_drush_drush_____sites_d8_core_ _php_ _264x50_and_testing_text_format___hrweb

#### Action config entity
__forks_drush_drush_____sites_d8_core_ _php_ _264x50

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -18,7 +18,8 @@ function cli_drush_command() {
  */
 function drush_cli_core_cli() {
   $boris = new \Boris\Boris(drush_sitealias_bootstrapped_site_name() . '> ');
-  $boris->setInspector(new \Boris\ColoredInspector());
+  // Use the Symfony var-dumper to inspect variables.
+  $boris->setInspector(new \Drush\Boris\VarDumperInspector());
 
   // Boris will never return control to us, but our shutdown
   // handler will still run after the user presses ^D.  Mark
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
     "php": ">=5.3.0",
     "d11wtq/boris": "*",
     "symfony/yaml": "~2.2",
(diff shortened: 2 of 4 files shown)


dawehner:
Best thing!!!!

LionsAd:
+1 to this change

weitzman:
1. We use Kint library to do this in devel. It would be nice to standardize on Kint or VarDumper. Kint can do CLI. Has anyone compared them? I'd like to hear more about this before commit.
1. How are you guys using Boris? I'd love a blog post or a tweet about it. I dont think enough people know about it.

damiankloip:
I tried Kint::dump() at the cli too and from what it looks like (IMO), not as good as VarDumper. It should similar information but in a way less concise way.

Same examples as PR description, but with `Kint::dump()`:

drush_____sites_d8_core_ _php_ _264x50

LionsAd:
I think VarDumper has the better syntax output, too. Kint is way too verbose for quickly seeing what is relevant.

dawehner:
IMHO the CLI environment has so many more restrictions, (for example you can't collapse stuff).
Given that, I don't think we need consistency in terms of using the same library.

Yes, it would be nice if both libraries would work more together

weitzman:
OK, I'm fine with this PR. Needs reroll as changes were made to composer.lock.

wimleers:
AWESOME! :)

Issue #1105, opened 2015-01-20
Drupal 8 root check means we can't move/rename some core files anymore


stefanruijsenaars:
As of (https://github.com/drush-ops/drush/commit/9da9a52acd22c15d1510dc76b2eebd32512d1c5c), Drush does the following check to validate a correct drupal installation:

```php
function drush_valid_drupal_root($path) {
  if (!empty($path) && is_dir($path) && file_exists($path . '/index.php')) {
    // Drupal 8 root. Additional check for the presence of core/composer.json to
    // grant it is not a Drupal 7 site with a base folder named "core".
    $candidate = 'core/includes/common.inc';
    if (file_exists($path . '/' . $candidate) && file_exists($path . '/core/misc/drupal.js') && file_exists($path . '/core/core.services.yml')) {
      return $candidate;
    }
```

This means that we will break a couple of drush installs with a cryptic "no root found" error in case we rename/move any of:

* core/core.services.yml
* core/includes/common.inc
* core/misc/drupal.js

The first 2 are unlikely to change but **core/misc/drupal.js** we are looking to move to **core/assets/js/drupal.js** as per this issue: https://www.drupal.org/node/1663622

theodoreb:
What's the reason for the drupal.js check in the first place?

ruijse:
You're right, we probably don't need that check. In addition to **core/includes/common.inc** we used to just check for **composer.json** in the root (which is admittedly not very drupal specific), but now that we check for **core/core.services.yml**, further checks may not be needed.

From https://github.com/drush-ops/drush/commit/9da9a52acd22c15d1510dc76b2eebd32512d1c5c

```// Extra checks to insure that the candidate is in fact part of a Drupal root,
// and not some other non-Drupal directory
function _drush_valid_drupal_root_extra_checks($path, $candidate) {
 // We also require that there must be a file misc/drupal.js in a sibling
 // folder of the candidate file.```

ruijse:
I have posted a pull request to remove drupal.js from the root check over at https://github.com/drush-ops/drush/pull/1106

greg-1-anderson:
So, we have a basic problem that there is no official way to detect a Drupal 8 root directory, so we try to do the best we can by checking for "fingerprint" files, especially those that contain the name "drupal".  We want to avoid false positives and false negatives to the greatest extent possible.  We also want to avoid parsing text files during the bootstrap, if possible.

[...]

PR #1106, opened 2015-01-20
Remove drupal.js from D8 root check.

stefanruijsenaars:


ruijse:
See https://github.com/drush-ops/drush/issues/1105

greg-1-anderson:
The problem here is that we've weakened the Drush boostrap, making false positives more likely.  Without drupal.js, none of the files we check contain the name "drupal" any longer.

I see two good ways to move forward (lacking an official way to detect the root, as mentioned in #1105):
1. Temporarily remove drupal.js from the bootstrap test.  Put it back in later, once file locations are stable.  Add a comment with the new location, and a TODO to put the check back in later.
2. Check for both the new and the old location of drupal.js.  Put in a comment to remove the old location once we are ready to de-support old Drupal 8 betas (maybe after Drupal 8 stable release).

I'd slightly prefer #2.

stefanruijsenaars:
#2 seem good to me. The new location is core/assets/js/drupal.js

stefanruijsenaars:
Updated the pull request to check for both locations.

greg-1-anderson:
Looks good.  Thanks.

stefanruijsenaars:
Thanks. Can we do a new Drush release so this can be installed on
qa.drupal.org?

2015-02-09 15:56 GMT+01:00 Greg Anderson <notifications@github.com>:

greg-1-anderson:
Seems reasonable.  Discussion on releases is in #1123.

PR #1109, opened 2015-01-22
Add Composer's global bin directory to the system PATH for Mac OS

esod:
Mac OS uses a BSD sed which is different from the Unix BSD. This instruction:

sed -i '1i export PATH="$HOME/.composer/vendor/bin:$PATH"' $HOME/.bashrc

on the documentation page http://drush.org/en/master/install/, doesn't work for Mac OS users.

We can use sed in Mac OS this way:

sed -i '' '1i\
export PATH="$HOME/.composer/vendor/bin:$PATH"' $HOME/.bashrc

but this instruction still doesn't work if a user's ~/.bashrc is empty. When helping users install Drush 7, I often find their ~/.bashrc to be empty. Custom configuration is instead written to their ~/.bash_profile.

Mac OS users can still use ~/.bashrc for this. This patch makes sure the export PATH line gets written.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -26,10 +26,10 @@ Composer - One Drush for all Projects
         sed -i '1i export PATH="$HOME/.composer/vendor/bin:$PATH"' $HOME/.bashrc
         source $HOME/.bashrc
         
-        For Mac OS users whose $HOME/.bashrc file may be empty or non-existent.
-        touch $HOME/.bashrc
-        { echo 'export PATH="$HOME/.composer/vendor/bin:$PATH"'; cat $HOME/.bashrc; } | tee $HOME/.bashrc
-        source $HOME/.bashrc
+        For Mac OS users
+        touch $HOME/.bash_profile
+        { echo 'export PATH="$HOME/.composer/vendor/bin:$PATH"'; cat $HOME/.bash_profile; } | tee $HOME/.bash_profile
+        source $HOME/.bash_profile
 
 * To install Drush 6.x (stable):
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -25,6 +25,11 @@ Composer - One Drush for all Projects
 
         sed -i '1i export PATH="$HOME/.composer/vendor/bin:$PATH"' $HOME/.bashrc
         source $HOME/.bashrc
+        
+        For Mac OS users whose $HOME/.bashrc file may be empty or non-existent.
+        touch $HOME/.bashrc
+        { echo 'export PATH="$HOME/.composer/vendor/bin:$PATH"'; cat $HOME/.bashrc; } | tee $HOME/.bashrc
+        source $HOME/.bashrc
 
 * To install Drush 6.x (stable):
 


esod:
I don't know what 

Source file not found - core/themes/bartik/logo.png.

has to do with my patch (See https://travis-ci.org/drush-ops/drush/jobs/47953426) 

The sed command that my patch concerns itself with really doesn't work on Mac OS. When the command fails it makes users weary of upgrading. And we all know how developers are ;)

weitzman:
That logo.png bug was fixed this morning. Still a couple fails to fix on Drupal 8.

esod:
Great. Thanks. Now, how do I re-submit my patch so it gets tested again?

esod:
I did some more research and it turns out we actually can't use ~/.bashrc for Mac OS users (See http://apple.stackexchange.com/a/13019/110668 for why). I've amended my patch to write to the ~/.bash_profile file instead. 

If Mac OS users were to use ~/.bashrc, each time they logged in they'd have to source their ~/.bashrc. Otherwise they get the dreaded -bash: drush: command not found error.

weitzman:
We reference .bashrc elsewhere in the codebase. we should be consistent. This issue should probably just focus on the sed improvement, not filename change. as no longer recommending sed, i defer to @greg-1-anderson or anyone else to review.

weitzman:
I'd be OK removing our example code and simply telling users to edit their PATH.

greg-1-anderson:
I agree that we should get rid of all of the code, and just advise people to add the `export PATH` line to their bash configuration file.  Folks who use ~/.bash_profile should also add the `export PATH` line to their ~/.bashrc file, as this is the configuration file that is used when initializing the shell non-interactively.  Perhaps this never happens on desktop/laptop systems, though; for Drush, non-interactive shells are only initialized during remote operations.

esod:
I think it's important to make the change to the documentation. However you see fit will be fine. Remember that folks are writing about Drush 7 already and referring to the sed command. See https://drupalize.me/blog/201408/upgrading-drush-work-drupal-8 for an example. For Mac users the sed command returns no response but when they log back in to their computers, drush, which has probably worked flawlessly and consistently for them for years, is not working. Eeek!

esod:
Something simple like:

`echo "PATH=$PATH:~/.composer/vendor/bin" > ~/.bash_profile`

or `...> ~/.bashrc`

also works and gives the user the basics for setting up this configuration themself. 

@greg-1-anderson I think average Mac users only use their `~/.bash_profile`.

weitzman:
I removed the instructions entirely and linked to Composer. Send them a PR
if desired :)

esod:
Apologies if this is a nitpick, but that won't work either. I'm on a new account on my Mac right now. Composer is already installed globally, so I can run

`$ composer global require drush/drush:dev-master`

which runs fine. I then run

`$ drush --version` in the same terminal and 

`-bash: drush: command not found` is returned, which we know to expect. New users may not. 

I think it's necessary to tell folks to add `~/.composer/vendor/bin` to their paths, even if it's an instruction in the subsequent paragraph and not a markdown-ed command line instruction. I prefer the command line instruction. It's easy to spot for cutting and pasting and it breaks instructional paragraphs into easily understandable segments.

greg-1-anderson:
What if we fixed up the PATH automatically, in a composer post-install script?

If ~/.bash_profile does not exist, but ~/.bash_login or ~/.profile does exist, then create ~/.bash_profile, and make it source ~/.bash_login or ~/.profile.  (On Linux, also consider ~/.bashrc).  We want to do this so that we can consistently tell people that the PATH was set in ~/.bash_profile; however, if we just created a ~/.bash_profile, it would cause Bash to stop using ~/.bash_login and ~/.profile, as Bash only uses the first config file it finds.

Adjust the PATH at the top of both ~/.bash_profile and ~/.bashrc.  (If ~/.bash_profile already sources ~/.bashrc, then we can skip modifying ~/.bash_profile.)  We want to be sure to adjust ~/.bashrc to catch non-interactive invocations.  Even though most folks won't use this on a desktop system, it can cause a lot of heartache if not set up correctly, so better to be consistent.

Before modifying any file, the script should first check to see if there is already a line that contains both  'PATH=' and '.composer/vendor/bin', anywhere in the file, then we trust that the file was correctly set up, and do nothing.

The script might need to do something entirely different for Windows; a little bit of research would be necessary here.

Instructions for post-install scripts are here: https://getcomposer.org/doc/articles/scripts.md

Example snippet for Drush composer.json:
```json
{
    "scripts": {
        "post-update-cmd": "Drush\\Composer\\Installer::postUpdate",
        "post-install-cmd": "Drush\\Composer\\Installer::postInstall",
    }
}
```
Then we'd just need to add a class 'Installer" in lib/Drush/Composer/Installer.php, and define postUpdate and postInstall methods that calls the path-fixing script.

We'd have to continue to include the instructions on updating the PATH in the section on manual installation.

We'd have to continue to advise the user to type `source ~/.bash_profile` after running `composer install`, because there is no way for our subprocess to modify the environment variables of the running shell.

This would be a big service to Drush users, but I admit to some hesitation, as there are likely to be things other than drush in ~/.composer/vendor/bin.  I tend to think that it is likely to also be a big service to the users of the other commands there to have the PATH set correctly, but it would be annoying if Drush kept adding the PATH back on every update, if you didn't want it there.  We could mitigate this problem by having the script add a comment above the PATH modification, attributing the source of the line, and instructing the user that if they simply comment out the line (or perhaps remove the PATH line but leave the comment), then the line will stay gone, but if they delete everything, then it will come back.  Is that suitable good citizenship for this feature?

If this sounds suitably useful and awesome, then I will code up a proposal.

esod:
An all-in-one installer does sound awesome. For those who use a different shell, like zshell, let's assume they're sourcing their ~/.bash_profile in their ~/.zshrc. That's what I do.

Advising users to `source ~/.bash_profile` after running `composer install` seems perfectly reasonable to me. 

The comment above the PATH modification sounds like a good idea. I think commenting out the line will work better than removing the line and leaving the comment. If there is a comment and no line, I think the comment may get removed at some point in the future. You know, like I look at it three months from now and say "What's that doing there" and then remove it.

weitzman:
Please no composer post-install shipping with Drush. That cleverness belongs in Composer project, if anywhere. We just need docs here IMO.

esod:
At this moment, there is no instruction on the documentation page to add `$HOME/.composer/vendor/bin` to one's path. Composer users will probably already have their path variables set up, but the newer user may not. 

What about adding the following snippet above the `To update to a new version bullet`:

Now add Drush to your system path by placing this line:
export PATH="$HOME/.composer/vendor/bin:$PATH"
into your $HOME/.bash_profile (Mac OS users) or into your $HOME/.bashrc (Linux users).

greg-1-anderson:
+1 for @esod's  suggestion.

PR #1115, opened 2015-01-24
Fix unary operator expected error.

benjy:
I get the following error when I run Drush with xDebug enabled:

    Command stderr:
    /usr/local/drush/drush: line 129: [: -le: unary operator expected


greg-1-anderson:
See continuing discussion in #1119.

PR #1119, opened 2015-01-24
Re-submit #1115 to get Travis to re-test same change.

greg-1-anderson:
Again, this is working locally, but was causing the Travis build to break.  Re-submitting to get Travis to run it again.

### DIFF
--- a/drush
+++ b/drush
@@ -126,7 +126,7 @@ fi
 
 # If on PHP 5.3, disable magic_quotes_gpc and friends
 PHP_MINOR_VERSION="`\"$php\" -r 'print PHP_MINOR_VERSION;'`"
-if [ $PHP_MINOR_VERSION -le 3 ] ; then
+if [[ $PHP_MINOR_VERSION -le 3 ]] ; then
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Related to this, I think that once we create a branch for Drush 7, we should just go ahead and make Drush master depend on PHP 5.4, like Drupal 8 already does.  Then we can take out this comparison altogether, along with all of the magic quotes stuff.  Drush 7 will be plenty modern enough for folks still stuck on 5.3.

greg-1-anderson:
Yeah, this is already failing on Travis.  I'm not sure why.  Thoughts?  Should we proceed as suggested above?

benjy:
Removing them checks makes sense for the Drush 7 branch. +1 for that.

greg-1-anderson:
I'm going to presume that we will require php 5.4 in Drush master after we create the Drush 7.x branch.

Tagging this as "Drush7" so that this gets done, although the Drush 7 branch itself will still be php 5.3 (presumably).

greg-1-anderson:
Actually, I shouldn't hijack this PR, because we are not going to commit this.  Opened #1128 to track this going forward.

PR #1120, opened 2015-01-24
Support db-prefix when running the built-in sql-sanitize queries.

pdrakeweb:


weitzman:
This results in triple curly braces around the table name - `TRUNCATE TABLE {{{sessions}}};`. Is that desired?

pdrakeweb:
Two of the sets are removed during the string interpolation operations, leaving the final string with a single set.

Issue #595, opened 2014-04-21
Remove auto-disable of magic_quotes settings


weitzman:
We are forcibly disabling magic_quotes php settings since https://drupal.org/node/1748228. That issue has all the gory backstory.

I'm thinking that we can remove this feature for Drush7, and just document a fix. Thats roughly how we used to handle this. My motivation is that it annoys me to see all this noise in the logs when we redispatch.

I hesitate to remove this since it did cause some pain in the past and we haven't yet made php 5.4 a requirement. 

Any feedback from @greg-1-anderson and others?

PR #1121, opened 2015-01-25
Remove PHP 5.3 checks.

benjy:
Fix #1119

### DIFF
--- a/drush
+++ b/drush
@@ -124,12 +124,6 @@ if [ -n "$PHP_OPTIONS" ] ; then
   php_options="$php_options $PHP_OPTIONS"
 fi
 
-# If on PHP 5.3, disable magic_quotes_gpc and friends
-PHP_MINOR_VERSION="`\"$php\" -r 'print PHP_MINOR_VERSION;'`"
-if [ $PHP_MINOR_VERSION -le 3 ] ; then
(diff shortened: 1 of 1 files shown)


weitzman:
Not sure about this. We discussed this in #595 and the d.o. issue that is linked from there. Drush7 still works on PHP 5.3.

benjy:
Not sure either way, I guess Drush 7 only supporting 5.4 seems fair when Drupal 8 has a minimum requirement for 5.4. That said, this isn't a big issue, probably not the issue to finally drop 5.3.

The problem is with recent versions of xdebug, "php -r" gives a bus error. I've not tried to debug this.

Since this only started happening to me since I upgraded to 5.5 I think, how about a simple check like so?

    # If on PHP 5.3, disable magic_quotes_gpc and friends
    PHP_MINOR_VERSION="`\"$php\" -r 'print PHP_MINOR_VERSION;'`"
    if [ -z "$PHP_MINOR_VERSION" ]; then
      PHP_MINOR_VERSION=5
    fi

greg-1-anderson:
Drush 7 still supports Drupal 6 and Drupal 7, so it will continue to support php 5.3.

For Drush 8, we have decided to remove php 5.3 support, regardless of Drupal 6 & 7 support, since 5.3 is getting pretty old now.

weitzman:
There are a few more pieces that need updating. Grep the codebase for 5.3

Issue #365, opened 2014-01-09
mysql credentials leak in drush sqlc


durist:
I'm copying this verbatim from https://drupal.org/node/671906:

"drush sqlc" (and other commands that invoke the mysql cli) leak the database credentials. For example, after running "drush sqlc" I can see the password showing up in the "ps" listing.

My environment allows shell access to multiple Drupal vservers via a single file server, so this security bug renders drush practically unusable for my installation.

greg-1-anderson:
As mentioned in #1175, new versions of mysql complain when using the credential-leaking options in the mysql tool.

This issue might be easier to resolve now that sql-sync has been refactored.

weitzman:
@greg - this is issue is already closed and fixed.

greg-1-anderson:
Nice!  Didn't even notice.  :P

PR #1133, opened 2015-01-30
Ensure that --db-su works without --db-su-pw for site-install

hansfn:
In the follow up to #365 (commit ac050dd) the defaults-extra-file is only created if a password is given. This means that --db-su=root has no effect without the (optional) argument --db-su-pw being set when running site-install. The username is simply ignored.

### DIFF
--- a/lib/Drush/Sql/Sqlmysql.php
+++ b/lib/Drush/Sql/Sqlmysql.php
@@ -11,17 +11,15 @@ class Sqlmysql extends SqlBase {
   public function creds($hide_password = TRUE) {
     if ($hide_password) {
       // EMPTY password is not the same as NO password, and is valid.
-      if (isset($this->db_spec['password'])) {
-        $contents = <<<EOT
+      $contents = <<<EOT
 #This file was written by Drush's Sqlmysql.inc.
 [client]
 user="{$this->db_spec['username']}"
 password="{$this->db_spec['password']}"
 EOT;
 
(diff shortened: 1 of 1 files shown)


weitzman:
It looks like that commit also had `if (isset($this->db_spec['password'])) {` so this must be older code that you are removing. I'd like to see when it got added in order to better understand the implications.

hansfn:
Hi, weitman. Thx for reponding quickly.

The problem occurred when you changed the order of the if tests. The ```if (isset($this->db_spec['password'])) {``` made perfectly sense when it came first. Now it blocks writing of the defaults-extra-file if you don't provide a su password. Get it? (When developing locally many have empty root passwords for MySQL.)

weitzman:
I'm a little concerned that this PR will regress the my.cnf fix that we made in that commit. Dont have time to experiment.

hansfn:
Seriously, this is a no-brainer. The only thing that is changed is that the defaults-extra-file file is written independently of the password. This doesn't break the my.cnf fix, in addition it fixes the very serious problem that the su user is ignored. Example from before the fix:

```
`$ drush -d -d site-install standard --db-su=root --db-url=mysql://drush_si_test:1234@localhost/drush_si_test
...
Executing: mysql --database=information_schema --host=localhost --silent  < ".../druF3DF.tmp"
  ERROR 1044 (42000) at line 1: Access denied for user ''@'localhost' to database 'drush_si_test'
```

Notice that user has no username ...

PR #1135, opened 2015-01-30
Added user-edit command and db-drop-existing-tables to site-install - Drush6

dustinmoris:
Added a new option called "dp-drop-existing-tables" to the site-install
command. This option denotes whether the existing database should be
cleared before installing the Drupal tables.

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -273,6 +273,7 @@ function core_drush_command() {
         'description' => 'Password for the "db-su" account. Optional.',
         'example-value' => 'pass',
       ),
+      'db-drop-existing-tables' => 'Defaults to TRUE. Remove all tables in the target database before creating Drupal tables.',
       'account-name' => 'uid1 name. Defaults to admin',
       'account-pass' => 'uid1 pass. Defaults to a randomly generated password. If desired, set a fixed password in drushrc.php.',
       'account-mail' => 'uid1 email. Defaults to admin@example.com',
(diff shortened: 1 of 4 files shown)


dustinmoris:
I got the latest version of the 6.x branch and properly implemented this feature now. I will perform the change on the main branch soon as well.

weitzman:
Is there a substantial difference between this option and running sql-drop command before site-install? If it just a matter of convenience, I'm inclined to pass on this PR.

dustinmoris:
Actually this feature is about making it possible to install Drupal into an existing database without deleting tables. By default the site-install command deletes all tables from the database, which makes it difficult to co-exist Drupal with another application in the same database.

With my change nothing will change for existing integrations by default, but if needed you can change the behaviour by flicking the switch to false.

dustinmoris:
Hi, I have added one more powerful command which our ops team needed and I thought it would be a good command to add to Drush:

# Added a user-edit command to the user commands

This command enables me to edit any field of the user by specifying the
field as an option.

This is very powerful, because now I can not only edit any core user
field via Drush, but also any other field which has been introduced by
another community module (e.g. the password_policy module) and Drush
still remains without a dependency on community modules.

Example of usage:
>`drush user-edit MyUser1 --mail="myNewEmail@example.org"
--pass="new-password-1234" --force_password_change=0`

weitzman:
Please don't combine unrelated features into a single PR.

dustinmoris:
Makes sense, I reverted it and will commit it as a separate pull request.

weitzman:
We decided not to take user-edit into Drush core in #1222.

dustinmoris:
Okay, it's a pitty. What about the dp-drop-existing-tables? It is such a crucial thing not to delete all existing tables and I really would like to have it part of Drush core. I have reverted the changes in this pull request, so it is just the drop-existing-tables feature now. Why did this get closed?

PR #1136, opened 2015-01-30
Added db-drop-existing-tables to site-install

dustinmoris:
Added 'db-drop-existing-tables' switch to site-install command to optionally not delete existing tables in the target database, so that Drupal can co-habit a database with other applications.

This switch defaults to TRUE if not specified, so that normal functionality is unaffected.


dustinmoris:
I would like to get this little feature into the official branch after your re-factorings now :) 
Let me know if there is an issue with this! Thanks!

commands/core/site_install.drush.inc
```diff
@@ -181,7 +187,12 @@ function drush_core_pre_site_install($profile = NULL) {
     drush_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_SITE);
   }
 
-  $sql->drop_or_create($db_spec);
+  if ($db_drop_existing_tables) {
+    $sql->drop_or_create();
```
weitzman: I think we should remove  drop_or_create() method entirely. It is used in archive-restore only. Just copy its code to there.

commands/core/site_install.drush.inc
```diff
@@ -21,6 +21,7 @@ function site_install_drush_command() {
         'description' => 'Password for the "db-su" account. Optional.',
         'example-value' => 'pass',
       ),
+      'db-drop-existing-tables' => 'Defaults to TRUE. Remove all tables in the target database before creating Drupal tables.',
```
weitzman: Lets say move the 'Defaults to TRUE' to the end. Also, make the value an array and add key 'example-value' with value of 0 so usage is clearer. Thanks.

commands/core/site_install.drush.inc
dustinmoris: I have made this change! Good suggestion, now the code makes more sense when I read it as well!

commands/core/site_install.drush.inc
dustinmoris: Done, thanks :)

dustinmoris:
Hi, just wondering what the status is on this? An update would be great, thank you!

dustinmoris:
This PR hasn't received a response for a year now. Clearly no one is interested in this feature. I will close it now.

PR #1142, opened 2015-02-03
Add make-update command

jonhattan:
The purpose of this command is to improve and replace --lock-update. It uses the update_status engine to generate an updated makefile.


weitzman:
Would be great to merge this soon IMO.

weitzman:
Input from @jhedstrom and @ergonlogic would be great here.

ergonlogic:
This looks pretty solid. along with this, I'd suggest removing the `--lock-update` option from `drush make`. In retrospect, I think it was a bad design decision (on my part) to add an option that so significantly changed the `make` command.

Issue #1110, opened 2015-01-22
Make should generate YAML file


weitzman:
When writing a makefile via --lock, --lock-update, or make-generate command, make should be able to generate YAML in addition to its current ini format.

ergonlogic:
I'm not sure whether this is a valid approach, but I was thinking of converting the rendering of makefiles to use the output-format engine. I believe that this'd largely involve moving the ini-generating code into a new 'ini' output format. We'd probably want a validate hook to only allow 'yaml' and 'ini' as output formats for make-related commands, and remove 'ini' as an engine for most other commands.

I consider the 'ini'-style a largely legacy feature, that we should consider deprecating, and eventually removing. Isolating it in an engine should facilitate that. In addition, this'd probably be a step towards supporting other makefile formats, such as json, which'd be neat.

dreftymac:
+1

https://github.com/dreftymac/stackoverflow.questions.1715811

weitzman:
We could consider deprecating ini format here so that all we do is generate yaml. we just put make-convert command in drush so folks using ini can convert and then use all our tools in drush7 (i.e. generate-make, make-update, etc.)

miloskroulik:
I'm looking forward to this! Convert command would be nice!

weitzman:
make-convert is in. now the options that write ini need to learn to write yaml.

ergonlogic:
This is in now. I've added a `--format` option to the various commands that generate makefile. As per previous comments, we default to generating YAML.

weitzman:
Nice! I think this issue can be closed?

ergonlogic:
Sure

PR #1143, opened 2015-02-03
Adding command to convert legacy makefiles to yaml.

grasmash:
This is a useful command for converting from the legacy makefile format to YAML. 

I have a feeling I may not be using best practices by manually adding a `require` statement here. Open to feedback!

### DIFF
--- a/commands/make/make.drush.inc
+++ b/commands/make/make.drush.inc
@@ -118,6 +118,9 @@ function make_drush_command() {
   $items['make-convert-yaml'] = array(
     'bootstrap' => DRUSH_BOOTSTRAP_DRUSH,
     'description' => 'Convert a legacy makefile into YAML format.',
+    'arguments' => array(
+      'makefile' => 'Filename of the makefile to convert.',
+    ),
     'examples' => array(
       'drush makefile-convert-yaml example.make' => 'Convert example.make to example.make.yml',
     ),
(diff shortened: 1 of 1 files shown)


commands/make/make.drush.inc
```diff
@@ -115,6 +115,14 @@ function make_drush_command() {
     'aliases' => array('generate-makefile'),
   );
 
+  $items['make-convert-yaml'] = array(
+    'bootstrap' => DRUSH_BOOTSTRAP_DRUSH,
+    'description' => 'Convert a legacy makefile into YAML format.',
+    'examples' => array(
```
ergonlogic: We'll need an 'arguments' array here.

commands/make/make.drush.inc
```diff
@@ -281,6 +289,31 @@ function drush_make($makefile = NULL, $build_path = NULL) {
 }
 
 /**
+ * Drush callback; convert makefile to YAML.
+ */
+function drush_make_convert_yaml($makefile = NULL) {
```
ergonlogic: Presumably the input makefile is required. So why provide an argument default here?

commands/make/make.drush.inc
```diff
@@ -281,6 +289,31 @@ function drush_make($makefile = NULL, $build_path = NULL) {
...
+
+  drush_log(dt('Beginning to convert !makefile.', array('!makefile' => $makefile)), 'ok');
+
+  $file = $makefile . '.yml';
+  $info = make_parse_info_file($makefile);
+
+  require dirname(__FILE__) . '/../core/outputformat/yaml.inc';
+  $dumper = new drush_outputformat_yaml(array());
```
ergonlogic: Here you probably want to use Drush's engines, instead of a 'require' and a 'new'. Something like:

$dumper = drush_load_engine('outputformat', 'yaml');

commands/make/make.drush.inc
```diff
@@ -281,6 +289,31 @@ function drush_make($makefile = NULL, $build_path = NULL) {
...
+  $yaml = $dumper->format($info, array());
+
+  if (!$file) {
```
ergonlogic: When would this ever evaluate to true?

grasmash:
@ergonlogic I've made changes based on your review. All great comments!

grasmash:
Do you think it would be a good idea to ksort() the project? I'd personally love that... but I'm not certain it's something that we'd want to impose.

weitzman:
I'd like to implement this in such a way that it is trivial to do #1110. I'll leave it to the make maintainers to make sure that goes smoothly.

weitzman:
We can think about removing runtime support for ini style makefiles after this is in. Conversion would be trivial.

ergonlogic:
This looks much better. Thanks @grasmash!

As for `ksort()`'ing projects, it'd certainly make for more consistent makefiles. However, I don't think within this command would be the place to do it. Maybe in `make_parse_info_file()` itself, so that we're always working with a sorted array...

weitzman:
Merged. Also added a little cleanup in f1b75d0. Thanks @grasmash. Feel free to add that ksort in new PR.

Issue #1145, opened 2015-02-04
"Unable to copy" if drush make fails to set mtime of copied file


xurizaemon:
## Motivation

_drush_recursive_copy() fails if mtime preservation is unsuccessful; PHP touch() [won't touch a file if not owned by the current user](http://php.net/manual/de/function.touch.php#86781) and this can prevent drush make etc from succeeding where it otherwise could.

In [d.o#1914470](https://www.drupal.org/node/1914470) Drush introduced preserving mtime of copied files, and fails _drush_recursive_copy() if the touch() to set mtime is not successful.

The outcome is that PHP's touch() behaviour causes inconsistency between how Drush behaves and how other UNIX tools behave; two users who can otherwise operate on the same files by virtue of regular FS permissions are not able to use Drush make to update a shared codebase.

## Proposed resolution

Try to preserve mtime, but don't bail out if it fails. (Alternatively, exec UNIX touch instead of PHP touch()?)

## Demonstration

Users _chris_ and _gemma_ share primary group `www-editors` and are collaborating on a shared codebase. When the files are writable but not owned by the current user, downloading a module to overwrite works using `drush dl` but not when using `drush make`.

Confirm that users share a primary group:
```
chris:/tmp$ groups
www-editors chris
$ groups gemma
gemma : www-editors gemma
```

Download a couple of Drupal codebases as user _chris_.
```
chris:/tmp$ drush dl drupal --drupal-project-rename=example1
Project drupal (7.34) downloaded to /tmp/example1.                                                                                                                       [success]

chris:/tmp$ drush dl drupal --drupal-project-rename=example2
Project drupal (7.34) downloaded to /tmp/example2.                                                                                                                       [success]
```

Ensure they are group-writable.
```
chris:/tmp$ chmod -R g+rw example[12]
```

Download Views module to both example sites as user _chris_.
```
chris:/tmp$ cd example1
chris@fudge:/tmp/example1$ drush dl views
Project views (7.x-3.8) downloaded to /tmp/example1/sites/all/modules/views.                                                                                             [success]
Project views contains 2 modules: views_ui, views.

chris:/tmp/example1$ cd ../example2
chris:/tmp/example2$ drush dl views
Project views (7.x-3.8) downloaded to /tmp/example2/sites/all/modules/views.                                                                                             [success]
Project views contains 2 modules: views_ui, views.
```

Switch to user _gemma_.
```
chris:/tmp$ sudo su - gemma
gemma:/tmp$ cd /tmp
[...]

PR #1146, opened 2015-02-04
Issue #1145. Permit touch() to fail without consequences.

xurizaemon:


weitzman:
Please elaborate on why this is needed.

xurizaemon:
@weitzman I've updated the issue description with a real-world scenario that hopefully makes the issue clearer.

includes/filesystem.inc
```diff
@@ -264,9 +264,10 @@ function _drush_recursive_copy($src, $dest) {
   elseif (is_link($src)) {
     symlink(readlink($src), $dest);
   }
-  elseif (!(copy($src, $dest) && touch($dest, filemtime($src)))) {
+  elseif (!copy($src, $dest)) {
     return FALSE;
   }
+  touch($dest, filemtime($src));
```
weitzman: Lets add a comment about why we do it this way. Can be a link to #1146. Otherwise, this looks good. Thanks for the explanation.

PR #1161, opened 2015-02-09
Silence a warning when an alias record does not contian a 'root' element

greg-1-anderson:
Just putting this in the queue for Travis.

I did update .travis.yaml to get one branch tested, but then I had to put it back before merging with master. Is there a way to tell Travis to test a branch that contains a pattern, but NOT if it's on the master branch (UNLESS it IS a pull request)?  Probably not.  Just wondering if there's something we could live with that stayed in .travis.yaml all the time.

### DIFF
--- a/commands/core/sitealias.drush.inc
+++ b/commands/core/sitealias.drush.inc
@@ -94,11 +94,7 @@ function sitealias_site_set_complete() {
  * is the site specification for the given alias.
  */
 function _drush_sitealias_alias_list() {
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Wrong PR -- this one already merged. :P

greg-1-anderson:
This is the right branch, the commits are just strange.

greg-1-anderson:
All working now.

PR #1165, opened 2015-02-11
Autoload bootstrap

greg-1-anderson:
This is a continuation of #1156.  I started a new branch in case I needed to back up, but I'm pretty happy about how this is straightening out.

This patch no longer attempts to rely on Composer for initialization.  Doing Symfony Console-like stuff looks like a great idea, but we'll have to tackle this later, in Drush 8.  For now, I am maintaining the behavior of the current Drush bootstrap.  Commandfiles must exist in the usual magic locations (~/.drush, etc.)

The bootstrap is a lot cleaner now.  In drush.php, the main preflight phases are now:

  drush_preflight();
  $bootstrap = _drush_preflight_root();
  drush_preflight_site();

These preflight phases also appear in preflight.inc in this order, and their responsibilities are much more clearly defined.  valid_root() is moved to the Boot class.

- drush_preflight:  Load global config and global command files. Now also handles aliases and 'site-get'
- _drush_preflight_root:  Find the Drupal root, and decide which bootstrap class is needed.
- drush_preflight_site:  Find the site inside the Drupal root, and load config and command files associated with the root and site.

So, if you have a commandfile in ~/.drush, it can define a Boot class that is consulted when searching for the Drupal root from the cwd.  Once a Boot class is found, Drush remembers it for successive operations, so existing sitealias code & c. still works the same way it always has.

If this looks okay, I think that some of the code from the bootstrap.inc next to DrupalBoot could be moved into includes/bootstrap.inc.  Taking the "DRUPAL" out of some of these constants will have to be a Drupal-8 activity.

Oh, and Drush\Drush::autoload() still exists, but now the recommended time to call it is in your drush_hook_init().  I just realized that autoload() is still adding the commandfile; this is innocuous, but unnecessary.  I'll remove it.  The only purpose of this method will be to include the Drush extension's autoload.php file, if it was not already included as part of the Drupal site's autoload.php.


greg-1-anderson:
The tests passed when I ran on my branch: https://travis-ci.org/greg-1-anderson/drush/builds/50285664

If this is okay to commit, I will rebase and fix the merge conflicts.

includes/environment.inc
```diff
@@ -352,20 +352,18 @@ function drush_locate_root($start_path = NULL) {
...
+
+function drush_bootstrap_class_for_root($path) {
+  $candidates = drush_preflight_get_bootstrap_candidates();
+  foreach ($candidates as $candidate) {
+    if ($candidate->valid_root($path)) {
       return $candidate;
     }
   }
-  return FALSE;
+  return null;
```
weitzman: capitalize here, and above (in the diff)

lib/Drush/Boot/DrupalBoot.php
```diff
@@ -5,7 +5,26 @@
 class DrupalBoot implements Boot {
 
   function __construct() {
-    // @todo Find current version of Drupal and return more specific subclass. For now, there is only 1.
+  }
+
+  // @todo We could split this into separate classes for Drupal 8, and Drupal 7 + 6,
```
weitzman: Would be great to get to this. Not urgent though.

lib/Drush/Command/Commandfiles.php
```diff
@@ -0,0 +1,70 @@
...
+namespace Drush\Command;
+
+/**
+ * Default commandfiles implementation.
+ *
+ * This class manages the list of commandfiles that are active
+ * in Drush for the current command invocation.
+ */
+class Commandfiles implements CommandfilesInterface {
+  const FOO = 'my constant';
```
weitzman: classes can have constants :)

lib/Drush/Drush.php
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * The Drush API.
+ */
+class Drush {
+	static $commandfiles_cache;
+	
+	/**
+	 * Run a drush command in a new process
+	 */
+	static function drush($site, $command_name, $commandline_args = array(), $commandline_options = array(), $backend_options = TRUE) {
```
weitzman: Is this method and invoke() in use anywhere? In general, I'm a little hesitant to have a Drush class as that can become a grab bag of unrelated methods.

weitzman:
Left a few comments. I think this looks good. Might have to wait for Drush 8 though. Will chat with Greg.

drush.php
```diff
@@ -38,17 +38,15 @@ function drush_main() {
...
-   * - 'early' files can bootstrap when needed.
-   * - bootstrap constants are available.
-   */
-  $bootstrap_class = drush_get_option('bootstrap_class', 'Drush\Boot\DrupalBoot');
-  $bootstrap = new $bootstrap_class;
-  drush_set_context('DRUSH_BOOTSTRAP_OBJECT', $bootstrap);
-  $bootstrap->preflight();
+  // Find the selected site based on --root, --uri or cwd, and
+  // return the bootstrap object that goes with it.
+  $bootstrap = _drush_preflight_root();
```
greg-1-anderson: This is the part of the PR that is most important to me.  This and the corresponding code re-org that goes with it makes the Drush bootstrap much cleaner, and I'd rather not go keep the existing code around for all of the Drush 7 stable branch's life.

greg-1-anderson:
As I said offline, I'd like to get some of this stuff in for Drupal 7 branch for better maintainability.  Other parts, like making a Drush\Drush class, is not as important to me, and can wait until later.

The one thing we'd need to figure out, if we punt Drush\Drush and keep the rest, is how the existing procedural code should get a reference to the commandfile class.  This sort of thing is usually done with a singleton pattern, but that's not really a best practice in code that uses DI, and I don't know that we want to put in a dependency injection container in Drush 7.  That's why I settled for a static class (Drush\Drush) that held our other classes that use modern techniques.  Maybe we could just fake this with drush_set_context for now, if you don't want Drush\Drush.

greg-1-anderson:
In the commits above, I removed the Drush\Drush class, and went with simple procedural access functions.  Maybe we can put this in for Drush 7 just as a bootstrap  cleanup, and continue with DI stuff after we kick off Drush 8.

pirog:
Giving this a try with Backdrop so we can try to package in "Backdrush" with Backdrop apps for Kalabox2. The custom Boot class seems to load good and proper. 

Not as familiar with Drush or Backdrop as i would like to be... but it seems like after getting the custom class to load and writing a good valid_path method for backdrop the task is to write custom bootstrap.inc and command.inc files to actually handle the bootstrapping of backdrop. 

Does that approach sound on target to you?

pirog:
So i gave this a shot and it works decently (read: some basic stuff works like `drush status` or `drush uli`) well but we are still left with a couple problems. 

What i'm using right now is this custom Boot class plus associated files
https://github.com/kalabox/kalabox-app-examples/tree/master/backdrop/config/drush
(you can ignore the aliases)

Plus this hacky hack to get drush to recognize that Backdrop is Drupal 7 compatible.
https://github.com/pirog/drush/commit/5c3099a66d7b86d61333fb4f70288fd86b5cdf24

Some observations and thoughts here:

1. Right now this kind of spoofs drush into thinking that Backdrop is Drupal 7 which while somewhat accurate given Backdrops compatibility layer is most wholly accurate nor seems like a good idea.
2. This seems to produce some weird behavior like the correct "system.module" not getting loaded before drupal/environment.inc (and _7.inc) resulting in failure for a lot of commands like `drush pm-list`. It is likely this happens with other modules and commands as well. 
3. At the same time `drush_drupal_major_version` is used like 30 times throughout many command files which is also a problem unless we want to write a bunch of backdrop specific commands of those things which given how similar those commands are likely to be to D7 seems like not the right thing either.
4. Because of 1. we are able to use `drush_include_engine('drupal', 'environment')`  in our custom Backdrop  `bootstrap.inc` with some degree of success but as i read it maybe we actually want a  `drush_include_engine('backdrop', 'environment')` and/or other Backdrop specific engines. 

I'm more than happy to keep working on this but it would be good to get some feedback on these things and ideally a consensus-ish direction on how to proceed.

pirog:
Also let me know if you want me to create a new issue for this!

pirog:
**Update over here:**

So actually have got this working on Backdrop ... seemed like some commands just didn't have a bootstrap level. Not sure why that is but if you give them FULL BOOTSTRAP everything seems to work. 

Of course i dont think we want to blanket full bootstrap all commands that have no bootstrap level specified so going to investigate this a little more and see if there is a better way to add this data in.

greg-1-anderson:
Waiting for Moshe to get back from vacation, to review this and see if it we can get it in for Drush 7.

The Backdrop continuation work is looking good, but I think that this discussion should be continued at #342.

pirog:
Cool cool!

I'll kick these comments over there then!

On Sunday, February 22, 2015, Greg Anderson <notifications@github.com>
wrote:

> Waiting for Moshe to get back from vacation, to review this and see if it
> we can get it in for Drush 7.
>
> The Backdrop continuation work is looking good, but I think that this
> discussion should be continued at #342
> <https://github.com/drush-ops/drush/issues/342>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/drush-ops/drush/pull/1165#issuecomment-75471671>.
>


-- 
Cheers,

Mike Pirog
Kalamuna
www.kalamuna.com

weitzman:
This looks good to me. Please take one last look at my code comments and see if any are unaddressed. Then, feel free to merge.

Also, I think the title of this PR should be updated as it morphed over time.

greg-1-anderson:
Merged into master.  Thanks.

pirog:
Pumped to see this merged!

On Wed, Feb 25, 2015 at 12:42 PM, Greg Anderson <notifications@github.com>
wrote:

> Closed #1165 <https://github.com/drush-ops/drush/pull/1165>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/drush-ops/drush/pull/1165#event-242017281>.
>



-- 
Cheers,

Mike Pirog
Kalamuna
www.kalamuna.com

Issue #1168, opened 2015-02-11
Use drush_set_error in queue run


bjeavons:
I'm using jenkins to call drush queue-run but a build won't fail unless I throw an exception from worker callback which also keeps the queued item from being deleted. If drush_queue_run() could support an error return state from the worker it could call drush_set_error while also allowing the item to be deleted.

bjeavons:
PR #1169 is for 6.x. For master I'm not sure yet how to trigger an error within the queue class or if should be within the drush_queue_run().

weitzman:
In master, you can throw an exception in your queue class and it will be caught as an error by the drush invoke code that runs the command.

greggles:
Thanks for the insight, Moshe. That makes it sound like PR #1169 could be accepted to Drush 6.x without the need to port to another branch.

Any thoughts on that proposal?

weitzman:
Closing this in favor of #1169

PR #1169, opened 2015-02-11
1168: Use drush_set_error if queue worker callback returns false

bjeavons:
For #1168

### DIFF
--- a/commands/core/queue.drush.inc
+++ b/commands/core/queue.drush.inc
@@ -116,8 +116,11 @@ function drush_queue_run($queue_name) {
   $end = time() + (isset($info['cron']['time']) ? $info['cron']['time'] : 15);
   $queue = drush_queue_get($queue_name);
   while (time() < $end && ($item = $queue->claimItem())) {
-    $function($item->data);
+    $result = $function($item->data);
     $queue->deleteItem($item);
+    if (isset($result) && $result === FALSE) {
+      drush_set_error('DRUSH_QUEUE_RUN_ERROR', dt('Error during queue run.'));
+    }
   }
 }
 
--- a/commands/core/queue.drush.inc
+++ b/commands/core/queue.drush.inc
@@ -116,11 +116,15 @@ function drush_queue_run($queue_name) {
   $end = time() + (isset($info['cron']['time']) ? $info['cron']['time'] : 15);
   $queue = drush_queue_get($queue_name);
   while (time() < $end && ($item = $queue->claimItem())) {
-    $result = $function($item->data);
-    $queue->deleteItem($item);
-    if (isset($result) && $result === FALSE) {
-      drush_set_error('DRUSH_QUEUE_RUN_ERROR', dt('Error during queue run.'));
+    // A callback can report an error by returning FALSE. A callback can throw an
+    // exception to halt the run and keep the item from being deleted.
+    if (!$function($item->data)) {
+      $run_result = FALSE;
     }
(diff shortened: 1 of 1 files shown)


commands/core/queue.drush.inc
```diff
@@ -116,8 +116,11 @@ function drush_queue_run($queue_name) {
   $end = time() + (isset($info['cron']['time']) ? $info['cron']['time'] : 15);
   $queue = drush_queue_get($queue_name);
   while (time() < $end && ($item = $queue->claimItem())) {
-    $function($item->data);
+    $result = $function($item->data);
     $queue->deleteItem($item);
```
weitzman: Is it right for all use cases to unconditionally delete the item even if $function returned FALSE?

commands/core/queue.drush.inc
```diff
@@ -116,8 +116,11 @@ function drush_queue_run($queue_name) {
...
+    if (isset($result) && $result === FALSE) {
+      drush_set_error('DRUSH_QUEUE_RUN_ERROR', dt('Error during queue run.'));
```
weitzman: This is going to give the same error for each item that fails. Do we want one error? if not we should ideally state which item is failing. Also, if we rely on the exception technique in D8 we will only get one error message.

commands/core/queue.drush.inc
bjeavons: Worker callback return values are a new concept so I think we get to define what's right. Right now throwing an exception is the only way to keep the item from being deleted, which seems appropriate.

commands/core/queue.drush.inc
bjeavons: There aren't consistent IDs for queued items so I'm not sure how we can report on which item failed. I'd be fine with a single error message tho.

commands/core/queue.drush.inc
weitzman: Makes sense.

commands/core/queue.drush.inc
weitzman: If you get a chance please update the PR, moving this drush_set_error() out of the loop. Probably should be `return drush_set_error()`

bjeavons:
The issue was testQueue isn't expecting a status code of 1. I'll look into the aggregator_feeds run, it may be that this is an actual expected return value.
````
There was 1 failure:
1) QueueCase::testQueue
Unexpected exit code: /home/travis/build/drush-ops/drush/drush --root=/tmp/drush-sandbox/web --uri=dev --nocolor queue-run aggregator_feeds --yes
Failed asserting that 1 matches expected 0.
/home/travis/build/drush-ops/drush/tests/drush_testcase.inc:378
/home/travis/build/drush-ops/drush/tests/drush_testcase.inc:453
/home/travis/build/drush-ops/drush/tests/queueTest.php:37
````

weitzman:
Any progress here, @bjeavons?

davereid:
I don't think we can use this logic unless core does the same.

davereid:
We should be using exception handling only and not a FALSE return logic. There's no standard that queue running callbacks in core should return a FALSE value on failure, so this would likely cause problems with existing queues.

davereid:
I would vote this be closed in favor of https://github.com/drush-ops/drush/pull/1434, which I'm going to work on backporting through the various branches of Drush to ensure it handles exceptions properly.

davereid:
#1434 has been merged, so closing this in favor of the work there.

Issue #1163, opened 2015-02-09
Allow config-import to work with partial sets of config


jhedstrom:
The config UI allows one to import individual configs. The current `config-import` command only works with entire config sets. It would be nice to have an option to import only configs found in a target directory (if one tries this now, the import command will delete all configs not present).

This alternatively could be an option to the `config-edit` command to take a source file (thus skipping the interactive editing part) and make it the active config for that object.

weitzman:
Yeah, this would be useful.

jhedstrom:
Let's track further discussion in the PR #1170.

PR #1170, opened 2015-02-11
Config import partial

jhedstrom:
This is a start on #1163. It shouldn't be too much more work to allow for directory imports of partial config sets.


jhedstrom:
Failure looks to be unrelated (honeypot needs updating since `_l()` was removed from 8.x).

weitzman:
Looking good so far. Thanks for the test.

weitzman:
I really like the --partial feature for config-import. I'm ambivalent on `config-edit --file` but if you think its valuable lets go with it. This passed tests so I think it can go in once @jhedstrom is happy with it.

jhedstrom:
I've left the `config-edit --file` option in place for now.

Issue #1171, opened 2015-02-11
example.bashrc doesn't work when drush is symlinked


jhedstrom:
Commit 55deaf2 seems to have broken the example when drush is symlinked on OSX.

jhedstrom:
Possible fix in #1171. Needed to use regex and double brackets.

greg-1-anderson:
Existing commit works on my OSX system with a simlink--but I like the way yours looks better anyway, so we should commit this if it works in multiple platforms.  I just need to remember to test this a couple of places.  I seem to remember that some systems don't support the "[[ ]]" syntax, but I don't quite remember where the problem with this is.

What is the failure you are seeing with the current code?

PR #1172, opened 2015-02-11
example.bashrc doesn't work with symlinks

jhedstrom:
Fix for #1171.

### DIFF
--- a/examples/example.bashrc
+++ b/examples/example.bashrc
@@ -122,7 +122,7 @@ if [ -h "$d" ] ; then
   # Change `readlink` to `readlink -f` if your drush is a symlink to a symlink. -f is unavailable on OSX's readlink.
   d2="$(readlink "$d")"
   # If 'readlink' comes back as a relative path, then we need to fix it up.
-  if [ "x$d2" == "x/" ] ; then
+  if [[ "$d2" =~ ^"/" ]] ; then
     d="$d2"
   else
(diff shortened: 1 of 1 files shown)


jhedstrom:
Confirmed this doesn't work on linux either, but this pr fixes it.

greg-1-anderson:
Existing commit works on my OSX system with a simlink--but I like the way yours looks better anyway, so we should commit this if it works in multiple platforms. I just need to remember to test this a couple of places. I seem to remember that some systems don't support the "[[ ]]" syntax, but I don't quite remember where the problem with this is.

What is the failure you are seeing with the current code?

greg-1-anderson:
I just confirmed that the current checked-in code works both on OSX 10.10, and also on Ubuntu 12.04.

I also tested your proposed change, and found it worked in both places as well.  I wish I could remember where [[ was a problem.

I notice that complete.sh is using [[, so perhaps we can use it here as well.  Ah, looks like it was in #1115.

greg-1-anderson:
Travis does not source example.bashrc, so I think we can go ahead and merge this PR.

jhedstrom:
I detailed the failures in comments on commit 55deaf2 (not the clearest). The current code doesn't properly detect an absolute path, so assumes it is relative, resulting in the concatenation of the symlink + the real path, which then doesn't exist, so drush.complete.sh isn't sourced.

PR #1177, opened 2015-02-12
Add a --cache-uri option to drush runserver

greg-1-anderson:
This defaults to off when running rs directly, but is set by drush quick-drupal.  This allows folks using drush quick-drupal to easily re-run drush runserver and drush user-login as necessary when working with thier temporary sites.

The following workflow is supported:

$ drush qd --core=drupal-8.0.x --cache --package-handler=git_drupalorg
[Do some testing, decide to erase the database, so ^C]
$ drush qd --use-existing
[Press ^C to use the terminal, then resume]
$ drush rs /
[Do some testing, log out; then, from another terminal w/ the same cwd]
$ drush uli admin/index

These operations take a bit more typing and fiddling before this patch.

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -938,6 +938,9 @@ function drush_core_quick_drupal() {
   }
   if (!drush_get_option('no-server', FALSE)) {
     if ($server = drush_get_option('server', '/')) {
+      // Tell reunserver to save our host and port in ROOT/drush/drushrc.php 
+      // file for future reference.
+      drush_set_option('cache-uri', TRUE);
       // Current CLI user is also the web server user, which is for development
       // only. Hence we can safely make the site directory writable. This makes
       // it easier to delete and edit settings.php.
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -559,13 +559,13 @@ function _core_site_status_table($project = '') {
         if (!empty($db_spec['unix_socket'])) {
           $status_table['db-socket'] = $db_spec['unix_socket'];
         }
-        else {
+        elseif (isset($db_spec['host'])) {
           $status_table['db-hostname'] = $db_spec['host'];
         }
-        $status_table['db-username'] = $db_spec['username'];
-        $status_table['db-password'] = $db_spec['password'];
-        $status_table['db-name'] = $db_spec['database'];
-        $status_table['db-port'] = $db_spec['port'];
+        $status_table['db-username'] = isset($db_spec['username']) ? $db_spec['username'] : NULL;
+        $status_table['db-password'] = isset($db_spec['password']) ? $db_spec['password'] : NULL;
+        $status_table['db-name'] = isset($db_spec['database']) ? $db_spec['database'] : NULL;
+        $status_table['db-port'] = isset($db_spec['port']) ? $db_spec['port'] : NULL;
         if ($phase > DRUSH_BOOTSTRAP_DRUPAL_DATABASE) {
           $status_table['db-status'] = dt('Connected');
           if ($phase > DRUSH_BOOTSTRAP_DRUPAL_FULL) {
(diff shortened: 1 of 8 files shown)


commands/core/core.drush.inc
```diff
@@ -940,6 +940,9 @@ function drush_core_quick_drupal() {
   }
   if (!drush_get_option('no-server', FALSE)) {
     if ($server = drush_get_option('server', '/')) {
+      // Tell reunserver to save our host and port in ROOT/drush/drushrc.php
```
weitzman: type reunserver

commands/runserver/runserver.drush.inc
```diff
@@ -134,6 +135,43 @@ function drush_core_runserver($uri = NULL) {
...
+<?php
+
+// Inserted by `drush runserver`.  This allows `drush user-login`
+// and similar commands to work seemlessly.  Remove if using 
+// Drupal multisite feature.
+\$options['uri'] = "$server";
+EOT;
+    if ($create_file) {
+      drush_mkdir($dir);
+      file_put_contents($target_file, $content);
```
weitzman: Is there a reason not to use Drush's cache API for this?

weitzman:
I dont use quick-drupal so I'm not too clear on the use case. Ping to @grugnog for review.

commands/runserver/runserver.drush.inc
greg-1-anderson: The purpose of setting $options['uri'] here is to literally set the --uri option, so that it is correct in whatever context it might be needed.  If we used the Drush cache API, then we'd need to also add special-checking to the user-login command that used the cache value instead of --uri.  Then we also have to make sure that --uri overrides the cache value iff it was user-specified, and doesn't if it is a default.

grugnog:
I think I see the use case, although I wonder if creating a file is perhaps overkill. In most cases, couldn't you just create an @rs alias then run the subsequent uli with that, or via site-set if you prefer? If you use different ports for multiple concurrent runservers, you could create multiple (@rs1, @rs2 etc) and even encode the default ports (e.g. :8001, :8002 etc) into each alias's option presets.

greg-1-anderson:
Yes, it is true that you can set up aliases to keep track of your rs sites; however, the big advantages of the rs command is that it is very lightweight in terms of the prerequisites a user needs to satisfy before typing 'drush rs'.  Setting up an alias is just one more thing you need to do (think: on a new system, or someone else's system, what is the minimum viable command to get a test site up?). Not that it's bad to set up an alias, or too hard, but I think it's still worth considering how to streamline the non-alias case.

In contrast, the execution of 'drush rs' is a fairly heavyweight operation; it already creates a bunch of files, like settings.php.  Since the site created cannot be used with a default --uri, to me it seems to make a lot of sense to just cache the value while you have it on-hand.  To this end, storing this information in a drushrc.php file is a standard way to keep this configuration.

If someone sets up an alias later, the drushrc.php file won't hurt anything, as the alias values will take precedence where set.

greg-1-anderson:
As an additional clarification of the use-case above, making the on-ramp to using drush qd in the patch contribution workflow is shortened with this patch, as users do not need to learn how to make aliases to do iterative development with the same qd site.

c.f. https://events.drupal.org/losangeles2015/sessions/improving-patch-contribution-workflow-drush-qd-and-drush-iq

weitzman:
Thought a bit more about this and I don't love Drush writing a drushrc file as part of running a regular command. Perhaps we build out those alias CRUD commands as a weak substitute for this PR.

greg-1-anderson:
I admit that there is a certain amount of oddity to writing a drushrc.php during the runserver command.  This facility is really useful for Drush qd, though, and the Drush qd command creates an entire scratch Drupal site.  What if we moved the drushrc.php file creation to the qd command, and leave rs out of it?  I admit there is some possibility of angst if someone takes the files output by Drush qd and then uses them to create a site served from a webserver, but we can commend the drushrc.php file contents, and drush_log its creation in qd.  The typical use-case for qd is to make a scratch site, and it sure would be nice if it worked iteratively.

greg-1-anderson:
I wanted to start over on this, with the drushrc.php creation strictly in the qd command, so I rebased with master, which of course screwed up this PR.  I'm going to close this, and make a new clean PR.

PR #1185, opened 2015-02-13
Switch TRUE to FILE_EXISTS_OVERWRITE for drush_copy_dir. This fixes #1184

### LINKED ISSUE #1184: Various places call drush_copy_dir with wrong argument (TRUE instead of FILE_EXISTS_OVERWRITE)
In https://www.drupal.org/node/1539076, drush_copy_dir's argument was changed, but various places call it with TRUE which bypasses the overwrite handling.

To see this in action, have two conflicting make files, one that defines a project with a version, the other that gets the project from git. Instead of deleting the overwritten pm downloaded, the project is merged. Our issue is that both definitions applied the same patch so patch was unable to be applied because it already was.

hefox:


jhedstrom:
Thanks!

PR #1186, opened 2015-02-14
Move to cmi overrides.

benjy:
Thanks to @chx for suggesting that we can simply use CMI overrides. Simplifies the code and allows us to support more than just overriding source/destination config.


chx:
You should probably merge in the original settings overrides

benjy:
Sorry for the delay on this, I fly out for DrupalSouth today but i'll take a look at this in the sprint on Saturday.

benjy:
OK, i took a look at this and based on Config::getOriginal() it would seem that the merging is already handled.

benjy:
OK, this should be good.

chx:
Yes, this looks very nice.

PR #1194, opened 2015-02-18
Allow drush make to run composer install, for better integration of exis...

greg-1-anderson:
This PR is a proof-of-concept. It has been lightly tested, but needs some work.

These are just some ideas.  We are in a situation where it is becoming really useful to use Composer wherever dependency management is needed (i.e., modules that need to require external php libraries from github / packagist); however, Drush make is firmly ingrained in the infrastructure of a lot of workflows, so it makes sense to provide a transition plan to help composer adoption.

The basic feature in this patch is that users can reference a composer file to use in their makefile via a make option override.  This feature is enabled by default, but may be turned off via --allow-overrides.  Those who are already using --allow-overrides (e.g. drupal.org) will find composer install turned off by default.

The benefit of just this much of the code is not huge; it saves folks from having to type `composer install` after `drush make`.  This is a minor service, but it ties in with other integrations.

We also allow `drush make composer.json` -- that is, if the user substitutes a composer.json file for their makefile, then make will just build it.  The advantage here is that it makes it easier to enable Composer use with existing tools that are already integrated with make.  For example, `drush qd --makefile=composer.json`.  We could also just enhance `drush qd`; the question is, how many other tools or services could we integrate with if we did it this way?

While we are here, we should also consider what the right way to manage other assets, such as Javascript libraries is.  Make handles this; it seems that Composer does not (packagist being for php code).  [Bower](http://bower.io/) is commonly used for this outside of Drupal, but does not presently seem to be commonly used with Drupal.  Perhaps the best interim solution would be make + composer?  This PR does not address js files.

Something else to consider is the current Drupal user who is managing sites with Drush make.  What happens when an existing module adds a dependency from packagist, and includes a new composer.json at the root of the project?  drush make could warn about this situation.  It might even be possible to move the module from the makefile to the project's composer.json; however, in instances like this, it's probably better to just move all of the modules to composer.json, and leave the makefile for managing just the non-module assets (js libraries, et. al.).  In any event, this PR does not attempt to handle this.

### DIFF
--- a/commands/make/make.drush.inc
+++ b/commands/make/make.drush.inc
@@ -79,6 +79,10 @@ function make_drush_command() {
         'description' => 'Restrict the make to this comma-separated list of libraries. To specify all libraries, pass *.',
         'example-value' => 'tinymce',
       ),
+      'composer' => array(
+        'description' => 'Run "composer install" on the specified composer file before processing makefile.',
+        'example-value' => 'composer.json',
+      ),
       'allow-override' => array(
         'description' => 'Restrict the make options to a comma-separated list. Defaults to unrestricted.',
       ),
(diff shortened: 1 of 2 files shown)


weitzman:
So this only runs a single composer.json, right? It does not recursively process the composer.json that might be included in a downloaded project.

greg-1-anderson:
I am guessing that we do not want to do this.  Folks can just use `composer install` instead of Drush make if they wish composer support.  Support for composer can be added directly to `drush qd`.  That is the only usecase in Drush core.

PR #1204, opened 2015-02-25
Support entity updates in drush updb

Berdir:
- Show them in the list
- Execute them first as a single batch (not very pretty, but exactly like core does it now)
- Also implemented a missing function needed for drush updbst, copy from 7.x + showing the change summary there as well.

### DIFF
--- a/commands/core/drupal/update.inc
+++ b/commands/core/drupal/update.inc
@@ -106,17 +106,27 @@ function update_main() {
 
   $start = array();
 
+  $change_summary = \Drupal::service('entity.definition_update_manager')->getChangeSummary();
+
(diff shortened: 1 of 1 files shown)


weitzman:
Change looks good. I dunno whats up with Travis. I restarted the build but it is not going anywhere.

Berdir:
Hm. 10 is failing because honeypot is broken it seems. No idea how 7 is related. 9 looks somewhat related, trying to figure out how to run this locally at the moment.

Berdir:
Ok, I've managed to run the tests now, that test is also failing on master, so I don't think any of the test fails are related to my changes.

PR #1213, opened 2015-02-27
Fix unescaped command line parameters.

david-garcia-garcia:


weitzman:
[See how mysql does](http://api.drush.org/api/drush/lib%21Drush%21Sql%21Sqlmysql.php/function/Sqlmysql%3A%3Acreds/master) it with params_to_option() method. That handles escaping.

david-garcia-garcia:
Fixed to use params_to_option().

Also fixed a bug where calling drush "the hardcore way":

"C:\Program Files (x86)\PHP\php5.6\php.exe" -c "D:\REPOSITORIOS_SABENTIS\drupal7" "D:\DRUPAL OFICIAL\drush\drush.php" --php="C:\Program Files (x86)\PHP\php5.6\php.exe"

and when the the paths have whitespaces internally triggered drush proc_opens where failing.

weitzman:
I think I fixed the backend.inc problem mentioned here. Not 100% sure. If not we'll grab the quoting proposed here. The sqlsrv improvements look good but they have conflicts. If anyone can reroll, I will merge in quickly.

PR #1216, opened 2015-02-28
Add 'DRUSH_XDEBUG' environment variable option, setting XDEBUG_CONFIG immediately before invoking drush.php

bendoh:
If I simply feed XDEBUG_CONFIG into the drush executable, the debugger will fire twice, and getting the "real" debugger instance (where drush.php actually executes) becomes a timing dance.

$ XDEBUG_CONFIG="IDEKEY=vim" drush ...

Will trigger the debugger once and close immediately (without breaking)

With this new change, introducing the DRUSH_XDEBUG option

$ DRUSH_XDEBUG="vim" drush ...

The debugger executes correctly in drush.php as the XDEBUG_CONFIG variable is exported
only immediately before the drush PHP executable is invoked.


weitzman:
I have been debugging for years without this. There are many tutorials on the web which don't suggest mention this. Are you launching debugging from your IDE? I usually put my IDE in listen mode and then run a drush command from my terminal as normal.

bendoh:
Yeah, so this could be specific to my environment, not sure.

My setup is:

MacOS 10.10.1
Vdebug 1.4.2
MacVim 73
PHP 5.6.3-dev + Xdebug 2.3.1-dev.

The same behavior does not appear in Codebug, but I haven't tried any other debuggers, as Vdebug is my mainstay. My guess is that somewhere in the drush bootstrap shell script, PHP is invoked (really don't know how, would need to pull it apart), and if I feed XDEBUG_CONFIG to the drush shell script, it will trigger the debugger and then immediately close, without ever breaking on a line.

I think other debugging front-ends may immediately enable after a quick connection hits them, and so the second connection breaks as it should. Vdebug shuts off entirely after a connection is made and completed, needing user intervention to listen again.

This may be a bug in Vdebug, but this is a very simple fix in drush. It won't export the XDEBUG_CONFIG until the very moment that flag is needed by PHP, in the main drush.php invocation.

The vdebug log shows this:

```
- [Info] {Sat 28 2015 13:27:15} Searching for win by name /Users/bendoh/vc/drush/includes/preflight.inc
- [Info] {Sat 28 2015 13:27:15} Win 1, name /Users/bendoh/vc/drush/includes/preflight.inc
- [Info] {Sat 28 2015 13:27:15} Returning window number 1
- [Info] {Sat 28 2015 13:27:15} Found connection from ('127.0.0.1', 61987)
- [Debug] {Sat 28 2015 13:27:15} Command: context_names -i 1
- [Debug] {Sat 28 2015 13:27:15} Response: <?xml version="1.0" encoding="iso-8859-1"?>
<response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="context_names" transaction_id="1"><context name="Locals" id="0"></context><context name="Superglobals" id="1"></context><context name="User defined constants" id="2"></context></response>
- [Debug] {Sat 28 2015 13:27:15} Available context names: {0: 'Locals', 1: 'Superglobals', 2: 'User defined constants'}
- [Debug] {Sat 28 2015 13:27:15} Command: step_into -i 2
- [Debug] {Sat 28 2015 13:27:15} Response: <?xml version="1.0" encoding="iso-8859-1"?>
<response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="http://xdebug.org/dbgp/xdebug" command="step_into" transaction_id="2" status="stopping" reason="ok"></response>
- [Info] {Sat 28 2015 13:27:15} Debugging session has ended
- [Info] {Sat 28 2015 13:27:15} Closing the connection
- [Debug] {Sat 28 2015 13:27:15} Closing the socket
```

weitzman:
Environment variables get inherited by subprocesses so I dont see why we would need to pass this along explicitly. Anyway, we can reopen this when others state that they need it. Until then, this is considered a local issue or too edge case.

PR #1220, opened 2015-03-03
Don't run cache clear after remote-add/remove-perm if no permission was added/removed

PLPeeters:
This prevents the cache clear from running after modifying permissions if no permission was modified, since it isn't needed in that case.

### DIFF
--- a/commands/core/role.drush.inc
+++ b/commands/core/role.drush.inc
@@ -161,13 +161,21 @@ function drush_role_add_perm($rid, $permissions = NULL) {
   else {
     $perms = _convert_csv_to_array($permissions);
   }
+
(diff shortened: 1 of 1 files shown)


commands/core/role.drush.inc
```diff
@@ -161,13 +161,21 @@ function drush_role_add_perm($rid, $permissions = NULL) {
   else {
     $perms = _convert_csv_to_array($permissions);
   }
+
+  $added_perm = false;
```
weitzman: booleans should be uppercase per coding standards (e.g. FALSE).

PLPeeters:
Whoops, my bad. Corrected.

PR #1222, opened 2015-03-04
6.x user edit

dustinmoris:
#### Added a new command called "user-edit". This command allows me to edit any user field for an existing user.

The fields can be specified as an option on the command. e.g.:
`drush user-edit MyUser1 --mail="new-email@example.org" --status=1 --pass="new-password"`

The power behind this command is that there is no limitation to the fields. I can even edit fields, which have been introduced by another module. e.g.:
`drush user-edit MyUser1 --force_password_change=0`

Drush remains independent and doesn't need to know which modules are installed. It simply passes all the fields to the user_save function and if it doesn't exist, then it gets ignored by Drupal anyway.

This command was much desired by our ops team, which is able to change a user's password policy settings from an automated script now.

Hope this can make its way to Drush. If you are willing to accept the pull request, then I will request a separate pull request for the master branch as well. Thank you!

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -273,6 +273,7 @@ function core_drush_command() {
         'description' => 'Password for the "db-su" account. Optional.',
         'example-value' => 'pass',
       ),
+      'db-drop-existing-tables' => 'Defaults to TRUE. Remove all tables in the target database before creating Drupal tables.',
       'account-name' => 'uid1 name. Defaults to admin',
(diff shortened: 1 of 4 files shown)


weitzman:
I think this is too edge case and error prone for core drush. 

I think there should be a set of entity/field commands in Contrib. This could be part of it, along with moving drush's existing field-* commands to it.

dustinmoris:
It is just a command to edit a user, what is edge case? I think I don't quite get your comment about the entity/field commands in Contrib either. Could you explain this a bit more?

Issue #691, opened 2014-06-26
drush cr broken after D8 bootstrap refactor


yched:
After http://www.drupal.org/node/2016629, drupal_rebuild() requires 2 arguments : 
- a ClassLoader
- a Request

Issue #961, opened 2014-11-06
Bootstrap level for updatedb-batch-process


helmo:
I don't understand why commit b45184a02b914a8a1055d37733d5f4fad2373d29 upped the bootstrap level for updatedb-batch-process to DRUSH_BOOTSTRAP_DRUPAL_LOGIN.

Especially while https://github.com/drush-ops/drush/blob/master/commands/core/drupal/update.inc#L98 mentions BOOTSTRAP_DRUPAL_FULL not _LOGIN.

It crashed an upgrade process from Drupal 6.33 to 7.33 just before starting the very first hook_update_n().

I'm documenting my findings on https://www.drupal.org/node/2370275#comment-9316769
#827 seems related ...

Issue #110, opened 2013-09-16
PEAR install: make drush look for php.ini in the script path as well as in the installation path


James-Oakley:
Please reference Drupal.org support issue: https://drupal.org/node/1671896

When Drush is installed through PEAR, the $SCRIPT_PATH variable in bin/drush is hard-coded to a directory such as /usr/local/lib/php/drush.

When $conf_dir is being formed, it only looks in $SELF_PATH. However the Drush documentation suggests that custom php.ini files should go in the lib/drush folder, not the bin folder. This is also in line with Linux directory-naming convention - custom ini files should not go in an application's bin.

I have written a patch that I apply whenever I install or update Drush using PEAR. This patch may not be the correct solution for the central Drush repository, but it's a starting point. Any changes to the way the PEAR installer modifies the Drush code need to be non-breaking for existing installations, who may well have their php.ini file in the lib directory.

damienmckenna:
I suspect this won't be fixed as the PEAR installer no longer works.

jonhattan:
yeah, PEAR no longer supported.

PR #1223, opened 2015-03-04
Attempt to fix D6 to D7 upgrades, see https://www.drupal.org/node/238266...

helmo:
After #691 was fixed we still had a problem running the Aegir test suite on drush master. 

On D.o: https://www.drupal.org/node/2382663

Summary: 
When upgrading a site from D6 to D7 we get the " Base table or view not found: ... 'blocked_ips'" error.

I've git bisected this to one commit in drush: b45184a02b914a8a1055d37733d5f4fad2373d29 

After much debugging I came to this change that lets out tests pass again.

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -336,7 +336,7 @@ function core_drush_command() {
       'batch-id' => 'The batch id that will be processed',
     ),
     'required-arguments' => TRUE,
-    'bootstrap' => DRUSH_BOOTSTRAP_DRUPAL_FULL, // https://github.com/drush-ops/drush/issues/961
+    'bootstrap' => DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION, // https://github.com/drush-ops/drush/issues/961
   );
   $items['core-global-options'] = array(
     'description' => 'All global options',
@@ -1137,9 +1137,6 @@ function drush_core_batch_process($id) {
  */
 function drush_core_updatedb_batch_process($id) {
   drush_include_engine('drupal', 'update');
-  require_once DRUSH_DRUPAL_CORE . '/includes/install.inc';
-  // require_once DRUSH_DRUPAL_CORE . '/includes/update.inc';
-  drupal_load_updates();
   _update_batch_command($id);
 }
 


weitzman:
b45184a got changed later in #961. This bootstrap level has been moving around a lot and just not sure what will break if we move it back to CONFIGURATION. Site to site updates are hard to unit test for obvious reasons. Not sure how to proceed here.

helmo:
Darn, my reference to #691 should have been to #961 (what a coincidence that they are both about bootstraplevels).

The aegir tests are pretty simple, see the three lines staring from: http://cgit.drupalcode.org/provision/tree/provision-tests/provision_tests.drush.inc#n81

From b45184a02b914a8a1055d37733d5f4fad2373d29 I can understand the changes to update.inc, but not adding drupal_load_update().  That function is already called in update.inc, update_6.inc and update_7.inc.

helmo:
@ergonlogic could comment on this?

ergonlogic:
So, if I follow, it looks like `drupal_is_denied()` is being called (and failing) prior to `system_update_7002()`, where the `blocked_ips` table is created. Any bootstrap level at or beyond _FULL would likely cause this. Chicken and egg, AFAICT. I think we'll need to bite the bullet and find a way to lower the req'd bootstrap level.

We should be able to require the include files directly instead of `module_load_install()`, if that's why we're using the higher bootstrap level. That appears to be what the docs for [module_load_include()](https://api.drupal.org/api/drupal/includes%21module.inc/function/module_load_include/7) (called by `module_load_install()`) seems to suggest.

helmo:
I'm actually removing the `require_once DRUSH_DRUPAL_CORE . '/includes/install.inc';
` line in this PR, as I see no need for it.

Leaving it in does not seem to harm my test, so I don't mind.

Leaving in `drupal_load_updates();` does harm my tests. And as I said in the second comment here "` drupal_load_update()` is already called in update.inc, update_6.inc and update_7.inc."

helmo:
I've done some initial work to add a testcase that does these upgrades within Drush. See #1250

helmo:
@weitzman please consider committing this.

helmo:
Yippee, build fixed!

08:29 -aegirjenkins:#aegir- Project P_Aegir_Puppet_Module_functional_test_Aegir3-drush-master-dev build #110: FIXED in 10 min: http://ci.aegirproject.org/job/P_Aegir_Puppet_Module_functional_test_Aegir3-drush-master-dev/110/

Thanks.

Issue #1218, opened 2015-02-28
Surprising command-line parsing for user-add-role and user-delete-role


kenahoo:
Observe the following behavior:
```
% drush user-remove-role "choir member" --uid 946
Removed the choir member role from uid 1
Removed the choir member role from uid 946 
```

Apparently the `--uid` argument isn't attached to the `946`, so it gets the value `1`, and the `946` argument is treated as a standalone arg.  To get the expected behavior the user has to do `--uid=946` instead.

It would be very nice to let the user omit the `=` in the arg specification, since loads of command line programs work that way (it's the way the [C getopt utility](http://man7.org/linux/man-pages/man3/getopt.3.html) works).

Furthermore, it would be great to add a `--test` or `-n` option so that when searching by name or email, the user could preview what user is matched by the query.

weitzman:
You can use a space with short options but not long. In your case, `-u 946` works fine, as does `--uid=946`. I think Drush is standard in this respect but I'd be happy to learn otherwise.

kenahoo:
I think there are really two separate issues:

1) `--uid` by itself returns `1`, which is really never what the user intends;
2) whether the `=` may be omitted for long options.

Regarding the second point, the 'getopt' docs that I linked to say "A long option may take a parameter, of the form --arg=param or --arg param."  There's no POSIX standard that I know of for command-line parsing of long arguments, but the 'getopt' library is as close to one as it gets, I think.

There's also a Python example showing the same idea here https://docs.python.org/2/library/getopt.html (where it says "Using long option names is equally easy:").

weitzman:
OK, I pushed a commit which marks --uid and similar options as having a required value. Today, that only affects whether we show the example value with braces or brackets in Help. Perhaps we should actually enforce this requirement in order to fix bugs like this. @greg-1-anderson  - you in favor of checking for required option values?

greg-1-anderson:
Fixed!

weitzman:
As a post-mortem, there isn't consistency "2) whether the = may be omitted for long options.". See http://stackoverflow.com/questions/10818443/short-long-options-with-option-argument-is-this-some-sort-of-convention

PR #1224, opened 2015-03-04
Fail with an error message if the user omits a required option value

greg-1-anderson:
#1218 was not the first instance of someone using "--option value" where "--option=value" is expected.  This PR adds sanity checking, and reports an error if an option that is expected to have a value is not provided one.


weitzman:
Looks great. Thanks.

Issue #1219, opened 2015-03-03
Proposal for commands to merge configuration changes between two sites.


greg-1-anderson:
After talking with @weitzman and @davidstrauss, and trying some experimental merges manually, I have come up with the following proposal for configuration management commands.

## Problem Statement
Sometimes, site developers make configuration changes on dev / staging that they wish to deploy to live; however, the site administrators may have, at the same time, made configuration changes of their own on the production server.  It then becomes necessary to merge the prod configuration into the dev configuration.

## Proposed Solution
`config-merge` and `config-push` commands.  No code exists yet; this is just a proposal.  The code seems easy enough to write, though.

### Configuration Merge

`$ drush @dev config-merge @live label`

The `label` defaults to `staging`, just as it does in the `drush cex` and `drush cim` commands.  The command will fail if @dev is not under git revision control.  If it is, then the following operations will be done:

1. A new branch `live-config` in the git repository for @dev is created to hold configuration from @live.  If this branch already exists, it will be re-used.
1. `drush @live cex label` is executed via drush_invoke_process().
1. Copy the exported configuration from @live to @dev, probably via rsync
1. Commit the new changes to the branch prepared for @live.  Exit with "nothing to do" if there are no changes to be committed.
1. Run `drush @dev cex label`
1. Commit this exported configuration. (QUESTION: Create a temp branch `dev-config`, or just use the current branch?)
1. `git checkout live-config && git rebase dev-config`
1. At this point, the `dev-config` branch may be deleted, if it is a temp branch.
1. If there are MERGE CONFLICTS: prompt the user and run 3-way diff tool.

At this point, there are a lot of different situations that can arise.  There might be merge conflicts, or there might not.  There might be changes from @live, or @dev, or none.  The developer will need to test before anything else is done.  In order to test, it will be necessary to run `drush @dev cim label`, but this cannot be done until the merge is successfully completed.  Maybe config-merge just advises the user to run this command, but does not attempt to run it.

Similarly, configuration deployment might be done in a number of different ways, depending on the developer's workflow.  The config-merge command therefore should not attempt to do any deployment.

It should be okay to run config-merge over and over again, with or without running `drush config-push` in between, and with or without changes to the @dev and @live sites, without causing any problems (e.g. accidental "removals" in the diff that should not be there) with the 3-way merge.

### Configuration Push

[...]

PR #1228, opened 2015-03-05
#1219: Drush config-merge command.

greg-1-anderson:
This is a new Drush command, `drush config-merge`.  This is just a prototype, but it's pretty solid.

Usage:

1. Make a Drupal 8 site; configure it, export the configuration, and commit the exported files to git.
1. `git tag deploy1.0`
1. Deploy from @dev to @live via your favorite deployment methodology.
1. Make more configuration changes on @dev 
1. Make some configuration changes on @live too
1. `drush config-merge @live --base=deploy1.0 --tool=kdiff3`
1. Wait for a sec; if there are merge conflicts, kdiff3 will launch.
1. Resolve all merge conflicts in kdiff3 and save.
1. Find your merged configuration changes committed to your local git repo, and imported on @dev 
1. Test.
1. If satisfied, repeat from '2.' with a new deployment tag.

So, what could possibly go wrong?

1. As mentioned in #1219, if you use the wrong --base, you can erase any committed config changes made on @dev after the last deploy.  Config changes on @live, and uncommitted config changes are safe.
1. If you abort the `kdiff` or `meld` merge tool without saving, then `git mergetool` will hang forever.  When you press control-C, you will exit back to the shell with a dirty branch.
1. If `drush config-merge` exits abruptly for any reason (c.f. '2', above), then you are on your own to clean up your dirty branch.  `git rebase --abort && git clean -d -f && git checkout ORIGINALBRANCHNAME` will usually do the trick.  If `drush config-merge` exits cleanly (including exiting via a rollback), then it will clean up for you.

Provides a hook for smart policy files that might (?) find --base for you.  Has a small number of options to allow users to fine-tune behavior.

- Still needs docs.
- Needs more testing, and tests.
- Seems to work pretty well under light testing.
- Way easier than merging by hand.

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -160,7 +160,7 @@ function config_drush_command() {
       'comment' => 'Commit comment for the merged configuration.',
       'no-commit' => 'Do not commit the fetched configuration; leave the modified files unstaged.',
       'tool' => 'Specific tool to use with `git mergetool`.',
-      'prepare-only' => "Don't run `git mergetool`; fetch all configuration changes from both sites, and merge them onto the tracking branch.  May result in unresolved merge conflicts.",
+      'fetch-only' => "Don't run `git mergetool`; fetch all configuration changes from both sites, and merge them onto the tracking branch.  May result in unresolved merge conflicts.",
     ),
     'examples' => array(
       'drush @dev config-merge @production' => 'Merge configuration changes from the production site with the configuration changes made on the development site.',
@@ -688,8 +688,8 @@ function drush_config_merge($alias = '', $config_label = 'staging') {
     }
 
     // Stop right here if the user specified --merge-only.
-    if (drush_get_option('prepare-only', FALSE)) {
-      drush_log(dt("Specified --prepare-only, so stopping here after the merge.  Use `git checkout !b` to return to your original branch.", array('!b' => $original_branch)), 'ok');
+    if (drush_get_option('fetch-only', FALSE)) {
+      drush_log(dt("Specified --fetch-only, so stopping here after the merge.  Use `git checkout !b` to return to your original branch.", array('!b' => $original_branch)), 'ok');
       return TRUE;
     }
 
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -482,25 +482,6 @@ function drush_config_edit($config_name = '') {
 }
 
 function drush_config_merge($alias = '', $config_label = 'staging') {
-  // Is the current site under git revision control?  If not, fail.
-  $result = drush_shell_exec('git rev-parse --abbrev-ref HEAD');
-  if (!$result) {
-    return drush_set_error('DRUSH_CONFIG_MERGE_NO_GIT', dt("The drush config-merge command requires that the current site be under git revision control."));
-  }
-  $output = drush_shell_exec_output();
-  $original_branch = $output[0];
-  drush_log(dt("Original branch is !branch", array('!branch' => $original_branch)), 'notice');
-  // Fail if there are any uncommitted changes on the current branch
-  // inside the configuration path.
-  $result = drush_shell_exec('git status --porcelain');
-  if (!$result) {
-    return drush_set_error('DRUSH_CONFIG_MERGE_FAILURE', dt("`git status` failed."));
-  }
-  $uncommitted_changes = drush_shell_exec_output();
-  if (!empty($uncommitted_changes)) {
-    return drush_set_error('DRUSH_CONFIG_MERGE_UNCOMMITTED_CHANGES', dt("Working set has uncommitted changes; please commit or discard them before merging.  `git stash` before config-merge, and `git stash pop` afterwards can be useful here."));
-  }
-
   // Allow the user to provide a specific tracking branch to do live work on.
   $tracking_branch = drush_get_option('tracking', FALSE);
   $tracking_branch_is_tmp = FALSE;
@@ -516,7 +497,6 @@ function drush_config_merge($alias = '', $config_label = 'staging') {
     'commit' => !drush_get_option('no-commit', FALSE),
     'tool' => drush_get_option('tool', ''),
     'config-label' => $config_label,
-    'original-branch' => $original_branch,
     'live-site' => $alias,
     'dev-site' => '@self',
     'live-config' => $tracking_branch,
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
So, I was also thinking I should update this to use a git library.

The standard one is [libgit2/php-git](https://github.com/libgit2/php-git).  It has the advantage of being able to work directly on the repository, without affecting the working copy.  I didn't use it, though, because it is written in "C" and cannot be installed via Composer.  We need to switch the active branch of the working copy anyway, to run `git mergetool`, so I didn't think it was worth the large complication to the Drush installation process that this library would add.

I think I might try [gitonomy/gitlib](http://gitonomy.com/doc/gitlib/master/); it is just a php wrapper around the git tool, but it can be [installed via composer](http://gitonomy.com/doc/gitlib/master/installation/#using-composer).

weitzman:
I find the shell code perfectly readable and functional. I don't see much
value in a wrapper library.

greg-1-anderson:
I did some testing today, and feel that with the commits above, this command is pretty usable now.

Tomorrow I am going to move the import operation before the commit, and run it interactively.  This will give folks a chance to say yes or no before the commit goes in; if someone gets --base wrong, then they will see the import deleting a bunch of config from the development side that should not be removed, and will have an opportunity to say "no", and have the merge get backed out.

greg-1-anderson:
The latest commit adds a --git option to drush config-merge, to allow the exported configuration to be transferred via either git or rsync, per the caller's preference.

commands/core/config.drush.inc
```diff
@@ -146,6 +156,31 @@ function config_drush_command() {
       'drush --bg config-edit image.style.large' => 'Return to shell prompt as soon as the editor window opens.',
     ),
   );
+  $items['config-merge'] = array(
+    'description' => 'Combine configuration data from this site with the configuration data of another site; if there are conflicts, open an interactive three-way merge tool comparing the changes with the base revision.',
```
weitzman: We should add an abbreviated description in hook_help. That gets used in the command listing, as opposed to this full description which would be used for single command help.

commands/core/config.drush.inc
```diff
@@ -146,6 +156,31 @@ function config_drush_command() {
...
+    'aliases' => array('cm'),
+    'required-arguments' => 1,
+    'arguments' => array(
+      'site' => 'Alias for the site containing the other configuration data to merge.',
+      'label' => "A config directory label (i.e. a key in \$config_directories array in settings.php). Defaults to 'staging'",
+    ),
+    'options' => array(
+      'base' => 'The commit hash or tag for the base of the three-way merge operation.  This should be the most recent commit that was deployed to the site specified in the first argument.',
+      'tracking' => 'A tracking branch to use when doing the configuration merge (advanced). Default is to use a temporary branch.',
+      'comment' => 'Commit comment for the merged configuration.',
```
weitzman: I think the word 'message' is more common than 'comment' when referring to commit text.

commands/core/config.drush.inc
```diff
@@ -316,7 +351,68 @@ function drush_config_export($destination = NULL) {
...
+      if (!$comment) {
+        $comment = "Default cex comment."; // TODO: do it better.
+      }
+      $result = drush_shell_cd_and_exec($destination_dir, 'git commit -m %s', $comment);
+      if (!$result) {
+        return drush_set_error('DRUSH_CONFIG_EXPORT_FAILURE', dt("`git commit` failed."));
+      }
+      if ($remote) {
+        // Remote might be FALSE, if --push was not specified, or
+        // it might be TRUE is --push was not given a value.
```
weitzman: s/is/if

commands/core/config.drush.inc
```diff
@@ -461,9 +557,438 @@ function drush_config_edit($config_name = '') {
...
+    'dev-site' => '@self',
+    'live-config' => $tracking_branch,
+    'dev-config' => 'drush-dev-config-temp',
+    'autodelete-live-config' => $tracking_branch_is_tmp,
+    'autodelete-dev-config' => TRUE,
+  );
+  $result = drush_command_invoke_all_ref('drush_config_merge_info_alter', $merge_info);
+  if (array_search(FALSE, $result, TRUE) !== FALSE) {
+    return FALSE;
+  }
```
weitzman: In Drush7, an exception that gets thrown during a command callback gets handled by drush_invoke(). So we don't need to handle it here. I think thats what these 4 lines are doing.

commands/core/config.drush.inc
```diff
@@ -461,9 +557,438 @@ function drush_config_edit($config_name = '') {
...
+
+  // Find the current configuration path
+  // TODO: this throws an exception if $config_label does not exist.
```
weitzman: Se comment above. Exception is OK.

commands/core/config.drush.inc
```diff
@@ -461,9 +557,438 @@ function drush_config_edit($config_name = '') {
...
+  $merge_info['configuration_path'] = $configuration_path;
+
+  // Is the selected configuration directory under git revision control?  If not, fail.
+  $result = drush_shell_cd_and_exec($configuration_path, 'git rev-parse --abbrev-ref HEAD');
+  if (!$result) {
+    return drush_set_error('DRUSH_CONFIG_MERGE_NO_GIT', dt("The drush config-merge command requires that the selected configuration directory !dir be under git revision control.", array('!dir' => $configuration_path)));
+  }
+  $output = drush_shell_exec_output();
+  $original_branch = $output[0];
+  drush_log(dt("Original branch is !branch", array('!branch' => $original_branch)), 'notice');
```
weitzman: Should be 'debug' level, IMO.

commands/core/config.drush.inc
```diff
@@ -461,9 +557,438 @@ function drush_config_edit($config_name = '') {
...
+  // file supplied one, give the user a firm warning.
+  if (!$merge_info['base'] && $merge_info['autodelete-live-config']) {
+    $result = drush_shell_cd_and_exec($configuration_path, 'git rev-parse HEAD');
+    if (!$result) {
+      return drush_set_error('DRUSH_CONFIG_MERGE_NO_BASE', dt("--base not specified, and could not determine sha-hash of HEAD."));
+    }
+    $output = drush_shell_exec_output();
+    $merge_info['base'] = $output[0];
+    $short_sha_hash = substr($merge_info['base'], 0, 6);
+    if (!drush_confirm(dt("No base commit specified.  It is STRONGLY recommended that you re-run this command with --base=TAG, where TAG is either a tag or an sha-hash of the content that was deployed to !live.  If you do not know what TAG was deployed, you risk losing some of your local configuration changes.\n\nWould you like to continue, POTENTIALLY ERASING YOUR LOCAL CONFIGURATION CHANGES?", array('!sha' => $short_sha_hash, '!live' => $merge_info['live-site'], '!dev' => $merge_info['dev-site'])))) {
```
weitzman: I think the ALL CAPS is a bit too much warning.

commands/core/config.drush.inc
```diff
@@ -461,9 +557,438 @@ function drush_config_edit($config_name = '') {
...
+
+  // Run config-export on the live site.
+  $values = drush_invoke_process($merge_info['live-site'], 'config-export', array($merge_info['config-label']), $export_options);
+  if ($values['error_status']) {
+    return drush_set_error('DRUSH_CONFIG_MERGE_CANNOT_EXPORT', dt("Could not export configuration for site !site", array('!site' => $merge_info['live-site'])));
+  }
+
+  // Copy the exported configuration from 'live-site', either via git pull or via rsync
+  $commit_needed = FALSE;
+  if($merge_info['git-transport']) {
```
weitzman: missing space

weitzman:
We need a topic to go along with config-merge. Note that we can't use PHP to get fancy in our topics anymore, and they are commonly read at http://docs.drush.org/en/master/

commands/core/config.drush.inc
greg-1-anderson: Currently, running `drush config-merge @qd.prod undefined`, where "undefined" in an undefined configuration label, then the result is:
```        
exception 'Exception' with message 'The configuration directory type [error]
<em class="placeholder">undefined</em> does not exist.' in ... [stack trace]
```

Should this be caught somewhere else, e.g. in the command handler?

commands/core/config.drush.inc
greg-1-anderson: drush_command_invoke_all_ref calls the specified hook on all commandfiles, and returns the result from each aggregated into a single array.  The array_search() pattern allows any hook to abort the current operation by returning FALSE.  Any other return value is ignored.

This pattern is also used in pm-download.

weitzman: OK we can proceed with this. I just wanted to communicate that a hook can throw an exception instead of returning false and a similar fate will occur.

commands/core/config.drush.inc
weitzman: I would think that it gets caught by https://github.com/drush-ops/drush/blob/7.0.0-alpha9/includes/command.inc#L362

weitzman:
1. Should be easy to additionally accept a site specification as 'site' argument. See how archive-dump uses drush_sitealias_resolve_sitespecs(). Lets document this ability in argument help.
1. The option is still called 'message' in command definition
1. Do we need a --format-patch option or can the command figure this out for us?
1. Would be good if we had example values options that take values. Mark option values as required if applicable.

greg-1-anderson:
I wrote a test for config-merge, but in order to get it to work, I had to rebase with master (because the base commit for this branch does not work with today's Drupal 8).

If I push this to this PR, that will basically destroy our ability to review it, because github doesn't handle rebasing PRs correctly.  So, I can either commit this as-is, and handle the remaining issues with followup PRs, or abandon this PR and make a new one.  Probably the later is the better option.

greg-1-anderson:
Continued at #1317

PR #1230, opened 2015-03-05
Workaround for removed drupal bootstrap constants.

Berdir:
I guess it needs more refactoring, but for now, I just switched to use the drush constants and cleaned it up, because nothing was calling it with anything else but configuration or full. Also nothing uses the return value.

Not sure what a proper fix would look like...

### DIFF
--- a/lib/Drush/Boot/bootstrap.inc
+++ b/lib/Drush/Boot/bootstrap.inc
@@ -526,7 +526,7 @@ function _drush_bootstrap_drupal_configuration() {
     drupal_bootstrap(DRUPAL_BOOTSTRAP_CONFIGURATION);
   }
   else {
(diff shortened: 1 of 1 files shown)


Berdir:
Ah, maybe the boot() call needs to move to full as well, lets see what travis has to say :)

jhedstrom:
This gets drush working again with the latest HEAD. Since these drush-specific constants already exist, this seems like a reasonable fix for the time being.

pfrenssen:
It's working well for me too.

geerlingguy:
Ha! I just ran into this error (pasting below so Google searches might help the few people getting the error) on drupal 8 head + drush head... then ran updates again and the error went away. Nice!

```
Use of undefined constant DRUPAL_BOOTSTRAP_CONFIGURATION - assumed 'DRUPAL_BOOTSTRAP_CONFIGURATION'
```

Issue #324, opened 2013-12-13
Difference with remote hosts when single or group alias is used


jurgenhaas:
The following two commands are creating different commands when calling drush_build_drush_command() because they pass different set of variables:

drush @group.alias status
drush @group status

The first one is OK, because it comes from drush_do_command_redispatch() which determines the command options by calling drush_redispatch_get_options() where the options "php" and "php-options" get unset. That makes a lot of sense because we never know what's required on a remote host.

The second one gets it wrong, because it comes from drush_preflight_command_dispatch() which determines the command options by calling drush_get_context('cli') and that doesn't unset the options "php" and "php-options".

FreekVR:
Ran across this issue today, though I'm surprised more people haven't. At any rate, it really only occurs when your own system has PHP installed on a different path than a remote group - in my case, using a remote alias for a multisite install on OSX didn't work until I made the change in drush.inc as suggested by jurgenhaas.

Created a new pull request for the current master branch.

greg-1-anderson:
Now this has been committed.  Sorry for the delay.

PR #363, opened 2014-01-08
Bug fix for #324

jurgenhaas:
Fix the multiple-site invoke to take care not to pass "php" and "php-options" in, so that they are not forwarded to remote systems.

PR #1233, opened 2015-03-06
Fix for #324, don't pass PHP, PHP-options for drush alias groups

FreekVR:
Fixed issues with remote drush groups with different PHP paths. See issue #324.


greg-1-anderson:
Duplicate of #363.  I'll interpret this as an RTBC.  :)

PR #1235, opened 2015-03-06
Improve the confirmation question of database manipulation before site-iinstall

Sweetchuck:
During the `site-install` there is a confirmation question like this:

> You are about to DROP all tables in your `'d7_localhost'` database. Do you want to continue? (y/n):

Which contains the name of the database that we are currently using, but without any relevant information such as host name and port number.
In that case when somebody using more than one database server this confirmation message is not obvious enough which database will be affected.

This pull request change the confirmation message.

> You are about to DROP all tables in your `'mysql://msandbox:****@127.0.0.1:3310/d7_localhost'` database. Do you want to continue? (y/n):


weitzman:
This is more verbose for many people at the expense of very few who would benefit. Furthermore, it would often have a local IP instead of an alias name or public IP of the web server.

PR #1238, opened 2015-03-08
Automatically detect a distribution installation profile when installing Drupal 8

tstoeckler:
Instead of always falling back to `standard` this allows lazy people to just type `drush si` and Drush will install
a distribution installation profile if there is one. This makes `drush si` consistent with the behavior of `install.php`

### DIFF
--- a/commands/core/drupal/site_install.inc
+++ b/commands/core/drupal/site_install.inc
@@ -4,11 +4,25 @@
  * Install Drupal 8+
  */
 function drush_core_site_install_version($profile, array $additional_form_options = array()) {
-  if (!isset($profile)) {
-    $profile = 'standard';
-  }
-
   require_once DRUSH_DRUPAL_CORE . '/includes/install.core.inc';
+  $class_loader = drush_drupal_load_autoloader(DRUPAL_ROOT);
+
+  if (!isset($profile)) {
+    // If there is an installation profile that acts as a distribution, use that
+    // one.
+    $install_state = array('interactive' => FALSE) + install_state_defaults();
+    try {
+      install_begin_request($class_loader, $install_state);
+      $profile = _install_select_profile($install_state);
+    }
+    catch (\Exception $e) {
+      // This is only a best effort to provide a better default, no harm done
+      // if it fails.
+    }
+    if (!isset($profile) || !$profile) {
+      $profile = 'standard';
+    }
+  }
 
   $sql = drush_sql_get_class();
   $db_spec = $sql->db_spec();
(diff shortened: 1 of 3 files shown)


commands/core/drupal/site_install.inc
```diff
@@ -4,12 +4,26 @@
...
+    $install_state = array('interactive' => FALSE) + install_state_defaults();
+    try {
+      install_begin_request($class_loader, $install_state);
+      $profile = _install_select_profile($install_state);
+    }
+    catch (\Exception $e) {
+      // This is only a best effort to provide a better default, no harm done
+      // if it fails.
+    }
+    if (!isset($profile) || !$profile) {
```
weitzman: Is this the same as `if (empty($profile))` ?

weitzman:
Looks good. Just had one simplification suggestion.

tstoeckler:
Heh, I was so proud of myself that I considered both conditions. But you are of course correct that the check can be collapsed.

I also checked and Drupal 7 allows a similar feature, there you can mark an installation profile as `exclusive`. And only 5 require_once's later it actually worked. :-) I verified again that the old and new code paths both work in Drupal 7 and Drupal 8

weitzman:
Thanks. Lets document the new behavior in the help text for the 'profile' argument.

tstoeckler:
Done

commands/core/site_install.drush.inc
```diff
@@ -4,7 +4,7 @@ function site_install_drush_command() {
   $items['site-install'] = array(
     'description' => 'Install Drupal along with modules/themes/configuration using the specified install profile.',
     'arguments' => array(
-      'profile' => 'the install profile you wish to run. defaults to \'default\' in D6, \'standard\' in D7+',
+      'profile' => 'the install profile you wish to run. defaults to \'default\' in D6, \'standard\' in D7+, unless an install profile is marked as exclusive (or as a distribution in D8+ terminology) in which case that is used',
```
weitzman: Sorry to be a stickler, but help uses sentence case. So, starts with a capital and ends with a period. Also, could we add code comment that documents how an install profile is marked as a distribution or 'exclusive'?

tstoeckler:
No reason to be sorry :-)

Will fix the sentence case. I had just tried to match the current way as closely as possible before...

Regarding the code comment: Sure, more docs is always good :-), but where specifically should this be documented? I.e. should we print out example code in help? Or do you mean in the actual site_install.inc's where the "magic" happens? Or somewhere else?

weitzman:
I think a code comment just above the argument that links off to some other place is best. If there is no place on the web where 'exclusive' is documented, then please add that to the right page on d.o. handbook or to an issue which adds docs to core.

tstoeckler:
I updated the relevant pages on Drupal.org with the information and added some inline docs.

PR #1239, opened 2015-03-09
Set env-vars even for non-subprocess executions.

valkum:
#955 adds support for env-vars for remote executions or local subprocess executions.
This patch adds support for non-subprocesses by setting the env for this execution via putenv in preflight phase.
This way you can store database cred in your drush and vhost file and add them to settings via getenv for example.


weitzman:
I'm not crazy about using alias files as an alternative to writing own scripts that set local env variables. However, this does provide consistency with aliases that have remote hosts. So, I'll let @greg-1-anderson decide on this.

valkum:
One point to discuss is my ```substr(drush_escapeshellarg($value, $os), 1, -1)```.
drush_escapeshellarg wraps the $value into '. Putenv adds all characters after = into the env. var. Even the '. Anyone got a better fix for that problem than substr the ' out?

greg-1-anderson:
I think the best style would be to factor a function out of drush_escapeshellarg that escapes the string without wrapping it in quotes.  Then, drush_escapeshellarg should call this function, and then wrap it in quotes.

In practice, it will be a little messier than this. Note that drush_escapeshellarg does not wrap anything in quotes if the string is simple, and it always uses ' on Linux, and always uses " on Windows, so you will need to actually factor out two functions from the different branches of drush_escapeshellarg, and then make a new API function to escape without quotes.

valkum:
I changed the drush_escapeshellarg to have one more parameter. The new parameter $wrap is default set to true. If set to fasle the underlaying function won't wrap the $arg with quotes.
It seemed to be the better way for me. An extra function which is nearly a copy seems to much. 

This change should not affect anything else.

includes/environment.inc
```diff
@@ -865,3 +865,19 @@ function drush_get_drupal_core_compatibility() {
...
+/**
+ * Set Env. Variables for given site-alias.
+ */
+function drush_set_environment_vars() {
+  if ($alias = drush_get_context('DRUSH_TARGET_SITE_ALIAS')) {
+    $site_record = drush_sitealias_get_record($alias);
+    $os = drush_os($site_record);
+    if (isset($site_record['#env-vars'])) {
+      foreach ($site_record['#env-vars'] as $var => $value) {
+        $env_var = drush_escapeshellarg($var, $os, false) . '=' . drush_escapeshellarg($value, $os, false);
```
greg-1-anderson: Code style: FALSE

greg-1-anderson:
I think I'd prefer to see the polarity of this flag reversed:  $raw = FALSE is the default, and call it with $raw = TRUE to remove the outer quotes.  I think that this, combined with a comment in the docblocks explaining when this can be used, is okay with me.

The comment should be clear that the caller cannot use $raw = TRUE and then add quotes later, as the caller won't know what sort of quotes to add, so $raw should only be requested in those rare (singular?) instances where quotes are not needed.

To match existing code conventions, use TRUE and FALSE.

greg-1-anderson:
Okay, this is pretty good, but I am going to burden you with one more request.  We just did a big cleanup of the Drush bootstrap, but this PR is still using `drush_get_context('DRUSH_TARGET_SITE_ALIAS')`.  Let's avoid that.

1. Change drush_sitealias_check_arg() so that it returns the full alias record for the alias it finds, or an empty array instead of FALSE if it finds nothing.

2. Change drush_preflight_site() so that it returns the result of drush_sitealias_check_arg().

3. Move the call to the function drush_set_environment_vars() to drush_preflight(), immediately after the call to drush_preflight_site().  Merge together the '#env-vars' elements from the alias records returned by (1) and (2), and set any variables found.

Usually, only one or the other of those calls will find an alias record, so merging the env-vars is probably the simplest way to handle the two sources.  Avoiding the access to the global DRUSH_TARGET_SITE_ALIAS will keep the bootstrap clean.

Thanks for your work on this.

greg-1-anderson:
It would be nice to have at least one unit test that showed that an '#env-vars' record actually sets a variable.  Should someone accidentally break this feature, a unit test would catch that, and then we'd be sure to fix it.

greg-1-anderson:
Reroll this now that #1267 is in.  $alias_record is now nicely available for you in drush_preflight(); you can call your function from there; just before drush_shell_alias_replace() might be a good place.

valkum:
I've got some problems with implementing the suggestions. Calling the drush_sitealias_get_record() function in drush_sitealias_check_arg() returns an empty record. the check_arg function is called a second time in preflight_site.
I suggest to return a record by the preflight_site function by using the global context.

greg-1-anderson:
@valkum you don't need to look up the site alias record yourself any more.  All of the data you need (the contents of the alias the user selected) will be in the array $alias_record in the function drush_preflight().  Lead check arg alone; call your function from drush_preflight somewhere around where it calls drush_shell_alias_replace(), and pass it $alias_record.  Look for $alias_record['#env-vars'] in your function.

valkum:
Ah sry. Didn't have the newest version of preflight.inc.
I checked twice against drush-ops/drush/master i didn't changed anything in preflight but the call for drush_set_environment_vars($alias_record); with the $alias_record in preflight. My call ist after the global_options call. 
The problem is the $alias_record varaible is TRUE. I don't know what is wrong here as `drush @test status ` returns the right paths to my test site.
I will debug some more but if you have an answer to that, that would be helpful.

greg-1-anderson:
There could be a bug in my code.  Push your latest work to this PR, and I'll take a look at it.

valkum:
_drush_sitealias_set_context_by_name(...) does not return the $site_alias_settings. It returns TRUE or FALSE in HEAD of master.

greg-1-anderson:
@valkum: you do not need to call this function any longer.  You'll find your alias record in the $alias_record variable in function drush_preflight().  Call your routine from drush_preflight.  Pass it $alias_record.  Check for '#env-vars'.

Sorry I haven't had time to pull your code -- very busy this weekend.

valkum:
The call in your part of the code. I resolved it by fixing your version of  _drush_sitealias_set_context_by_name

greg-1-anderson:
Perfect.  Thanks for working through this.

PR #1240, opened 2015-03-09
#64:  Use a per-user tmp directory with the 'use' command.

greg-1-anderson:
This PR changes the directory that the 'use' environment is stored in to be a per-user filename, so that multiple users on the same machine can use the 'use' command.


greg-1-anderson:
This just became a dup of #64.  Note that the only failure here is because I did not update the test for site-set, which #64 did.

PR #1242, opened 2015-03-10
Add new Boot classes for Drupal 6/7/8 and move phase callbacks there

weitzman:
Pushing this along a bit, as per a @todo for valid_root(). 

In a future PR I'd like to see the other bootstrap phases split into version specific implementations. I'm not sure yet how that should be architected. ideas welcome there.


greg-1-anderson:
Glad you did this!  Travis says that completion is broken for D8 with this PR, although I am not sure how.

weitzman:
Tough one to track down. I had to learn more about conditional breakpoints in PHP to get this far. Perhaps @greg-1-anderson can actually squash the problem. 

Our woes happen during drush_complete_rebuild(). This function immediately does `drush_bootstrap_max(DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION);` which bootstraps up to CONFIG without this PR, but fails to validate ROOT with this PR. Because we can't bootstrap even to ROOT level, our site specific commandfile does not load and testComplete() fails.

I think the reason that this fails to validate root is that Drush has chosen D6 as its bootstrap class, as that is the last candidate in the new array. The site is actually D8 and so valid_root() fails later. I'm unclear why D8 is not chosen initially, given that --root is provided by an alias. 

1. _drush_preflight_root() also has a suspicious comment which is not fully accurate after this PR:
    
    // TEMPORARY: if we have not found a bootstrap class by this point,
    // then take the last one and use it.  This should be Drupal.
1. I wonder if the completion bootstrap is not properly getting --root from the @dev alias. 

I'm available by phone or chat if this is unclear.

greg-1-anderson:
That's perfectly clear.  I'll take a look, I'm pretty confident I can figure this out and fix it easily.  We should be able to use valid_root() to find the right bootstrap class in time for things to work right.  I'll figure out why it's not doing that already.

greg-1-anderson:
So, the issue here is that complete has its own parallel bootstrap code.  As mentioned in the "TEMPORARY" comment you quoted above, the bootstrap class is latched as soon as it's found, so it's "stuck" by the time complete gets around to doing its special business of processing the alias in the first arg, if any.

I am really tempted to fix complete bootstrapping, so that it uses the regular bootstrap, at least up to the point that it wants it (i.e., don't actually bootstrap the site--although apparently there's already a bootstrap_max() in there, which seems like a mistake, we only want the commandfiles, not a bootstrap--but that's another issue). 

I'm worried about the impact of doing that, though, so I think that instead I am going to work on making the bootstrap class not latched until it actually finds a real Drupal root.  To do this, we'll need a "null" bootstrap object that we can use as a proxy (to avoid checking for $bootstrap == FALSE everywhere we want to use it to call a method).

PR coming right up.

greg-1-anderson:
Fixed complete, but broke other stuff. Will keep working on it; getting closer.

greg-1-anderson:
Drush core-quick-drupal isn't working, because it uses drush_invoke to call site-install.  There was no Drupal root when qd bootstrapped, but there is at drush_invoke time.  This used to work, but does not with the current bootstrap.  I think the easiest / best solution here it to use drush_invoke_process instead.  No time to do that today, but I think that might be the last issue.

greg-1-anderson:
The 'pm' tests broke on Travis with 16c3862, but they are working fine when I run the same tests on my local machine.  Transient failure, or subtle behavior change?

The quick-drupal tests are passing now.

greg-1-anderson:
The 'pm' tests broke because a module we depend on changed state.  I fixed the test in fe3fdff, and cherry-picked it to master.

drush.php
```diff
@@ -43,11 +43,16 @@ function drush_main() {
 
   // Find the selected site based on --root, --uri or cwd, and
   // return the bootstrap object that goes with it.
```
weitzman: we dont use a return object anymore

includes/bootstrap.inc
```diff
@@ -26,6 +26,12 @@ define('DRUSH_BOOTSTRAP_MAX', -2);
  */
 define('DRUSH_BOOTSTRAP_DRUSH', 0);
```
weitzman: In a followup, I'd like to get rid of this constant. We don't "bootstrap" Drush anymore. We preflight Drush, and we bootstrap Drupal. Lets work on making that distinction complete.

greg-1-anderson:
Okay, I will fix the comment and commit this, and work on fixing the bootstrap constants in a separate PR.

includes/preflight.inc
```diff
@@ -247,6 +232,7 @@ function drush_preflight_site() {
   drush_sitealias_create_self_alias();
 
   // If applicable swaps in shell alias value (or executes it).
+  // TODO: This should probably come later in bootstrap.
```
weitzman: "Later in the *preflight*".

lib/Drush/Boot/DrupalBoot.php
```diff
@@ -2,115 +2,471 @@
...
+      DRUSH_BOOTSTRAP_DRUSH                  => 'bootstrap_drush',
+      DRUSH_BOOTSTRAP_DRUPAL_ROOT            => 'bootstrap_drupal_root',
+      DRUSH_BOOTSTRAP_DRUPAL_SITE            => 'bootstrap_drupal_site',
+      DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION   => 'bootstrap_drupal_configuration',
+      DRUSH_BOOTSTRAP_DRUPAL_DATABASE        => 'bootstrap_drupal_database',
+      DRUSH_BOOTSTRAP_DRUPAL_FULL            => 'bootstrap_drupal_full',
+      DRUSH_BOOTSTRAP_DRUPAL_LOGIN           => 'bootstrap_drupal_login');
+  }
+
+  function bootstrap_init_phases() {
```
weitzman: How is this used, again? Perhaps add some Doxygen

lib/Drush/Boot/DrupalBoot.php
```diff
@@ -2,115 +2,471 @@
...
+    if (!$command) {
+      $command = $this->drush_command_belongs_to_disabled_module();
+    }
+    parent::report_command_error($command);
   }
 
   function command_defaults() {
     return array(
+      'drupal dependencies' => array(),
       'bootstrap' => DRUSH_BOOTSTRAP_DRUPAL_LOGIN,
```
weitzman: It occurs to me that the fact that we do LOGIN by default is a holdover from ancient spaghetti Drupal. We might be able to change that for D7/D8. Something to consider for the future.

lib/Drush/Boot/DrupalBoot.php
```diff
@@ -2,115 +2,471 @@
...
-            if ($return === TRUE) {
-              $return = '';
-            }
-
-            if (drush_get_context('DRUSH_DEBUG') && !drush_get_context('DRUSH_QUIET')) {
-              // @todo Create version independant wrapper around Drupal timers. Use it.
-              drush_print_timers();
-            }
-            break;
+  function commandfile_searchpaths($phase, $phase_max = FALSE) {
```
weitzman: This still has a lot of variation by major version. Would be worth splitting this up in the future, IMO.

lib/Drush/Boot/DrupalBoot8.php
```diff
@@ -0,0 +1,107 @@
...
+   *   drupal_bootstrap(DRUPAL_BOOTSTRAP_FULL);
+   * @endcode
+   *
+   * @param $phase
+   *   A constant telling which phase to bootstrap to. Possible values:
+   *   - DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION: Initializes configuration and
+   *     kernel.
+   *   - DRUSH_BOOTSTRAP_DRUPAL_FULL: Boots the kernel.
+   */
+  function drupal8_bootstrap($phase = NULL) {
```
weitzman: Lets refactor this into the phase methods above. Furthermore, $kernel should be a property in this class so that later code can call $kernel->terminate(). See #1253

includes/preflight.inc
greg-1-anderson: I made that note because the comment above said that drush_shell_alias_replace() might exec shell aliases, and execution should not happen this early in the preflight.  However, it turns out that drush_shell_alias_replace() does not exec here, so this is a good place to call it.  I updated the comment.

drush.php
greg-1-anderson: These different preflight phases were split up, because dirty legacy stuff was happening in drush_main() between them.  Now, all that stuff has been cleaned  up, so I rolled all of these into drush_preflight, and made it return the bootstrap object again.

lib/Drush/Boot/DrupalBoot.php
greg-1-anderson: There is still a lot of work to be done on commandfile handling in general, to figure out which commands are relevant to the given site.  This will take some thought and discussion.  We could split up the commandfile searching as it currently exists as you suggest as a first step.

lib/Drush/Boot/DrupalBoot8.php
greg-1-anderson: Yes, let's do this as a follow-up, as you suggest in #1252.

includes/bootstrap.inc
greg-1-anderson: Follow-up issue is #1254

lib/Drush/Boot/DrupalBoot.php
greg-1-anderson: Follow-on is #1255.

lib/Drush/Boot/DrupalBoot.php
greg-1-anderson: Follow-on is #1256

greg-1-anderson:
Squashed and merged!  Let's keep the bootclasses branch around for a while, in case we need to examine it later.

Issue #800, opened 2014-09-08
URI can't change during SITE bootstrap phase - Drush7


weitzman:
The fix is to re-set the $SERVER vars during SITE phase, as we do during Drush6. Probably should factor out the server vars stuff.

greg-1-anderson:
In Drush 7, drush_load_config('drupal') and drush_load_config('site') are called from drush_preflight_site(), prior to the SITE bootstrap phase.  Seems this was broken in alpha-4, but I am depending on it for [#1177], for example, so I think it's working fine now.

morvans:
AFAICT, it does not currently work in master/HEAD. I'll try to have a look but I'm not used to the drush codebase.

greg-1-anderson:
Post your exact use-case that is not working.  If you set uri in a site-alias, this might behave differently than if uri is unset, for example.

weitzman:
I think the case that used to work is to set uri option from a drushrc.php that is located in a drupal site folder.

morvans:
More precisely : located in the sites/default folder. But it doesn't work either when located in drupal root folder.

@greg-1-anderson, see #1231 for the details of my use case.

greg-1-anderson:
Ah, I see, I was actually depending on drush_load_config('drupal').  The issue we have at drush_preflight_site() time is, if --uri is not specified, then Drush does not know where the sites folder is.  We could easily fix this particular use case by providing a default value of 'sites/defaut' if --uri is not set.

If --uri is set, then we have sort of a problem.  The priority of the contexts are:

- cli
- alias
- site
- drupal

So, if you specify --uri on the commandline, or in an alias file, then the value you set there will have priority over any 'uri' set in the 'site' or 'drupal' context.  How should Drupal decide that the 'uri' in the site context is better than the 'uri' in the cli or alias context?

Does Drush 6 really handle this?  How does it decide which 'uri' is best?  I don't remember any code that worked around the config priority during bootstrap. _drush_bootstrap_selected_uri() just uses drush_get_option('uri') in Drush 6.

838a741 looks the same as Drush 6.  838a741 crashes on my site, so I can't test right now. I suppose I could spin up D7 with qd, but need to go afk right now.

greg-1-anderson:
If 'sites/default' is the only thing that ever was supported, and the only thing desired, then that's easy to support again, and I can make a PR.

greg-1-anderson:
[...]

PR #1243, opened 2015-03-10
#800: Allow sites/[site]/drushrc.php to change the value of 'uri'.

greg-1-anderson:
This patch insures that our selected site is set prior to the site bootstrap phase, so that we don't have to go back and fix up anything if uri changes in the sites drush config file.


weitzman:
Looks good, if tests are happy.

greg-1-anderson:
All green -- merged.  Thanks!

PR #1244, opened 2015-03-10
Drush autoload backported to 6.x

greg-1-anderson:
We should recommend that Drush extensions that have composer.json files should call drush_autoload in their hook_init.  This will allow us the flexibility to support extensions in the vendor directory, once we have an efficient mechanism for doing that, while still allowing folks to write extensions that can still go in the usual places.

Today, extension writers have to put their own logic for finding and including their autoload file, which is not optimal.  We already have drush_autoload() in Drush 7; it would be low-impact to have it in Drush 6 as well, which will help folks who want to be compatible with the recommended Drush 7 mechanism without becoming incompatible with Drush 6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,8 +330,10 @@ current working directory without using the -l option, but you need your
 base_url to be set correctly, you may force it by setting the uri in
 a drushrc.php file located in the same directory as your settings.php file.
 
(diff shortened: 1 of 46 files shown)


greg-1-anderson:
Wrong branch for the PR!  Argh.

Issue #1252, opened 2015-03-13
$kernel->terminate() is not called


dawehner:
For various reasons drush should call <code>$kernel->terminate</code>when drush is done with its processing.

One example is https://www.drupal.org/node/356399 which would require drush to rebuild the router.

dawehner:
I'm trying to solve that, but its a bit tricky, because the $kernel is currently just a local static variable.

Berdir:
I noticed that too, we should move that to a function or something.

+1 to this, because it will also allow to rely on terminate() for things like sending out an aggregated cache tag invalidation list instead of doing it immediately in an cache tag invalidation listener.

greg-1-anderson:
This should be easier now that #1242 has landed.

greg-1-anderson:
Fixed by #1259

PR #1253, opened 2015-03-13
$kernel->terminate() is not run()

dawehner:
Solves #1252


weitzman:
Thanks. Lets redo this after #1242.

greg-1-anderson:
Continued at #1259

PR #1257, opened 2015-03-13
Insure that we have reached the right bootstrap level in drush_dispatch()

greg-1-anderson:
This is in case a command with a low bootstrap requirement invokes another command with a higher bootstrap requirement.  The only example we have of this today is drush quick-drupal, and qd must still explicitly set the selected site and call bootstrap, because it wants to bootstrap to a site that didn't even exist.

I think it's good to put this in for clarity, but at the moment it is a redundant step.  Let me know what you think.


weitzman:
Looks like an improvement to me.

greg-1-anderson:
Merged; thanks.

Issue #1256, opened 2015-03-13
Refactor commandfile_searchpaths()


greg-1-anderson:
There is version-specific code here that should be moved into specific DrupalBoot classes.

Follow-on to #1242

PR #1258, opened 2015-03-14
#1256: Split out contrib-paths from commandfile_searchpaths()

greg-1-anderson:
Defined separate methods to return the locations where modules and themes can be found.

### DIFF
--- a/lib/Drush/Boot/DrupalBoot.php
+++ b/lib/Drush/Boot/DrupalBoot.php
@@ -74,6 +74,28 @@ abstract class DrupalBoot extends BaseBoot {
     );
   }
 
+  /**
+   * @return array of strings - paths to directories where contrib
+   * modules can be found
+   */
+  function contrib_modules_paths() {
+    return array(
+      conf_path() . '/modules',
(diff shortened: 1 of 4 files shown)


weitzman:
In general this is good. We may want to distinguish 'sitewide' paths from site specific paths. Even in a category, we should order by recommendedness. For example, in d8 /modules is recommended over sites/all/modules (only there for legacy).

greg-1-anderson:
Why would we want to distinguish sitewide paths from site-specific paths?  We are always bootstrapping just a single site, and modules in both locations become available at the same time in the bootstrap.  We could put this in just for the sake of organization / documentation, but I don't think it makes any practical difference.

According to the README.txt in the Drupal 8 modules directory, /sites/all/modules takes precedence over the /modules directory.  Of course, we cannot guarantee that the module that is actually enabled is the one with the highest precedence, so we could put these in a different order (e.g. /modules first) if you prefer.

weitzman:
The distinction matters in Unish test writing but you are right, it doesn't matter in Drush. Merged.

PR #1259, opened 2015-03-14
Insure that kernel->terminate is called at the end of commands.

greg-1-anderson:
Also refactored drupal8_bootstrap().

### DIFF
--- a/lib/Drush/Boot/DrupalBoot8.php
+++ b/lib/Drush/Boot/DrupalBoot8.php
@@ -7,6 +7,11 @@ use Drupal\Core\DrupalKernel;
 
 class DrupalBoot8 extends DrupalBoot {
 
+  /**
+   * @var \Drupal\Core\DrupalKernelInterface
+   */
+  protected $kernel;
+
   function valid_root($path) {
     if (!empty($path) && is_dir($path) && file_exists($path . '/index.php')) {
       // Additional check for the presence of core/composer.json to
@@ -43,7 +48,9 @@ class DrupalBoot8 extends DrupalBoot {
   }
 
   function bootstrap_drupal_configuration() {
-    $this->drupal8_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION);
+    $request = Request::createFromGlobals();
+    $classloader = drush_drupal_load_autoloader(DRUPAL_ROOT);
+    $this->kernel = DrupalKernel::createFromRequest($request, $classloader, 'prod');
 
     // Unset drupal error handler and restore drush's one.
     restore_error_handler();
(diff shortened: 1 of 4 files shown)


lib/Drush/Boot/DrupalBoot8.php
```diff
@@ -43,7 +48,9 @@ function bootstrap_drupal_database() {
   }
 
   function bootstrap_drupal_configuration() {
-    $this->drupal8_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION);
+    $request = Request::createFromGlobals();
+    $classloader = drush_drupal_load_autoloader(DRUPAL_ROOT);
```
dawehner: Honestly, this looks wrong, because this is the bootstrap for a regular request.

lib/Drush/Boot/DrupalBoot8.php
```diff
@@ -64,44 +73,13 @@ function bootstrap_drupal_full() {
...
-            $kernel->boot();
-            $kernel->prepareLegacyRequest($request);
-            break;
-        }
-      }
-      $boot_level = $phase;
+  public function terminate() {
+    parent::terminate();
+
+    if ($this->kernel) {
```
dawehner: I guess the if was too conservative?

dawehner:
Thank you for working on it!

lib/Drush/Boot/DrupalBoot8.php
greg-1-anderson: $this->kernel is not initialized until DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION, but it is possible that a DrupalBoot8 might be constructed and terminated without ever reaching this phase, so I think the if is necessary.

greg-1-anderson:
All of Drush's tests pass with this, but I didn't add any tests to see if the call to terminate is behaving as expected.

I am also wondering if these calls nest properly.  I think it should be okay; in any one process you should only ever have one DrupalBoot8, which means that only one 'kernel' will ever be created (drush won't ever run the bootstrap code for any given bootstrap level more than once), and every kernel that is created should be destroyed in terminate().  The only time you get nesting is when Drush calls drush_invoke_process(), but in that instance, the nested call to create & terminate another kernel happens in a separate process.  So, it all seems right to me, but I wanted to call it out in case there was something I wasn't thinking about.

lib/Drush/Boot/DrupalBoot8.php
greg-1-anderson: This is just the existing Drush bootstrap code for Drupal 8, refactored--unless I made a mistake.  :)   In any event, if there is anything that could be improved or made more efficient here, we'd be glad to make adjustments.

dawehner: Ah, good to know

dawehner:
This looks great for me!

RTBM!

PR #1260, opened 2015-03-15
Respect enabled/disabled state of themes

greg-1-anderson:
When searching for commandfiles, respect the enabled / disabled state of themes, just like modules.

Incomplete - still needs Drupal 8 support.

### DIFF
--- a/commands/core/drupal/environment.inc
+++ b/commands/core/drupal/environment.inc
@@ -121,6 +121,14 @@ function drush_module_list() {
 }
 
 /**
+ * TODO: Returns a list of enabled themes.
+ */
+function drush_theme_list() {
+  return array();
+}
+
+
+/**
  * Installs a given list of modules.
  *
  * @see \Drupal\Core\Extension\ModuleInstallerInterface::install()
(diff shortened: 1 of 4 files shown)


greg-1-anderson:
Do we want this for Drush 7, or defer to 8?  Should be in rc1 if we're going to take it.

weitzman:
We have a challenge with D8 - the active theme and admin theme are stored in Config and thats not yet available during DRUSH_BOOTSTRAP_DRUPAL_SITE. Perhaps we just skip searching in themes for these early bootstrap commands. Sound OK?

We can easily get the active themes during FULL level. See drush_theme_get_default() and drush_theme_get_admin().

weitzman:
Just committed a simpler fix for this. We still unconditionally search themes when doing a SITE bootstrap but when doing FULL, we only load from the default and admin themes. 

I imagine that some people were using commandfiles from disabled themes (especially parent themes). That wont work anymore. You have to include the theme with --include or copy the commandfile into one of our search locations like ~/.drush

B-Prod:
@weitzman copying the command file to another location could be a solution, but this causes issues with themes in profile (whatever they are enabled or not). It is possible to launch profile themes commands, but it is no more possible to make a "drush -l" because the command files are included twice.

I also tried with the following content in the sites/all/drush/drushrc.php file : 

    $options['include'] = array('profiles/myprofile/themes');

But this does not work either.

PR #1267, opened 2015-03-18
Improve site alias resolution in bootstrap

greg-1-anderson:
We will give a separate error for 'alias not found', and not simply say "@alias command" could not be found.

If the tests are okay here, there is one more change I am going to make in command.inc.

### DIFF
--- a/drush.php
+++ b/drush.php
@@ -71,7 +71,9 @@ function drush_main() {
       }
     }
   }
-  $bootstrap->terminate();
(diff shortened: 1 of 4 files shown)


greg-1-anderson:
Merged this.  Later, I might get rid of $target_alias_name in drush_parse_args(), as it seems mostly pointless.  Don't want to spend time on it now, though.

PR #1271, opened 2015-03-20
Make runserver log in by default

greg-1-anderson:
It sort of drives me nuts that runserver opens a browser, but does not log you in unless you specify --user=1.  I periodically think that this is a bug, and think rs is broken.

This PR changes the default to log in with a browser.  Doesn't seem to hurt anything.  Any reason we shouldn't make this the default?

### DIFF
--- a/commands/runserver/runserver.drush.inc
+++ b/commands/runserver/runserver.drush.inc
@@ -96,9 +96,10 @@ function drush_core_runserver($uri = NULL) {
   // We pass in the currently logged in user (if set via the --user option),
   // which will automatically log this user in the browser during the first
   // request.
-  if (drush_get_option('user', FALSE)) {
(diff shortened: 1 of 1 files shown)


weitzman:
Makes sense to me. Any comment, @grugnog ?

grugnog:
This sounds good - I think we need to update the docs to indicate the default.

greg-1-anderson:
Merged with this documentation:
```
 --user                                    If opening a web browser,            
                                           automatically log in as this user    
                                           (user ID or username). Default is to 
                                           log in as uid 1.
```

Let me know if you'd like this documented anywhere else.

greg-1-anderson:
Marking closed since merged; re-open with any further requests.

PR #1272, opened 2015-03-20
Move uri cache from runserver to qd

greg-1-anderson:
This is a continuation of #1177, with the creation of the drushrc.php file moved from the runserver command to the quick-drupal command.

It is very useful to record the port used when running qd, as it helps subsequent commands to be correctly targeted, should they need to use the 'uri' for any reason.

I would like to see this in the 7.0-rc1 release, if possible.

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -936,8 +936,13 @@ function drush_core_quick_drupal() {
      return drush_set_error('QUICK_DRUPAL_PROJECT_ENABLE_FAIL', 'Project enable failed.');
     }
   }
+  $server = drush_get_option('server', '/');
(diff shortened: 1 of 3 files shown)


weitzman:
Makes sense to me. Lets give @grugnog a couple days to chime in. If unavailable, @greg-1-anderson is welcome to proceed here.

greg-1-anderson:
Tomorrow is Drupal 8-beta8 and Drush 7.0-rc1.  I'll leave this open for a few more hours, and if there are no objections or requests raised before then, I'll go ahead and merge this in.

greg-1-anderson:
Merged, but just let me know if any adjustments are desired here.

PR #1273, opened 2015-03-20
Respect --lock=drupal in pm-updatecode.

greg-1-anderson:
For some reason, the --lock feature of pm-updatecode explicitly excludes core.  This PR enables --lock with core, which is very useful if you are in a managed hosting environment with a custom core (e.g. pressflow at Pantheon).


jonhattan:
Reason is that it never worked, and in 9b33b123 I fixed the warning it was giving.

greg-1-anderson:
Thanks!

Issue #975, opened 2014-11-09
Simplification ideas for site aliases and backend.inc


weitzman:
- [ ] Remove concurrency
- [ ] Remove shell alias replacements, e.g. `{{@target}}` - #1369.
- [ ] Remove site alias lists, e.g. `@foo,@bar`
- [x] Remove 'parent' inheritance in site aliases - #1279
- [x] Remove relative aliases @site/@peer. https://github.com/drush-ops/drush/commit/db8166def32e64b2c75b10dc884eddaa5f1b5c03 
- [x] Add [an alter hook](http://api.drush.org/api/drush/drush.api.php/function/hook_drush_sitealias_alter/master), so aliases can be adjusted in a [policy file](https://github.com/drush-ops/drush/blob/c129ea6ce51f6387d244f57debd04335b9c07dc1/examples/policy.drush.inc).
- [ ] Change site aliases to .yml. Dynamic changes still possible with [hook_sitealias_alter](http://api.drush.org/api/drush/drush.api.php/function/hook_drush_sitealias_alter/master). Segregate config, site selection (root+uri), and other parts of the array better. Replace array with one or more objects. - #1609 
- [ ] Refactor remote sites to be a first class object that gets passed around. Distinct from drush_get_option()

greg-1-anderson:
I am thinking we should make these changes in master, after we have spun off Drush 7, rather than try to make these changes at this time.

ergonlogic:
In Aegir and a number of related projects, we use much of backend.inc extensively. I'm all for simplifying it as much as possible.

In core, concurrency only appears to be used by make. We could move the relevant logic over there in a simplified manner, presumably. Are there any other projects that we know of making extensive use of concurrency (or the other proposed removals, for that matter)?

More OOP, YML and support for remote sites will benefit just about all the projects I'm working on.

greg-1-anderson:
@juampy72 recently used concurrency in one of his projects.

juampy72:
This command uses drush_invoke_concurrent() when GNU-parallel is not
available: http://drupal.org/project/syncdb.

http://drupal.org/project/concurrent_queue uses it too.

greg-1-anderson:
I added a PR #1279 to deprecate the 'parent' element and add an alter hook.  This is a first step toward supporting YAML site alias files.

ccharlton:
"Remove shell alias replacements" -- you mean the custom tokens like `%my_custom_thing`?

[...]

PR #1279, opened 2015-03-24
Add a hook_drush_sitealias_alter, with documentation.

greg-1-anderson:
Deprecate 'parent' element in site aliases, for eventual removal (probably in Drush 8).

### DIFF
--- a/examples/example.aliases.drushrc.php
+++ b/examples/example.aliases.drushrc.php
@@ -201,11 +201,7 @@
  *   standard port, alternative identity file, or alternative
  *   authentication method, ssh-options can contain a string of extra
  *   options that are used with the ssh command, eg "-p 100"
- * - 'parent': The name of a parent alias (e.g. '@server') to use as a basis
- *   for this alias.  Any value of the parent will appear in the child
- *   unless overridden by an item with the same name in the child.
- *   Multiple inheritance is possible; name multiple parents in the
- *   'parent' item separated by commas (e.g. '@server,@devsite').
+ * - 'parent': Deprecated.  See "altering aliases", below.
  * - 'db-url': The Drupal 6 database connection string from settings.php.
  *   For remote databases accessed via an ssh tunnel, set the port
  *   number to the tunneled port as it is accessed on the local machine.
@@ -295,6 +291,30 @@
  * ),
  * @endcode
  *
+ * Altering aliases:
+ *
+ * Alias records are written in php, so you may use php code to alter
+ * alias records if you wish.  For example:
+ *
+ * @code
+ * $common_live = array(
+ *   'remote-host' => 'myserver.isp.com',
+ *   'remote-user' => 'www-admin',
+ * );
+ *
+ * $aliases['live'] = array(
+ *   'uri' => 'mysite.com',
+ *   'root' => '/path.to/root',
+ * ) + $common_live;
+ * @endcode
+ *
+ * If you wish, you might want to put $common_live in a separate file,
+ * and include it at the top of each alias file that uses it.
+ *
+ * You may also use a policy file to alter aliases in code as they are
+ * loaded by Drush.  See policy_drush_sitealias_alter in
+ * `drush topic docs-policy` for details.
+ *
  * Some examples appear below.  Remove the leading hash signs to enable.
  */
 
@@ -354,11 +374,6 @@
 #    'os' => 'Linux',
 #  );
 #$aliases['live'] = array(
-#    'parent' => '@server,@dev',
 #    'uri' => 'mydrupalsite.com',
-#     'target-command-specific' => array (
-#       'sql-sync' => array (
-#         'skip-tables-list' => 'comments',
-#       ),
-#     ),
-#  );
+#    'root' => $aliases['dev']['root'],
+#  ) + $aliases['server'];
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
'parent' might be a seldom-used feature.  If we think folks aren't using it much, we could just flat-out remove it in Drush 7 and be done with it.

nerdcorenet:
I am using the 'parent' elemtn quite extensively in drush aliases. It allows us to define a server environment and have all sites residing on that server inherit things like the remote-host and root values.

esolitos:
Does the 'parent' complicate the code a lot?
Because in our case it's a "must" feature of drush, using multiple remote servers we're using it quite extensively.

greg-1-anderson:
To follow up on my tweet:

If you are using the 'parent' feature, read the patch and see if the policy file solution works well for your use-case.

The thing we are trying to change here is translating alias files to yaml rather than loading them in code.

davereid:
Acquia uses `@parent` in all of their Drush alias files for Cloud sites. Might want to get someone from their infrastructure team's input?

greg-1-anderson:
Probably deprecate this in Drush7, remove in Drush8, I'm guessing, then.

We'll also have to consider if the submitted patch is too noisy for folks.  :)

greg-1-anderson:
I mis-spoke above:  there is no real requirement to remove 'parent' if we switch to yaml alias files.  We could keep 'parent', and commit just the alter hook.  But I know that moshe has wanted to get rid of 'parent' for a while (see #975)

weitzman:
I really like this PR. It moves the complexity of parents to a policy. It is valuable, independant of the transition to YAML site alias files. 

The core issue is that greg is the sole person who works on backend and site aliases. Much of the code others could maintain but some is gnarly and rarely used; I think 'parent' is in that category. Drush project is too crucial to ignore this SPOF problem.

greg-1-anderson:
Shall we commit it as-is, then, with 'parent' deprecated and noisy, but still functional?

weitzman:
I think we can turn down the log level a notch and then merge.

examples/example.aliases.drushrc.php
```diff
@@ -201,11 +201,7 @@
  *   standard port, alternative identity file, or alternative
  *   authentication method, ssh-options can contain a string of extra
  *   options that are used with the ssh command, eg "-p 100"
- * - 'parent': The name of a parent alias (e.g. '@server') to use as a basis
- *   for this alias.  Any value of the parent will appear in the child
- *   unless overridden by an item with the same name in the child.
- *   Multiple inheritance is possible; name multiple parents in the
- *   'parent' item separated by commas (e.g. '@server,@devsite').
+ * - 'parent': Deprecated.  See "altering aliases", below.
```
kenorb: *Depreciated

includes/sitealias.inc
```diff
@@ -775,7 +775,9 @@ function _drush_sitealias_add_inherited_values(&$aliases) {
 }
 
 function _drush_sitealias_add_inherited_values_to_record(&$alias_value) {
+  drush_command_invoke_all_ref('drush_sitealias_alter', $alias_value);
   if (isset($alias_value['parent'])) {
+    drush_log(dt("Using deprecated 'parent' element '!parent' in '!name'.", array('!parent' => $alias_value['parent'], '!name' => $alias_value['#name'])), 'warning');
```
kenorb: *depreciated

examples/policy.drush.inc
```diff
@@ -29,6 +29,34 @@ function drush_policy_sql_sync_validate($source = NULL, $destination = NULL) {
...
 /**
+ * Implements hook_drush_sitealias_alter
+ *
+ * Alter alias record data in code.
+ */
+function policy_drush_sitealias_alter(&$alias_record) {
+  // A duplicate of the old implementation of the 'parent' element.
+  // Keep this if you want to keep using 'parent', but do not want
+  // to be nagged (or worse, break when it is removed).
+  if (isset($alias_record['parent'])) {
```
kenorb: It's usually advised to use array_key_exists() for arrays instead.

examples/policy.drush.inc
```diff
@@ -29,6 +29,34 @@ function drush_policy_sql_sync_validate($source = NULL, $destination = NULL) {
...
+    // Fetch and merge in each parent
+    foreach (explode(',', $alias_record['parent']) as $parent) {
+      $parent_record = drush_sitealias_get_record($parent);
+      unset($parent_record['#name']);
+      unset($parent_record['#file']);
+      unset($parent_record['#hidden']);
+      $array_based_keys = array_merge(drush_get_special_keys(), array('path-aliases'));
+      foreach ($array_based_keys as $array_based_key) {
+        if (isset($alias_record[$array_based_key]) && isset($parent_record[$array_based_key])) {
```
kenorb: No big deal, but it's usually advised to use array_key_exists() for arrays instead.

greg-1-anderson:
Deprecated: http://en.wiktionary.org/wiki/deprecated

greg-1-anderson:
I turned the log level down to 'debug' and merged.

End effect:  'parent' element is now deprecated in Drush 7, although it still works just fine, and the deprecation warnings are only visible in "debug" mode.

PR #1281, opened 2015-03-24
Fixes #1241: Make Drush better at finding site aliases in the sites folder

### LINKED ISSUE #1241: Site aliases not working
Site aliases stopped working sometime between 7.0.0-alpha3 and 7.0.0-alpha4. Please see my example output at https://gist.github.com/mikeyp/641f78ef0060cbdb2d74.

I can try to bisect if that would help figure out what caused this regression.

greg-1-anderson:
Drush had trouble finding aliases in the 'sites' folder early in the bootstrap.

This is a simple bugfix, that brings the behavior of `drush @foo command` in line with the documentation, and `drush sa @foo`, for aliases defined in ROOT/sites/default/aliases.drushrc.php.

I'll merge this if the tests come up clean.

### DIFF
--- a/includes/sitealias.inc
+++ b/includes/sitealias.inc
@@ -408,6 +408,14 @@ function drush_sitealias_alias_path($alias_path_context = NULL) {
   if (!empty($drupal_root)) {
     $site_paths[] = $drupal_root . '/drush';
     $site_paths[] = $drupal_root . '/sites/all/drush';
+    $uri = drush_get_context('DRUSH_SELECTED_URI');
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This seemed like the easiest way to fix this, but I think it is actually better to just remove this location from consideration, even though that involves more changes.

weitzman:
I'd be OK with the PR as is.

greg-1-anderson:
Okay, merging to restore consistency.  We can remove sites/default as a valid location in Drush 8.

PR #1282, opened 2015-03-24
Fixes #1265:  Define an exit status code for user aborts.

### LINKED ISSUE #1265: drush_user_abort() does not set an error exit code
If a drush command, for example `sql-drop`, is aborted by a user drush will return a success exit code. This makes scripting interactive solutions very difficult, as stderr must be captured and parsed.

The specific use case I am running in to is that I am scripting some drush commands as part of a build process. There are some specific commands that I would like the user to give confirmation to before proceeding (hence no yes-to-all). Drush currently provides no easy way (that I can find) for me to know a user abort vs regular execution path completed without scraping text.

I can understand the logic behind having a user abort treated as a success (the script did not fail). I think an `--abort-is-failure` argument would be very useful. If there is already some hidden command that provides this behavior feel free to let me know.

greg-1-anderson:
Here's a proposed patch that defines an exit code (75) that Drush returns if the user cancels.  Also defines a mechanism for Drush to set other exit codes, if desired.  (Currently, only "0" or "1" is ever returned.)

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -675,15 +675,15 @@ function drush_core_config($filter = NULL) {
   $exec = drush_get_editor();
   if (count($all) == 1) {
     $filepath = current($all);
-    return drush_shell_exec_interactive($exec, $filepath, $filepath);
   }
   else {
     $choice = drush_choice($all, 'Enter a number to choose which file to edit.', '!key');
-    if ($choice !== FALSE) {
-      $filepath = $all[$choice];
-      drush_shell_exec_interactive($exec, $filepath, $filepath);
+    if (!$choice) {
+      return drush_user_abort();
     }
+    $filepath = $all[$choice];
   }
+  return drush_shell_exec_interactive($exec, $filepath, $filepath);
 }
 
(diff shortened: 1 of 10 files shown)


greg-1-anderson:
These tests fail due to a change in Drupal 8; they will all pass again when merged with master.

Let me know if you want to try this in Drush 7; it shouldn't hurt anything, although nonstandard.

weitzman:
Well it would hurt any scripts that were expecting a success code after an abort. Do you think thats so rare or wrong that it is ignorable?

greg-1-anderson:
I can't think of a situation where a script would want to continue after the user said abort.  If there was such a script, it could trap the error.  Of course, we break these scripts, but we do not guarentee that we won't across major version upgrades.  I think that the benefits outweigh the costs, but if you think that scripts that want to continue after an abort is more than an edge case, we can skip this.

weitzman:
Makes sense. Feel free to merge this, when you are sure the test suite passes.

greg-1-anderson:
Retested and merged.  Thanks.

PR #1283, opened 2015-03-25
Fixes Drush with D8 user signatures change

greg-1-anderson:
This is related to the commit of https://www.drupal.org/node/1548204, which removed user signatures.


greg-1-anderson:
And there was much rejoicing.

Issue #1295, opened 2015-03-27
Projects using a major release of '0' fail to download


ergonlogic:
If a project uses a release like `0.4-rc1`, it will fail to offer any releases in `pm-download`.

ergonlogic:
It looks like it's in Project::getRecommendedOrSupportedRelease(). Pull request forthcoming.

ergonlogic:
Follow up in the PR

PR #1296, opened 2015-03-27
Allow major versions of '0'.

ergonlogic:
ref: https://github.com/drush-ops/drush/issues/1295


ergonlogic:
Test suite passed. I'm going to go ahead and merge, since this is pretty trivial.

I dug into the history a bit, but I couldn't find a commit to this code that explicitly disallows major versions of '0'. I suppose it's possible that this was a policy decision. If so, feel free to revert.

jonhattan:
There wasn't any intention to disallow 0. It is a collateral effect of using empty().

PR #1298, opened 2015-03-28
String::decodeEntities() moved to Html::decodeEntities() (PHP7)

Berdir:


greg-1-anderson:
Confirmed -- both work today, but String::decodeEntities is deprecated, and will be removed.

PR #1303, opened 2015-03-30
Use the right boostrap object

### LINKED ISSUE #1297: Enabling a module does not make its routes available
See https://www.drupal.org/node/2451665 - with this change, enabling a module through drush that defines a route does not make the route available, you need to do a drush cr afterwards. Would be nice if drush did the router rebuild.

dawehner:
Fixes #1297

### DIFF
--- a/drush.php
+++ b/drush.php
@@ -71,7 +71,7 @@ function drush_main() {
       }
     }
   }
-  if ($bootstrap) {
+  if ($bootstrap = drush_get_context('DRUSH_BOOTSTRAP_OBJECT')) {
     $bootstrap->terminate();
   }
   drush_postflight();


weitzman:
The problem that this fixes is that $bootstrap->kernel is NULL when we call terminate. So @dawehner proposes to just fetch the populated Boot object. Leaving this for @greg-1-anderson to review. Looks fine to me.

greg-1-anderson:
The original intention of the code here was that drush_preflight() should always return the right bootstrap object, which should always be the same thing that is in the global context DRUSH_BOOTSTRAP_OBJECT.  Unfortunately, later, to maintain compatibility / flexibility, the bootstrap code later changed such that the bootstrap object can sometimes be selected during bootstrap, instead of preflight.  This allows us to have a command declare itself DRUSH_BOOTSTRAP_DRUSH, then select a drupal root, and then call drush_bootstrap_max().

I'd like to get rid of references to globals such as DRUSH_BOOTSTRAP_OBJECT wherever possible, but I don't know that we're ready to stop supporting commands that might set the root and then bootstrap some more.  If it is only quick-drupal that needs to do this, maybe we can do this for Drush 7, but I think this might be a Drush 8 change.

I'm just going to merge in the PR as submitted, because I think that supporting the desired behavior for Drupal 8 is more important than getting rid of the global -- and the existing code does need to be as shown in this PR to be correct, without more rework to other parts of the bootstrap.

greg-1-anderson:
Two more things of note here:

a) drush_get_bootstrap_object() would be the better function to call here.

b) it seems to me that it should be really rare for $bootstrap to be different than the latched value in the global.

This will need some further investigation later.

Issue #1307, opened 2015-04-01
Defaults are not working


claudiu-cristea:
I'm using the latest from master. Here's my make file that fails because of `defaults:`:

```yaml
core: 7.x
api: 2

defaults:
  projects:
    download:
      type: file

projects:
  drupal:
    download:
      url: ../.resource/make/vendor/drupal/drupal-7.35.tar.gz
```
If I move the `type: file` under `drupal[download]` works well.

I don't know if it's a bug or the docs are not clear for me https://github.com/drush-ops/drush/blob/master/docs/make.md#defaults.

claudiu-cristea:
Pull request #1308.

claudiu-cristea:
#1308 was merged and is fixing the bug.

PR #1308, opened 2015-04-01
Make defaults apply in deep

claudiu-cristea:
Right now the defaults are not applied in deep. If a project has a level 1 key, the will override the default without checking in deep. See #1307.


claudiu-cristea:
I don't understand the test failures:

```
Unable to determine mime type for unish_dev.sql. 
```
I haven't touched that file.

jhedstrom:
Test failures are unrelated.

Thanks for this!

PR #1309, opened 2015-04-01
Add MSSQL support for sql_drush_sql_sync_sanitize.

hswong3i:


weitzman:
Would be good to eventually move sanitize bits to database specific sql classes. And maybe version specific classes as well.

PR #1313, opened 2015-04-02
Shorthand for download[url] in make file

claudiu-cristea:
Implement a shorthand for the case we have only the `url` key under project/library `download`.

```yaml
projects:
  mymodule:
    download: http://example.com/modules/mymodule.git
libraries:
  thirdpartylib:
    download: http://example.com/libs/thirdpartylib.min.js
```

is equivalent to:

```yaml
projects:
  mymodule:
    download:
      url: http://example.com/modules/mymodule.git
libraries:
  thirdpartylib:
    download:
      url: http://example.com/libs/thirdpartylib.min.js
```


jhedstrom:
Thanks!

Issue #1314, opened 2015-04-02
drush core-requirements does not output descriptions in render arrays


pfrenssen:
In [issue #2296929: Clean up system_install() SafeMarkup::set() use](https://www.drupal.org/node/2296929#comment-9764547) it was discovered by [herved](https://www.drupal.org/u/herved) that `drush rq` does not output descriptions that are rendered with render arrays correctly. Instead it shows `Array`.

This is because the `concatenate-columns` option is used to format the table. This treats the input data as strings and simply concatenates them.

The table formatter actually expects all input data to be raw strings. To make it possible to use arrays and objects in the input data we could provide a callback function that preprocesses the data of particular columns before trying to format it. This would solve it in a generic way.

AdamPS:
Thanks @weitzman I think this is now fixed and ready to close.

PR #1315, opened 2015-04-02
Issue #1314: Fix drush core-requirements does not output descriptions in render arrays

pfrenssen:
Fix for issue #1314.

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -200,6 +200,7 @@ function core_drush_command() {
       'fields-default' => array('title', 'severity', 'description'),
       'column-widths' => array('severity' => 8),
       'concatenate-columns' => array('description' => array('value', 'description')),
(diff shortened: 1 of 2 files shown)


weitzman:
Looks sane. Would be good if @greg-1-anderson looked at ouputformat code.

greg-1-anderson:
Actually, I was kind of ambivalent about this, and forgot to come back and comment.

The problem with this implementation is that, if there's a render array in the output, then the output will contain complete junk (perhaps not from Drupal's perspective, but as far as any Drush client is concerned) that is only readable when selecting some table-based output format that uses drush_rows_of_key_value_to_array_table().

I think it would be better if core-requirements found and fixed the render arrays in its result before returning the data for output.

greg-1-anderson:
Closing for lack of activity.

PR #1317, opened 2015-04-02
Config merge 2

greg-1-anderson:
Continuation of #1228

### DIFF
--- a/tests/Unish/UnishTestCase.php
+++ b/tests/Unish/UnishTestCase.php
@@ -52,7 +52,8 @@ abstract class UnishTestCase extends \PHPUnit_Framework_TestCase {
    * Runs after all tests in a class are run. Remove sandbox directory.
    */
   public static function tearDownAfterClass() {
-    if (file_exists(UNISH_SANDBOX) && empty(getenv('UNISH_DIRTY'))) {
+    $dirty = getenv('UNISH_DIRTY');
+    if (file_exists(UNISH_SANDBOX) && empty($dirty)) {
       unish_file_delete_recursive(UNISH_SANDBOX, TRUE);
     }
     self::$sites = array();
--- a/tests/Unish/UnishTestCase.php
+++ b/tests/Unish/UnishTestCase.php
@@ -52,7 +52,7 @@ abstract class UnishTestCase extends \PHPUnit_Framework_TestCase {
    * Runs after all tests in a class are run. Remove sandbox directory.
    */
   public static function tearDownAfterClass() {
-    if (file_exists(UNISH_SANDBOX)) {
+    if (file_exists(UNISH_SANDBOX) && empty(getenv('UNISH_DIRTY'))) {
       unish_file_delete_recursive(UNISH_SANDBOX, TRUE);
     }
     self::$sites = array();
@@ -202,7 +202,11 @@ abstract class UnishTestCase extends \PHPUnit_Framework_TestCase {
     return $pass;
   }
 
-  function webroot() {
+  function webroot($env = "dev") {
+    $sites = self::getSites();
+    if (isset($sites[$env]['root'])) {
+      return $sites[$env]['root'];
+    }
     return UNISH_SANDBOX . '/web';
   }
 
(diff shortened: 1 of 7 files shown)


tests/configMergeTest.php
```diff
@@ -0,0 +1,72 @@
...
+* @file
+*  Test config merge, to copy configuration from one site to another.
+*/
+
+namespace Unish;
+
+/**
+ *  @group slow
+ *  @group commands
+ *  @group sql
```
weitzman: c/p bug

tests/configMergeTest.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+/**
+* @file
+*  Test config merge, to copy configuration from one site to another.
```
weitzman: Not really a copy. It merges.

tests/configMergeTest.php
```diff
@@ -0,0 +1,72 @@
...
+ *  @group sql
+ */
+class configMergeTest extends CommandUnishTestCase {
+
+  /**
+   * Covers the following responsibilities.
+   *   - A user created on the source site is copied to the destination site.
+   *   - The email address of the copied user is sanitized on the destination site.
+   *
+   * General handling of site aliases will be in sitealiasTest.php.
```
weitzman: c/p bug

tests/configMergeTest.php
```diff
@@ -0,0 +1,72 @@
...
+    // Run config-merge to copy the configuration change from 'stage' to the 'dev' site
+    $this->drush('config-merge', array('stage'), $dev_options);
+
+    // Verify that the configuration change we made on 'stage' now exists on 'dev'
+    $this->drush('config-get', array('system.site', 'name'), $dev_options);
+    $this->assertEquals("'system.site:name': config_test", $this->getOutput(), 'Config set, merged and fetched.');
+  }
+
+  protected function createGitRepository($dir) {
+    exec("cd $dir && git init && git add . && git commit -m 'Initial commit.'", $output, $return);
```
weitzman: can we use $this->execute()?

greg-1-anderson:
Added a 'topic' command called 'docs-cmi'.  'docs-configuration' was already taken.  I know cmi == "configuration management initiative", which is just "configuration management" once the "initiative" is done, but it seems folks are still calling it cmi.  If this is going to be confusing in the long term, we can rename this.

The one open recommendation from the previous PR is:

> Should be easy to additionally accept a site specification as 'site' 
> argument. See how archive-dump uses drush_sitealias_resolve_sitespecs(). 
> Lets document this ability in argument help.

Other suggestions welcome; I'm going to try to resolve all of the last issues shortly so that we can get this merged.

weitzman:
Ideally we test more options and code paths beyond the simplest case.

greg-1-anderson:
Oh, yeah, I forgot to mention that.  The current unish framework only sets up multisites, and only the simplest case works on multisites.  I started enhancing the setup code so that it would set up multiple single sites, so we can config-merge between them.  I'll finish this up and add more tests before we are done.

weitzman:
An alternative is to setup one site and then clone it elsewhere via archive-dump+archive-restore.

weitzman:
One legit test fail - https://travis-ci.org/drush-ops/drush/jobs/57108640#L191

greg-1-anderson:
Oh, yeah, I have to configure git for travis.  Thanks.

I fixed the mime-type thing, but I don't think it's worth it to rebase again to get the fix. :p

greg-1-anderson:
There are now unit tests for all of the primary modes of operation, and all are passing.

Just a little more cleanup to do, and this will be ready to go.  If anyone has any good ideas at how config-export could generate a good and meaningful commit comment, let me know.  config-import gives a nice list of changed fields, but this isn't available at export time on the remote system.

greg-1-anderson:
I was under the mis-impression that the handy diff done by config-import could only be done against the active configuration in the database, but it turns out that Drupal 8 CMI is way cooler than that.  You can also compare the other way, or you can even compare one set of exported files against another set.  Because of that, I was able to generate awesome commit messages in config-export and config-merge, that list the summary of changes made in the commit.

greg-1-anderson:
We're all green now!

I forgot to mention earlier, drush_sitealias_resolve_sitespecs() is not appropriate for use in config-merge, because it processes lists of aliases, and config-merge can merge with at most one other Drupal site at a time.  I did fix up the code so that site specifications can be used here, though.

Unless there are more comments or requests, I think this is about ready to go in.

tests/Unish/UnishTestCase.php
```diff
@@ -52,7 +52,8 @@ public static function setUpFreshSandBox() {
    * Runs after all tests in a class are run. Remove sandbox directory.
    */
   public static function tearDownAfterClass() {
-    if (file_exists(UNISH_SANDBOX)) {
+    $dirty = getenv('UNISH_DIRTY');
```
weitzman: Document that env variable in tests/README.md

tests/Unish/UnishTestCase.php
```diff
@@ -276,12 +281,58 @@ function setUpDrupal($num_sites = 1, $install = FALSE, $version_string = UNISH_D
     return self::$sites;
   }
 
-  function fetchInstallDrupal($env = 'dev', $install = FALSE, $version_string = UNISH_DRUPAL_MAJOR_VERSION, $profile = NULL) {
-    $options = array();
-    $root = $this->webroot();
-    $site = "$root/sites/$env";
+  function setUpSingleSiteDrupal($env = 'dev', $install = FALSE, $version_string = UNISH_DRUPAL_MAJOR_VERSION, $profile = NULL) {
```
weitzman: This looks like a lot of duplicated code. Did you consider my proposal to use archive-dump/restore to get a 2nd site? Would also save on awkward sql-sync and uuid fiddling during test.

tests/configMergeTest.php
```diff
@@ -0,0 +1,270 @@
...
+*  Test config-merge, that merges configuration changes from one site to another.
+*/
+
+namespace Unish;
+
+/**
+ *  @group slow
+ *  @group commands
+ */
+class configMergeTest extends CommandUnishTestCase {
```
weitzman: Do we need to test for format-patch?

tests/Unish/UnishTestCase.php
greg-1-anderson: I like the way that alias creation, db-url generation, and site placement is encapsulated in the UnishTestCase class.  I'd have to do that in each test if we just used archive restore.  There's only one test that uses single site right now, but single site is A Thing, so I am guessing that we might use it in some other tests at some point.  I'll see if I can refactor the code in UnishTestCase, or otherwise make it easier to duplicate single sites in the tests.

Oh, the other thing about archive-restore is that we'd then have one single site "dev" with a sites folder named "dev", and another single site named "stage" with a sites folder named "dev".  Again, considering the use-case of using this in more tests, I thought that might get confusing, so the current code uses "default" for the sites folder when making single sites.

tests/configMergeTest.php
greg-1-anderson: Yeah, it's already tested.  See "Part 1" (git), "Part 2" (format-patch) and "Part 3" (rsync) in the tests.  I deliberately made non-isolated tests for the different modes, because it's supposed to be a feature of config-merge that you can keep running one merge after another, so if one part breaks the next one, then that's a bug.  (Except format-patch and rsync do break git, if you do one of those first -- that's a limitation, not a bug, so we test git first.)

greg-1-anderson:
The last two commits take care of everything discussed in this PR.  I just need to remove the --format-patch feature, and then I will put this in.

greg-1-anderson:
Merged.

PR #1319, opened 2015-04-03
Use explicit 'env' for remote command execution

### LINKED ISSUE #1305: Drush with remote alias with a prefix COLUMNS=### breaks remote execution
I'm using the latest dev release of drush 7 and when I'm calling ```drush @pweb.default status -v``` there is the following output:

```
Loaded alias @pweb.default from file /etc/drush/pweb.aliases.drushrc.php
Begin redispatch via drush_invoke_process(). 
Calling proc_open(ssh -o PasswordAuthentication=no pweb 'COLUMNS=163 drush  --verbose --uri=www.paragon-es.de --root=/var/www/live/htdocs  core-status 2>&1' 2>&1);
```

The command being executed is ```COLUMNS=163 drush ...```which doesn't seem to make sense, does it?

jurgenhaas:
This is a PR to resolve #1305

### DIFF
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -479,6 +479,7 @@ function drush_build_drush_command($drush_path = NULL, $php = NULL, $os = NULL, 
 
   // Add environmental variables, if present
   if (!empty($environment_variables)) {
+    $prefix .= ' env';
     foreach ($environment_variables as $key=>$value) {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Looks good to me.  'env' should be in everyone's PATH.  It's often in '/bin', which should always be available; on both Mac and Ubuntu Linux, it is in '/usr/bin', which is the same location as other required executables such as 'tput' and 'which'.

The one test failure is unrelated to this PR.

PR #1328, opened 2015-04-08
Add a drush-directory command

RobLoach:
Add a `drush-directory` command to retrieve the [realpath()](http://php.net/manual/en/function.realpath.php) of the currently executing version of Drush.

---

Going to use this for sourcing Drush completion without knowing Drush's path:

```
source $(drush drush-dir)/drush.complete.sh
```


weitzman:
How about using `drush eval "echo DRUSH_BASE_PATH"`

RobLoach:
Thanks Moshe! Good call.

PR #1331, opened 2015-04-09
List modules not enabled due to other failing.

clemens-tolboom:
Having a list of modules to enable drush only informs about the failing module but is not telling the others are not enabled either.

### DIFF
--- a/commands/pm/pm.drush.inc
+++ b/commands/pm/pm.drush.inc
@@ -1059,12 +1059,9 @@ function drush_pm_enable_validate() {
       // Check to see if the module can be installed/enabled (hook_requirements).
       // See @system_modules_submit
       if (!drupal_check_module($module)) {
+        unset($modules[$key]);
         drush_set_error('DRUSH_PM_ENABLE_MODULE_UNMEET_REQUIREMENTS', dt('Module !module doesn\'t meet the requirements to be enabled.', array('!module' => $module)));
         _drush_log_drupal_messages();
-        if (count($modules) > 1) {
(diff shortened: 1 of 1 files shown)


commands/pm/pm.drush.inc
```diff
@@ -1059,9 +1059,11 @@ function drush_pm_enable_validate() {
       // Check to see if the module can be installed/enabled (hook_requirements).
       // See @system_modules_submit
       if (!drupal_check_module($module)) {
-        unset($modules[$key]);
```
clemens-tolboom: As we are not continu enabling other modules no need to unset.

commands/pm/pm.drush.inc
```diff
@@ -1059,9 +1059,11 @@ function drush_pm_enable_validate() {
...
         drush_set_error('DRUSH_PM_ENABLE_MODULE_UNMEET_REQUIREMENTS', dt('Module !module doesn\'t meet the requirements to be enabled.', array('!module' => $module)));
         _drush_log_drupal_messages();
+        if (count($modules) > 1) {
+          drush_set_error('DRUSH_PM_ENABLE_MODULE_UNMEET_REQUIREMENTS', dt('None of these !modules is enabled', array('!modules' => join(', ', array_keys($modules)))));
+        }
         return FALSE;
```
clemens-tolboom: Should we return or continu enabling the other modules. This line is from

> 68684e83 (Greg Anderson       2011-01-05 06:28:44 +0000 1067)         return FALSE;

Can we continue enabling the others?

clemens-tolboom: Keeping `unset($modules[$key]);` intact and removing the `return FALSE` thus testing the other flow trying to enable the other modules fails. Not sure why.

greg-1-anderson:
Uncertain about the status of this code. `pm-*` is not getting much attention. Closing.

PR #1332, opened 2015-04-09
Require /autoload.php instead of /core/vendor/autoload.php

webflo:


webflo:
Follow-up https://www.drupal.org/node/2458045 and https://www.drupal.org/node/2406681

includes/drupal.inc
```diff
@@ -15,7 +15,7 @@ function drush_drupal_load_autoloader($drupal_root, $new_autoloader = NULL) {
   }
   if (!$autoloader) {
     // Should this just be 'require'?  See drush_preflight_prepare()
```
weitzman: Lets remove this comment.

greg-1-anderson:
Agree that comment should go.  There is also another change needed at the top of preflight.inc.

I am looking into the test failure.  When I run `drush cache-get schema` on a Drupal 8 site that I installed from the commandline, it works just fine -- but the unit test is failing for me as well.  I'm not sure what's wrong, but it is happening on the master branch as well, without this change, so I suggest we go ahead and merge it.  Give me a couple more minutes to fix the call in preflight.inc.

greg-1-anderson:
I was wrong about preflight -- it's fine the way it is.  I'm going to remove the commit and merge.

greg-1-anderson:
Merged, thanks.

PR #1337, opened 2015-04-15
Call Drupal Console commands natively from Drush

greg-1-anderson:
In this PR, we build Drush command records for all Console commands, so that they can be displayed in Drush's help output, and executed. 

- [x] Require Console in Drush's composer.json, so it is always available.
- [x] Run Console commands via Drush: `drush @site x-generate-module`.
- [x] Bootstrap Drush-style, using Console to create kernel and later bootstrap it.
- [ ] Decide if we want to support recommending instead of requiring Console
- [ ] Decide if we want commands to be generate:module or generate-module
- [ ] Decide on a strategy to resolve conflicts between similarly-named or similarly-featured commands
- [ ] Make Console return structured output to Drush, so that Drush-styled output formatters can be used.
- [ ] Make a 1.0 stable release of Console, and bump the major version every time a breaking change is made.

This PR, while still early, is an exciting step forward, and meets many of the goals discussed in #88.  It allows folks to write symfony console component commands, and call them from Drush utilizing Drush features (e.g. site aliases, shell aliases, command-specific options, etc.).  In the future, the construction of the Symfony application object, currently done in an adapter file (consoleadapter.drush.inc) could instead become part of the core Drush bootstrap, and used natively by Drush commands.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,8 @@
     "d11wtq/boris": "~1.0",
     "symfony/yaml": "~2.2",
     "symfony/var-dumper": "2.6.3",
-    "pear/console_table": "~1.2.0"
+    "pear/console_table": "~1.2.0",
+    "drupal/console": "dev-master"
   },
   "require-dev": {
     "phpunit/phpunit": ">=3.5",
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "symfony/yaml": "~2.2",
     "symfony/var-dumper": "2.6.3",
     "pear/console_table": "~1.2.0",
-    "drupal/console": "dev-master"
+    "drupal/console": "~0"
   },
   "require-dev": {
     "phpunit/phpunit": ">=3.5",
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "symfony/yaml": "~2.2",
     "symfony/var-dumper": "2.6.3",
     "pear/console_table": "~1.2.0",
-    "drupal/console": "~0"
+    "drupal/console": "dev-master"
   },
   "require-dev": {
     "phpunit/phpunit": ">=3.5",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "be3f9714724eb03579970f09fb725596",
+    "hash": "71173c7386a8df209ed4f5197aae2b84",
     "packages": [
         {
             "name": "composer/installers",
@@ -139,16 +139,16 @@
         },
         {
             "name": "drupal/console",
-            "version": "0.7.9",
+            "version": "dev-master",
             "source": {
                 "type": "git",
                 "url": "https://github.com/hechoendrupal/DrupalAppConsole.git",
-                "reference": "5dceda77e2538efa494d36d21b5437ed6f87bf56"
+                "reference": "3eb00b790ec687b8bcecd13ff18ccd48b5860ff3"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/hechoendrupal/DrupalAppConsole/zipball/5dceda77e2538efa494d36d21b5437ed6f87bf56",
-                "reference": "5dceda77e2538efa494d36d21b5437ed6f87bf56",
+                "url": "https://api.github.com/repos/hechoendrupal/DrupalAppConsole/zipball/3eb00b790ec687b8bcecd13ff18ccd48b5860ff3",
+                "reference": "3eb00b790ec687b8bcecd13ff18ccd48b5860ff3",
                 "shasum": ""
             },
             "require": {
@@ -162,6 +162,7 @@
                 "symfony/dom-crawler": "2.6.*",
                 "symfony/event-dispatcher": "2.6.*",
                 "symfony/finder": "2.6.*",
+                "symfony/http-foundation": "2.6.*",
                 "symfony/translation": "2.6.*",
                 "symfony/yaml": "2.6.*",
                 "twig/twig": "1.16.*"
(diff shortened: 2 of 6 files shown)


greg-1-anderson:
Drat.  I deliberately did not include composer.lock in the original PR, to avoid a huge number of irrelevant lines in the diff, but doing that caused Travis to have an existential crisis.  So, I've added composer.lock to make the tests run -- sorry for the size of the diff.

greg-1-anderson:
I didn't need to worry about including composer.lock in the PR; github is smart enough to hide this file in the diff by default.

There are a couple of spurrious failures above, related to cloning the cck_signup project.  Everything else is green; onward and upward.

jhedstrom:
I like that this is the `1337` PR :+1:

composer.json
```diff
@@ -30,7 +30,8 @@
     "d11wtq/boris": "~1.0",
     "symfony/yaml": "~2.2",
     "symfony/var-dumper": "2.6.3",
-    "pear/console_table": "~1.2.0"
+    "pear/console_table": "~1.2.0",
+    "drupal/console": "~0"
```
jibran: May I ask why is this required? I think it should be optional or recommended.

greg-1-anderson: If Drupal Console is required, then it gives us a mechanism to use its code to start migrating Drush commands to the Symfony console.  At the moment, though, this PR is just a proof-of-concept.  At first, there won't be any Drush commands that require Drupal Console, so we could certainly start off with this as a recommended package.

weitzman:
Myself and Jesus from Drupal Console plan to discuss this PR and general collaboration approach during upcoming Drupalcon (mid May). I'll post here after that discussion.

webchick:
Not sure exactly where best to cross-link this, but trying this one (nice issue ID btw :D).

Note that there's an RTBC issue in the Drupal core queue at https://www.drupal.org/node/2493807 to add Symfony Console to core. This is soft-blocking the critical issue about testing 8.x -> 8.x upgrade paths at https://www.drupal.org/node/2447573 which is why it's being considered at this late date for 8.0.0.

Not sure if this helps or hurts your efforts but maybe chime in there if you have thoughts!

chx:
What I said in https://www.drupal.org/node/2391219 stands: console shouldn't exist because it will force contrib authors to ship console AND drush integration both. As I mentioned there, core hardwires certain things to stop such idiocy -- modules can only ship default extension templates so that noone can demand from contrib authors to ship several templates. Configuration install storage is no longer a service, enforcing everyone to ship YAML configuration. Alas, the project has no tech leadership to enforce such splinter does not happen outside of core. Short of that and short of the webmasters having two brain cells to shut the console project down, this is an absolute critical so that when people file issues to ship console integration the contrib authors can won't fix them and point to drush integration and that's it.

greg-1-anderson:
Well, I admit it is kind of regrettable that there are two tools here, but the current state of affairs being what they are, most folks will not be able to help but to use both Drush *and* Drupal Console when working with Drupal 8.  This is going to cause some problems in contrib, but the problem is just going to be that contrib authors are going to have to decide whether to use Drush or Console to implement their cli tools; no one is going to need to write a Drush interface **and** a Console interface.

Still, there are lots of reasons why we should merge Console and Drush.  The main ones are:

 - There should be one cli bootstrap for D8. It's actually a lot of work to get this right, and this work needs to be duplicated.

 - No one wants to maintain two sets of site aliases. Drush has some pretty nifty remote command handling features; Console is just starting on remote execution, and it will be a big hassle for everyone if it works differently than Drush.

 - Drush has done a lot of work to eliminate the "dependency hell" problems that can arise with Composer-managed Drupal projects.  Console should leverage this work, not duplicate it -- or ignore it.  Console has been getting along just fine ignoring the problem, and rev-locking one version of Console to one version of Drupal, but this won't work well once people start using more than one version of Drupal 8 at a time.

The solution here isn't to "stop Console".  We all owe a lot to the Console project.  They have made some really useful tools, and the decision to decouple allowed them to make faster progress than if they had fixed the problems in Drush.  As I have mentioned before, we have had the idea / desire to use the Symfony Console component in Drush since before the Drupal Console existed, but no one has been available to finish the hard work that would make that possible.  This issue has come a long way to making that a reality, but it is unclear whether, in the end, we as a community are going to do the right thing, or whether we're going to keep doing the easier thing, and make small independent changes to each project.  I think that in the beginning, it was the right decision to work independently, but only because we lacked the volunteers to do the hard work, and we needed the capabilities that could be easily provided to the community by making a separate Symfony-based tool. The time has come now, though, to start paying down some of the technical debt, and work on better integration for the tools.

There are two more points I want to make in closing:

1. This issue is not the only way forward. We can also meet some of our goals by creating separate libraries used by both tools.  I have started some work in this direction at https://github.com/drush-ops/drush-aliases.

2. Even if we do merge Drush + Drupal Console into a single tool, we are still going to have both the legacy Drush command dispatcher, and the Symfony command dispatcher, and contrib authors will need to choose which one they use.

So, fragmentation is inevitable. The solution is not to try to stop folks from writing awesome tools, but rather to do more of the less-glamorous maintenance work to keep them working well together.

We're going to be sprinting at BADCamp; anyone who wants to pitch in is welcome to join us.

develCuy:
@greg-1-anderson ++

The spirit of Free Software is to let authors contribute with freedom, we should not read (yes, I'm using "should" @chx) "don't duplicate efforts" literally. I prefer to say: "don't waste efforts". Should we afford to rewrite Drush for Drupal 8? Drush was a fantastic tool but now is time for a new project: DrupalConsole. Letting Drush integrate with the future is an smart move, less hassle IMHO.

My 2c.

PS: http://cgit.drupalcode.org/drush/commit/?id=92611ecbcbfe7f6722882bcfa50e6cbcc915c2a5

mglaman:
I think this is a great move forward. Console supports calling Drush, Drush supports calling Console. Having two tools is a bit maddening, but it's better to piggyback one another than work on the same thing. Drush doesn't provide code scaffolding and Drush does many things Console does not.

Fragmentation is a repercussion of progress. At least efforts are being made to simplify the future and abstract components into shared libraries.

@greg-1-anderson++

greg-1-anderson:
See also #2103

weitzman:
Drush's master branch can now execute Symfony commands. I think thats as close as we are going to get on this one.

Issue #1339, opened 2015-04-16
Add "Drush script" to drush status for clarifying redispatch


derhasi:
In some situations it may be unclear to a drush user, if he using the global drush or was redispatched to another binary. Therefore it would be helpful to add "Drush script" as output to `drush status`.

This issue came to my mind, when testing https://github.com/drupal-composer/drupal-project/pull/16.

greg-1-anderson:
Fixed per suggestion.

PR #1341, opened 2015-04-16
#1339: Add path to Drush script to 'drush status' command.

greg-1-anderson:
Per #1339.  Seems like a good idea.


greg-1-anderson:
Okay, fixed the tests.  I think this is a good addition to drush status; if there is general agreement, I think it's ready to go in.

PR #1343, opened 2015-04-19
Strategies for managing your site-local Drush

greg-1-anderson:
If you 'use' a Drupal site with a site-local Drush, your PATH will be adjusted such that the site-local Drush appears first.  'use @none' or using a site with no site-local Drush will reset your PATH back to the way it was.

### DIFF
--- a/commands/core/sitealias.drush.inc
+++ b/commands/core/sitealias.drush.inc
@@ -398,8 +398,7 @@ function drush_sitealias_site_set($site = '@none') {
       // exists, in $HOME/.drush/use/__PPID__/bin and
       // $HOME/.drush/use/bin, or remove both of those
       // if there is no site-local Drush.
-      $use_dir = drush_server_home() . '/.drush/use';
-      $terminal_specific_link = drush_sitealias_create_site_specific_symlinks($use_dir, $site);
+      $terminal_specific_link = drush_sitealias_create_site_specific_symlinks($site);
       if ($terminal_specific_link === FALSE) {
         return FALSE;
       }
@@ -427,41 +426,44 @@ function drush_sitealias_site_set($site = '@none') {
  *   path to the terminal-specific symlink created, or an empty
  *   string if a symlink is instead removed.
  */
-function drush_sitealias_create_site_specific_symlinks($use_dir, $site) {
-  drush_mkdir($use_dir);
-  $terminal_specific_link = $use_dir . '/' . posix_getppid() . '/bin';
-  $global_link = $use_dir . '/bin';
+function drush_sitealias_create_site_specific_symlinks($site, $use_dir = FALSE) {
+  // Get the path to the link files.
+  $terminal_specific_link = drush_get_terminal_specific_use_link_path($use_dir);
+  $global_link = drush_get_global_use_link_path($use_dir);
 
   $alias_record = drush_sitealias_get_record($site);
   $root = sitealias_find_local_drupal_root(array($alias_record));
-  $site_local_drush_dir = FALSE;
+  $site_local_drush_target = FALSE;
   if (!empty($root)) {
-    $search_locations = array('.', 'core', '..');
-    foreach (array('vendor/bin', 'vendor/drush/drush') as $bin_dir) {
-      foreach ($search_locations as $loc) {
-        $drush_test = "$root/$loc/$bin_dir";
-        if (!$site_local_drush_dir && file_exists($drush_test . '/drush')) {
-          $site_local_drush_dir = $drush_test;
-          break 2;
-        }
-      }
-    }
+    $site_local_drush_target = find_wrapper_or_launcher($root);
   }
-  if ($site_local_drush_dir) {
+  if ($site_local_drush_target) {
     drush_mkdir(dirname($terminal_specific_link));
-    if (!@symlink($site_local_drush_dir, $terminal_specific_link)) {
+    if (!@symlink($site_local_drush_target, $terminal_specific_link)) {
       return drush_set_error('DRUSH_SYMLINK_ERROR', dt("Could not create symlink at !link", array('!link' => $terminal_specific_link)));
     }
-    if (!@symlink($site_local_drush_dir, $global_link)) {
+    drush_mkdir(dirname($global_link));
+    if (!@symlink($site_local_drush_target, $global_link)) {
       return drush_set_error('DRUSH_SYMLINK_ERROR', dt("Could not create symlink at !link", array('!link' => $global_link)));
     }
 
     return $terminal_specific_link;
   }
   else {
+    // If there is no site-specific Drush script associated
+    // with this alias, then remove the terminal-specific and
+    // global "use" symlinks.
     if (is_file($terminal_specific_link)) {
       if (!@unlink($terminal_specific_link)) {
         return drush_set_error('DRUSH_SYMLINK_ERROR', dt("Could not remove symlink at !link", array('!link' => $terminal_specific_link)));
+      }
+      // If there is a terminal-specific link (it will be at
+      // the same path as the global link if not), then also
+      // remove the PPID directory, and the 'bin' directory
+      // inside it.
+      if ($terminal_specific_link != $global_link) {
+        @rmdir(dirname($terminal_specific_link));
+        @rmdir(dirname(dirname($terminal_specific_link)));
       }
     }
     if (is_file($global_link)) {
--- a/commands/core/sitealias.drush.inc
+++ b/commands/core/sitealias.drush.inc
@@ -64,6 +64,9 @@ function sitealias_drush_command() {
       'drush site-set -' => 'Go back to the previously-set site (like `cd -`).',
       'drush site-set' => 'Without an argument, any existing site becomes unset.',
     ),
+    'options' => array(
+      'path' => 'Print out a modified PATH that places the site-local Drush at the head if it exists in the used site.',
+    ),
   );
   return $items;
 }
@@ -380,12 +383,37 @@ function drush_sitealias_site_set($site = '@none') {
       $success_message = dt("Site set to !site", array('!site' => $site));
       if ($site == '@none') {
         if (drush_delete_dir($filename)) {
-          drush_print($success_message);
+          drush_log($success_message, 'ok');
         }
       }
       elseif (drush_mkdir(dirname($filename), TRUE)) {
         if (file_put_contents($filename, $site)) {
-          drush_print($success_message);
+          drush_log($success_message, 'ok');
+        }
+      }
+      // See if user requested PATH adjustment
+      if (drush_get_option('path')) {
+        $alias_record = drush_sitealias_get_record($site);
+        $root = sitealias_find_local_drupal_root(array($alias_record));
+        $new_path = array();
+        if (!empty($root)) {
+          $search_locations = array('.', 'core', '..');
+          foreach ($search_locations as $loc) {
+            $drush_test = "$root/$loc/vendor/drush/drush";
+            if (is_file($drush_test . '/drush')) {
+              $new_path[] = $drush_test;
+            }
+          }
+        }
+        $path = getenv("PATH");
+        foreach (explode(':', $path) as $path_item) {
+          if (strstr($path_item, "/vendor/drush/drush") === FALSE) {
+            $new_path[] = $path_item;
+          }
+        }
+        $set_path = implode(':', $new_path);
+        if (trim($set_path) != trim($path)) {
+          drush_print($set_path);
         }
       }
     }
--- a/commands/core/sitealias.drush.inc
+++ b/commands/core/sitealias.drush.inc
@@ -391,34 +391,71 @@ function drush_sitealias_site_set($site = '@none') {
           drush_log($success_message, 'ok');
         }
       }
+      // Create a symlink to the site-local Drush, if it
+      // exists, in $HOME/.drush/use/__PPID__/bin and
+      // $HOME/.drush/use/bin, or remove both of those
+      // if there is no site-local Drush.
+      $use_dir = drush_server_home() . '/.drush/use';
+      $terminal_specific_link = drush_sitealias_create_site_specific_symlinks($use_dir, $site);
+
       // See if user requested PATH adjustment
       if (drush_get_option('path')) {
-        $alias_record = drush_sitealias_get_record($site);
-        $root = sitealias_find_local_drupal_root(array($alias_record));
-        $new_path = array();
-        if (!empty($root)) {
-          $search_locations = array('.', 'core', '..');
-          foreach ($search_locations as $loc) {
-            $drush_test = "$root/$loc/vendor/drush/drush";
-            if (is_file($drush_test . '/drush')) {
-              $new_path[] = $drush_test;
-            }
-          }
-        }
-        $path = getenv("PATH");
-        foreach (explode(':', $path) as $path_item) {
-          if (strstr($path_item, "/vendor/drush/drush") === FALSE) {
-            $new_path[] = $path_item;
-          }
-        }
-        $set_path = implode(':', $new_path);
-        if (trim($set_path) != trim($path)) {
-          drush_print($set_path);
-        }
+        drush_sitealias_manage_path($use_dir, $terminal_specific_link);
       }
     }
     else {
       return drush_set_error('DRUPAL_SITE_NOT_FOUND', dt("Could not find a site definition for !site.", array('!site' => $site)));
     }
+  }
+}
+
+function drush_sitealias_create_site_specific_symlinks($use_dir, $site) {
+  drush_mkdir($use_dir);
+  $terminal_specific_link = $use_dir . '/' . posix_getppid() . '/bin';
+  $global_link = $use_dir . '/bin';
+
+  $alias_record = drush_sitealias_get_record($site);
+  $root = sitealias_find_local_drupal_root(array($alias_record));
+  $site_local_drush_dir = FALSE;
+  if (!empty($root)) {
+    $search_locations = array('.', 'core', '..');
+    foreach (array('vendor/bin', 'vendor/drush/drush') as $bin_dir) {
+      foreach ($search_locations as $loc) {
+        $drush_test = "$root/$loc/$bin_dir";
+        if (!$site_local_drush_dir && file_exists($drush_test . '/drush')) {
+          $site_local_drush_dir = $drush_test;
+        }
+      }
+    }
+  }
+  if ($site_local_drush_dir) {
+    drush_mkdir(dirname($terminal_specific_link));
+    @symlink($site_local_drush_dir, $terminal_specific_link);
+    @symlink($site_local_drush_dir, $global_link);
+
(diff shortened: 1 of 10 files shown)


commands/core/sitealias.drush.inc
```diff
@@ -365,12 +368,37 @@ function drush_sitealias_site_set($site = '@none') {
...
+          foreach ($search_locations as $loc) {
+            $drush_test = "$root/$loc/vendor/drush/drush";
+            if (is_file($drush_test . '/drush')) {
+              $new_path[] = $drush_test;
+            }
+          }
+        }
+        $path = getenv("PATH");
+        foreach (explode(':', $path) as $path_item) {
+          if (strstr($path_item, "/vendor/drush/drush") === FALSE) {
```
weitzman: Any reason not to choose vendor/bin/drush? That should have execute perm.

greg-1-anderson:
I deliberately used vendor/drush/drush instead of vendor/bin so that I could remove the path I added with relative confidence.  If I added vendor/bin, it would have the side effect of not just selecting Drush, but also favoring all of the executables associated with the Drupal site you just used.  This might actually be a *desirable* side effect, but again, I am not sure how I would differentiate between the paths added by Drush, from vendor/bin paths added by the user.  Conceivably, a user might be using composer to manage a collection of executables anywhere.  In the current PR, I remove all of the vendor/drush/drush paths I find, on the assumption that the user's PATH elements are all vendor/bin.

I'd be happy to go with vendor/bin if you had a good idea how to remove the path some other way.

weitzman:
Brilliant! 

I would love for this to work for folks who don't use our bashrc suggestions And don't pass --path. This pr is a bit too conservative IMO. 

I also like the side effect you describe but I'm ok with either approach.

greg-1-anderson:
I decided that maybe I don't like my idea.  Folks will look at the PATH we write, and then decide that vendor/drush/drush is the right PATH to use, and set vendor/drush/drush in their PATH, and get upset with Drush when we remove their paths.  Maybe this is not a serious issue, but I have a better idea anyway, that incorporates the desirable side-effect described above.

My other thought was to set the PATH to __HOME__/.drush/__PPID__/bin, and make that path a symlink to the vendor/bin directory in the currently-`use`d Drupal site.  If a different site is `use`d, then we're use the same PATH element, and just move the symlink.  For `use @none`, just remove the symlink.

For folks who do not use our bashrc suggestions, we can also make a symlink __HOME__/.drush/use/bin, and set it every time we set __HOME__/.drush/__PPID__/bin, and remove it every time `use @none` is used.  Then, we put into our documentation instructions that folks must manually add __HOME__/.drush/use/bin at the head of their path (but only if they do NOT include our example bashrc file).  These users do not get to see which site they used on their prompt, and whichever site they site-set will be effective across all terminal windows.  They can figure out which site they are using via `which drush`.  n.b. you cannot set an environment variable from a subprocess, so there really isn't a better option than this if our bashrc is not being sourced.

greg-1-anderson:
I'll have time to code this up & fix the unit test later this evening.

greg-1-anderson:
Here's one step closer, but we're not quite there yet.

In this PR, you can operate in one of two modes.

1. via the "use" bash function in examples.bashrc -- full functionality, each terminal can 'use' a different Drupal site.

2. simple mode -- add __HOME__/.drush/use/bin to the head of your $PATH.  Then, `drush site-set @site` will insure that the site-local Drush comes first in your path whenever it selects a site.  It does this by modifying the symlink in __HOME__/.drush/use/bin.  This is global, for all terminals.

There are two problems here:

A: If you call the site-local drush via `drush`, by managing your PATH per 1. or 2., then the --local option is not set.  This can cause global commandfiles to load, which may be undesirable.

B: When operating in the simper mode, it is possible that you might use site-set to select a different site in each terminal; however, there will be only ONE site-specific Drush on your PATH for all terminals, so you might end up calling the wrong site-local Drush.  This would be bad.

I see some potential solutions to these problems.

Proposed solution for A: We already figure out which autoloader to call, by looking for autoload.php in different locations first thing in the preflight.  Based on the location of the autoload.php file, we could detect that our autoloader is shared with a Drupal site, and put ourselves into site-local Drush mode automatically.  This mode could just disable loading global commandfiles, and --local could turn off configuration, for folks who want to be very strict about team-specific settings.  We could potentially change the order the config files are loaded in, so that folks could set --local in the drushrc.php file of their Drupal site.  This, of course, would only work when your site-local Drush is on your PATH.

Proposed solution for B: Site-set could detect when __HOME__/.drush/use/bin is on your PATH, and revert to one-site-for-all-terminal modes when it detects this, so that `site-set` works the same way as the site-local-Drush selection mode.  Note that when using `use` to manage your PATH, then the path Drush writes is __HOME/.drush/use/__PPID__/bin; site-set could also detect this, and go back to one-site-per-terminal mode if this path is set.  To make it easier for folks who already have their prompt set (because they are using site-set to select different sites in each terminal), we could move the 'use' function from examples.bashrc to drush.complete.sh.  Note that the later is sourced by the former, so if you have the example bash functions, you always have drush.complete.sh, but the reverse is not true.

I think that with these adjustments, then site-set / use will be fairly easy to configure, and should also provide a UX that folks can understand and use correctly.

greg-1-anderson:
It's actually hard to find the Drupal root given the location of the autoloader.  There's no standardization on where the vendor directory goes; it might be found in:

  core/vendor
  vendor
  sites/all/vendor

Some people put it in the directory above the root, which is a very good place, IMO, but introduces extra difficulties if you want to find the root given the autoloader.

If there was some executable component in vendor/bin that was placed somewhere in the Drupal root, then we could easily find the root by following the symlink.  Drupal does not include anything like this, though (no executables in Drupal's vendor/bin -- more is the pity), and we wouldn't want to require Drush to be in the Drupal root.

I've made a proposed PR that autodetects the site-local Drush in #1340.

commands/core/sitealias.drush.inc
```diff
@@ -365,17 +368,79 @@ function drush_sitealias_site_set($site = '@none') {
...
+  $alias_record = drush_sitealias_get_record($site);
+  $root = sitealias_find_local_drupal_root(array($alias_record));
+  $site_local_drush_dir = FALSE;
+  if (!empty($root)) {
+    $search_locations = array('.', 'core', '..');
+    foreach (array('vendor/bin', 'vendor/drush/drush') as $bin_dir) {
+      foreach ($search_locations as $loc) {
+        $drush_test = "$root/$loc/$bin_dir";
+        if (!$site_local_drush_dir && file_exists($drush_test . '/drush')) {
+          $site_local_drush_dir = $drush_test;
```
weitzman: should we call `break` once we find one?

commands/core/sitealias.drush.inc
```diff
@@ -365,17 +368,79 @@ function drush_sitealias_site_set($site = '@none') {
...
+        }
+      }
+    }
+  }
+  if ($site_local_drush_dir) {
+    drush_mkdir(dirname($terminal_specific_link));
+    @symlink($site_local_drush_dir, $terminal_specific_link);
```
weitzman: maybe  should log when there is a problem with symlink (debug is OK). error suppression has bitten us in the past.

commands/core/sitealias.drush.inc
```diff
@@ -365,17 +368,79 @@ function drush_sitealias_site_set($site = '@none') {
...
+  }
+  $path = getenv("PATH");
+  foreach (explode(':', $path) as $path_item) {
+    if (strstr($path_item, $use_dir) === FALSE) {
+      $new_path[] = $path_item;
+    }
+  }
+  $set_path = implode(':', $new_path);
+  if (trim($set_path) != trim($path)) {
+    drush_print($set_path);
```
weitzman: Should we return our value instead of print?

Also, new functions should get Doxygen.

weitzman:
Want to record an external email conversation here:

* Moshe: Your site-set PR is intriguing, but it requires folks to remember
about site-set and somewhat hassles them to use our bashrc code. The
phpenv is nice because you put the shim dir in your PATH and then all
can get handled for you with files like
https://github.com/phpenv/phpenv#phpenv-local. i'm kinda hand waving
here. no time to dig in.
* Greg:  I'm in the middle of writing something that works just like that, with a shim directory that Drush maintains (writes a symlink to the right site-local drush). It still requires the user to use site-set, and it works better if they use our bashrc code (because you can get per-term Drush), but I think it's good ... Oh, I see, phpenv also writes a file to the cwd.  I don't think I like that idea; I'd prefer it if the Drush I selected "followed me around", and I'd also prefer it if site-set did not leave scattered dot files around (that might later select a Drupal site for me without me noticing).

I'll think more about the tradeoffs here. Not sure.

commands/core/sitealias.drush.inc
greg-1-anderson: On the one hand, print vs. return does not matter much, since you can't call site-set remotely, and there is only one output mode supported.  However, I switched to 'return' instead of 'print' as suggested to reduce side-effects in the code.

greg-1-anderson:
Okay.  I added a commit to fix some of the issues so far.  We still need to decide about [A] and [B] above.

greg-1-anderson:
Here are some comments that describe how we could use a "finder" script in addition to our existing "wrapper" (`examples/drush`) and "launcher" (existing `drush`) scripts to route to the correct site-local Drush.  This could be in addition to, or instead of the $PATH management features in this PR.

The scripts are simpler than the explanation, but I decided it would be best to get your take on the strategy before perfecting the scripts.  Let me know what you think about this.
```
#
# OVERVIEW OF DRUSH FINDER / WRAPPER / LAUNCHER SCRIPTS
#
# When the user types "drush", up to three scripts might be
# involved in the initial launch:
#
#   "drush finder" -> "drush wrapper" -> "drush launcher".
#
# DRUSH FINDER
#
# - The "drush" script on the user's global $PATH
# - It's goal is to find the correct site-local Drush to run.
#
# The Drush finder will locate the correct site-local Drush to use
# by examining:
#
#   a) The --root option
#   b) The site set via `drush site set` in the current terminal
#   c) The cwd
#
# If no site-local Drush is found, then the global Drush will be
# used.  The Drush finder assumes that the global Drush is the
# "Drush launcher" found in the same directory as the Drush finder itself.
#
# If a site-local Drush is found, then the Drush finder will call
# either the "Drush wrapper", if it exists, or the "Drush launcher" if
# there is no wrapper script.  If the "Drush launcher" is called, then
# the Drush Finder will add the --local flag.  If the --local flag was
# already passed in when the Drush Finder is called, then it will always
# call the "Drush launcher", even if there is a Drush Wrapper, to prevent
# possible infinite loops.
#
# DRUSH WRAPPER
#
# - The "drush" script that the user copies to their Drupal __ROOT__
# - Its goal is to allow the user to add config locations when --local is in use
#
# The "drush wrapper" is found in examples/drush, and may optionally
# be copied to __ROOT__/drush by the user.  It will call the "drush launcher"
# for the same site that it is located in.  It adds the --local flag; the
# user is encouraged to add other options to the "drush wrapper", e.g. to set
# the location where aliases and global commandfiles can be found.
# We must always call the "drush wrapper" if it exists, since it
# can be customized.  However, if the user does not want to customize
# the site-local Drush's early options (site-alias locations, etc.),
# then the wrapper does not need to be used.
#
# DRUSH LAUNCHER
#
# - The "drush.launcher" script in vendor/bin
# - The script formerly called "drush"
#
# The "drush launcher" is the traditional script that identifies PHP and
# sets up to call drush.php.  It is called by the "drush wrapper", or
# directly by the "drush launcher" if there is no "drush wrapper" in use.
#
# SUMMARY OF WHERE THESE SCRIPTS ARE LOCATED
#
#   "drush finder"  :   __ROOT__/vendor/bin/drush           (composer install)
#                       __DRUSH__/drush                     (source)
#
#   "drush wrapper" :   __ROOT__/drush                      (copied by user)
#                       __DRUSH__/examples/drush            (source)
#
#   "drush laucher" :   __ROOT__/vendor/bin/drush.launcher  (composer install)
#                       __DRUSH__/drush.launcher            (source)
#
# WHAT ABOUT BACKEND CALLS?
#
# Backend calls should be set up to call the "drush" script in the $PATH,
# or perhaps some might call __ROOT__/vendor/bin/drush directly, by way
# of the "drush-script" element in a site alias.  In either event, this will
# be the "drush finder".
#
# The backend call will always set --root.  The "drush finder" script
# always favors the site-local Drush stored with the site indicated by the
# --root option, if it exists.  If there is no site-local Drush, then the
# "drush finder" will behave as usual (i.e., it will end up calling the
# "drush launcher" located next to it).
#
# This should always get you the correct "Drush" for local and remote calls.
#
# WHAT IF THE DRUSH WRAPPER CALLS THE DRUSH FINDER BY MISTAKE?
#
# Old versions of the "drush wrapper" call vendor/bin/drush instead of
# vendor/bin/drush.launcher, so this introduces the possibility of causing
# an infinite loop if someone has installed this wrapper, and then
# upgrades to the new Drush that utilizes this launch scheme.
#
# In that instance, the --local flag will be set by the "drush wrapper",
# because the wrapper always adds this option.  If the "drush finder" sees that
# this flag has been set, then it will call the "drush launcher" directly,
# bypassing the "drush wrapper" to avoid the infinite loop.
#
# If a user copies an old "drush wrapper" and removes --local, then an
# infinite loop will result.
#
# WHAT SYMLINKS ARE MAINTAINED BY SITE-SET?
#
# The 'site-set' command will write the following symlink:
#
#  .drush/use/bin/drush -> drush (maybe no longer necessary?)
#  .drush/use/__PPID__/bin/drush -> drush
#  .drush/use/__PPID__/root -> __ROOT__
#
# [This location was choosen to make it easier to allow users to place
# this directory in their $PATH; however, with the "Drush finder" script,
# we probably will not recommend this.  We could therefore move these to
# the same tmp directory that "site-set" currently uses for Aliases.]
#
# If there is a "drush wrapper" in the "use"d site, then the "drush" symlink
# will point to it, and the "root" symlink will be deleted.
#
# If there is no "drush wrapper" in the "use"d site, then the "drush" symlink
# will point to the "drush launcher" in the local site, and the "root" symlink
# will point to the Drupal root.
#
# If there is no site-local Drush at all (or if the "use"d site is remote),
# or if the user types "drush site-set @none", then site-set will delete all
# of these symlinks.
#
# WHAT "USE"d ALIASES ARE MAINTAINED BY SITE-SET
#
# The 'site-set' command caches the "used" alias name in the following file:
#
#  __TMP__/drush-env-__USER__/drush-drupal-site-__PPID__
#
# This file affects the site that Drush will select during bootstrap.
# None of the Drush scripts parse or evaluate this file in any way;
# The aliases are used instead.
#
# HOW DOES DRUSH MANIPULATE THE $PATH?
#
# If the user has source'ed our example.bashrc, and used the
# 'use' bash function rather than calling `drush site-set`
# directly, then their $PATH will be updated dynamically to
# include .drush/use/__PPID__/bin; in that instance, when they
# run 'drush', it will directly execute the site-local drush,
# and this script will not be involved.
#
# If, on the other hand, they do not use the 'use' command,
# then they should have ~/.composer/vendor/bin in their $PATH.
# This, as shown above, will be the "Drush finder".
#
# DOES THE SITE-LOCAL DRUSH *ALWAYS* OPERATE ON THE SITE IT IS STORED WITH?
#
# No.
#
# Example:
#
#   $ drush use @local
#   $ drush @remote status
#
# In this instance, the second call to "drush" will end up dispatching
# to the site indicated by @local, which will be used to find and dispatch
# to the @remote site.  Note that it is possible that @remote might be
# available only in site alias files enabled by options set in the
# "Drush wrapper" script.
#
# IS IT POSSIBLE TO USE THE GLOBAL DRUSH AFTER "USE"ing A SITE-LOCAL DRUSH?
#
# Yes.
#
#   $ drush @none status
#
# The "Drush Finder" will recognize the "@none" parameter, and will then
# ignore the site-local Drush, and launch the global Drush instead.
#
# [n.b. This does not work if Drush is manipulating the $PATH, because the
# selected site-local Drush will be called directly.  The @none will cause
# it to not bootstrap the local site, but the local Drush, rather than the
# global one, will be used.]
#
# CAN THIS BE ANY SIMPLER?
#
# Yes.
#
# - Eliminate the Drush wrapper.  This would simplify a number of conditionals
#   in the "Drush finder", and would be one less concept for users to have
#   to worry about.  The downside here is that there wouldn't be any good way
#   to specify custom site-local configuration "early"; you'd have to put a
#   config file in a default location, and use it to define other valid locations.
#   Probably not a good idea to remove this.
#
# - Eliminate the $PATH manipulation in the "use" Bash alias.  The Bash alias
#   is more efficient and simpler than the "Drush finder"; however, as noted
#   above, it is less flexible, and not necessary if we have the Drush Finder.
#   Probably a good idea to remove iff we commit the Drush Finder, just to
#   reduce the number of different environments we'd need to deal with in
#   support requests.
#
```

greg-1-anderson:
A section that got written only in my head, forgotten above, and typed up below:
```
#
# HOW DOES DRUSH SET THE "LOW-PRIORITY" DRUPAL ROOT?
#
# The "Drush wrapper" script now includes the following directive:
#
#    cd "$(dirname $0)"
#
# Setting the current working directory to the Drupal Root for the local site has
# the desired side effect of specifying the "default" or "low-priority" Drupal root
# to use, based on existing site-selection logic in Drush.
#
# When there is no "Drush wrapper" script, then the "Drush finder" will call
# the "Drush launcher" directly; in this instance, it also sets the cwd to the
# Drupal root first, which it always knows.
#
```

weitzman:
I think this is sane enough that it makes sense to keep developing the system and then we can see how much we like/dislike when it is a mature PR.

greg-1-anderson:
Webflo created this: https://github.com/webflo/drushenv generated from the "anyenv" project.

The advantages of this is that it's simple to build, and gives the user pretty cool control over selecting the global version of Drush that they want to use.  The downside over the currently-suggested approach is that it requires the user select the "site" version of Drush; if the user selects "Drush 8" or "Drush 7" instead, then the appropriate version of Drush associated with a given Drupal site is not automatically selected.

Perhaps the best solution would be to move forward with the "Drush Finder" script for auto-selection of the correct site-local Drush, and allow users to optionally also install Drush env if they want to select their global version of Drush -- which of course still will go wrong if you explicitly select an older version of Drush that does not have the "Drush Finder" script, but at least the default case ("stock" Drush users on a modern version) are covered.

weitzman:
Yeah, lets keep developing the finder script independant of drushenv. I bet many folks will find little need to mess with shims for selecting a global drush.

Let us know when the PR is ready for review.

greg-1-anderson:
Working on merging this with #1595.  Still needs work; some tests are failing.

greg-1-anderson:
This PR breaks the site-set command when there is **not** a site-specific Drush.  I'll fix this later (soon).

greg-1-anderson:
Pushed this forward a little bit.  I'm not sure that this needs to be finished, though; #1595 might solve all open problems on its own without this infrastructure.

weitzman:
Closing in favor #1595

Issue #1344, opened 2015-04-20
drush site-install breaks if minimal.profile is not present


marcvangend:
There is a patch in the Drupal 8 issue queue which works fine from within Drupal, but breaks `drush site-install`: [Do not use SERVER_NAME as default value for the site name](https://www.drupal.org/node/2473709) (especially comment # 12 and below).

As far as I can see, this is a drush bug. Maybe it can be fixed in Drush before that patch goes into core (assuming that it will).

greg-1-anderson:
Casual inspection of the code does not reveal the source of this dependency; I'll have to trace through it and see why Drush si is behaving differently than the GUI installation.  I don't see a dependency on the minimal installation profile in any .info.yml file in Core, so this is a bit of an oddity.

marcvangend:
Thanks for looking into this Greg. I did some more tests to figure out exactly when this problem occurs. TL;DR: It seems to be caused by the `$settings['install_profile']` variable in settings.php.

Long version:

First, I tried to install Drupal with a completely clean codebase (except for the patch from https://www.drupal.org/node/2473709 of course), specifying the database credentials in the command:

    $ drush si standard --db-url=mysql://dbuser:password@localhost:3306/drupal8

That worked.
Second, I re-installed with the standard profile, using the db credentials from settings.php:

    $ drush si

No problems there.
Next, I re-installed with the minimal profile, again using the db credentials from settings.php:

    $ drush si minimal

Again no errors.
Finally, I tried to re-install with the standard profile, which failed:

    $ drush si 
      You are about to DROP all tables in your 'drupalcore' database. Do you want to continue? (y/n): y
      Starting Drupal installation. This takes a while. Consider using the --notify global option.            [ok]
      exception 'Drupal\Core\Installer\Exception\InstallerException' with message 'Required modules: Required [error]
      modules not found.
      
      The following modules are required but were not found. Move them into the appropriate modules
      subdirectory, such as <em>/modules</em>. Missing modules: <span class="admin-missing">Minimal</span>' in
      /home/marc/http/public/drupalcore/core/includes/install.core.inc:2268

In an attempt to figure out where this reference to Minimal was stored, I deleted all config files with `rm -rf sites/default/files/*`, dropped all tables in the database and tried again. The error still occurred. Then I deleted the following line from settings.php:

    $settings['install_profile'] = 'minimal';

[...]

PR #1345, opened 2015-04-20
Refuse re-install to a different install profile due to Drupal bug

greg-1-anderson:
If settings.php contains a cached value holding the installation profile, and drush site-install is ran again to install a different profile, then we must update the settings value, or the install will fail.

Anyone have any objection to Drush performing this service to insure that the install goes okay?  Should we write-protect settings.php again afterward modifying it?

### DIFF
--- a/commands/core/site_install.drush.inc
+++ b/commands/core/site_install.drush.inc
@@ -155,7 +155,20 @@ function drush_core_pre_site_install($profile = NULL) {
     drush_log(dt('Sites directory @subdir already exists - proceeding.', array('@subdir' => $conf_path)));
   }
 
-  if (!drush_file_not_empty($settingsfile)) {
+  // If there is an existing settings.php file, and if it records the installation profile used,
+  // then update it to insure that it matches the installation profile currently being installed.
+  // See https://github.com/drush-ops/drush/issues/1344
+  if (drush_file_not_empty($settingsfile)) {
+    $selected_profile = $profile ?: "standard";
+    // If the settings.php file exists, we need to make sure that the $settings['install_profile'] line is
+    // removed, if it exists.
+    $settings_contents = file_get_contents($settingsfile);
+    $settings_contents = preg_replace('/(\$settings\[.install_profile.\] = ).*/m', '\1 ' . "'$selected_profile';", $settings_contents);
+    @chmod(dirname($settingsfile), 0777);
+    @chmod($settingsfile, 0777);
+    file_put_contents($settingsfile, $settings_contents);
+  }
+  else {
     if (!drush_op('copy', 'sites/default/default.settings.php', $settingsfile) && !drush_get_context('DRUSH_SIMULATE')) {
       return drush_set_error(dt('Failed to copy sites/default/default.settings.php to @settingsfile', array('@settingsfile' => $settingsfile)));
     }


weitzman:
This has been an issue since forever. I kinda think we should just leave it alone. Fiddling with a read-only settings.php gives me a bad feeling. We also have to log when we fail to edit the file.

greg-1-anderson:
Drush `site-install` time is one place where I think it's reasonable to modify settings.php, since the Drupal installer does as much itself.  It's a real head-scratcher when this just fails; if we don't fix it up ourselves, then the user thinks that Drush is broken, since the site install works when run in the browser.  See https://www.drupal.org/node/2473709 and #1344.

Maybe we should only fix settings.php if it's already modifiable, and abort if settings.php's profile value doesn't match the user-selected value.  Then at least we could tell the user what was wrong, rather than just let it blow up in a hard-to-diagnose way.

weitzman:
Couldn't we just call this a Drupal bug that it does not append the new
profile upon re-install?

greg-1-anderson:
My inclination is to report and abort (at a minimum), to avoid letting people run down the rabbit hole. If Drupal fixed the problem, then the abort would prevent us from running a working use case.

I'm skeptical that this will be fixed in core, though, so I'd rather see it detected or fixed in Drush.

weitzman:
I found https://www.drupal.org/node/2156401 which was marked a dupe of
https://www.drupal.org/node/1245714 which was fixed for D8 a month ago. So
maybe we don't take the PR anymore, since D7/D7 users have suffered with
this for a while? Requires some experimentation.

greg-1-anderson:
When I was testing this last week, I was using D8.  I think my experience was the same as reported in the issues I quoted above -- worked in the installer, failed in Drush.

weitzman:
I'm pretty sure this is a Drupal bug, if still exists. Maybe Drupal's non-interactive install was not fixed ... We can refuse to proceed, but I dont want to write install_profile to settings.php ourselves. We got out of that business as of https://www.drupal.org/node/1245714

greg-1-anderson:
Hm, yes, this situation is very similar to the problem in the issue you reference, ironically enough.

I'll change this PR to refuse to proceed if the install_profile setting has changed.  If I can perhaps let the install run, and detect that Drupal did not fix the install_profile, that would be even better.  I'll try it; then, things will be cool once (if) the problem is fixed in Drupal.

weitzman:
I reactivated this Drupal bug - https://www.drupal.org/node/2156401

weitzman:
Also, lets refuse the reinstall *before* dropping existing tables and such.

pjcdawkins:
Make sure you read https://www.drupal.org/node/2451363 because it's fixed (in D8) and it *seems* related

weitzman:
Yes, that Drupal issue fixed a problem but caused head aches when settings.php is read-only. [Here are some Pantheon customers feeling the pain](https://github.com/pantheon-systems/drops-8/issues/36). I submitted [a Drupal patch which mitigates the issue](https://www.drupal.org/node/2156401). /cc @greg-1-anderson

greg-1-anderson:
Fixed in Drupal. Solution here was kind of hacky anyway.

PR #1350, opened 2015-04-22
Use pcntl_exec to use process replacement when redispatching

greg-1-anderson:
If you are not selecting your site-local Drush via your PATH environment variable (c.f. #1343), then you can cause Drush to redispatch to the site-local Drush by setting an option in __ROOT__/drush/drushrc.php:

```$options['drush-script'] = __DIR__ . "/../core/vendor/drush/drush/drush";```

This much already works.  (n.b. this file is included before Drupal is bootstrapped, so DRUPAL_ROOT is not available here.)

When operating in this mode, this PR alters the Drush redispatch to use pcntl_exec to do the redispatch.  This means that the initial 'launcher' Drush process dies, and is replaced by the process running the site-local Drush, removing the need for backend invoke to shuffle stdin and stdout (and avoiding stderr being redirected to stdout).

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -688,6 +688,12 @@ function _drush_backend_adjust_options($site_record, $command, $command_options,
  *            dispatched using the alias name on the command line, instead of
  *            the options from the alias being added to the command line
  *            automatically.
+ *          'process-replacement'
+ *            Optional. Defauts to FALSE. Only usable locally.
+ *            If set, then backend invoke with use execve to do a full process
+ *            replacement, overlaying the new process on top of the current
+ *            process, terminating it. In this mode, the call to backend
+ *            invoke will never return.
  * @param common_options
  *    Optional. Merged in with the options for each invocation.
  * @param backend_options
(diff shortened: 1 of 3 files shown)


weitzman:
* [PCNTL is not enabled by default and not available on Windows](http://php.net/manual/en/pcntl.installation.php). Not sure if those are deal killers but its close. if we proceed we should handle those situations.
* We are trying to move toward using Symfony Process component in #673. Would this be a step backward?

greg-1-anderson:
Does the Symfony Process component support pcntl_exec?  I didn't check, but suspect not.

In any event, this is a performance enhancement (or, in our case, a stderr-handling enhancement), an optional bit of code that can be disabled if it's not desired. As such, I don't think it's a step backwards.  I'll modify it so that it is silently ignored if pcntl_exec is not available.

Maybe we should go ahead and commit #673, if its tests are still passing.  I still don't think we should tackle a full backend invoke refactor yet, but some minor process improvements such as this could be considered.

greg-1-anderson:
My point above being, is that it will be easier to review the impact of PRs such as this one if #673 is already committed, so let's put it in if nothing is blocking that.  Then we can reroll here if it causes conflicts, but I suspect the issues are orthogonal.

PR #1355, opened 2015-04-27
Don't access uid directly, as it differs between drupal versions.

### LINKED ISSUE #1354: drush user-create problem with D8
```bash
$ drush user-create pepe --mail="pepe@example.com" --password="pepe"
```
```
explode() expects parameter 2 to be string, object given drush.inc:203                                                                                                                                                     [warning]
array_filter() expects parameter 1 to be array, null given drush.inc:203                                                                                                                                                   [warning]
array_map(): Argument #2 should be an array drush.inc:203                                                                                                                                                                  [warning]
Invalid argument supplied for foreach() UserList.php:102
```

penyaskito:
Use UserVersion->getUid() instead.

Fixes  #1354

### DIFF
--- a/commands/user/user.drush.inc
+++ b/commands/user/user.drush.inc
@@ -311,8 +311,7 @@ function drush_user_create($name) {
   );
   if (!drush_get_context('DRUSH_SIMULATE')) {
     if ($account = $userversion->create($new_user)) {
-      $userlist = new UserList($account->uid);
-      return $userlist->each('info');
+      return array($account->id() => $account->info());
     }
     else {
       return drush_set_error("Could not create a new user account with the name " . $name . ".");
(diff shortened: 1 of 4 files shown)


jonhattan:
There's no need to add a new method. This seems to me the way to go:

```
$account = $userversion->create($new_user);
$userversion->setCurrentUser($account);
$usersingle = $userversion->getCurrentUserAsSingle();
$usersingle->id();
```

penyaskito:
Discussed with @jonhattan on IRC. If `create()` returns a `UserSingle` instance, this would be much cleaner and we could just call `id()`.

jonhattan:
.. and we keep on the drush abstraction layer.

PR #1356, opened 2015-04-27
Remove semicolon from invalid array declaration

mjl3475:


greg-1-anderson:
A comma would have been even better here, but since this is an improvement over the typo in the file, I went ahead and merged it as-is.

PR #1364, opened 2015-05-04
Allow overriding default Cache class in custom config files.

### LINKED ISSUE #1036: Allow altering of cache class even if site alias is not used.
Is there any way to force a specific drushrc.php config file for a drush command in versions of drush greater than 6.3?

Once you removed these lines: https://github.com/drush-ops/drush/blob/6.2.0/includes/sitealias.inc#L279-L282, 
combined with this line in bootstrap.inc:
https://github.com/drush-ops/drush/blob/6.x/includes/bootstrap.inc#L505
right before the line that loads the config... you avoided any way in setting a default cache class for drush.

That's because at some point you are looking for cached aliases, so **drush_cache_get** is called and cache object is initialized, that's a static variable:
https://github.com/drush-ops/drush/blob/6.x/includes/cache.inc#L40-L51
Having no config loaded... default DrushJSONCache class is used to create cache object.

Previously a default config file was set in sitealias.inc (sites/default/drushrc.php) when you looked up for aliases, see again: https://github.com/drush-ops/drush/blob/6.2.0/includes/sitealias.inc#L279-L282

Now... even if I alter the default cache class in drushrc.php, by using 

```
$options['cache-default-class'] = 'CustomCacheClass';
```

or

```
$options['cache-class-default'] = 'CustomCacheClass';
```

the default class will be DrushJSONCache initially.

Do you see any way to alter the cache class even if I don't use site aliases?

E.g. By running: **drush** **sitefolder** **command** everything is fine, but if I run only **drush** **command** from root folder of drupal, the cache class is not altered early enough.

laurentiu1981:
This commit fixes #1036

### DIFF
--- a/includes/bootstrap.inc
+++ b/includes/bootstrap.inc
@@ -500,16 +500,16 @@ function _drush_bootstrap_drush() {
   // Check to see if we 'use'd a site alias with site-set
   drush_sitealias_check_site_env();
 
-  // If an alias was not created by drush_sitealias_check_arg()
-  // or drush_sitealias_check_site_env(), make one now.
-  drush_sitealias_create_self_alias();
-
   // Load the config options from Drupal's /drush and sites/all/drush directories,
   // even prior to bootstrapping the root.
   drush_load_config('drupal');
 
   // Similarly, load the Drupal site configuration options upfront.
   drush_load_config('site');
+
+  // If an alias was not created by drush_sitealias_check_arg()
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This must be done for the master branch first.

greg-1-anderson:
Bootstrapping is quite different in `master`. Not sure if the initial use-case is satisfied or not. Parties interested in resurrecting this effort should try reworking for the latest version of Drush.

PR #1378, opened 2015-05-12
Fix "drush: 129: [: -le: unexpected operator" errors when Xdebug is enabled

tstoeckler:
I don't exactly know why, but Xdebug messes up the `php -r` call in the Drush sh script. This commit simply disables Xdebug for the version check, where it should be unnecessary anyways. This avoids the nasty errors/failures.

### DIFF
--- a/drush
+++ b/drush
@@ -125,7 +125,9 @@ if [ -n "$PHP_OPTIONS" ] ; then
 fi
 
 # If on PHP 5.3, disable magic_quotes_gpc and friends
-PHP_MINOR_VERSION="`\"$php\" -r 'print PHP_MINOR_VERSION;'`"
+# If Xdebug is enabled it interferes with the following command, so disable it
+# for the version check (but not for the actual call to drush.php).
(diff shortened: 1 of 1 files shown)


tstoeckler:
See b515bd9f23a807b1de968fdfb0481f032f3ca423, 760d41b120a292d0911f45341d5f4abb3194098d and 00c0ab128364e77a74ac4db349a987d51fe47523 for previous art

weitzman:
I have xdebug enabled and have no trouble. We need to understand root cause a bit better imo.

tstoeckler:
Hmm... Debugging Xdebug internally is over my head unfortunately.

Maybe this is related to http://bugs.xdebug.org/bug_view_page.php?bug_id=00001102, which was fixed in XDebug 2.3.0. I have 2.2.6 locally, so in case you have a later version that would be a *possible* explanation for the problem and why you're not seeing it.

On the other hand, disabling Xdebug for the version check doesn't hurt anyone, right?

tstoeckler:
On the other hand that mentions a segfault, which is not what happens... Hmm...

weitzman:
No recent feedback.

PR #1381, opened 2015-05-13
Disable Drupal 8 support in the Drush 7.x branch.

greg-1-anderson:
Drupal 8 support will continue in the 'master' branch after the creation of the 7.x branch.

### DIFF
--- a/lib/Drush/Boot/DrupalBoot8.php
+++ b/lib/Drush/Boot/DrupalBoot8.php
@@ -68,6 +68,10 @@ class DrupalBoot8 extends DrupalBoot {
     return $core;
   }
 
+  function bootstrap_drupal_root_validate() {
+    return drush_set_error('DRUSH_DRUPAL_VERSION_UNSUPPORTED', dt('Drush !drush_version does not support Drupal 8.', array('!drush_version' => DRUSH_VERSION)));
+  }
+
   function bootstrap_drupal_database() {
     // D8 omits this bootstrap level as nothing special needs to be done.
     parent::bootstrap_drupal_database();


ergonlogic:
Could we perhaps add a flag to bypass this check?

We'd like to support D8 in Aegir3 and are looking at a release in mid-July (around the AegirSummit at NYCcamp). At this point, the plan is to require Drush 7, but this'd be a non-starter for us.

greg-1-anderson:
It would be futile to bypass the check with a flag; once Drush 7 no longer supports Drupal 8, it will (probably) quickly be broken by changes in Drupal 8.  It would benefit no one to allow a broken system to run, so we will be disabling it at bootstrap with this patch once the 7.x branch is created.  Also, we cannot "simply" continue to support Drupal 8 in the 7.x branch; per our policy on backwards compatibility support, allowing Drush 7.0-stable to support Drupal 8.0.0-beta* implies that Drush 7 will continue to support Drupal 8, which we do not plan to do in the 7.x branch.

See: https://pantheon.io/blog/what-version-drush-works-drupal-8

ergonlogic:
I get that, but the plan is to keep Drush7 dev going in the master branch until there's an API change, right? So how are we even going to be able to continue working on Drupal 8 support in the mean time?

greg-1-anderson:
This issue is a placeholder; it's not going to be committed until we split off the 7.x branch -- that is, this PR is for the 7.x branch, not the master branch.

ergonlogic:
Ok, we also talked about starting Drush 8 alpha release as soon as Drush 7.0.0 is released, so adventurous Aegir3 users can use Drupal 8 betas, etc.

weitzman:
This was just merged into the new 7.x branch - 8ebe1e8

Issue #1249, opened 2015-03-12
segmentation fault when make with yaml 


lussoluca:
Using the latest drush 7 dev version I get a segmentation fault when I try to build a site with drush make. It seem that there is a problem with the yaml length because if I split the file in more included files all works correctly.

weitzman:
I think this has been reported before. If the error is in the parser (likely), you would want to ask the Symfony folks about this. Thats who makes our YAML component.

lussoluca:
Hi Moshe and thanks for your reply,
I've done a little investigation and the problem is in the PCRE
implementation: in /commands/make/make.utilities.inc::571 we try to parse
the YAML file as INI and in /lib/Drush/Make/Parser/ParserIni.php::41 the
preg_match_all fails to parse big YAML file and throws the "Segmentation
fault" exception. Can we switch the test in _make_determine_format (YAML
first and then INI)? Or we need a clever way to determine if a file is in
YAML or INI format.

Thanks

Luca

On Thu, Mar 12, 2015 at 6:39 PM, Moshe Weitzman <notifications@github.com>
wrote:

> I think this has been reported before. If the error is in the parser
> (likely), you would want to ask the Symfony folks about this. Thats who
> makes our YAML component.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/drush-ops/drush/issues/1249#issuecomment-78542082>.
>



-- 
Lusso Luca

weitzman:
See #1386

PR #1386, opened 2015-05-14
Issue #1249: Test for YAML first to avoid segfault.

jhedstrom:
- On large files, preg_match_all can segfault. To avoid this, test for
  YAML first.

### DIFF
--- a/commands/make/make.utilities.inc
+++ b/commands/make/make.utilities.inc
@@ -608,13 +608,27 @@ function _get_working_copy_option($download) {
  */
 function _make_determine_format($data) {
   // Test YAML first to avoid segmentation faults from preg_match in INI parser.
-  if ($parsed = ParserYaml::parse($data)) {
-    $parsed['format'] = 'yaml';
-    return $parsed;
+  $yaml_parse_exception = FALSE;
+  try {
+    if ($parsed = ParserYaml::parse($data)) {
+      $parsed['format'] = 'yaml';
+      return $parsed;
+    }
   }
-  elseif ($parsed = ParserIni::parse($data)) {
+  catch (\Symfony\Component\Yaml\Exception\ParseException $e) {
+    // Note that an exception was thrown, and display after .ini parsing.
+    $yaml_parse_exception = $e;
+  }
+
+  // Try INI format.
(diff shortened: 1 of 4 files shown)


weitzman:
Looks like tests failed here. ping @jhedstrom

jhedstrom:
Hmm, so the parser throws an exception. If we catch and ignore, folks trying to use YAML won't get proper feedback on invalid YAML.

weitzman:
I am seeing multiple reports of this. Would be good to find a resolution.

jonhattan:
In #1479 I proposed to enforce using `---` to identify yaml makefiles.

jhedstrom:
This change wraps the YAML parsing in a try/catch.

@jonhattan if we go the `---` route, would we just check for that line at the beginning of a file? If this try/catch bit doesn't work I'd almost prefer to go back to the original way this worked, which was simply to check for the `.yml` file ending.

jhedstrom:
Oh, I forgot about the stdin detection which removed the requirement for file ending.

jhedstrom:
This is now green, any thoughts before it gets merged?

jonhattan:
I had another idea: any makefile must have `core` and `api` keys. We could grep one of those and check if it follows ini or yaml syntax. Expressed in bash:

```
CORE=$(grep ^core tests/makefiles/defaults.make.yml | tr -d '[[:space:]]')
[[ $CORE  == "core:"* ]]  # yaml (starts with core:)
```

jhedstrom:
I like that idea, but will it work with `stdin`?

jonhattan:
yes, [_make_parse_info_file()](http://api.drush.org/api/drush/commands!make!make.utilities.inc/function/_make_parse_info_file/master) obtains the makefile contents (file, stdin, remote..) via `make_get_data()`, and then calls `make_determine_format()`.

jhedstrom:
@jonhattan just as I started to make this change, I realized that *included* make files need not repeat the  core attribute. I will add this check anyway, then fallback to the try/catch logic in the current PR.

jhedstrom:
This adds the check for the core attribute, and a few tests.

weitzman:
Looks fine to me.

Issue #1226, opened 2015-03-04
Drush is converting Cygwin paths unnecessarily. If Cygwin PHP is being used, paths *should not* be converted.


anandpathaksharma:
## Summary
I've found an issue related to this on Stack Exchange > Drupal Answers: [Cygwin Drush throws error in Drupal folders](http://drupal.stackexchange.com/questions/143707/cygwin-drush-throws-error-in-drupal-folders).

## Background
My setup includes Cygwin's PHP port for better integration - the Cygwin PHP executable understands Cygwin paths (unix style) and *doesn't need path conversion*.

I've found the helper function that is causing this problem. My Drush version is 6.5.0...

### Code
Here's the code (file: PATH_TO_DRUSH\includes\environment.inc:194):
````
/**
 * Converts a Windows path (dir1\dir2\dir3) into a Unix path (dir1/dir2/dir3).
 * Also converts a cygwin "drive emulation" path (/cygdrive/c/dir1) into a
 * proper drive path, still with Unix slashes (c:/dir1).
 */
function _drush_convert_path($path) {
  $path = str_replace('\\','/', $path);
  $path = preg_replace('/^\/cygdrive\/([A-Za-z])(.*)$/', '\1:\2', $path);

  return $path;
}
````

## Proposed Solution
I don't have the time right now to provide a solution. I envision it would entail updating the function above to detect the location of the PHP executable in use. Maybe we could set or access a Global variable that indicates the path style that the executing PHP requires. If the PHP executable is Cygwin or Msys/MingW and understands Unix style paths, leave return the path as is. If the PHP executable was compiled natively for Windows and requires Windows style paths, then move forward with the conversion.

## Immediate Workaround
For those users who find themselves in my shoes, please open up the file and comment out the lines of code that perform the string/preg replacing... or wrap them in an appropriate condition.

### Code
````
/**
 * Converts a Windows path (dir1\dir2\dir3) into a Unix path (dir1/dir2/dir3).
 * Also converts a cygwin "drive emulation" path (/cygdrive/c/dir1) into a
 * proper drive path, still with Unix slashes (c:/dir1).
 */
function _drush_convert_path($path) {
  //$path = str_replace('\\','/', $path);
  //$path = preg_replace('/^\/cygdrive\/([A-Za-z])(.*)$/', '\1:\2', $path);

  return $path;
}
````

greg-1-anderson:
[...]

PR #1391, opened 2015-05-17
fix for use under cygwin's php

diddledan:
fixes issue #1226 
simple change to `_drush_convert_path()` function in `includes/environment.inc` to filter out alteration of `/cygdrive/` paths when the php is cygwin's php - I've not tested under cygwin with the standard non-cygwin-based php so this change may have broken that scenario.


weitzman:
merged to master and 7.x. thanks.

PR #1399, opened 2015-05-22
Add option to colorize Drush PS1 with DRUSH_PS1_SHOWCOLORHINTS

TravisCarden:
Git has a nice option to colorize your PS1 Git string using a `GIT_PS1_SHOWCOLORHINTS` environment variable. I would like to be able to color my PS1 Drush string as well, to give a visual indication of the environment my active Drush alias points to. This pull request accomplishes just that. Following the instructions at the top of the new `examples/example.prompt.sh` results in the prompt depicted below, for example:

drush-ps1

### DIFF
--- a/drush.complete.sh
+++ b/drush.complete.sh
@@ -24,6 +24,13 @@ __drush_ps1() {
     __DRUPAL_SITE="$DRUPAL_SITE"
   fi
 
+  # Set DRUSH_PS1_SHOWCOLORHINTS to a non-empty value and define a
+  # __drush_ps1_colorize_alias() function for color hints in your Drush PS1
+  # prompt. See example.prompt.sh for an example implementation.
+  if [ -n "${__DRUPAL_SITE-}" ] && [ -n "${DRUSH_PS1_SHOWCOLORHINTS-}" ]; then
+    __drush_ps1_colorize_alias
+  fi
+
   [[ -n "$__DRUPAL_SITE" ]] && printf "${1:- (%s)}" "$__DRUPAL_SITE"
 }
 
--- a/examples/example.prompt.sh
+++ b/examples/example.prompt.sh
@@ -1,0 +1,80 @@
+# -*- mode: shell-script; mode: flyspell-prog; ispell-local-dictionary: "american" -*-
+#
+# Example PS1 prompt.
+#
+# Example - rename this to .drush_prompt, and in your ~/.bash_profile add:
+#
+#    if [ -f ~/.drush_prompt ] ; then
+#        . ~/.drush_prompt
+#    fi
+#
+# Features:
+#
+# Displays Git repository and Drush alias status in your prompt.
+
+if [ "\$(type -t __git_ps1)" ] && [ "\$(type -t __drush_ps1)" ]; then
+
+  # This line enables color hints in your Drush prompt. Modify the below
+  # __drush_ps1_colorize_alias() to customize your color theme.
+  DRUSH_PS1_SHOWCOLORHINTS=true
+
+  # Git offers various prompt customization options as well as seen in
+  # https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh.
+  # Uncomment any or all of the following lines to enable the corresponding
+  # features:
+  #
+  # GIT_PS1_SHOWDIRTYSTATE=true
+  # GIT_PS1_SHOWSTASHSTATE=true
+  # GIT_PS1_SHOWUNTRACKEDFILES=true
+  # GIT_PS1_SHOWUPSTREAM=auto
+  # GIT_PS1_SHOWCOLORHINTS=true
+
+  # The following line sets your bash prompt according to this example:
+  #
+  #   username@hostname ~/working-directory (git-branch)[@drush-alias] $
+  #
+  # See http://ss64.com/bash/syntax-prompt.html for customization options.
+  export PROMPT_COMMAND='__git_ps1 "\u@\h \w" "$(__drush_ps1 "[%s]") \\\$ "'
+
+  # PROMPT_COMMAND is used in the example above rather than PS1 because neither
+  # Git nor Drush color hints are compatible with PS1. If you don't want color
+  # hints, however, and prefer to use PS1, you can still do so by commenting out
+  # the PROMPT_COMMAND line above and uncommenting the PS1 line below:
+  #
+  # export PS1='\u@\h \w$(__git_ps1 " (%s)")$(__drush_ps1 "[%s]")\$ '
+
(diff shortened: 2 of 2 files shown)


drush.complete.sh
```diff
@@ -24,9 +24,40 @@ __drush_ps1() {
...
+# Helper function meant to be called from __drush_ps1.
+__drush_ps1_colorize() {
+  if [[ -n ${ZSH_VERSION-} ]]; then
+    local red='%F{red}'
+    local green='%F{green}'
+    local blue='%F{blue}'
+    local clear='%f'
+  else
+    # Using \[ and \] around colors is necessary to prevent issues with command
+    # line editing/browsing/completion!
```
TravisCarden: I stole these color definitions from [git/git-prompt.sh](https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh), along with this comment, and I'm just taking their word for it.

drush.complete.sh
```diff
@@ -24,9 +24,41 @@ __drush_ps1() {
...
 
+__drush_ps1_colorize_alias() {
+  if [[ -n ${ZSH_VERSION-} ]]; then
+    local COLOR_RED='%F{red}'
+    local COLOR_GREEN='%F{green}'
+    local COLOR_BLUE='%F{blue}'
+    local COLOR_CLEAR='%f'
+  else
+    # Using \[ and \] around colors is necessary to prevent issues with command
+    # line editing/browsing/completion.
```
TravisCarden: I took these color codes from [git/git-prompt.sh](https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh), and I'm taking their word for it on this comment.

greg-1-anderson:
I'm not opposed to colorized prompts in principal; however, the problem here is that the names that folks give their environments, and the colors that they want to use to display each environment in are matters of individual preference.  Providing a single on/off switch therefore seems insufficient, and adding an entire color preference scheme seems too complicated for drush.complete.sh.

Perhaps this needs to be an "examples" file instead?

TravisCarden:
I agree with your assessment, @greg-1-anderson. I have definitely made assumptions about both conventions and preferences here. I initially just set out to do this in my own `.bashrc` file before deciding it would be a shame not to make it easy for others to enjoy. Do you think we could find a good compromise between simplicity and flexibility? Perhaps, for example, Drush could include a `__drush_ps1_colorize_alias()` that works well with the most common setups (the one I provided is compatible with Acquia Hosting alias-naming conventions) but allow an alternative function to be specified instead--say with a  `DRUSH_PS1_COLORHINTSFUNCTION` variable. Then there would be simple, out-of-the-box functionality for the 90% use case that's easy to override for edge cases and power users. What's more, the included `__drush_ps1_colorize_alias()` would serve as built-in documentation for how to write your own implementation.

greg-1-anderson:
Note that Pantheon uses "@site.live", "@site.test" and "@site.dev", platform.sh uses "@master", "@local" and "@NAMEYOUROWNBRANCH", and folks on Digital Ocean & c. can name their aliases however they'd like.  That's too much complexity to pile into drush.complete.sh.

Drush could certainly provide the examples you suggest, but it should do so in the "examples" folder, not in drush.complete.sh.

Note that example.bashrc looks for drush.complete.sh, and sources it if it finds it.  This file could be changed to also look for a "colorized" version of drush.complete.sh, which could be provided in the "examples" folder.  Folks could then copy and customize it to their liking, and example.bashrc (which is designed to allow the user to source it in-place) would pick it up if it was copied to the correct location.

greg-1-anderson:
GitHub isn't great for collaborating on PRs, so I'm going to do this the lazy way.  You can support multiple names for the same environment types like this:
```
  case "$__DRUPAL_SITE" in
    *live|*prod) local ENV_COLOR="$COLOR_RED" ;;
    *local) local ENV_COLOR="$COLOR_GREEN" ;;
    *) local ENV_COLOR="$COLOR_BLUE" ;;
  esac
```
I would use different colors, myself, but I will avoid bikeshedding this PR, and instead mention that if three colors are not enough, there are [a lot of bash colors available](http://misc.flogisoft.com/bash/tip_colors_and_formatting), and [zsh has a good selection as well](https://gabri.me/blog/custom-colors-in-your-zsh-prompt).

TravisCarden:
Thanks, @greg-1-anderson! How about this? (PR updated.)
- I moved `__drush_ps1_colorize_alias()` into example.bashrc.
- I enhanced the environment patterns per your suggestion and added one for test/stage colored yellow.
- I added several more color variables.

greg-1-anderson:
This PR is designed to be customized by the user, but example.bashrc is designed so that it can be ran in place.  Also, does example.bashrc happen to work in zsh today?  It is not advertised to be zsh-compatible, so I don't know that it makes sense to define zsh colors here.

I think it might make more sense to put the colorized example in a separate file; that way, it could be compatible with zsh and bash, and could be designed to be highly customized by the user.  If this stayed in example.bashrc, then we should probably drop zsh support, and add a way for the user to customize the color prompts without modifying the example script (e.g. with environment variables, or by including another file, etc.).  The separate file seems cleaner to me.

TravisCarden:
I think I understand you better now, @greg-1-anderson. How about this? I moved the whole thing into a new `examples/example.prompt.sh` file patterned after `examples/example.bashrc`. I added instructions for sourcing it and customizing it.

TravisCarden:
Okay, @greg-1-anderson. This has been rebased to the current `master`, and the tests pass again. If you require further adjustments before merging, I'm ready to address them.

greg-1-anderson:
@weitzman What do you think?  Do you want this in Drush core?

examples/example.prompt.sh
```diff
@@ -0,0 +1,81 @@
...
+# Features:
+#
+# Displays Git repository and Drush alias status in your prompt.
+
+if [ "\$(type -t __git_ps1)" ] && [ "\$(type -t __drush_ps1)" ]; then
+
+  # Uncomment the following line to enable color hints in your Drush prompt.
+  # Modify __drush_ps1_colorize_alias() to customize your color theme.
+  #
+  # DRUSH_PS1_SHOWCOLORHINTS=true
```
weitzman: Dont think this needs to be commented out. I would expect this file to work if I deliberately include it from .bashrc.

weitzman:
yes, i think this is good for core as an example. nicely done.

TravisCarden:
Awesome! Thanks, @weitzman. I've addressed your comment.

weitzman:
@greg-1-anderson is welcome to merge this when satisfied. Thanks @TravisCarden

greg-1-anderson:
Looks great -- thanks!

PR #1405, opened 2015-05-26
Do not check memory limit if it is set to unlimited

Berdir:
If you have the memory limit set to unlimited then there's a pretty confusing message about memory usage. This should avoid that.

batch_7.inc probably needs the same fix.


weitzman:
cherry picked onto master and 7.x. thanks.

PR #1406, opened 2015-05-26
Load install file before running update function

Berdir:
I was testing custom updates and they just silently didn't do anything.

I don't know why exactly but it looks like the .install file is (no longer?) included, so those functions weren't found and silently ignored.

I made sure that the necessary .install file is loaded and also added an error if a function can't be found.. it's not something that should ever happen (with my fix..), but if it does, then there is at least an error.


Berdir:
It looks like https://github.com/drush-ops/drush/pull/1223 removed the code to load the install files, I guess it still works in 7.x but not 8.x...

PR #1412, opened 2015-05-28
Allow config-merge --git to operate on commits already in the repository.

greg-1-anderson:
If you do not include a target site argument, then config-merge will assume that the configuration changes that you want to merge with have already been committed to the git repository.  You can include --branch=x if the configuration was committed to some other branch 'x'.

This PR also handles a common development workflow, where multiple developers are all working on the "master" branch, with each developer pushing new commits to master whenever they are ready.  This means that, at config-merge time, a given developer might be both ahead and behind of the master branch.  To allow the merge to happen with a single invocation of the merge tool, we move the local commits that have not yet been pushed to a temporary working branch, and then reset the master branch back to the last commit that is already on the central repository.  This allows the pull from master to happen without an error.  All of the commits end up back on the master branch at the end, after we merge the branches together and run the merge tool (if necessary).

If something goes wrong in the command, we are careful to roll back the master branch to the state it was in before we moved the un-pushed commits to another branch.

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -164,7 +164,6 @@ function config_drush_command() {
   $items['config-merge'] = array(
     'description' => 'Merge configuration data from two sites.',
     'aliases' => array('cm'),
-    'required-arguments' => 1,
     'arguments' => array(
       'site' => 'Alias for the site containing the other configuration data to merge.',
       'label' => "A config directory label (i.e. a key in \$config_directories array in settings.php). Defaults to 'staging'",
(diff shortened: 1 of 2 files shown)


weitzman:
I'm finding it hard to give meaningful reviews on config-merge. It is too complex for me. I am wondering if it is time to move it to contrib. It can be a suggested package in our composer.json if you want.

greg-1-anderson:
drush-ops/contrib-merge?

weitzman:
Think you mean config-merge. That works for me. If you are feeling big, you could do config-extra or something.

greg-1-anderson:
Yep, fixed my typo (was: contrib-merge, oops).

Overall, I'd rather just keep it in Drush core, but since one particular three-line section in this PR required four diagrams and a few paragraphs of explanation, I can't argue against the point that it's complicated.  Simple git workflows can arise to complicated requirements (e.g. moving some commits to a branch).

I can go with config-extra.  I'll set up the repository, and make a PR to remove config-merge from Drush core.

PR #1413, opened 2015-05-29
Automatically set prompt to the alias record for the current site

greg-1-anderson:
There are ~~two~~ *three* prerequisites for this patch:

1. The user must source examples/example.bashrc
2. The user must also define `alias cd=cddl`
3. *The user must use a PS1 that contains $(__drush_ps1)*

If ~~both~~ *all three* of these are true, then with this PR, any time the user 'cd's to a Drupal root that has a matching site alias, their prompt will switch to contain the name of the alias to the current site.

There is also a change in behavior here.  If the user 'cd's to a directory that does not contain a Drupal site, then Drush will automatically `use @none`.  This insures that your prompt matches your cwd; however, without this PR, you could formerly `use @alias`, and have your site selection persist when 'cd'ing to different non-Drupal directories.

It would be easy enough to change it around, so the current behavior (sticky 'use') is maintained; however, I suspect that 'auto' + 'sticky' might be a little annoying.  I don't use `use` much, so I'm not sure, but I like the auto-prompt switching feature.

Auto-prompt switching is particularly useful when using #1399.  If you are going to use color as an indicator of what type of site you're currently operating on, it's handy if the prompt picks up available site aliases automatically when switching sites via 'cd' instead of 'use'.

### DIFF
--- a/commands/core/sitealias.drush.inc
+++ b/commands/core/sitealias.drush.inc
@@ -348,6 +348,7 @@ function drush_sitealias_site_set_validate() {
  */
 function drush_sitealias_site_set($site = '@none') {
   if ($filename = drush_sitealias_get_envar_filename()) {
+    $current = is_file($filename) ? trim(file_get_contents($filename)) : "@none";
     $last_site_filename = drush_sitealias_get_envar_filename('drush-drupal-prev-site-');
     if ($site == '-') {
       if (file_exists($last_site_filename)) {
@@ -356,6 +357,19 @@ function drush_sitealias_site_set($site = '@none') {
       else {
         $site = '@none';
       }
+    }
+    if ($site == '@self') {
+      $path = drush_cwd();
+      $site_record = drush_sitealias_lookup_alias_by_path($path, TRUE);
+      if (isset($site_record['#name'])) {
+        $site = '@' . $site_record['#name'];
+      }
+      else {
+        $site = '@none';
+      }
+    }
+    if ($current == $site) {
+      return;
     }
     if (_drush_sitealias_set_context_by_name($site)) {
       if (file_exists($filename)) {
@@ -375,7 +389,7 @@ function drush_sitealias_site_set($site = '@none') {
       }
     }
     else {
-      drush_set_error('DRUPAL_SITE_NOT_FOUND', dt("Could not find a site definition for !site.", array('!site' => $site)));
+      return drush_set_error('DRUPAL_SITE_NOT_FOUND', dt("Could not find a site definition for !site.", array('!site' => $site)));
     }
   }
 }
--- a/commands/core/sitealias.drush.inc
+++ b/commands/core/sitealias.drush.inc
@@ -348,7 +348,7 @@ function drush_sitealias_site_set_validate() {
  */
 function drush_sitealias_site_set($site = '@none') {
   if ($filename = drush_sitealias_get_envar_filename()) {
-    $current = is_file($filename) ? trim(file_get_contents($filename)) : "@none";
+    $current = drush_sitealias_site_get() ?: "@none";
     $last_site_filename = drush_sitealias_get_envar_filename('drush-drupal-prev-site-');
     if ($site == '-') {
       if (file_exists($last_site_filename)) {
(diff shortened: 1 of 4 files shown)


weitzman:
This would get slow with hosts that have lots and lots of aliases, and have not yet built up their alias cache. If the slowness is limited to just that, then this is OK. If we slow down cd in any wider way, I cant see anyone using this.

greg-1-anderson:
This PR only affects the performance of the 'cddl' alias, so it will only affect those who alias 'cd' to 'cddl' and have a lot of individual alias files (esp. those who use a ton of individual 'site.alias.drushrc.php' files, rather than grouping aliases into single 'group.aliases.drushrc.php files).

There are probably only a very few users who fall into this category.  For their benefit, I have factored out a 'fastcddl' shell script function that behaves the same way that the old cddl did.  There should be no performance impact of using this function, so those who do not like the 'use' can use the faster function.

weitzman:
I tried to get this working on my machine, but it just wont show the alias automatically. I have a feeling that the problem is my environment. I did alias cd to cddl ... Will try again another day. Would be great if others would verify the PR.

greg-1-anderson:
Did you source your patched example.bashrc file in the terminal you were using?

I'll re-test the PR and see if it's still okay.

greg-1-anderson:
Just tried this again on MacOS X 10.10; expect it should work the same on other bash platforms as well, but have not tried.

Steps to verify (replace '@local' with a site alias on your system that points at a valid local Drupal site):

1. Check out this PR (branch: auto-set-prompt)
2. Make sure that the `drush` with this PR appears in your $PATH, and is the first drush that appears.
3. `source /path/to/drush/examples/example.bashrc`
4. `cd @local`
5. `cd ..`
6. `cd -`

After step 4, your prompt should contain '@local'.  After step 5, your prompt should not contain any Drupal site.  After step 6, your prompt should contain '@local' again.

If this still isn't working, `echo $PS1` and check to see if it contains `$(__drush_ps1)`.  Perhaps this PR is mis-named, as PS1 must be manually set by the user, or nothing will happen.  PS1 has always been considered a very personal thing, so Drush does not ever try to set it.  It provides `$(__drush_ps1)`, though; I guess I assume that all Drush users put this in their prompt.

This is how I set my prompt in my .bashrc.php:

```
if [ "`type __drush_ps1 2> /dev/null`" == "" ] ; then
  __drush_ps1() {
    return 0
  }
fi

if [ "x$TERM" != "xcygwin" ] ; then
  HOSTNAME=$(hostname -s)
  if [ "$HOSTNAME" == "vps" ] || [ "$HOSTNAME" == "server" ] ; then
    HOSTNAME=$(hostname -f | sed -e 's/^[^.]*\.//' -e 's/\..*//')
  fi
  PROMPT_COMMAND='_p=$(__prompt_info)'
  PS1='${_p:0:1}\[\e[$(__prompt_color)\]${_p:1:$((${#_p}-3))}\[\e[$c_reset\]${_p#"${_p%??}"}\[\e[$c_blue\]\u@'"$HOSTNAME"':\w\[\e[$c_reset\]\[\e[$c_green\]$(__drush_ps1)\[\e[$c_reset\]$ '
fi
``

weitzman:
Thanks for writing up those directions. I went through them, and the debugging steps, but still nothing. My prompt says `\w$(__git_ps1 " (%s)")$(__drush_ps1 "[%s]")\$`. I am pointing to the right Drush in my path, I have aliased cd=cddl. Still no dice. 

It would be good if someone else could verify that this works.

greg-1-anderson:
If you type `use @local` (with or without this PR), for any alias @local that points at a valid Drupal site that exists on your local system, does your prompt change?

If so, can you reset with `use @none`, then cd to the Drupal root for your site and type `use @self` to set the prompt?  (With this PR only.)

weitzman:
Yes, `use @local` does change my prompt. When I reset, cd to that drupal root, and then `use @self`, I don't see any prompt change. FYI, here is my alias:

    $aliases["d7"] = array (
      'root' => '/Users/moshe.weitzman/htd/d7',
    );

i do have cd=cddl bash alias.

greg-1-anderson:
Got it.  You need a 'uri' for this to work.  Try adding `'uri' => 'default'` in your alias.

That is, of course, a bug in my code.  I've reproduced your situation, and will fix it now.

weitzman:
It is working now. Feel free to merge when you are satisfied.

greg-1-anderson:
Fixed for aliases with no 'uri' and merged with master.

greg-1-anderson:
cc @TravisCarden

PR #1418, opened 2015-06-01
Also support objects in the YAML formatter.

### LINKED ISSUE #1417: drush vget does not output objects in its default (YAML) format
When a variable contains an object it is not output using the YAML formatter. Instead the user is told the variable contains NULL:

```
$value = (object) array('test' => TRUE);
variable_set('my_variable', $value);
```

```
$ drush vget my_variable
my_variable: null
```

Since the YAML formatter is the default this is perhaps a major bug.

pfrenssen:
Set the `objectSupport` flag to TRUE. Fixes #1417.


weitzman:
Nice find. Committed to master, 7.x, 6.x.

PR #1419, opened 2015-06-01
Add a UNISH_INCLUDE environment variable

greg-1-anderson:
This allows Drush extensions to use Unish functional tests to test their commands.

Already using this in [drush-ops/config-extra](https://github.com/drush-ops/config-extra/blob/master/runtests.sh), but let me know if you know of a better way to do this.  Without adding --include to `drush` callbacks, Drush will not be able to find the commandfiles it is supposed to be testing.

If this PR is okay, I can set up travis to run the config-extra functional tests on every commit once this is available.

### DIFF
--- a/tests/Unish/CommandUnishTestCase.php
+++ b/tests/Unish/CommandUnishTestCase.php
@@ -230,6 +230,13 @@ abstract class CommandUnishTestCase extends UnishTestCase {
     $hide_stderr = FALSE;
     $cmd[] = UNISH_DRUSH;
 
+    // If the UNISH_INCLUDE environment variable is set, then include  the specified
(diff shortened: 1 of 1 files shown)


weitzman:
Not sure it helps, but the test runner for Devel project is at http://cgit.drupalcode.org/devel/tree/run-tests-drush.sh. You can see that in the absence of arguments, it runs all the tests in /drush. You can see the test class at http://cgit.drupalcode.org/devel/tree/drush

weitzman:
So in the context of a global commandfile like config-extra, you could symlink it into unish/drush-sandbox instead of into the drupal site like devel does.

greg-1-anderson:
Yeah, something like that will work just as well.

PR #1420, opened 2015-06-01
Remove config-merge command

greg-1-anderson:
It is moving to https://github.com/drush-ops/config-extra

This PR removes the config-merge command itself, it's 'topic', and its functional tests.


weitzman:
Yeah, I think this is a good move. Feel free to add config-extra to our composer.json

PR #1425, opened 2015-06-02
Test after_success hook with 'secure' environment variable.  

greg-1-anderson:
Can we expose secrets with 'echo'?


greg-1-anderson:
Closed in favor of #1426 (source of the PR is important -- secure environment variables are not set for submissions from outside repositories).

PR #1426, opened 2015-06-02
Test children (dependent) repositories

greg-1-anderson:
This test uses a Travis secure environment variable to hold an "auth token" that will allow us to tell Travis to rebuild a different child repository.

The tokens in this PR are test values only.  Secure variables are not sent to tests started with pull requests -- a good security feature.


greg-1-anderson:
Okay, this PR is now working.  When the configTest passed, it restarted the most recent build in drush-ops/config-extra.  The result can be seen at the bottom of the log file:

https://api.travis-ci.org/jobs/65165639/log.txt?deansi=true

The credential used is for my Travis account, which mirrors my GitHub authorizations.  With this token, then, we could re-test any "child" repository in Drush-ops.

As explained above, Travis keeps the token secure by not setting it for any PR that originated from another repository.  We'd just have to be careful not to commit any PR that, as a side effect echo'ed $AUTH_TOKEN.  I don't think that the Travis auth token can be used to do anything on GitHub, though, so I think this is an acceptable risk.

To make this work, we need to add another key to env, "matrix".  This is the default for "env" -- the matrix is the list of all environment variables that define the builds that will be run.  The other kind of entry that can go into env is "global"; these environment variables are added to each entry in the "matrix".

The "after success" rule runs after every test in the matrix, so I defined the TEST_CHILDREN environment variable to define which child repositories should be retested after (if) the given matrix test succeeds.  For this PR, I retest drush-ops/config-extra if the configTest passes.  Currently, configTest also runs as part of the last test group; we could either make a new group just for the config tests, or we could just move TEST_CHILDREN="drush-ops/config-extra" to be attached to the last group of tests.  TEST_CHILDREN can contain more than one repository (space-separated), as long all of them are in a repository that our AUTH_TOKEN is authorized to update (drush-ops).

So, this is not perfect, but there's no way to put any configuration in config-extra to rebuild itself when drush-ops/drush changes, so this push-notification is the best we can do for now.  I think this level of testing will be really helpful to catch Drush changes that break (require updates to) config-extra, even if such failures are not reflected in the Drush test logs.

"children.sh" can be named something else, and placed somewhere else, maybe hidden somewhere in the "tests" folder.  Let me know if you have any opinions on the cosmetic parts of this.

greg-1-anderson:
Just for record-keeping, this is derived from https://github.com/ajdm/travis-build-children

weitzman:
Looks good. Note that the PR is adding in some indentation in .travis.yml file which I recently took out. Our commits crossed paths.

greg-1-anderson:
The indentation in this PR is for a different purpose.  I added "matrix" to env, which required another necessary level of indentation.  Just a coincidence that you recently removed unnecessary indentation.

I'll remove the redundant test group, move the TEST_CHILDREN to the last d8 test, move the children.sh script into 'tests' as test_children.sh, and then squash-commit.  Let me know if you want anything different than that or anything additional.

.travis.yml
```diff
@@ -15,33 +15,38 @@ php:
 sudo: false
 
 env:
+  matrix:
+    - UNISH_NO_TIMEOUTS=y UNISH_DRUPAL_MAJOR_VERSION=8 UNISH_DB_URL=mysql://root:@127.0.0.1 PHPUNIT_ARGS=tests/configTest.php TEST_CHILDREN="drush-ops/config-extra"
```
weitzman: What exactly does this line do? Most of the params are repeated below. Is this effectively a way to get TEST_CHILDREN added to each line below? If so, could we not do same for UNISH_TIMEOUTS? I have tried to use matrix before but always been stymied by https://github.com/travis-ci/travis-ci/issues/1519

greg-1-anderson:
As you mention (https://github.com/travis-ci/travis-ci/issues/1519), two-dimensional (and greater) matrixes do not work.  One-dimensional matrixes work just fine, though.  Why have a one-dimensional matrix, when they behave the same as an "env" section without a "matrix"?  The reason is that if you put all of your env values under "matrix", then you can add additional env values under a "global" section.  Everything in "global" is added to every environment setting in "matrix" for every test.  I tested this successfully previously with multiple "secure" values; the latest commit that I added factors out "UNISH_TIMEOUTS" and "UNISH_DB_URL" to the "globals" section.  If we ever wanted to turn back on the mysql and postgres tests, we'd have to move UNISH_DB_URL back out of the "globals" section, but it should work like this for now.

Travis tests are temporarily stalled.  We'll see if this works right once they start up again.

weitzman:
Looks good to me. Feel free to squash+merge once Travis recovers.

greg-1-anderson:
Merged via 63d901fab

PR #1430, opened 2015-06-05
Add a storage wrapper / storage filter to Drush config import and export

greg-1-anderson:
**Goal**

Keep "development" modules such as "devel" enabled on the development machine, but disabled on the production machine.  It should be possible to store configuration in git, and make configuration changes on both dev and prod, and merge them via config-merge without causing repetitious conflicts between the dev modules' state, and the configuration state stored in the repository.

**Why This is Hard**

The Drupal 8 configuration management system defines "staging" and "active" configurations, and allows us to make more if we need them.  Couldn't we just make a "development" configuration, and then, in a post-config-import hook, copy "staging" to "development", apply the development module configuration changes, and run config-import on the development configuration set?

Besides the minor issue of recursion, which could be worked out, the larger issue that we have is that the development configuration changes are now in the active configuration in the database; if we make other config changes, and want to export those to commit them to the git repository, then the config changes we want are mixed in with the config changes automatically applied to enable dev modules.  This is not desirable.

**What About settings.php?**

If you want to change a simple configuration value, and have it active only on a dev machine, you can easily add configuration overrides to settings.php, like this:

`$config['system.site']['name'] = 'Local Install of Awesome Widgets, Inc.';`

This works for most kinds of configuration settings that you'd like to adjust on the dev site only; unfortunately, core.extension cannot be altered like this, as enabling / disabling modules is done in hooks that run at value import / set time, and settings.php only allows values to be overridden.

(If someone can get settings.php config overrides to work with modules, that would be a cool alternative to this, but I don't think it's possible.)

**Proposed Solution**

This PR adds "StorageWrapper" and "StorageFilter" classes, and hooks them in to config-merge and config-export.  A "StorageWrapper" is simply a wrapper class for any StorageInterface object.  When a method of StorageWrapper is called, it simply calls through to its delegate storage object.  In the case of reading and writing configuration, it allows the "StorageFilter" object to adjust the configuration values before it is read or written.  To keep this class simple, it does not allow configuration files or collections to be added or removed; only the values they store can be adjusted.

Drush config-import and config-export will apply a StorageWrapper to the FileStorage object used to read or write (respectively) configuration to the file system, and filter them with any active StorageFilter objects.  The configuration is also filtered as necessary prior to being used by the StorageComparer, so that the user gets accurate reports  

The proposal on the table is to put at least this much into Drush core.

**Interim Implementation**

Just having the ability to apply filters is not enough, of course; it is also necessary to use the filters to do something.  As previously mentioned, configuration overrides in settings.php cover most of the desired use cases, so this PR provides just one simple filter designed only to enable and disable modules.

This one filter is currently hard-coded into config-import / config-export, but the intention is to hook it.  Also, the exact UX for exposing this feature is still preliminary; feedback on options and usage appreciated.

Finally, we also need to decide on packaging.  This could conceivably be placed in the "examples" folder, like the sync-enable policy, or maybe the implementation goes in config-extra.

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -106,7 +106,7 @@ function config_drush_command() {
         'example-value' => 'branchname',
       ),
       'destination' => 'An arbitrary directory that should receive the exported files. An alternative to label argument',
-      'module-adjustments' => 'A list of modules to always enable.',
+      'ignored-modules' => 'A list of modules to ignore during export (e.g. to avoid listing dev-only modules in exported configuration).',
     ),
   );
 
@@ -122,7 +122,7 @@ function config_drush_command() {
       ),
       'source' => 'An arbitrary directory that holds the configuration files. An alternative to label argument',
       'partial' => 'Allows for partial config imports from the source directory. Only updates and new configs will be processed with this flag (missing configs will not be deleted).',
-      'module-adjustments' => 'A list of modules to always enable.',
+      'ignored-modules' => 'A list of modules to ignore during import (e.g. to avoid disabling dev-only modules that are not enabled in the imported configuration).',
     ),
     'core' => array('8+'),
     'aliases' => array('cim'),
@@ -500,10 +500,10 @@ function drush_config_import_validate() {
 
 function _drush_config_get_storage_filters() {
   $result = array();
-  $module_adjustments = drush_get_option('module-adjustments');
+  $module_adjustments = drush_get_option('ignored-modules');
   if (!empty($module_adjustments)) {
     if (is_string($module_adjustments)) {
-      $module_adjustments = array_flip(explode(',', $module_adjustments));
+      $module_adjustments = explode(',', $module_adjustments);
     }
     $result[] = new CoreExtensionFilter($module_adjustments);
   }
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -8,6 +8,8 @@
 use Drupal\Core\Config\StorageComparer;
 use Drupal\Core\Config\ConfigImporter;
 use Drupal\Core\Config\FileStorage;
+use Drush\Config\StorageWrapper;
+use Drush\Config\CoreExtensionFilter;
 use Symfony\Component\Yaml\Parser;
 
 /**
@@ -104,6 +106,7 @@ function config_drush_command() {
         'example-value' => 'branchname',
       ),
       'destination' => 'An arbitrary directory that should receive the exported files. An alternative to label argument',
+      'module-adjustments' => 'A list of modules to always enable.',
     ),
   );
 
@@ -119,6 +122,7 @@ function config_drush_command() {
       ),
       'source' => 'An arbitrary directory that holds the configuration files. An alternative to label argument',
       'partial' => 'Allows for partial config imports from the source directory. Only updates and new configs will be processed with this flag (missing configs will not be deleted).',
+      'module-adjustments' => 'A list of modules to always enable.',
     ),
     'core' => array('8+'),
     'aliases' => array('cim'),
@@ -367,12 +371,30 @@ function _drush_config_export($destination, $destination_dir, $branch) {
   if (!$comment) {
     $comment = "Exported configuration.";
   }
+  $storage_filters = _drush_config_get_storage_filters();
   if (count(glob($destination_dir . '/*')) > 0) {
     // Retrieve a list of differences between the active and target configuration (if any).
     $target_storage = new FileStorage($destination_dir);
     /** @var \Drupal\Core\Config\StorageInterface $active_storage */
     $active_storage = Drupal::service('config.storage');
-    $config_comparer = new StorageComparer($active_storage, $target_storage, Drupal::service('config.manager'));
+    $comparison_source = $active_storage;
+
+    // If the output is being filtered, then write a temporary copy before doing
+    // any comparison.
+    if (!empty($storage_filters)) {
+      $tmpdir = drush_tempdir();
+      drush_copy_dir($destination_dir, $tmpdir, FILE_EXISTS_OVERWRITE);
+      $comparison_source = new FileStorage($tmpdir);
+      $comparison_source_filtered = new StorageWrapper($comparison_source, $storage_filters);
+      foreach ($active_storage->listAll() as $name) {
+        // Copy active storage to our temporary active store.
+        if ($existing = $active_storage->read($name)) {
+          $comparison_source_filtered->write($name, $existing);
+        }
+      }
+    }
+
+    $config_comparer = new StorageComparer($comparison_source, $target_storage, Drupal::service('config.manager'));
     if (!$config_comparer->createChangelist()->hasChanges()) {
       return drush_log(dt('There are no changes to export.'), 'ok');
     }
@@ -402,6 +424,10 @@ function _drush_config_export($destination, $destination_dir, $branch) {
   // Write all .yml files.
   $source_storage = Drupal::service('config.storage');
   $destination_storage = new FileStorage($destination_dir);
+  // If therea re any filters, then attach them to the destination storage
+  if (!empty($storage_filters)) {
+    $destination_storage = new StorageWrapper($destination_storage, $storage_filters);
+  }
   foreach ($source_storage->listAll() as $name) {
     $destination_storage->write($name, $source_storage->read($name));
   }
@@ -472,6 +498,18 @@ function drush_config_import_validate() {
   }
 }
 
+function _drush_config_get_storage_filters() {
+  $result = array();
+  $module_adjustments = drush_get_option('module-adjustments');
+  if (!empty($module_adjustments)) {
+    if (is_string($module_adjustments)) {
+      $module_adjustments = array_flip(explode(',', $module_adjustments));
+    }
+    $result[] = new CoreExtensionFilter($module_adjustments);
+  }
+  return $result;
+}
+
 /**
  * Command callback. Import from specified config directory (defaults to staging).
  */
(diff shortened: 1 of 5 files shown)


greg-1-anderson:
Oh, and one other thing -- this code worked for me when I tried it by hand, but the unit tests are failing.  I'm not sure what the problem is there yet, but I still wanted to get this proposal on the table so that it could be discussed.

weitzman:
I'd prefer if this "modules vary by environment" need is met outside of Drush. But I don't know of any existing solution for this. All I can think of is swapping the core config services (details TBD). Maybe @alexpott or @ Schnitzel can chime in.

lib/Drush/Config/CoreExtensionFilter.php
```diff
@@ -0,0 +1,82 @@
...
+    $this->adjustments = $adjustments;
+  }
+
+  public function filterRead($name, $data) {
+    if ($name != 'core.extension') {
+      return $data;
+    }
+    foreach($this->adjustments as $module => $weight) {
+      if (is_numeric($weight)) {
+        $data['module'][$module] = $weight;
```
yched: Ideally we'd only add the module if it is actually already enabled on the environment we're importing to - else the feature becomes "we will force-enable this list of development modules", rather than just "we won't disable them if they are present"

Not sure how easily that would be done. The method would need to receive the active store ?

greg-1-anderson: Yes, that's a good idea. If we require the module to be enabled already (or, conversely, require it to be DISABLED already, if it is enabled in prod), then we can drop the 'module weight' from the filter settings.  It should be a simple matter to access the active store to make this work right.

yched:
Not fully sure either where that code would better live - but the wrapper and filter classes are super helpful wherever they end up, Thanks a ton for jumping on this @greg-1-anderson :-)

It's indeed handy that drush cex/cim does all the instantiation for you based on a command option.
It also means that native / non-drush import/export won't have filtering logic, and would enable devel on prod if you use them.

I have a feeling that the same functionnality could be provided to all import/export (drush or not) if the same wrapper and filter classes were directly applied to an override of config.storage.staging defined in the local sites/default/services.yml, controlled by a list of dev modules defined in the "parameters" section ?

That's more setup work though. 

Also, if the wrapper and filter classes are not provided by drush, then I guess they are provided by a module - which would then be itself on the list of "development modules to leave out of exported config" :-). I don't think there's a risk of recursion there, though. Not sure if that module would then do the service override itself, or would just provide inactive code that you need to leverage in your custom settings/services (a bit like modules providing alternate cache backends, which stay inactive until you actually route some cache bins to them in your settings ?)

lib/Drush/Config/CoreExtensionFilter.php
yched: In fact, that affect the weights too - IMO they shouldn't be overwritten by those specified as part of the command options, but just kept as they are in the current config ?

--> you specify a list of modules you want to preserve, and they will be preserved, at their current weight, if they are enabled ?

greg-1-anderson:
The latest commit implements your suggestion -- the adjustments list becomes an "ignore" list, which simplifies things a bit.  The test is working for me locally now (fixed it), but is failing on Travis due to a syntax error on the main branch that shouldn't be in this PR.  I'm confused about that, but in any event, I fixed the typo on the master branch, so rebase with master if you have any problems with this locally.

I agree that it could work as described by @yched above. The module that causes import/export to ignore modules could indeed include itself in the list, and this wouldn't cause any problems.  I still tend to favor the in-tool implementation to altering the configuration manager configuration in Drupal, but welcome proposals that do it the other way if that's what folks prefer.

greg-1-anderson:
Oh, there is another problem with the "fix it in Drupal" (or otherwise outside of Drush) proposal.  Currently, Drush instantiates its own FileStorage and StorageComparer objects.  If you did implement a fancy in-Drupal solution, it would probably therefore be ignored by Drush config-export and config-import, which is the primary place folks are going to want this alteration to be active, in my mind.

This is not an outright showstopper, though; even without this PR, it would be a benefit if Drush called some Drupal API to set up the objects used during import and export.  I don't know if anything like that is currently available, though, so we might be looking at a multi-step process, to first enhance Drupal, then adjust Drush to use it, etc.; feedback from @alexpott or @Schnitzel or anyone else who might know how to do this more easily would be welcome.

Pragmatically, the in-tool hooks are less work to get started with.  If Drush core provided the StorageWrapper and StorageFilter definitions, plus the cex/cim code that used them, then the rest of the implementation could live elsewhere, if desired.

lib/Drush/Config/StorageWrapper.php
```diff
@@ -0,0 +1,259 @@
...
+  /**
+   * Returns whether a configuration object exists.
+   *
+   * @param string $name
+   *   The name of a configuration object to test.
+   *
+   * @return bool
+   *   TRUE if the configuration object exists, FALSE otherwise.
+   */
+  public function exists($name) {
```
yched: all those phpdocs for the public methods from the interface could be just {@inheritdoc} ;-)

commands/core/config.drush.inc
```diff
@@ -367,12 +371,30 @@ function _drush_config_export($destination, $destination_dir, $branch) {
...
+      drush_copy_dir($destination_dir, $tmpdir, FILE_EXISTS_OVERWRITE);
+      $comparison_source = new FileStorage($tmpdir);
+      $comparison_source_filtered = new StorageWrapper($comparison_source, $storage_filters);
+      foreach ($active_storage->listAll() as $name) {
+        // Copy active storage to our temporary active store.
+        if ($existing = $active_storage->read($name)) {
+          $comparison_source_filtered->write($name, $existing);
+        }
+      }
+    }
```
yched: Not too familiar with that part in config-export, but it seems that even if the in-drupal staging storage was directly configured to be the wrapped/filtered storage, drush config-export would still have to know that fact somehow because it needs to take some special steps for that case ?

greg-1-anderson: Yeah, Drush just instantiates its own FileStorage objects.  Drupal could in theory add filters on top of the objects the client provided, but we would need to be able to hook into this process somehow if we wanted to be able to do a diff and tell the user what was about to change.

Schnitzel:
Nice work!

And I definitely see this code usable, as we don't really know what we're gonna use CMI for in the bright future :)

But as far as I understand, we would like to have an environment based override of configuration (Example in the tests: do not enable tracker module even though it is in the main config).
But I'm a bit worried that this only happens in the case of writing/reading configuration from Drush, so it could still happen that another module enables tracker module, which Drush has no control over it.

But there is a Configuration Overwrite system in Drupal Core already, see https://www.drupal.org/node/1928898 "Providing overrides from modules".
Couldn't we implement a config environment module that then based on the environment Drupal is running in (dev, staging, production, whatever) provides some overrides to core? This would sound to me like the cleaner and more stable approach.

I could imagine having a config directory that contains the main configuration which is imported/exported with Drush and committed in GIT. 
And then there is a config directory for dev, staging, prod which contains only the YAML files that overwrite the main conf. 
A system that is already implemented by some people for settings.php files based on environments, like settings.php, settings.production.php, settings.local.php, etc.

But I'm actually also not sure if that config override system can override installed modules. It was implemented for Language Overrides with the mindset of also allowing that for other use cases, but enabling/disabling some modules was definitely not one of the use cases for languages. Maybe @alexpott knows some things here?

greg-1-anderson:
I tried to override enabled / disabled modules via settings.php, and it did not work. If there are other parts of the override system that do work for this, that would be fantastic.

commands/core/config.drush.inc
```diff
@@ -402,6 +424,10 @@ function _drush_config_export($destination, $destination_dir, $branch) {
   // Write all .yml files.
   $source_storage = Drupal::service('config.storage');
   $destination_storage = new FileStorage($destination_dir);
+  // If therea re any filters, then attach them to the destination storage
```
weitzman: typo

commands/core/config.drush.inc
```diff
@@ -104,6 +106,7 @@ function config_drush_command() {
         'example-value' => 'branchname',
       ),
       'destination' => 'An arbitrary directory that should receive the exported files. An alternative to label argument',
+      'ignored-modules' => 'A list of modules to ignore during export (e.g. to avoid listing dev-only modules in exported configuration).',
```
weitzman: FYI, _drush_add_commandfiles() already honors an option of the same name. We might have some unintended consequences if we reuse that here. If we do reuse it, I think this should be declared with 'global option' instead of 'option'

weitzman:
The system here is hard wired to only care about core.extension. I think people will want to change non-extension config, and they wont be excited about using the config override system to do so. Would we provide another filter for standard environment overrides, or simply not cater to these requests in Drush core?

I'm still undecided about this PR. If it goes in, we need some more docs, especially a use case talk-through.

greg-1-anderson:
I just noticed that I never responded to [weitzman on Jun 15](https://github.com/drush-ops/drush/pull/1430#issuecomment-112078223) -- sorry.

The CoreExtensionFilter class is in fact designed only to work with core.extension configuration. Making a general-purpose data-driven filter that met all of the needs of every client would be very complicated.  The intention is to make _drush_config_get_storage_filters() plugable, so that those who needed to change something other than code.extension, but cannot use overrides in settings.php (which hopefully covers the majority of the use cases) could write their own filter.

If we do agree that this solution is okay as an interim solution, until core is ready to roll it in (or something similar), then I'll add the hook and write up the docs.

greg-1-anderson:
Another possibility is that we could put just the hooks in Drush core, and put CoreExtensionFilter in config-extras.

weitzman:
It looks like core is not going to cater to this feature, nor do I see anyone else stepping up. So lets put it in Drush. This needs a topic.

greg-1-anderson:
I had to rebase with the latest master to continue, and therefore need to continue this on a different PR, as rebasing confuses GitHub.

The continuation is on #1522.

PR #1434, opened 2015-06-09
Improve queue running (exception handling, time limits, summary messages)

davereid:
* Adds exception handling to queue running (including the SuspendQueueException logic from Drupal 8). An exception should not prevent the rest of the queue items from being processed.
* Adds a summary message of how many items were processed. Currently you get absolutely no output from the queue-run command unless there was an exception.

### DIFF
--- a/commands/core/queue.drush.inc
+++ b/commands/core/queue.drush.inc
@@ -20,6 +20,9 @@ function queue_drush_command() {
       'queue_name' => 'The name of the queue to run, as defined in either hook_queue_info or hook_cron_queue_info.',
     ),
     'required-arguments' => TRUE,
+    'options' => array(
+      'time-limit' => 'The maximum number of seconds allowed to run the queue',
+    ),
   );
   $items['queue-list'] = array(
     'description' => 'Returns a list of all defined queues',
@@ -46,16 +49,13 @@ function queue_drush_command() {
  * Validation callback for drush queue-run.
  */
 function drush_queue_run_validate($queue_name) {
-  $queue = drush_queue_get_class();
-  $queue->validate('queue-run');
-}
-
-/**
- * Validation callback for drush queue-list.
- */
-function drush_queue_list_validate() {
-  $queue = drush_queue_get_class();
-  $queue->validate('queue-list');
+  try {
+    $queue = drush_queue_get_class();
+    $queue->getInfo($queue_name);
+  }
+  catch (\Drush\Queue\QueueException $exception) {
+    return drush_set_error('DRUSH_QUEUE_RUN_VALIDATION_ERROR', $exception->getMessage());
+  }
 }
 
 /**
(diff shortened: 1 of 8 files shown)


davereid:
Some more things I'd like to add:
* Improve the result message with time elapsed, and rate per minute of items processed.
* Add a way to override how much time should be used to process the queue. It would be nice if we could default the queue-run command to a better time limit instead of the time limit we assume has to be short due to being run along with cron.

weitzman:
We are getting rid of concurrency within drush so please remove that part.
We also have #1169 pending for a while.

We really need a queue maintainer for Drush. These commands dont require
much oversight but PRs like this are hard to judge for the current team.

davereid:
I'd love to help with maintaining queue.drush.inc. I've been active in queue running, would seek to make sure the current code matches core's queue running.

Is there an issue where the concurrency is getting removed? I've found that incredibly useful in several production projects and would have to look to other projects if it were gone.

weitzman:
Thanks for the offer to maintain queue.drush.inc!

Concurrency discussion is #975. @juampy already mentioned concurrent_queue.

davereid:
Ok I'll rescope this to remove queue-run-all for now

commands/core/queue.drush.inc
```diff
@@ -46,16 +46,13 @@ function queue_drush_command() {
...
- * Validation callback for drush queue-list.
- */
-function drush_queue_list_validate() {
-  $queue = drush_queue_get_class();
-  $queue->validate('queue-list');
+  try {
+    $queue = drush_queue_get_class();
+    $queue->getInfo($queue_name);
+  }
+  catch (\Drush\Queue\QueueException $exception) {
```
weitzman: Is this really needed? In modern Drush, the command callback is wrapped in a try/catch so any exception should get caught there.

davereid: One outputs a huge long exception message, one outputs a nice error statement. I think the latter is better for usability.

davereid:
Looks like I need to look into some tests. Although it's odd since it seems to be that aggregator module doesn't seem to be enabled properly, otherwise we wouldn't be seeing this failure.

weitzman:
I'm fine with this PR once tests are passing. Thanks @davereid.

Also, congrats to @davereid on becoming our newest Drush co-maintainer!

commands/core/queue.drush.inc
davereid: Compare:
```
drush queue-run invalid_queue
exception 'Drush\Queue\QueueException' with message 'Could not find the invalid_queue queue.' in /home/davereid/Sync/opt/drush/lib/Drush/Queue/QueueBase.php:36                                     [error]
Stack trace:
#0 /home/davereid/Sync/opt/drush/lib/Drush/Queue/Queue8.php(48): Drush\Queue\QueueBase->getInfo('invalid_queue')
#1 /home/davereid/Sync/opt/drush/commands/core/queue.drush.inc(76): Drush\Queue\Queue8->run('invalid_queue')
#2 [internal function]: drush_queue_run('invalid_queue')
#3 /home/davereid/Sync/opt/drush/includes/command.inc(359): call_user_func_array('drush_queue_run', Array)
#4 /home/davereid/Sync/opt/drush/includes/command.inc(210): _drush_invoke_hooks(Array, Array)
#5 [internal function]: drush_command('invalid_queue')
#6 /home/davereid/Sync/opt/drush/includes/command.inc(178): call_user_func_array('drush_command', Array)
#7 /home/davereid/Sync/opt/drush/lib/Drush/Boot/BaseBoot.php(62): drush_dispatch(Array)
#8 /home/davereid/Sync/opt/drush/drush.php(70): Drush\Boot\BaseBoot->bootstrap_and_dispatch()
#9 /home/davereid/Sync/opt/drush/drush.php(11): drush_main()
#10 {main}
```

with:
```
drush queue-run invalid_queue
Could not find the invalid_queue queue.                                                                                  [error]
```

davereid:
Looks like the remaining test failure is not related to queues, and currently broken in master: https://travis-ci.org/drush-ops/drush/branches

davereid:
I've also pushed this work to 7.x and 6.x as well.

PR #1439, opened 2015-06-11
Better bootstrap

jaylynnw:


weitzman:
These commits went in already. Seems like this PR is a mistake or WIP?

Issue #1441, opened 2015-06-13
drush sqlq --db-prefix doesn't work any more.


greg-1-anderson:
Drush 6 behavior (correct):
```
$ drush sqlq 'select * from users;'
0|||||||0|0|0|0|||0||
1|admin|...|admin@example.com||||1434161704|1434161729|1434161729|1|UTC||0|admin@example.com|b:0;
$ drush sqlq 'select * from {users};'
Error: near line 1: unrecognized token: "{"
Query failed.                                                        [error]
$ drush sqlq 'select * from {users};' --db-prefix
0|||||||0|0|0|0|||0||
1|admin|...|admin@example.com||||1434161704|1434161729|1434161729|1|UTC||0|admin@example.com|b:0;
```

Drush 7 and Drush 8 behavior:
```
$ drush sqlq 'select * from users;'
0|||||||0|0|0|0|||0||
1|admin|...|admin@example.com||||1434161704|1434161729|1434161729|1|UTC||0|admin@example.com|b:0;
$ drush sqlq 'select * from {users};' 
Query failed.                                                        [error]
$ drush sqlq 'select * from {users};' --db-prefix

Fatal error: Class 'Drush\Sql\Database' not found in /Users/ga/.composer/vendor/drush/drush/lib/Drush/Sql/SqlBase.php on line 201
```

Just leaving this here.  I'm not really fond of db prefixes myself, but ideally this would still work.

weitzman:
I see an attempt to prefix in the code. My hunch is that we only don't prefix when an input file is provided (as opposed to a string). I think that other commands use sql-query internally and provide and input file or --file option so they would suffer this same issue. The example given would not fall into this category so no idea on the example shown above.

greg-1-anderson:
I had a [better solution to this problem](https://github.com/greg-1-anderson/utiliscripts/blob/master/remove-dbprefix).

Ideally, Drush would still work with db prefixes, but I don't know if anyone will take the time to circle back here.

drzraf:
same (blocking) issue `drush sql-sanitize` which, sadly, will just NOT work without `--db-prefix`
What to do?

PR #1442, opened 2015-06-13
Allow Drush to bootstrap a Drupal (8, 7, 6) site that has a valid but empty database.

greg-1-anderson:
This situation originally cropped up in issue [1996004](https://www.drupal.org/node/1996004) on drupal.org.  The fix committed at that time worked well enough to allow `drush status` and `drush sql-sync` to work on sites with empty databases.  With the new bootstrap and sql class changes, though, these things no longer work.

The original fix suggested in that issue was to just wrap the database bootstrap phase in a try / catch, which seemed to work okay, but raised issues of stability, since it allowed Drush to continue working after a partially-completed database bootstrap.  The potential side effects for that were (and remain) unknown.

This PR provides a better option; it tests to see if one of the database tables needed in Drupal 7's database bootstrap exists in Drush's bootstrap database validate method; if that table does not exist, then validate returns FALSE, which cleanly stops the database bootstrap before it begins.  This is correct behavior; we'll see what Travis has to say about the results.

Note, however, that there is a weakness in this PR in that it does not work with databases that use a db-prefix; see related issue #1441, which interferes somewhat with supporting db-prefixes here.  While I tend to dislike db-prefixes, it would probably be undesirable to make it impossible to bootstrap any site that used them.  We could potentially fix bootstrapping without fixing --db-prefixes by assembling the table name ourself rather than using `{blocked_ips}`, but that would be a little hack-ish.

Supporting operations on sites with empty databases is important in scenarios where you might, for example, want to have Travis automatically run tests on a scratch Drupal site every time a commit is pushed to a GitHub repository.  I'm working on such a thing right now (extending from the [Badcamp behat example repository](https://github.com/arithmetric/badcamp2014-behat-travisci)); the site-install use case works fine, but this bug makes it harder to support the sql-sync use case (e.g. to copy an existing site to serve as the starting point for the tests).

Let me know what you think about the db-prefix stuff.

### DIFF
--- a/lib/Drush/Boot/DrupalBoot.php
+++ b/lib/Drush/Boot/DrupalBoot.php
@@ -454,6 +454,35 @@ abstract class DrupalBoot extends BaseBoot {
     return TRUE;
   }
 
+  // This is a bootstrap helper function designed to be called
+  // from the bootstrap_drupal_database_validate() methods of
+  // derived DrupalBoot classes.  If a database exists, but is
+  // empty, then the Drupal database bootstrap will fail.  To
+  // prevent this situation, we test for some table that is needed
+  // in an ordinary bootstrap, and return FALSE from the validate
+  // function if it does not exist, so that we do not attempt to
+  // start the database bootstrap.
+  function bootstrap_drupal_database_has_table($required_tables) {
+    try {
+      $sql = drush_sql_get_class();
+      $spec = $sql->db_spec();
+      $prefix = $spec['prefix'];
+      $tables = $sql->listTables();
+      foreach ((array)$required_tables as $required_table) {
+        if (!in_array($prefix . $required_table, $tables)) {
+          return FALSE;
+        }
+      }
+    }
+    catch (Exception $e) {
+      // Usually the checks above should return a result without
+      // throwing an exception, but we'll catch any that are
+      // thrown just in case.
+      return FALSE;
+    }
+    return TRUE;
+  }
+
   /**
    * Boostrap the Drupal database.
    */
--- a/lib/Drush/Boot/DrupalBoot.php
+++ b/lib/Drush/Boot/DrupalBoot.php
@@ -467,9 +467,13 @@ abstract class DrupalBoot extends BaseBoot {
       $sql = drush_sql_get_class();
       $spec = $sql->db_spec();
       $prefix = $spec['prefix'];
+      if (!is_array($prefix)) {
+        $prefix = array('default' => $prefix);
+      }
       $tables = $sql->listTables();
       foreach ((array)$required_tables as $required_table) {
-        if (!in_array($prefix . $required_table, $tables)) {
+        $prefix_key = array_key_exists($required_table, $prefix) ? $required_table : 'default';
+        if (!in_array($prefix[$prefix_key] . $required_table, $tables)) {
           return FALSE;
         }
       }
--- a/lib/Drush/Boot/DrupalBoot7.php
+++ b/lib/Drush/Boot/DrupalBoot7.php
@@ -49,6 +49,32 @@ class DrupalBoot7 extends DrupalBoot {
     return $core;
   }
 
+  function bootstrap_drupal_database_validate() {
+    if (!parent::bootstrap_drupal_database_validate()) {
+      return FALSE;
+    }
+    // This is like drush_valid_db_credentials, but
+    // for Drupal 7, we also want to know if the
+    // {blocked_ips} table exists, as the bootstrap will
+    // fail if it does not.  If that is the situation,
+    // then we'll return FALSE here in validate, so that
+    // we do not attempt to start the database bootstrap.
+    try {
+      $sql = drush_sql_get_class();
+      $result = $sql->query('SELECT * from blocked_ips limit 1;');
+      if ($result === false) {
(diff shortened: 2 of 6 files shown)


greg-1-anderson:
The tests passed, except for the [d8 migrate tests](https://travis-ci.org/drush-ops/drush/jobs/66634888).  I haven't investigated the cause; this PR only affects D7 bootstraps, and this test looks like it's a D6 -> D8 migration, so perhaps it is unrelated.

lib/Drush/Boot/DrupalBoot7.php
```diff
@@ -49,6 +49,32 @@ function bootstrap_drupal_core($drupal_root) {
...
+    // This is like drush_valid_db_credentials, but
+    // for Drupal 7, we also want to know if the
+    // {blocked_ips} table exists, as the bootstrap will
+    // fail if it does not.  If that is the situation,
+    // then we'll return FALSE here in validate, so that
+    // we do not attempt to start the database bootstrap.
+    try {
+      $sql = drush_sql_get_class();
+      $result = $sql->query('SELECT * from blocked_ips limit 1;');
+      if ($result === false) {
```
weitzman: capitalize to FALSE

weitzman:
The approach looks good to me. Note that we have listTables() method in the sql classes. Perhaps it doesn't buy us much. I'm nearly ready to drop db-prefix support for Drush8. But I guess it would be easy enough to handle the prefix manually here as you suggest.

Any reason why this is D7 only? Seems appropriate for all Drupal versions. D8 needs a different table to check.

greg-1-anderson:
Yes, this needs D6 and D8 versions, but starting with a proof-of-concept to work out how we deal with database prefixes.

As much as I do not like table prefixes, I just cannot seem to convince myself that it would be okay to drop support for them.

Presuming that we do handle db-prefix here, I think it is best to manually assemble the table names from the sql credentials.  The handling for {table} is typically done by bootstrapping Drupal to the BOOTSTRAP_DATABASE, and we can't really do that when validating that bootstrap phase.

greg-1-anderson:
Added a commit to handle database prefixes on Drupal 7.  Will test with Drupal 6 and Drupal 8 next, and see if similar validation is needed for those versions.

greg-1-anderson:
In Drupal 8, the first table checked in the bootstrap code is `key_value`.  On Drupal 6, the first table is 'cache'.  I'll refactor the code so that each version validates with the same function, but uses a different table name in the test.

greg-1-anderson:
Code looking pretty good; I'll add a couple of simple tests for this now.

greg-1-anderson:
Added a test to insure that sql-create will erase the database, and that core-status will not error out when the database is empty.  All of these new tests save for the last (core-status) pass without this PR; all pass with this PR.

I added my tests to the existing sqlConnectTest.php, to avoid spending extra time installing a Drupal site only to immediately erase the database again.  It wouldn't really be a good test if we did not create a real site first.  I suppose we could break this into a separate test file if you really wanted, but it might be best to just rename the file to sqlConnectCreateTest.php.

lib/Drush/Boot/DrupalBoot.php
```diff
@@ -454,6 +454,39 @@ function bootstrap_drupal_database_validate() {
...
+  // empty, then the Drupal database bootstrap will fail.  To
+  // prevent this situation, we test for some table that is needed
+  // in an ordinary bootstrap, and return FALSE from the validate
+  // function if it does not exist, so that we do not attempt to
+  // start the database bootstrap.
+  function bootstrap_drupal_database_has_table($required_tables) {
+    try {
+      $sql = drush_sql_get_class();
+      $spec = $sql->db_spec();
+      $prefix = $spec['prefix'];
```
weitzman: lets add a code comment about why we are doing our own prefixing here

lib/Drush/Boot/DrupalBoot.php
```diff
@@ -454,6 +454,39 @@ function bootstrap_drupal_database_validate() {
...
 
+  // This is a bootstrap helper function designed to be called
+  // from the bootstrap_drupal_database_validate() methods of
+  // derived DrupalBoot classes.  If a database exists, but is
+  // empty, then the Drupal database bootstrap will fail.  To
+  // prevent this situation, we test for some table that is needed
+  // in an ordinary bootstrap, and return FALSE from the validate
+  // function if it does not exist, so that we do not attempt to
+  // start the database bootstrap.
+  function bootstrap_drupal_database_has_table($required_tables) {
```
weitzman: Needs Doxygen for input param and return value

weitzman:
The testing approach looks great. I agree with renaming the test file. I added minor code comments. Feel free to merge after that.

greg-1-anderson:
Added comments and merged into master and 7.x.

greg-1-anderson:
If anyone else is wondering why Drush is not working with an empty but valid database, perhaps it is because you are using Domain Access.

In your settings.php file:
```
try {
  include DRUPAL_ROOT . '/sites/all/modules/contrib/domain/settings.inc';
}
catch(Exception $e) {}
```
Without the `try` / `catch`, Drush will fail to bootstrap Drupal when settings.php supplies a valid set of database credentials that point at an empty database.

PR #1448, opened 2015-06-19
Avoid warning in make_normalize_info()

arnested:
A warning was thrown in make_normalize_info() if a make file didn't
contain any projects. A typical use case of this is modules supplying
a make file (for recursive) use with only libraries in them.


arnested:
It will also make sense to backport it to `7.x` (just do a `git cherry-pick 4e6945569f1d451001a0375249a036013c427563`).

weitzman:
make is deprecated

Issue #1269, opened 2015-03-19
drush make --overwrite should respect --projects=?


hanoii:
I like the fact that --overwrite removes files that are no longer in the modules, as opposed to the default merge, however, I just found out that if you run --overwrite with some specific --projects=.. it also removes everything else from your contrib destinations, like if you do --projects=rules --overwrite, it will properly process the makefile for rules, but removes everything else. Is it by design or a bug? If the former, is there a way for me to make sure the new download removes anything that's not there?

hanoii:
Hello, just wondering if there's any idea on this or if can be put in a roadmap.

I am now manually removing directories before running make of projects but I would rather rely on drush.

I tried to work on this tu submit a PR but I couldn't find a way to only overwrite what was download and also making sure that it won't break existing functionality. At one point, I had an idea of storing downloaded paths right on ```DrushMakeProject::make()```, which is where the download_location is being resolved, and then use than info in ```make_move_build()``` only to overwrite those directories. I don't think it's a bad idea, but I was first storing it in a context only to realize I was storing it in a different drush process, so the context in make_move_build() was above the context in DrushMakeProject::make().

I then tried to rely on a temp file but all "drush" ways of storing data on a temp file rely on passing data to child processes, not to parent processes. I could hack my way around it by custom code but probably not the way to go.

While I am writing this, I thought of another alternative, which is to have DrushMakeProject::make() put a hidden file in the project top level dir, like .drush-make-toplevel or something like that, then ```make_move_build()``` can recursively scan directories until it reach a top-level one and move that one with the overwrite option. This might even work for every case.

I might try to give that idea a try, but if you share any points, ideas, etc. will be greatly appreciated.

PR #1450, opened 2015-06-22
Make sure top-level files are also merged

hanoii:
I was noticing that when you do a

```drush make xxx.make --projects=drupal```

On a current site so that it's updated, the top-level files are not copied, this is because an attempt to prevent removing top-level directories possible when overwriting files.

Probably not the best way of doing all this, see #1269 , but this is a fix following the same logic.


jhedstrom:
Thanks!

PR #1452, opened 2015-06-22
Fix fatal error running updb in Drupal 8

neclimdul:
Fixes the following fatal error when running `drush updb`
```
Call Stack:
    0.0002     232896   1. {main}() /usr/share/php/drush/drush.php:0
    0.0002     233456   2. drush_main() /usr/share/php/drush/drush.php:11
    0.2346    7335992   3. Drush\Boot\BaseBoot->bootstrap_and_dispatch() /usr/share/php/drush/drush.php:70
    0.2624    6914752   4. drush_dispatch() /usr/share/php/drush/lib/Drush/Boot/BaseBoot.php:62
    0.3865    9087960   5. call_user_func_array() /usr/share/php/drush/includes/command.inc:178
    0.3865    9088128   6. drush_command() /usr/share/php/drush/includes/command.inc:178
    0.3868    9089112   7. _drush_invoke_hooks() /usr/share/php/drush/includes/command.inc:210
    0.3875    9122824   8. call_user_func_array() /usr/share/php/drush/includes/command.inc:359
    0.3876    9122992   9. drush_core_updatedb() /usr/share/php/drush/includes/command.inc:359
    0.3895    9292792  10. update_main() /usr/share/php/drush/commands/core/core.drush.inc:460
    0.3976    7685944  11. drupal_load_updates() /usr/share/php/drush/commands/core/drupal/update.inc:108
Error: Call to undefined function drupal_get_installed_schema_version() in /var/www/d8/core/includes/install.inc, line 80
```

### DIFF
--- a/commands/core/drupal/update.inc
+++ b/commands/core/drupal/update.inc
@@ -103,6 +103,7 @@ function update_main() {
   // In D8, we expect to be in full bootstrap.
   drush_bootstrap_to_phase(DRUSH_BOOTSTRAP_DRUPAL_FULL);
 
+  require_once DRUPAL_ROOT . '/core/includes/schema.inc';
   require_once DRUPAL_ROOT . '/core/includes/install.inc';
   require_once DRUPAL_ROOT . '/core/includes/update.inc';
   drupal_load_updates();


davereid:
Is there a script we can compare to in the current D8 code for what updatedb should be running? Last I knew if was being re-added but wasn't complete yet.

neclimdul:
eep, that could be a problem if we start requiring update hooks Wednesday. https://groups.drupal.org/node/472163

neclimdul:
Tim set me straight, there's a implementation in `\Drupal\system\Controller\DbUpdateController`

davereid:
So I'm looking at https://api.drupal.org/api/drupal/core!modules!system!src!Controller!DbUpdateController.php/8 and don't see it specifically loading schema.inc. Do we need to fix this in the Drush bootstrap process instead?

neclimdul:
So... that raises an interesting point. The reason this isn't available is that we are not doing a full bootstrap for updb. DRUSH_BOOTSTRAP_DRUPAL_SITE to be exact.

In d8, "update.php" assumes that you have a fully functional site and a full bootstrap because its run as a full site callback. In previous versions of Drupal this was not the case which allowed it to fix caches and stuff. The later is the reason rebuild.php exists in d8.

We could change the bootstrap level but that would affect other versions of Drupal which I don't think makes sense so for now, I'd suggest just keeping the logic here personally.

weitzman:
I think we have changed the bootstrap level in the past for updatedb and its burned us. It makes sense to me that we do a `drush_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_FULL)` at the beginning of the D8 implementation of this command.

neclimdul:
I'm not sure what the next steps are here...

jhedstrom:
Working on #1504 I didn't encounter this issue. Can other reproduce it?

weitzman:
Please comment if still an issue.

jrockowitz:
I just ran into this exact error message when running `drush updb` and the issue has nothing to do with Drupal 8 or Drush. The /tmp directory on my server had ran out of disk space, Doh!. 

Hopefully someone can learn from my mistake.

PR #1462, opened 2015-06-24
Make a local $container variable available in drush cli

damiankloip:
It would be cool to have a local `$container` variable available in the REPL.

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -21,6 +21,9 @@ function drush_cli_core_cli() {
   // Use the Symfony var-dumper to inspect variables.
   $boris->setInspector(new \Drush\Boris\VarDumperInspector());
 
+  // Set a local $container variable.
+  $boris->setLocal(['container' => \Drupal::getContainer()]);
+
   // Boris will never return control to us, but our shutdown
   // handler will still run after the user presses ^D.  Mark
   // this command as completed to avoid a spurious error message.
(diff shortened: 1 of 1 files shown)


weitzman:
Seems reasonable, though I dunno how common it will be to mess with container. I don't want to open the floodgates for lots of local variables. 

I'll leave this open for a few days, and then commit if there are no objections.

damiankloip:
Fair, I think this is pretty common to be honest. If not one of the most common things when using `drush php`, you need the container to do almost anything useful. E.g load entities, entity type data, namespaces, load config, check cache items, path/route related stuff, the list goes on :)

dawehner:
> Seems reasonable, though I dunno how common it will be to mess with container. I don't want to open the floodgates for lots of local variables.

Yeah it would be a huge win DX wise!

Go damian!

weitzman:
OK, lets do it.

* This needs to run only for D8 sites
* Ideally we document this local variable and core-cli in general by adding a new topic. If someone can post text here, I can massage it into a topic. Just give some examples of what you can do with this command.

weitzman:
Ping @damiankloip or @dawehner, for finishing this off if desired.

damiankloip:
@weitzman Sorry about the delay, Just pushed a new commit to add this for D8 only.

As far as examples go, there are some above, and in the issue I opened to use Symfony VarDumper (https://github.com/drush-ops/drush/pull/1102). What else do you think would be good? We could change these examples to use the $container variable?

dawehner:
Nice!

dawehner:
Ready to be committed!

damiankloip:
Thanks @weitzman

PR #1477, opened 2015-07-06
Drush make docs: use quotes for version strings.

### LINKED ISSUE #1470: YAML makefile quotes gotcha
If you create a makefile with a line like this, you would expect to get version 1.0 of visitor_actions.
````
package:
  visitor_actions: 1.0
````
Instead, you will actually get 1.1 every time.

I'm guessing what's happening here is that the YAML parser is casting this version to an integer and dropping the trailing 0, so drush make downloads the latest version instead.

One of two things should be done to fix this: either tweak the parser to treat version numbers as strings (no idea if that's possible or not), or change the documentation (for instance [here](http://www.drush.org/en/master/make/)) to always put version numbers in quotes.

danepowell:
Fixes #1470 by updating the drush make documentation to specify that version strings should be in quotes.

### DIFF
--- a/docs/make.md
+++ b/docs/make.md
@@ -66,7 +66,7 @@ specified as the key.
 
     projects:
       drupal:
-        version: 7.33
+        version: "7.33"
 
 Do not use both types of declarations for a single project in your makefile.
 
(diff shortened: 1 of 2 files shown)


jonhattan:
`examples/example.make.yml` also needs updates

jhedstrom:
Thanks!

PR #1480, opened 2015-07-07
Allow Drush to search recursively inside any specified alias path.

greg-1-anderson:
Currently, Drush only considers alias files that appear immediately inside any given alias searchpath directory.  This PR expands the search to include any directory that is anywhere deep inside any specified searchpath directory.

For the most part, this should not cause Drush to search for alias files too deeply.  The current default alias search path only includes folders that are specific to Drush.  For example:
```
  /etc/drush
  $HOME/.composer/vendor/drush/drush
  $HOME/.drush
  $ROOT/drush
  $ROOT/sites/all/drush
  $ROOT/../drush
```
I'm not sure why Drush searches its own directory for aliases, but it does.

These directories do, for the most part, only contain Drush files, and therefore should not be too crowded with deeply-nested folders.  The one exception here is $HOME/.drush, which might contain a 'vendor' directory if folks use the recommended `cd $HOME/.drush && composer include ...` technique to get global Drush commandfiles that are managed by composer.  In this case, there might be some very deep hierarchies that we'd rather not search.

To work around this, this PR further restricts searching to inside a folder 'aliases', if it exists in any search location.  If a folder named 'aliases' does not exist, then the entire search location is searched.

I suppose an alternate idea would be to deeply search $LOCATION/aliases, and also consider alias files contained directly inside $LOCATION.  In this instance, though, the --alias-path location should probably always be searched recursively.  This is slightly awkward vis-a-vis the current implementation of alias file searching, but not insurmountably so.

### DIFF
--- a/tests/siteAliasTest.php
+++ b/tests/siteAliasTest.php
@@ -114,4 +114,38 @@ EOD;
   public function testBadAlias() {
     $this->drush('sa', array('@badalias'), array(), NULL, NULL, self::EXIT_ERROR);
   }
+
+  /**
+   * Ensure that Drush searches deep inside specified search locations
+   * for alias files.
+   */
+  public function testDeepAliasSearching() {
+    $aliasPath = UNISH_SANDBOX . '/aliases';
+    mkdir($aliasPath);
+    $deepPath = $aliasPath . '/deep';
+    mkdir($deepPath);
+    $aliasFile = $deepPath . '/baz.aliases.drushrc.php';
+    $aliasContents = <<<EOD
+  <?php
+  // Writtne by Unish. This file is safe to delete.
+  \$aliases['deep'] = array(
+    'remote-host' => 'fake.remote-host.com',
+    'remote-user' => 'www-admin',
+    'root' => '/fake/path/to/root',
+    'uri' => 'default',
+    'command-specific' => array(
+      'rsync' => array(
+        'delete' => TRUE,
+      ),
+    ),
+  );
+EOD;
+    file_put_contents($aliasFile, $aliasContents);
+    $options = array(
+      'alias-path' => $aliasPath,
+      'simulate' => TRUE,
+    );
+
+    $this->drush('sa', array('@deep'), $options);
+  }
 }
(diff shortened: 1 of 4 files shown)


weitzman:
I think the proposed implementation looks good. 

1. I prefer to longer site-aliases folder name. That disambiguates from command aliases and shell aliases.
1. Lets update any docs that you can find on this topic.
1. I think it is unwise to backport this but I'm open to discussion if desired.

greg-1-anderson:
I agree that we should not backport this.  If someone strongly desires this behavior in an older version of Drush, we could consider turning it on with an option.  At that point, though, you might as well set $options['alias-path'] in your drushrc.php, and explicitly list all of the folders you want searched. So, probably no backport.

I made the changes you suggested in #1 and #2 in the commit above.

tests/siteAliasTest.php
```diff
@@ -114,4 +114,38 @@ public function testSAList() {
   public function testBadAlias() {
     $this->drush('sa', array('@badalias'), array(), NULL, NULL, self::EXIT_ERROR);
   }
+
+  /**
+   * Ensure that Drush searches deep inside specified search locations
+   * for alias files.
+   */
+  public function testDeepAliasSearching() {
+    $aliasPath = UNISH_SANDBOX . '/aliases';
```
weitzman: should this be site-aliases?

tests/siteAliasTest.php
```diff
@@ -114,4 +114,38 @@ public function testSAList() {
...
+    mkdir($aliasPath);
+    $deepPath = $aliasPath . '/deep';
+    mkdir($deepPath);
+    $aliasFile = $deepPath . '/baz.aliases.drushrc.php';
+    $aliasContents = <<<EOD
+  <?php
+  // Writtne by Unish. This file is safe to delete.
```
weitzman: typo: Written

weitzman:
Ready to merge after you address the minor line comments above.

tests/siteAliasTest.php
greg-1-anderson: No; this is passed as --alias-path, which is always searched recursively, regardless of whether there is a 'site-aliases' directory or not (since the path was explicitly provided by the user).

I could change this to '/site-alias-directory' or something like that, for the avoidance of doubt.

Issue #894, opened 2014-10-17
Security: Disallow anything but TLS 1.0+


davidstrauss:
Drush performs multiple operations (downloading modules, etc.) using libcurl, but it does not currently restrict connections to TLS 1.0+. This could allow tampering or intercepting transferred data.

I'm not submitting this privately because POODLE is a generally known vulnerability that affects all SSL/TLS applications, and Drush is no more vulnerable than Chrome or Firefox right now.

jhedstrom:
Drupal.org switched from Edgecast to Fastly today, and `dl` or `pm-releases` fails with this error:

```Executing: wget -q --timeout=30 -O /tmp/download_filegZW5p5 https://updates.drupal.org/release-history/drupal/8.x
simplexml_load_file(): I/O warning : failed to load external entity "" Project.php:72 [0.39 sec, 6.3 MB]
```

Running the wget command manually:

```
Resolving updates.drupal.org (updates.drupal.org)... 23.235.47.249, 199.27.79.249
Connecting to updates.drupal.org (updates.drupal.org)|23.235.47.249|:443... connected.
ERROR: no certificate subject alternative name matches
	requested host name ‘updates.drupal.org’.
To connect to updates.drupal.org insecurely, use `--no-check-certificate'.
```

Invalid cert in action: https://updates.drupal.org/release-history/drupal/8.x

greg-1-anderson:
If you look at one of those updates.drupal.org URLs in a web browser, you can examine the certificate and see that the name in the cert is not correct.

greg-1-anderson:
Maybe the cert is correct, but wget does not interpret it correctly?

c.f. https://docs.fastly.com/guides/ssl/why-do-i-see-ssl-certificate-errors-when-using-wget

ryanaslett:
Drupal.org does not support https on updates.drupal.org. It was an accident of edgecast before, but not officially supported. We'll have it soon, but the api for updates.drupal.org is http only at the moment.

greg-1-anderson:
--no-check-certificate does appear to happen to work at the moment (an accident of fastly?).  Maybe it's better to use that than to go all the way back to http?

grigarr:
[...]

PR #1481, opened 2015-07-07
#894: Try downloading via https when http URLs are provided.

greg-1-anderson:
Try https anywhere the caller requests http.  This will increase privacy for sure, but perhaps this is only security-theater vis-a-vis MITM attacks, as the attacker could simply interfere with the https connection, and then take advantage of the fact that we fall back to the insecure connection.

It probably wouldn't be a good idea to universally reject http in Drush, though, so I am uncertain whether this PR is enough of an improvement to be worthwhile.


weitzman:
I think drush_download_file needs to just do as its told. So, nothing to do here. The packaging script needs to specify https for ftp.drupal.org when it generates update xml.

PR #1487, opened 2015-07-10
Require php 5.5.9 or later.

greg-1-anderson:
Because modern.  And supported.  And https://groups.drupal.org/node/473473

Master branch only.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "drush.complete.sh"
   ],
   "require": {
-    "php": ">=5.4.5",
+    "php": ">=5.5.9",
     "d11wtq/boris": "~1.0",
     "symfony/yaml": "~2.2",
     "symfony/var-dumper": "^2.6.3",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "0839439a1593f281a4ae602ee11a1cf7",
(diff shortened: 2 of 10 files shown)


weitzman:
There are a couple more places. See https://github.com/drush-ops/drush/pull/1424/files. 

I'm not sure those reasons are compelling. We could be making Drush8 unusable by a lot of Drupal 6 and 7 sites with this change.

davereid:
On the other hand, those people won't be able to use Drush 8 with Drupal 8 at all?

greg-1-anderson:
Dreamhost supports php 5.3, 5.4, 5.5 and 5.6.  Hostgator supports php 5.3 through 5.5.  Seems that non-obsolete versions of php are at least available on low-end hosts.

Does Drupal 6 break under php 5.5?  In a few months, php 5.4 will be EOL (http://php.net/supported-versions.php).  In light of that, it seems reasonable to me to have folks on older versions of php use Drush 7, especially considering that you can use the site-local Drush feature to mix-and-match your Drush versions to your site.

greg-1-anderson:
5.5. enters "security fixes only" tomorrow, for one more year.  I guess we could keep this issue open until September, and commit it once 5.4 is actually EOL.

On the flip side, we still have 80% of all users on php 5.4 or later. 

http://w3techs.com/technologies/details/pl-php/5/all

Kinda depressing.

greg-1-anderson:
Of course, 80% is on the server.  There's no reason why devs can't upgrade their work machines to a supported php, and run Drush 7 on the server.

weitzman:
Not going to bump this now since we still target D6 and D7, and those folks are stubbornly not updating. Thinking that we do this for drush9.

PR #1504, opened 2015-07-22
Update hooks are run prior to automatic entity definition updates.

jhedstrom:
- Core change: https://www.drupal.org/node/2535082


weitzman:
Looks good. Feel free to merge it.

PR #1511, opened 2015-07-28
fix(mimetype) Add mime type for .sql files

### LINKED ISSUE #1292: Drush archive restore “Unable to determine mime type”
I created a backup with drush archive-dump --description, but when I try to restore with drush arr ./drush-backups/archive-dump/\<path\>.tar.gz --db-url=mysql://\<db info\> --destination=/home/\<dest path\> I get an error "Unable to determine mime type for \<unpacked tmp.sql\>". After the error, the command continues and appears to correctly restore the doc dir, but the db remains empty.

Also asked [here](http://drupal.stackexchange.com/q/152223/14073).

minorOffense:
Fix #1292 

Adds a default mime type for .sql files since finfo returns application/octet-stream for that file type.

### DIFF
--- a/includes/drush.inc
+++ b/includes/drush.inc
@@ -902,6 +902,7 @@ function drush_mime_content_type($filename) {
     $filename = basename(current(explode('?', $filename, 2)));
     $extension_mimetype = array(
       '.tar'     => 'application/x-tar',
+      '.sql'     => 'application/octet-stream',
     );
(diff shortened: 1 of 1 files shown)


weitzman:
Pushed to master, 7.x, and 6.x.

We prefer PRs against master branch, if possible. Thanks for the contribution!

minorOffense:
Oh right, sorry about that. Thanks for reviewing it so quickly!

PR #1516, opened 2015-07-29
Add an update finished callback.

### LINKED ISSUE #1509: Caught exception errors not displayed during update process
The core entity schema updates catch exceptions, but drush is not displaying the error:

```php
    function update_entity_definitions(&$context) {
      try {
        \Drupal::service('entity.definition_update_manager')->applyUpdates();
      }
      catch (EntityStorageException $e) {
        watchdog_exception('update', $e);
        $variables = Error::decodeException($e);
        unset($variables['backtrace']);
            // The exception message is run through
        // \Drupal\Component\Utility\SafeMarkup::checkPlain() by
        // \Drupal\Core\Utility\Error::decodeException().
        $ret['#abort'] = array('success' => FALSE, 'query' => t('%type: !message in %function (line %line of %file).', $variables));
        $context['results']['core']['update_entity_definitions'] = $ret;
        $context['results']['#abort'][] = 'update_entity_definitions';
      }
    }
```

I think this is related to the recent fix in #1504.

jhedstrom:
- Displays any error (or informational) messages set during update
  hooks.
- If an error is set during updates, updb will now exit non-zero.
- Fixes #1509

### DIFF
--- a/commands/core/drupal/update.inc
+++ b/commands/core/drupal/update.inc
@@ -230,9 +230,18 @@ function drush_get_update_list() {
 /**
  * Process and display any returned update output.
  *
+ * @see \Drupal\system\Controller\DbUpdateController::batchFinished()
  * @see \Drupal\system\Controller\DbUpdateController::results()
  */
 function drush_update_finished($success, $results, $operations) {
+
+  if (!drush_get_option('cache-clear', TRUE)) {
+    drush_log(dt("Skipping cache-clear operation due to --cache-clear=0 option."), 'warning');
+  }
+  else {
+    drupal_flush_all_caches();
+  }
+
   foreach ($results as $module => $updates) {
     if ($module != '#abort') {
       foreach ($updates as $number => $queries) {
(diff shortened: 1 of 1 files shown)


weitzman:
Do we not need to run \Drupal\system\Controller\DbUpdateController::batchFinished? That gets removed in this PR AFAICT.

jhedstrom:
The `batchFinished()` callback was commented out. Looking at the code in there:

```
    // No updates to run, so caches won't get flushed later.  Clear them now.
    drupal_flush_all_caches();

    $_SESSION['update_results'] = $results;
    $_SESSION['update_success'] = $success;
    $_SESSION['updates_remaining'] = $operations;

    // Now that the update is done, we can put the site back online if it was
    // previously in maintenance mode.
    if (isset($_SESSION['maintenance_mode'])) {
      \Drupal::state()->set('system.maintenance_mode', FALSE);
      unset($_SESSION['maintenance_mode']);
```
it doesn't do the display of these messages, and relies on `$_SESSION`. We could either call that, or replicate the logic in our own function.

weitzman:
Seems like we should implement this ourselves, as you propose. That way, people can avoid the drupal_flush_caches_call as per #936.

weitzman:
I should be more clear - I'm fine with the PR as is. Feel free to merge.

jhedstrom:
Done.

PR #1520, opened 2015-07-31
Prevent dependency hell when users try to use multiple autoloaders

greg-1-anderson:
When using a composer-managed Drupal site, including any Drupal 8 site, enormous, difficult-to-diagnose problems can arise if multiple autoloaders are used.  This patch detects this situation, and fails with an error if the user tries to use a global Drush command with a foreign autoloader (that is, any autoload.php file that is different than the autoload.php file for the Drupal site being used).  I spent a whole day the first time this happened to me; as Composer usage becomes more common, this sort of mistake will happen more frequently.  This patch will probably save a lot of people from some significant agony.

In the event that we see a user is trying to use a command that requires a foreign autoloader, we print out a sequence of commands to help the user fix the problem, by running `composer require a/b` in the context of their Drupal site, for all projects `a\b` that have their own autoload.php file, and also for `drush/drush`, since you shouldn't find any global Drush commands if there is already a site-local Drush in use.

If the user is trying to use a command that does NOT require a foreign autoloader on a Composer-managed Drupal site, then we will unset any other global commandfile that requires a foreign autoloader, preventing its command hooks from running.

This PR has no affect on global commandfiles that do not require a foreign autoloader, nor does it have any affect on global commands that are run without a Drupal site, or that are run with a Drupal site that is not composer-managed.  This does allow for the slight possible for dependency hell if conflicts arise between an extension and Drush itself; however, since the extensions must be run with Drush during development, they are unlikely to have problems.  I think we can start off with it semi-loose like this, and then tighten it down a bit later. In particular, I'd like to backport this as a bugfix to Drush 7, and then perhaps tighten it down for Drush 8 once that is done.

### DIFF
--- a/lib/Drush/Command/Commandfiles.php
+++ b/lib/Drush/Command/Commandfiles.php
@@ -16,24 +16,45 @@ namespace Drush\Command;
 class Commandfiles implements CommandfilesInterface {
   protected $cache;
   protected $deferred;
+  protected $needs_autoloader;
+  protected $foreign_autoloaders;
 
   function __construct() {
     $this->cache = array();
     $this->deferred = array();
+    $this->needs_autoloader = array();
+    $this->foreign_autoloaders = array();
   }
 
+  /**
+   * @return the list of all cached commandfiles that were loaded.
+   */
   function get() {
   	return $this->cache;
   }
 
+  /**
+   * @return the list of all commandfiles that were added, but have
+   * not been loaded yet (waiting for a higher bootstrap level).
+   */
   function deferred() {
   	return $this->deferred;
   }
 
+  /**
+   * Sort all of the cached commandfiles.  This is called once
+   * Drush has finished calling 'add' for this bootstrap phase.
+   */
   function sort() {
   	ksort($this->cache);
   }
 
+  /**
+   * Add a commandfile to the cache.  The commandfile might not be
+   * immediately added; it may be saved in the 'deferred' list, and
+   * then added on a later bootstrap phase, once it can be determined
+   * to be acceptable.
+   */
   function add($commandfile) {
 	  $load_command = FALSE;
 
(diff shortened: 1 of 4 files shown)


lib/Drush/Command/Commandfiles.php
```diff
@@ -53,17 +74,134 @@ function add($commandfile) {
 	      	$load_command = TRUE;
 	      }
 	      else {
-		    // Signal that we should try again on
-		    // the next bootstrap phase.
-		    $this->deferred[$module] = $commandfile; 	
+          // Signal that we should try again on
+          // the next bootstrap phase.
+          $this->deferred[$module] = $commandfile;
```
greg-1-anderson: Hm, looks like my editor's tab preferences are not set up correctly.

weitzman:
Despite your excellent issue description, I really can't follow the code without very careful scrutiny. We need to simplify Drush, and not complicate it further. Lets discuss by phone.

lib/Drush/Command/Commandfiles.php
```diff
@@ -53,17 +74,134 @@ function add($commandfile) {
...
 	    }
 	  }
 	  return $load_command;
   }
+
+  /**
+   * Load a commandfile once it has been determined to be acceptable.
+   */
+  protected function load($key, $path) {
+    $this->cache[$key] = $path;
```
weitzman: Lets add doxygen for new methods.

greg-1-anderson:
#1595 is a good start toward replacing this PR, but we still have a couple of use cases with that patch that can cause the user to encounter a Dependency Hell scenario.

We do **not** automatically add --local in any instance with that patch, as it currently stands.  We are conservative about this because the simple existence of a site-local Drush does not necessarily mean the user wants --local.  It is still possible to use global commandfiles that do not use any composer libraries; it is also possible to use global commandfiles that **do** use composer libraries with a Drupal 7 or 6 site that are not managed by Composer.  Some people might even try to continue to use global commandfiles that use composer libraries on a Composer-managed Drupal site, and this will even work sometimes (but is not recommended, as any upgrade could result in Dependency Hell).

The next question, then, is, do we want to consider returning to the issue of having Drush load autoloaders, or perhaps address the above in some other way (e.g. try to "guess" when to add --local, perhaps in a different PR)?

weitzman:
maybe if there is a site-local Drush, we just add --local by default. user can avoid this with --local=0 as an override.

greg-1-anderson:
We'd have to add support for --local=0 in the Drush Finder script; it currently ignores the value.

Beyond this, I think that --local disables too much to add it as a default; a lot of people might want global configuration files and aliases to still be available with their site-local Drush.  Maybe we need to add an additional flag --site-local to disable just global commandfiles.  That would probably be enough safety to allow us to close this PR.

greg-1-anderson:
I tried to add --site-local from the Drush "finder" script, but it's actually hard to detect whether the Drush that was found is site-local or not.  Perhaps what we need to do instead is check to see if drush.php is (a) inside the Drupal root, or (b) inside the "vendor" directory that contains the autoloader.  In those instances, we know that we have a site-local Drush, and we could automatically turn off global commandfile locations.  (The user can still add in locations they deem safe via a drushrc.php file.)

Another thing we might want to consider doing is abort with an error if Drush is NOT site-local, and the Drupal site is version 8 or later, as Drush itself can cause Dependency Hell, and this could happen at any time with a future D8 upgrade.

weitzman:
Just learned that npm has solved this for node projects - http://blog.timoxley.com/post/20772365842/node-js-npm-reducing-dependency-overheads. Sigh.

torstenzenk:
Hi,
this error driove me nuts for a few weeks now, so i wanted to tell others how i solved this:
```
Error: Call to undefined method Symfony\Component\HttpKernel\Event\PostResponseEvent::setDispatcher() in                        [error]
Drupal\Component\EventDispatcher\ContainerAwareEventDispatcher->dispatch() (line 94 of
/var/www/html/drustack/core/lib/Drupal/Component/EventDispatcher/ContainerAwareEventDispatcher.php).
Error: Call to undefined method Symfony\Component\HttpKernel\Event\PostResponseEvent::setDispatcher() in /var/www/html/drustack/core/lib/Drupal/Component/EventDispatcher/ContainerAwareEventDispatcher.php on line 94
Error: Call to undefined method Symfony\Component\HttpKernel\Event\PostResponseEvent::setDispatcher() in Drupal\Component\EventDispatcher\ContainerAwareEventDispatcher->dispatch() (line 94 of /var/www/html/drustack/core/lib/Drupal/Component/EventDispatcher/ContainerAwareEventDispatcher.php).
```
I use the following setup:
 drush status
 Drupal version                  :  8.1.1
 PHP executable                  :  /usr/bin/php
 PHP configuration               :  /etc/php/7.0/cli/php.ini
 PHP OS                          :  Linux
 Drush script                    :  /home/mine/.composer/vendor/drush/drush/drush.php
 Drush version                   :  8.0.2
 Drush temp directory            :  /tmp
 Drush configuration             :
 Drush alias files               :
 Install profile                 :  drustack
 Sync config path                :  sites/default/files/something/sync

the same happend when using 8-dev.

Now as there was no drush configuration file i copied an example.drushrc.php in the corresponding .drush folder:
```
cp ~/.composer/vendor/drush/drush/examples/example.drushrc.php ~/.drush/drushrc.php
drush cc drush
drush status
```
The ContainerAwareEventDispatcher Error went away and drush started working again.

greg-1-anderson:
We really need to tell folks to **not** install Drush via `composer global require`.  This is currently recommended in http://docs.drush.org/en/master/install-alternative; however, anyone who uses this recommended installation technique will end up mixing the autoloaders for Drush, Drupal 8, and **every other project they installed via `composer global require`** into the same runtime space.  This can cause a wide variety of symptoms, and the workaround shown above cannot be counted on to fix all of the problems that might occur.

greg-1-anderson:
The Drush 8 extension mechanism is deprecated.

PR #1521, opened 2015-08-01
Only apply automatic entity updates if option is set.

jhedstrom:
- Corresponding Drupal core issue: https://www.drupal.org/node/2542748

This can probably be merged prior to the core issue.

### DIFF
--- a/commands/core/drupal/update.inc
+++ b/commands/core/drupal/update.inc
@@ -14,7 +14,8 @@
  */
 require_once DRUSH_DRUPAL_CORE . '/includes/update.inc';
 
-
+use Drupal\Core\Utility\Error;
+use Drupal\Core\Entity\EntityStorageException;
 /**
  * Perform one update and store the results which will later be displayed on
  * the finished page.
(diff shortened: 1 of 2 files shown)


weitzman:
Is there a reason one would not want entity updates? And if so, should the default be 'no entity updates'? That looks unintuitive. Is that how update.php works?

D8 only options get unhidden in a help_alter hook. See sql_drush_help_alter().

jhedstrom:
The core critical is discussing not running them in update.php. However, the premise of that issue is  based on a [bug that was causing them to silently fail](https://www.drupal.org/node/2544954). I think drush should eventually follow core, and if core removes automatic updates from update.php, drush should add an option to run them.

commands/core/drupal/update.inc
```diff
@@ -216,6 +220,26 @@ function drush_update_batch() {
   \Drupal::service('state')->set('system.maintenance_mode', FALSE);
 }
 
+/**
+ * Apply entity schema updates.
+ */
+function drush_update_entity_definitions(&$context) {
```
plach79: IMO it would be great to have a also a standalone command for this.

jhedstrom: That makes sense to me.

jhedstrom:
I've added this as a separate command (`entity-updates`) as well.

jhedstrom:
https://www.drupal.org/node/2542748 has been committed. Any last thoughts before I merge this?

weitzman:
Nope - go for it.

plach79:
RTBM on my end

commands/core/drupal/update.inc
```diff
@@ -216,6 +220,26 @@ function drush_update_batch() {
...
+  try {
+    \Drupal::service('entity.definition_update_manager')->applyUpdates();
```
plach79: Minor thing: now we have `\Drupal::entityDefinitionUpdateManager()` :)

jhedstrom: Good to know! I've updated the PR to use that, and will merge after tests run.

jhedstrom:
The failing test looks to be an unrelated issue with the libraries module. Committing this now.

PR #1522, opened 2015-08-01
Add a storage wrapper / storage filter to Drush config import and export (2)

greg-1-anderson:
This is a continuation of #1430, re-opened on a new branch because GitHub.

* I renamed --ignored-modules to --skip-modules to avoid conflict with existing option of the same name.
* I made the storage filter system hookable, so other commandfiles can add additional filters.
* I started writing documentation, split into two topics (using and extending).

The documentation isn't done yet, but I wanted to put this up. I should be able to finish it up on Monday.


greg-1-anderson:
I managed to finish up the documentation this morning.  Let me know if you'd like to see anything else before this goes in.

commands/core/docs.drush.inc
```diff
@@ -65,6 +65,22 @@ function docs_drush_command() {
     'callback' => 'drush_print_file',
     'callback arguments' => array($docs_dir . '/examples/example.drushrc.php'),
   );
+  $items['docs-filter-config'] = array(
```
weitzman: lets do docs-config-filter and docs-config-exporting. that will sort the commands closer and will help with disambiguation versus drupal filter system.

commands/core/config.drush.inc
```diff
@@ -104,6 +106,7 @@ function config_drush_command() {
         'example-value' => 'branchname',
       ),
       'destination' => 'An arbitrary directory that should receive the exported files. An alternative to label argument',
+      'skip-modules' => 'A list of modules to ignore during export (e.g. to avoid listing dev-only modules in exported configuration).',
```
weitzman: we should add an Example on import and export commands.

docs/exporting-config.md
```diff
@@ -0,0 +1,61 @@
...
+based on the environment, how to have a different set of enabled modules
+in different configurations without affecting your exported configuration
+values, and how to make more complex changes.
+
+## Simple Value Changes
+
+It is not necessary to alter the configuration system values to 
+make simple value changes to configuration variables, as this may be
+done by the configuration override system.
+
```
weitzman: Lets link to https://github.com/gherkins/regexpbuilderphp. Thats a really good doc page. We might even summarize its contents here (I'm thinking that module provided overrides can be relevant).

docs/filtering-config.md
```diff
@@ -0,0 +1,54 @@
...
+details; note, however, that it is not necessary for your commandfile
+to implement hook_drush_command(), or any other hook besides the storage
+filter hook.
+
+You will need a composer.json file as well, in order to define where
+your StorageFilter class is defined.  Make sure that Drush and your
+custom commandfile are required from the composer.json file of any
+Drupal site that you plan on using your filter with.
+
+## Implementing the Storage Filter Hook
```
weitzman: Give the existence of module overrides (https://www.drupal.org/node/1928898), I'm no longer sure if this hook system is needed. Perhaps the support burden is not worth it?

greg-1-anderson:
The advantage of implementing a configuration override filter in a Drush commandfile instead of in a module is that it is easier to provide options to Drush commands.  If there was a good way to do the same thing using a module-based configuration filter (that is, provide a value via a Drush option to `drush cex` or `drush cim` that is used in the module configuration filter), then perhaps we could get rid of the hook here.

weitzman:
I'm thinking that if you are coding a module for your own site's workflow, you can hard code behavior so that options are rarely/never needed.

greg-1-anderson:
What about testing for environment - dev / test / live / CI?  Sometimes you can rely on environment variables (e.g. 'PANTHEON_ENVIRONMENT'); is that good enough for the general case?

If you think so, I could just remove the 'filtering' docs (w/ a slight enhancement to the config exporting docs), and remove the hook.

weitzman:
Yes, I think we can assume the presence of environment variables that indicate Dev/Stage/Prod

greg-1-anderson:
Took it out, and added a comment to help put it back in, should it ever be needed.

weitzman:
Thanks ... Feel free to merge once tests come back green

ericduran:
@greg-1-anderson @weitzman I came here looking for `drush_command_invoke_all('drush_storage_filters');`

I'm working on a solution to allow a multi-site set up to share a majority of configs (Entity/Bundles/Fields/etc..) but keep certain specific settings to be unique per site such as `system.site.yml` in its own dir.

Right now it'll need to write new drush commands as it stands to selective import/export certain configurations into certain directories. That being said wit the addition of `drush_command_invoke_all('drush_storage_filters');` I can easily implement my own storage filters to figure out what goes where.

Thoughts? Concern?

Should I open a PR? Move the discussion to a new issue?

greg-1-anderson:
There are issues with the storage filter when configuration entities have dependencies; see #1820. Not sure that you will run into this problem for your use case, but you might, so be aware.

Other than that, your use-case looks interesting. It would be a nice bonus if the implementation did not require a multisite setup, and would also work for sharing partial configuration between different distinct sites as well.

weitzman:
I'm interested in the use case too. @ericduran - please open an issue here or link us to elsewhere where we can discuss. Thanks!

Issue #1474, opened 2015-07-06
Using composer installed drush


welly:
Hello,

Having a slight issue with installing a site-specific composer installed version of Drush. The problem is that it appears to not be able to recognise the Drupal installation.

I have a site install structure like:

```
/home/web/example
/home/web/example/build
/home/web/example/build/bin/drush
/home/web/example/build/docroot
/home/web/example/build/vendor
/home/web/example/build/composer.json
/home/web/example/build/composer.lock
```

When I run /home/web/example/build/bin/drush status, I get the following returned:

```
 PHP executable         :  /usr/bin/php
 PHP configuration      :  /etc/php.ini
 PHP OS                 :  Linux
 Drush script           :  /home/web/example/build/vendor/drush/drush/drush.php
 Drush version          :  7.0.0
 Drush temp directory   :  /tmp
 Drush configuration    :
 Drush alias files      :
```

I've tried creating an alias file in /home/web/example/build/docroot/sites/default/drushrc.php and using:

```
/home/web/example/build/bin/drush --local $@ status
```
which returns

```
 Drupal version         :  7.37
 Site URI               :  http://default
 PHP executable         :  /usr/bin/php
 PHP configuration      :  /etc/php.ini
 PHP OS                 :  Linux
 Drush script           :  /home/web/example/build/vendor/drush/drush/drush.php
 Drush version          :  7.0.0
 Drush temp directory   :  /tmp
 Drush configuration    :  /home/web/example/build/docroot/sites/default/drushrc.php
 Drush alias files      :
 Drupal root            :  /home/web/example/build/docroot
 Site path              :  sites/default
```

So basically I'm pretty stumped and have not found anything that would suggest what I'm doing is incorrect. Any help/advice would be appreciated.

Cheers

weitzman:
Drush can't yet find Drupal root automatically. Pass --root option. You can copy the script at examples/drush to wherever you'd like - and edit as needed.

welly:
Have given it the --root option and I'm still not having any joy:

```
-bash-4.1$ /home/web/example/build/bin/drush --root=/home/web/example/build/docroot/ status
 Drupal version         :  7.38
 Site URI               :  http://default
 PHP executable         :  /usr/bin/php
 PHP configuration      :  /etc/php.ini
 PHP OS                 :  Linux
 Drush script           :  /home/web/example/build/vendor/drush/drush/drush.php
 Drush version          :  7.0.0
 Drush temp directory   :  /tmp
[...]

PR #1440, opened 2015-06-11
Avoid PHP bug in 7.x branch

SchnWalter:
I have recently upgraded Drush to 7.0.0 in order to use the improved version of `drush make`, and I stumbled upon a small bug on the staging server.

I would like to kindly ask you to merge this pull requests. The two changes made avoid [this](https://bugs.php.net/bug.php?id=52854) PHP 5.3.3 bug. The bug has been fixed in PHP 5.3.4.

I know PHP 5.3 has [reached EOL](https://php.net/eol.php) in 14 Aug 2014. But some distributions that ship with 5.3, like RHEL6 (and SL6) are still supported for a while. ([October 31, 2016](https://access.redhat.com/support/policy/updates/errata/#Life_Cycle_Dates), if you don't count the extended life support which goes up to November 30, 2020)

Thank you.

Issue #544, opened 2014-03-21
7.x does not work with PHP 5.3.3


MadMind:
There is a bug in PHP 5.3.3 https://bugs.php.net/bug.php?id=52854 that causes drush to fail with the following error:

```
ReflectionException: Class Drush\Sql\Sql7 does not have a constructor,
so you cannot pass any constructor arguments in
ReflectionClass->newInstanceArgs()
(line 104 of drush/includes/drush.inc).
```

heddn:
I'm getting the above error as well using the official RHEL PHP distribution. https://github.com/drush-ops/drush/issues/544#issuecomment-62539791 fixed the issue.

chirale:
[diarmy solution](https://github.com/drush-ops/drush/issues/544#issuecomment-62630721) works fine in my case on php 5.3.3.

jakeman2048:
https://github.com/drush-ops/drush/issues/544#issuecomment-62539791 worked for me as well.

Can we either have drush7 minimum required PHP version changed to 5.3.4+ or patched in order fully work with PHP 5.3.0+?

norwegianblue-nz:
"Pretty sure this came up before and we are replying to folks that they should upgrade their PHP"
Poor answer. Not all users have the ability to upgrade their php. esp if on shared hosting for instance.
As there seems to be a soln (according to comments re a patch in this thread) why can it not by incorporated into the official code.

taz77:
Not a poor answer. PHP 5.3 has been end-of-life for 10 months now. If you are on a host using 5.3 then I think it would be prudent to look for a new host because running anything end of life on the web is a bad idea - no security patches, no updates!  

https://secure.php.net/eol.php

heddn:
While it isn't supported by php.net, redhat does support it on its LTS. Which is why it is important to me. My client wants to use RHEL.

heddn:
I hadn't noticed, but I guess I'll need to use an older version of PHP. This is just one of the many fun things about supporting an old version of PHP.

<code>includes/preflight.inc:  define('DRUSH_MINIMUM_PHP', '5.4.5');</code>

taz77:
Time to upgrade! :-)

heddn:
I can't upgrade php but I can downgrade drush. These clients insist that a supported version of php because of RHEL  is a supported version. I personally think that using php 5.5 or 5.6 is a much better solution. But I gotta pick my battles.

AndreBaumeier:
[...]

PR #1527, opened 2015-08-03
Fixes PHP 5.3.3 bug for RHEL, SL6, etc

heddn:
Mentioned in #544 & #1440 & #1474. However, those were all closed. I wouldn't bother with yet another patch on this, but I keep running across older sites that I support that the client isn't able to upgrade for a variety of reasons. These are running RHEL and tell me that they still have support.


weitzman:
Still Won't Fix as per #544.

taz77:
PHP 5.3.x went end of live 11 months ago. https://secure.php.net/eol.php
Remember how you how you hate Internet Explorer 8 as a web developer because it was old but still hanging around. Let PHP 5.3 die, upgrade, and move on. Despite what OS vendors may be doing, it is not supported by the PHP Group.

Issue #1528, opened 2015-08-03
drush rsync with remote-host but different username does not work


Schnitzel:
I've got some situations like that:

Local Server Hostname: backend.zh2.single.amazee.io
Local User: sitename_com_dev
Files folder: /var/www/sitename_com_dev/public_html/sites/default/files
File Permission of files folder: 700

Remote Server Hostname: backend.zh2.single.amazee.io
Remote User: sitename_com_prod
File Permission: 700
Files folder: /var/www/sitename_com_prod/public_html/sites/default/files
File Permission of files folder: 700

The site aliases is setup like that:
```
$aliases["prod"] = array (
  'remote-host' => 'backend.zh2.single.amazee.io',
  'remote-user' => 'sitename_prod',
);
```

Now I'm trying to run an rsync:
```
sitename_com_dev@backend.zh2.single.amazee.io:~/public_html$ drush rsync @prod:%files default:%files
```

But it fails because it tries to run an rsync like that:
```
Calling system(rsync -e 'ssh' -akzv --exclude=".git" --exclude=".gitignore" --exclude=".hg" --exclude=".hgignore" --exclude=".hgrags" --exclude=".bzr" --exclude=".bzrignore" --exclude=".bzrtags" --exclude=".svn" --stats --progress -d /var/www/sitename_com_prod/public_html/sites/default/files/ /var/www/sitename_com_dev/public_html/sites/default/files);
```
The error:
`rsync: change_dir "/var/www/sitename_com_prod/public_html/sites/default/files" failed: Permission denied (13)`

the problems comes from https://github.com/drush-ops/drush/blob/master/includes/sitealias.inc#L1939 which determines that if the remote hostname is the same as the local hostname it assumes that an rsync can be made without a remote hostname in there, which in a case of a different username will fail.

jonhattan:
I think drush_is_local_host() is correct here. This avoids undesired ssh connections when you're using the same alias definition in local and remote (so you want remote-host in local but it is useless in the remote env).

IMO a better fix is to add a condition like `drush_get_username() !=  $site_alias_settings['remote-user']`

jonhattan:
@greg-1-anderson I wonder if the check for username matching should be another condition in the if statement (proposed in #1636), or maybe it should be done in `drush_is_local_host()`.

[...]

PR #1529, opened 2015-08-03
Remove drush_is_local_host check from drush_sitealias_evaluate_path

Schnitzel:
as it breaks rsync, see https://github.com/drush-ops/drush/issues/1528

### DIFF
--- a/includes/sitealias.inc
+++ b/includes/sitealias.inc
@@ -1936,7 +1936,7 @@ function drush_sitealias_evaluate_path($path, &$additional_options, $local_only 
     drush_sitealias_set_alias_context($site_alias_settings);
 
     // Use 'remote-host' from settings if available; otherwise site is local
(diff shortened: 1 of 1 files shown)


includes/sitealias.inc
```diff
@@ -1936,7 +1936,7 @@ function drush_sitealias_evaluate_path($path, &$additional_options, $local_only
     drush_sitealias_set_alias_context($site_alias_settings);
 
     // Use 'remote-host' from settings if available; otherwise site is local
```
davereid: Don't we need to adjust this comment?

greg-1-anderson:
I agree with [Jonhattan's comment](https://github.com/drush-ops/drush/issues/1528#issuecomment-132916754); drush_is_local_host is necessary here. A PR that also checked the username would be welcome.

PR #1530, opened 2015-08-03
Remove drush_is_local_host check from drush_sitealias_evaluate_path

Schnitzel:
as it breaks rsync, see https://github.com/drush-ops/drush/issues/1528


jonhattan:
Closing along with #1529

PR #1537, opened 2015-08-06
Prevent site alias discovery from recursing into the files directory.

jhedstrom:
This is an alternative fix for #1536, which broke the ability to specify aliases in `sites/default/foo.aliases.drushrc.php`.

Now with tests!

### DIFF
--- a/includes/sitealias.inc
+++ b/includes/sitealias.inc
@@ -625,11 +625,15 @@ function _drush_sitealias_find_alias_files($aliasname = NULL, $alias_path_contex
     $alias_files[] = '/' . preg_quote($aliasname, '/') . '\.alias\.drush(' . DRUSH_MAJOR_VERSION . '|)rc\.php$/';
   }
 
-  // Search each path in turn
+  // Do not scan into the files directory.
+  $blacklist = array_merge(array('files'), drush_filename_blacklist());
+
+  // Search each path in turn.
   foreach ($alias_path as $path) {
     // Find all of the matching files in this location
     foreach ($alias_files as $file_pattern_to_search_for) {
-      $alias_files_to_consider = array_merge($alias_files_to_consider, array_keys(drush_scan_directory($path, $file_pattern_to_search_for, drush_filename_blacklist(), 0, TRUE)));
+      drush_log(dt('Scanning into @path for @pattern', array('@path' => $path, '@pattern' => $file_pattern_to_search_for)), 'sitealias');
+      $alias_files_to_consider = array_merge($alias_files_to_consider, array_keys(drush_scan_directory($path, $file_pattern_to_search_for, $blacklist, 0, TRUE)));
     }
   }
 
--- a/tests/siteAliasTest.php
+++ b/tests/siteAliasTest.php
@@ -19,7 +19,7 @@ class saCase extends CommandUnishTestCase {
    */
   function testDispatchStrictOptions() {
     $aliasPath = UNISH_SANDBOX . '/site-alias-directory';
-    mkdir($aliasPath);
+    file_exists($aliasPath) ?: mkdir($aliasPath);
     $aliasFile = $aliasPath . '/bar.aliases.drushrc.php';
     $aliasContents = <<<EOD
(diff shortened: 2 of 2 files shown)


tests/siteAliasTest.php
```diff
@@ -19,7 +19,7 @@ class saCase extends CommandUnishTestCase {
    */
   function testDispatchStrictOptions() {
     $aliasPath = UNISH_SANDBOX . '/site-alias-directory';
-    mkdir($aliasPath);
+    file_exists($aliasPath) ?: mkdir($aliasPath);
```
jhedstrom: Note, locally when running these, `UNISH_SANDBOX` was not torn down between each *function* in a class. I have no idea how these are passing on Travis or elsewhere since the sandbox is only cleaned up between each *class*.

greg-1-anderson:
What is our strategy here?  In a way, #1536 seemed like a better solution -- we really do not want to recursively search through everything in the `files` directory, so it seems to me it was a mistake to consider sites/default to be a search path *and* to recursively search there.

If #1536 is undesirable, and we really want to maintain `sites/default/foo.aliases.drushrc.php` as a valid location for an alias file, maybe we need both recursive and non-recursive alias locations?

Maybe we should re-do the recursive algorithm, so each path `foo` searches `foo` non-recursively (only immediate files considered) and `foo/aliases` recursively (deep searching).  Or something like that.

jhedstrom:
I'm up for either this, or #1536--I only closed that one since the test for `sites/default/bla.aliases.drushrc.php` had been in place for a very long time, so removing that ability might break things for folks.

includes/sitealias.inc
```diff
@@ -625,11 +625,15 @@ function _drush_sitealias_find_alias_files($aliasname = NULL, $alias_path_contex
...
-  // Search each path in turn
+  // Do not scan into the files directory.
+  $blacklist = array_merge(array('files'), drush_filename_blacklist());
+
+  // Search each path in turn.
   foreach ($alias_path as $path) {
     // Find all of the matching files in this location
     foreach ($alias_files as $file_pattern_to_search_for) {
-      $alias_files_to_consider = array_merge($alias_files_to_consider, array_keys(drush_scan_directory($path, $file_pattern_to_search_for, drush_filename_blacklist(), 0, TRUE)));
+      drush_log(dt('Scanning into @path for @pattern', array('@path' => $path, '@pattern' => $file_pattern_to_search_for)), 'sitealias');
```
weitzman: We should remove this before merge? Perhaps adds more noise than helpfulness.

jhedstrom: If we're certain this won't happen again, I'm fine removing that--when I was debugging this though, having that info would have saved an hour at least of manual debugging on a remote server w/o xdebug.

Schnitzel:
I would vote for that, or actually to be honest just have a solution rather earlier then later, as this issue affects our whole system quite heavy

jhedstrom:
@greg-1-anderson do you have thoughts on which way we should go? This PR has a test, but that could be moved over to the other one too.

greg-1-anderson:
I'm not opposed to merging this PR as a stopgap measure, but I've seen some folks put large dirs such as 'private' or 'tmp' in sites/default, so I don't think this is the final solution. I think the best solution would be to have some historic paths, such as sites/default, where Drush considers only the immediate contents, and other paths, such as */drush, where Drush will search deep.  Maybe predicate searching deep on whether or not the path contains "drush"?

Not sure what @weitzman will think about the complexity of having both deep and shallow searching, though.

jhedstrom:
I've merged this fix. We can remove the added verbosity if it proves too noisy. We can also follow-up with deep/shallow searching.

Schnitzel:
nice :) thanks a lot

Schnitzel:
btw before merge a normal `drush status` took:
```
real	0m19.825s
user	0m7.772s
sys	0m8.333s
```

after the merge:
```
real	0m5.007s
user	0m1.140s
sys	0m0.294s
```

PR #1555, opened 2015-08-17
Register the runtime assertion handler for PsySH

damiankloip:
This is just getting ready for when https://www.drupal.org/node/2536560 is committed into 8.0.x . Once that is in, this will work, then we have a tiny patch that will actually fix the problem consistently by registering that handler in drush php : https://www.drupal.org/node/2550641 .

So we need to hold off until the first issue in done at least.

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -20,10 +20,13 @@ function drush_cli_core_cli() {
   $shell = new Psy\Shell();
 
   if (drush_drupal_major_version() >= 8) {
+    // Register the assertion handler so exceptions are thrown instead of errors
+    // being triggered. The plays nicer with PsySH.
+    \Drupal\Component\Assertion\Handle::register();
     $shell->setScopeVariables(['container' => \Drupal::getContainer()]);
   }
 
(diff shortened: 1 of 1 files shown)


commands/core/cli.drush.inc
```diff
@@ -20,10 +20,13 @@ function drush_cli_core_cli() {
   $shell = new Psy\Shell();
 
   if (drush_drupal_major_version() >= 8) {
+    // Register the assertion handler so exceptions are thrown instead of errors
+    // being triggered. The plays nicer with PsySH.
```
weitzman: typo. should be This.

damiankloip: Thanks. fixed!

damiankloip:
https://www.drupal.org/node/2536560 is now in, just need https://www.drupal.org/node/2550641 now.

weitzman:
Those 2 issues are in. This is now ready for merge?

damiankloip:
@weitzman This is good to go. I also just double checked with 8.0.x HEAD and this branch all :+1:

damiankloip:
Awesome!

Issue #1561, opened 2015-08-21
drush arr not restoring correctly.


sherakama:
PHP: 5.5
Mysql: 5.6
Drush: 7.0.0

When calling drush arr when the database does not exist drush incorrectly tries to import into the 'information_schema' table. When the database does exist prior to calling drush arr everything works as it should.

Steps to reproduce:
1. Backup a site using `drush ard`
2. Run the below drush command but change the database_name to something new so that drush arr has to create a db
3. Review debug output.

Command: 
drush arr restore.tar.gz --db-url=mysql://root:root@localhost/database_name --destination=/site-name --debug --overwrite

Debug output:
Executing: mysql --defaults-extra-file=/private/tmp/drush_EcqCxQ --database=information_schema --host=localhost --silent  < /private/tmp/drush_9NRP8Z
Executing: mysql --defaults-extra-file=/private/tmp/drush_7Cuj8m --database=information_schema --host=localhost --silent  < /tmp/drush_tmp_1440188457_55d788299ec51/dbfilename.sql

Looks like it defaults here: https://github.com/drush-ops/drush/blob/7.x/lib/Drush/Sql/Sqlmysql.php#L35

sherakama:
It looks like line https://github.com/drush-ops/drush/blob/7.x/lib/Drush/Sql/SqlBase.php#L361 is removing the database option. Not sure why this is.

sherakama:
Ah, it seems to empty the database setting because we do not want to connect to a database when the sql statement is executed. Instead it defaults to information_schema database which doesn't allow write permissions.

PR #1565, opened 2015-08-24
ISSUE #1561: Fixes create and restore database with drush arr

sherakama:
Suggested fix for issue #1561

### DIFF
--- a/lib/Drush/Sql/SqlBase.php
+++ b/lib/Drush/Sql/SqlBase.php
@@ -239,11 +239,22 @@ class SqlBase {
    *   in a Windows shell. Set TRUE if the CREATE is not running on the bash command line.
    */
   public function createdb($quoted = FALSE) {
+    $original_spec = $this->db_spec['database'];
     $dbname = $this->db_spec['database'];
     $sql = $this->createdb_sql($dbname);
+
     // Adjust connection to allow for superuser creds if provided.
     $this->su();
-    return $this->query($sql);
+
+    // Perform the query.
+    $result = $this->query($sql);
+
+    // the su() command can change the database name to an empty string which
+    // can effect further operations. Restore the original db_spec to continue.
+    $this->db_spec = $original_spec;
+
+    // Return the result of the query.
+    return $result;
   }
 
   /**
@@ -257,10 +268,7 @@ class SqlBase {
       $this->drop($this->listTables());
     }
     else {
-      $original_spec = $this->db_spec;
       $this->createdb();
-      // Restore the original spec as create db and su() can modify the spec.
-      $this->db_spec = $original_spec;
     }
   }
 
@@ -362,6 +370,10 @@ class SqlBase {
     $create_db_target = $this->db_spec;
 
     $create_db_target['database'] = '';
+    // Set the new-database option to prevent including the database on the
+    // sql connection parameters.
+    $create_db_target['new-database'] = TRUE;
+
     $db_superuser = drush_get_option('db-su');
     if (isset($db_superuser)) {
       $create_db_target['username'] = $db_superuser;
--- a/lib/Drush/Sql/SqlBase.php
+++ b/lib/Drush/Sql/SqlBase.php
@@ -257,7 +257,10 @@ class SqlBase {
       $this->drop($this->listTables());
     }
     else {
+      $original_spec = $this->db_spec;
       $this->createdb();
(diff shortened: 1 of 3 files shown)


lib/Drush/Sql/SqlBase.php
```diff
@@ -257,7 +257,10 @@ public function drop_or_create() {
       $this->drop($this->listTables());
     }
     else {
+      $original_spec = $this->db_spec;
       $this->createdb();
+      // Restore the original spec as create db and su() can modify the spec.
+      $this->db_spec = $original_spec;
```
weitzman: I didnt investigate, but it is a bit odd to wrap createdb here. Can this dance not be done inside createdb()?

lib/Drush/Sql/Sqlmysql.php
```diff
@@ -32,7 +32,9 @@ public function creds($hide_password = TRUE) {
 
     // Some drush commands (e.g. site-install) want to connect to the
     // server, but not the database.  Connect to the built-in database.
-    $parameters['database'] = empty($this->db_spec['database']) ? 'information_schema' : $this->db_spec['database'];
+    if (!empty($this->db_spec['database'])) {
```
weitzman: I'm a little hesitant to remove information_schema. I think we need to specify a DB in certain cases (see early parts of site-install command). Needs investigation.

lib/Drush/Sql/SqlBase.php
sherakama: That might be a better approach. I'll take a look to see.

lib/Drush/Sql/Sqlmysql.php
sherakama: information_schema cannot be written to. Read is fine, but I wonder if we can protect this from defaulting to information_schema when it is a write operation.

weitzman: In that case the caller should have provided a database. I'm not too concerned about this.

sherakama:
Ok, moved the db_spec restore to the createdb method and added a flag for when creating a new database to avoid adding the --database parameter to the mysql statement. 

Debug output now looks like:

```
Executing: mysql --defaults-extra-file=/private/tmp/drush_OlVRJz --database=mydatabase --host=localhost --silent  < /private/tmp/drush_L8XnDW > /dev/null

Executing: mysql --defaults-extra-file=/private/tmp/drush_dgBIe0 --host=localhost --silent  < /private/tmp/drush_NnPnGb

Executing: mysql --defaults-extra-file=/private/tmp/drush_YEIDG2 --database=mydatabase --host=localhost --silent  < /tmp/drush_tmp_1440615699_55de0d13b289c/ds_jumpstart_cars.sql
```

You can see that the second mysql statement does not include the database parameter as that one has the create new db statement in it.

lib/Drush/Sql/SqlBase.php
```diff
@@ -239,11 +239,26 @@ public function createdb_sql($dbname, $quoted = FALSE) {}
...
+
+    // Create the query.
+    $result = $this->query($sql);
+
+    // the su() command can change the database name to an empty string which
+    // can effect further operations. Restore the original db_spec to continue.
+    $this->db_spec = $original_spec;
+
+    // Return the result of the query.
+    return $result;
```
weitzman: This hunk looks good to me.

lib/Drush/Sql/Sqlmysql.php
```diff
@@ -34,6 +34,12 @@ public function creds($hide_password = TRUE) {
     // server, but not the database.  Connect to the built-in database.
     $parameters['database'] = empty($this->db_spec['database']) ? 'information_schema' : $this->db_spec['database'];
 
+    // If we are creating a new database we don't want to include the database
+    // parameter for the mysql statement.
+    if (isset($this->db_spec['new-database']) && $this->db_spec['new-database']) {
+      unset($parameters['database']);
+    }
+
```
weitzman: Maybe creds() should always return a full array and it is up to the caller to remove 'database' if needed?

sherakama: Outside of creating or dropping a database I don't see a need to remove the database option. Perhaps changing the key from 'new-database' to 'no-database' or something similar?

sherakama:
Moved this to: https://github.com/drush-ops/drush/pull/1872

Issue #1491, opened 2015-07-14
make_download_copy is gone, why?


jurgenhaas:
Why is the make_download_copy gone in drush 7.x? I sued to use that for custom modules that were available in a local directory and should be copied over to the Drupal tree. Is there a better way available for that now?

weitzman:
I'm not sure why, but we have a few copy features in includes/filesystem.inc. There are also native php and bash commands for copying.

jurgenhaas:
Sorry, those don't help. In a make file something like this was possible in 6.x:

```
projects[crc][type] = "module"
projects[crc][download][type] = "copy"
projects[crc][download][url] = "file:///path/to/dev/code"
projects[crc][subdir] = "custom"
```

This is now silently ignored in 7.x, it just creates the directory but doesn't copy the files inside that directory anymore.

jhedstrom:
Hmm, local patches still work (as there are tests for that).

charlietoleary:
The above example use case fits well: I would say most people using the 'copy' functions are using in-development code.

Some potential issues that I would say including copy functions fixes:
* There is no way to achieve the 'copy' functionality using drush make without doing custom scripting or extending drush.
* Migrating drush make 6.x -> 7.x will deprecate all copy calls, creating a barrier to upgrade.
* 'file' downloading only works with single files.
* git and bazaar types rely on those respective source-management software suites to work.
* When using git (and potentially bzr aswell, i have never used it), your latest branch will be checked out by default, disregarding local changes and making it extremely hard to test in-development projects. If you do not wish to commit changes yet or are simply testing new changes: bad luck.

jonhattan:
Long story short: #522 was merged only in 6.x branch, not in 7.x neither master.

I've cherry picked d7bce8f1eb9ec7b763015f4e7324d86e663b9f5c and 1d516580873c8cd0d870654f2df04aac7e5218c6

PR #1566, opened 2015-08-25
Issue #1491 workaround - make_download_copy functions missing from 7.x

charlietoleary:
make_download_copy functions are missing from the 7.x branch.

included them as-is from the 6.x branch for compatibility: the functions still seem to work fine.


weitzman:
Please work with Make maintainers in #1491 to justify this. We dont just "put back a function" without understanding why it left and what the use case is. Thanks.

We can reopen this PR if needed.

Issue #1560, opened 2015-08-20
Provide better info when The external command could not be executed due to an application error.


greggles:
I was debugging a drush site/command combination that was getting the infamous error:

"The external command could not be executed due to an application error."

Which is not so useful.

What I found was that several people are able to debug this problem by using either --verbose or --debug. 

Eventually I ended up going into backend.inc and printing out the ```$proc``` which helped me see that the problem in the ```$proc['output']``` was a fatal error in requiring a file.

So...
1. The error message could be amended to suggest using --debug or --verbose to help diagnose the problem.
2. If someone has used --debug when invoking the command AND this error is about to be set, it seems worthwhile to log the ```$proc['output']```

greg-1-anderson:
Generally, when this error message appears, Drush does not have much additional info to go on, and `$proc['output']` should, I think, already be printed.  Adding advice to run again with --debug sounds reasonable, though.

greggles:
The error I hit is caught and displayed in newer versions of drush.

danlobo02:
I had this error. The first thing I noticed it was caused by a memory error. After some investigation i found that passing:

$backend_options = ['invoke-multiple' => '1']; through to drush_invoke_process() when calling a migration script resolved the issue for me. I do not understand why this is necessary but managed to reduce my memory consumption by a factor of 8. Strangely the same was true when passing though as a default of 4.

PR #1567, opened 2015-08-25
Issue #1560: improved help and debugging info when a command fails

greggles:
For https://github.com/drush-ops/drush/issues/1560

I didn't make a branch for this - sorry :/ I can do that if you'd like.

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -1022,7 +1022,8 @@ function _drush_backend_invoke($cmds, $common_backend_options = array(), $contex
     $ret = array();
     foreach ($procs as $site => $proc) {
       if (($proc['code'] == DRUSH_APPLICATION_ERROR) && isset($common_backend_options['integrate'])) {
-        drush_set_error('DRUSH_APPLICATION_ERROR', dt("The external command could not be executed due to an application error."));
+        drush_set_error('DRUSH_APPLICATION_ERROR', dt("The external command could not be executed due to an application error. Try running the command again with the --debug option."));
+        drush_log(dt('The raw command and output is !proc', array('!proc' => var_export($proc, TRUE))), 'debug');
       }
 
       if ($proc['output']) {


jonhattan:
The test fail is unrelated. The cause is libraries module with latest core https://www.drupal.org/node/2557419

includes/backend.inc
```diff
@@ -1022,7 +1022,8 @@ function _drush_backend_invoke($cmds, $common_backend_options = array(), $contex
     $ret = array();
     foreach ($procs as $site => $proc) {
       if (($proc['code'] == DRUSH_APPLICATION_ERROR) && isset($common_backend_options['integrate'])) {
-        drush_set_error('DRUSH_APPLICATION_ERROR', dt("The external command could not be executed due to an application error."));
+        drush_set_error('DRUSH_APPLICATION_ERROR', dt("The external command could not be executed due to an application error. Try running the command again with the --debug option."));
+        drush_log(dt('The raw command and output is !proc', array('!proc' => var_export($proc, TRUE))), 'debug');
```
weitzman: 1. $proc can be giant, with log messages, return value, errors, and output. I think this should be $proc['output']? Paging @greg-1-anderson.
1. Furthermore, there is a line later which attempts to log the output. Would be good to understand why that didn't help you. See `$ret = drush_set_error('DRUSH_FRAMEWORK_ERROR', dt("The command could not be executed successfully (returned: !return, code: !code)", array("!return" => $proc['output'], "!code" =>  $proc['code'])));`

greg-1-anderson:
Could you provide the exact situation you encountered that prompted you to submit this?

Note that in the case of a DRUSH_APPLICATION_ERROR, this usually indicates that Drush itself did not start at all, so $proc might be smaller than usual inside this particular conditional.  Not sure that this is always the case, though, and in any event, as Moshe said, `$proc['output']` should have the info you need already.  Let's chase this down before deciding what to do here.

includes/backend.inc
greggles: I debated about whether to dump all of $proc or not. I'd be fine with just output, but I feel like using --debug is asking for "all the data".

greggles:
I can describe the situation, but I'm not sure it will be reproducible on other systems. I can reproduce it locally and in the test environment (this runs as part of a script in test for me).

* sites/default/settings.php had a fatal error in it - require_once of a nonexistent file.
* the alias I was using set keys for root and uri that should not have loaded sites/default/settings.php and instead should have loaded sites/example.backup/settings.php
* I was running `sql-sanitize -y` and oddly enough that command succeeded...but right before it printed the list of things it was going to do...it printed this error message and nothing else
* I was running the command from outside of the docroot

Here's a gist of the debug output from my local environment https://gist.github.com/greggles/2fbb8464cb7eacf63bda

The error is on line 21 and there's no additional output.

I think it makes sense to add the sentence about --debug regardless of outputting `$proc`.

weitzman:
This bug report applies to Drush6. I'll leave it to other interested parties to determine the value of this PR. Feel free to reopen if new info arises.

greggles:
My memory of why I closed this was that we discussed it in irc (or similar) and the feeling by drush maintainers that it wasn't appropriate.

I just helped someone in the #drush slackroom who had a silent fail of drush. They had to go look in syslog to find the problem.

It still seems odd to me that drush silently hides errors instead of showing it to people.

PR #1589, opened 2015-09-02
Bring back sqlite support

### LINKED ISSUE #1336: drush site-install and relative path for SQLite not working
Documentation (http://www.drushcommands.com/drush-7x/site-install/site-install) shows this example to use with SQLite:
```
drush site-install --db-url=sqlite://sites/example.com/files/.ht.sqlite
```
But when I use the above command, I get an error:
```
drush si -y --db-url=sqlite://sites/all/files/.ht.sqlite
You are about to CREATE the '/sites/all/files/.ht.sqlite' database. Do you want to continue? (y/n): y
Directory / exists, but is not writable. Please check directory permissions.                                                        [error]
SQLITE: Cannot create /sites/all/files   
```
^^ it looks like drush is trying to write to the root directory? 
When I use the full path:
```
drush si standard -y --db-url=sqlite://Users/alimac/drupal/sites/all/files/.ht.sqlite
```
installation works. 

Thing is, the relative path worked for me a couple days ago. I did `composer global update` and updated drush yesterday. Can anyone else reproduce this issue?

jonhattan:
* Fix sqlite support. Reported in #1336. Broke with 8027687a96fdebc0d71527cbc8702e3dad5bc0ea
 * Add support for absolute urls. We did never supported it before: parse_url() is unable to parse them.
 * Prevent drush_sql_get_class() from trying to load a class if db_spec is not valid ("Unable to load class Drupal\Driver\Database\\Drush")

### DIFF
--- a/commands/sql/sql.drush.inc
+++ b/commands/sql/sql.drush.inc
@@ -570,14 +570,16 @@ function drush_sql_get_class($db_spec = NULL) {
   $target = drush_get_option('target', 'default');
 
   // Try a few times to quickly get $db_spec.
-  if ($db_spec) {
+  if (!empty($db_spec)) {
     return drush_get_class(array('Drush\Sql\Sql', 'Drupal\Driver\Database\\' .  $db_spec['driver'] . '\Drush'), array($db_spec), array($db_spec['driver']));
   }
   elseif ($url = drush_get_option('db-url')) {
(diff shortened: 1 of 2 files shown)


weitzman:
The fails are unrelated. This looks fine to merge. Thanks.

PR #1592, opened 2015-09-04
Add a path for commandfiles that is loaded during locate_root

greg-1-anderson:
Allow potential Drush-compatible systems to put their bootstrap class hooks at __ROOT__/drush/bootstrap.

For backdrop/backdrop-issues#47

### DIFF
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -323,11 +323,6 @@ function drush_locate_root($start_path = NULL) {
     if ($follow_symlinks && is_link($path)) {
       $path = realpath($path);
     }
-    // Check to see if we have any bootstrap classes in this location.
-    $bootstrap_class_dir = $path . '/drush/bootstrap';
-    if (is_dir($bootstrap_class_dir)) {
-      _drush_add_commandfiles(array($bootstrap_class_dir), DRUSH_BOOTSTRAP_DRUSH);
-    }
     // Check the start path.
     if (drush_valid_root($path)) {
       $drupal_root = $path;
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -323,6 +323,11 @@ function drush_locate_root($start_path = NULL) {
     if ($follow_symlinks && is_link($path)) {
       $path = realpath($path);
     }
+    // Check to see if we have any bootstrap classes in this location.
+    $bootstrap_class_dir = $path . '/drush/bootstrap';
+    if (is_dir($bootstrap_class_dir)) {
+      _drush_add_commandfiles(array($bootstrap_class_dir), DRUSH_BOOTSTRAP_DRUSH);
+    }
     // Check the start path.
     if (drush_valid_root($path)) {
       $drupal_root = $path;
--- a/includes/bootstrap.inc
+++ b/includes/bootstrap.inc
@@ -160,7 +160,12 @@ function drush_get_bootstrap_candidate_classnames() {
   return $candidates;
 }
 
+/**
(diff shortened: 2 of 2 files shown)


jenlampton:
:+1:

jenlampton:
Well, I tried placing both the file containing the hook `backdrop_bootstrap_candidates()` and the file containing the class in both `drush/bootstrap` and `core/drush/bootstrap` but in neither situation was the class loaded. Not sure how to debug why not.

greg-1-anderson:
I fixed my typo -- "drush" was what I meant, not "drupal".

You should be using the master branch for this; I don't know if 7.x works.

includes/environment.inc
```diff
@@ -323,6 +323,11 @@ function drush_locate_root($start_path = NULL) {
     if ($follow_symlinks && is_link($path)) {
       $path = realpath($path);
     }
+    // Check to see if we have any bootstrap classes in this location.
+    $bootstrap_class_dir = $path . '/drush/bootstrap';
+    if (is_dir($bootstrap_class_dir)) {
+      _drush_add_commandfiles(array($bootstrap_class_dir), DRUSH_BOOTSTRAP_DRUSH);
```
weitzman: 1. The better constant is DRUSH_BOOTSTRAP_NONE.
1. Now this function does two things. Would be better to move this commandfile search elsewhere. I don't know where offhand.

greg-1-anderson: Loading the bootstrap object is an integral part of finding the root; we couldn't really move it somewhere else. The best we could do is to refactor drush_locate_root into two functions, and make the first one drush_locate_bootstrap_class.  This would have a performance impact, since we'd be walking up the tree twice instead of once. Perhaps this would not be significant due to filesystem caching.  I can do the refactor if you think it significant.

Also, I could refactor _drush_add_commandfiles so that we had another entrypoint into it that searched for a more limited pattern of commandfiles, e.g. foo.drush-bootstrap.inc, or whatever.  This would avoid accidentally loading general-purpose commandfiles that were mis-placed in this directory.  We could also improve performance by limiting recursive searching at this point -- which would only be a problem if a user made the mistake of misplacing a large folder in a 'bootstrap' directory.

jonhattan:
`drush_bootstrap_class_for_root($path)` is a better place to try and load the bootstrap commandfile.

greg-1-anderson:
Jonhattan is right. I moved the function as suggested.  This means that it is logically still part of locate_root, but perhaps that is okay with this implementation.

I also latched the function, so that we don't try to search the same location more than once.  Maybe this optimization should be pushed down to _drush_add_commandfiles().

weitzman:
Looks ok to me.

greg-1-anderson:
Merged -- thanks.

jenlampton:
Hm. I was under the impression that -master wouldn't work with Backdrop and we needed to use the 7.x branch instead.  Sorry about that, will try master :)

jenlampton:
Aha, it's working :)  Classes loaded, woohoo :) thanks!

greg-1-anderson:
The path you specify is correct, and was working last time I tried it.  Using ~/.drush to make progress is a good strategy.  Perhaps we can look at this together sometime, in IRC or whatever.

jenlampton:
@greg-1-anderson sounds good. Maybe we can tackle this at BADCamp :) My Drush branch for Backdrop is at https://github.com/jenlampton/backdrop/tree/47-drush.

PR #1595, opened 2015-09-05
Split 'drush' script into finder, wrapper and launcher stages.

greg-1-anderson:
Related to #1343.  These should probably be combined together, if we decide we want to go with both strategies.

- [x] Test this call chain.
- [x] More end-user focused docs. install.md is already pretty complex, so it will be a bit challenging to work this in smoothly. 
- [x] Moshe needs to test with site aliases, shell aliases, site-set, bash completion (i.e. --early handling)
- [x] Decide about Windows support.


drush
```diff
@@ -1,131 +1,350 @@
-#!/usr/bin/env sh
+#!/bin/sh
+
+#
+# This is the Drush "finder" script, which is one part of the
+# Drush dispatching strategy.  This is the script that
```
weitzman: lets use 'chain' instead of strategy

drush
```diff
@@ -1,131 +1,350 @@
...
+# should appear in your global $PATH, or, if using composer
+# (as usually the case), will be found in your vendor/bin directory.
+#
+#
+# NEVER copy this script to your Drupal root. Use examples/drush instead.
```
weitzman: Lose the CAPS and use list syntax starting with dashes.

drush
```diff
@@ -1,131 +1,350 @@
...
+# the Drush Finder will add the --local flag.  If the --local flag was
+# already passed in when the Drush Finder is called, then it will always
+# call the "Drush launcher", even if there is a Drush Wrapper, to prevent
+# possible infinite loops.
+#
+#
+# DRUSH WRAPPER
+#
+# - The "drush" script that the user copies to their Drupal __ROOT__
+# - Its goal is to allow the user to add config locations when --local is in use
```
weitzman: The term 'config locations' here seems imprecise.

drush
```diff
@@ -1,131 +1,350 @@
...
+# there is no wrapper script.  If the "Drush launcher" is called, then
+# the Drush Finder will add the --local flag.  If the --local flag was
+# already passed in when the Drush Finder is called, then it will always
+# call the "Drush launcher", even if there is a Drush Wrapper, to prevent
+# possible infinite loops.
+#
+#
+# DRUSH WRAPPER
+#
+# - The "drush" script that the user copies to their Drupal __ROOT__
```
weitzman: The "drush.wrapper" script?

drush
```diff
@@ -1,131 +1,350 @@
...
-  # On MSYSGIT, we need to use "php", not the full path to php
-  if [ -n "$MINGW" ] ; then
-    php="php"
-  else
-    # Default to using the php that we find on the PATH.
-    # We check for a command line (cli) version of php, and if found use that.
-    # Note that we need the full path to php here for Dreamhost, which behaves oddly.  See http://drupal.org/node/662926
-    php="`which php-cli 2>/dev/null`"
+# We will look for a "drush wrapper" script that might
+# be stored in the root of a Drupal site.  If there is
```
weitzman: We have been working towards moving references to Drupal to boot classes and commandfiles. Lets omit Drupal here.

weitzman:
I like what I see, and first read. This is a huge change, so I'll want to spend more time with it.

It occurs to me that it would be helpful if the finder script were php. That way, we could additionally package drush as a phar. Maybe thats less attractive in the future as Composer built sites become the norm.

greg-1-anderson:
Drush does some work in the "launcher" script to allow the user to select the version of php to use. If we eliminate this capability, and assume that the user will employ other tools to select the appropriate php executable on their PATH, then it might make sense to go the phar route.

greg-1-anderson:
Wow, this breaks every test in the deck!  Seems that Travis is using a different shell or something, and is getting a syntax error that I'm not seeing on my local system.

If we were using Circle CI, I could run the test again with ssh, and see what was up with the tests. Do you have any strong feelings about sticking with Travis?

greg-1-anderson:
Okay, sh-ified the scripts so they'd run in more diverse environments, just like our existing script.  Tests all green now.

greg-1-anderson:
The site-set code over at #1343 needs to use the same exact algorithm as the Drush finder in order to locate the same site-local Drush that the finder does.  So, yes, I think you are right that the finder should be written in php.

greg-1-anderson:
Here is a straight php port for the drush finder script.  Note that we use pcntl_exec here, so that we can exec the next script with process replacement, and we do not need to worry about escaping the options as we would with exec() or shell_exec().  However, because we cannot rely on PHP populating $_ENV (it isn't on Travis CI, for example), we sometimes have to do a little work to generate a working environment.  Between the trade-offs, I think I prefer it the way it is, with pcntl_exec.

Next step is to factor out the re-usable parts into a separate include file.

greg-1-anderson:
This is working pretty well at this point, except `drush @localsite cmd` does not pass through the drush wrapper for the site @localsite.  It would have to redispatch to do so, and currently does not.

#1343 is more advanced in terms of path management, but still does not solve the `drush @localsite cmd` problem.  Perhaps we can work out all of our issues here, and not need the other PR.  Without #1343, redispatches would always go to the global Drush, but it could figure out the site-specific script to launch on its own.

greg-1-anderson:
`drush @localsite cmd` now correctly passes through the Drush wrapper script.  All of the tests are green, so I think this is "done", but it could still use some healthy ad-hoc testing. I have set up the Drush finder as my global Drush, and will just use it for a while.  Hopefully we can merge this soon, though.

I think we can just close #1343, but please take a quick look at it and see if there is anything there that you think is worth doing.  I think our next steps can move in another direction, though.

greg-1-anderson:
Seems to be working fairly well for me so far. Let me know if you've found any issues.

weitzman:
From http://php.net/manual/en/pcntl.installation.php, "Currently, this module will not function on non-Unix platforms (Windows)."

We need a launching strategy for Windows. I'd rather not drop it cold turkey. Dunno if we can monkey with the .bat file to get something working.

greg-1-anderson:
My opinion is that Drush currently does not support Windows.  We have been trying for years, but we don't have enough Windows maintainers to make it more than sorta-kinda work.

At this juncture, I think we should either recruit a Windows maintainer who is willing to do the launcher / wrapper work for Windows, or finally call it a day and drop support for Windows in Drush.  Maybe someone would actually find time to prioritize Windows support for Drush after we drop it; nothing we have done up to now has been successful.

greg-1-anderson:
Okay, I [tweeted about this](https://mobile.twitter.com/greg_1_anderson/status/642024018803949568). Hopefully folks will boost the signal.

Please limit the discussion on this thread to pcntl and its equivalent support on Windows. If we are going to move forward and seriously consider dropping support for Windows, we will open a separate thread for folks to put in their "I use Drush on Windows, but don't have time/experience to help support it" type comments.  Thanks.

greg-1-anderson:
So, it is also possible, although unlikely, that a Linux / MacOS platform might not have pcntl support. I've always seen this compiled in on these platforms, but it does not have to be.

If we wanted to support non-pcntl systems on Linux / MacOS, I would recommend checking `function_exists("pcntl_exec")` at the end of the Drush finder, and falling back to one of the standard exec / passthru commands if it is not available.  We would then have to escape all of the arguments in order to build up a command string for execution here.  The big irony is that shell escaping does not work well on Windows PHP, so we'd need to do extra work if we wanted to support this on Windows.  We have some code in Drush to do this; we'd just need to move some of it to includes/startup.inc.

drush
```diff
@@ -1,131 +1,300 @@
...
-if [ -n "$PHP_OPTIONS" ] ; then
-  php_options="$php_options $PHP_OPTIONS"
-fi
-
-# Pass in the path to php so that drush knows which one to use if it
-# re-launches itself to run subcommands.  We will also pass in the php options.
-# Important note: Any options added here must be removed  when Drush processes
-# a #! (shebang) script.  @see drush_adjust_args_if_shebang_script()
-exec "$php" $php_options "$SCRIPT_PATH" --php="$php" --php-options="$php_options" "$@"
+#!/usr/bin/env php
```
weitzman: PHPStorm refuses to recognize this file as PHP as it doesn't understand shebangs. Thats inconvenient but whats worse is that I can't step set breakpoints in non-php files. Could we rename this to drush.finder.php or something and then symlink it as 'drush' in vendor/bin? That would also help people get used to executing the /bin/drush file

greg-1-anderson:
Composer does not give us a facility to rename or symlink a script as it is copied into the bin directory.  It does not seem good to add another script just to allow for this, and forcing a foreign symlink into Composer's bin directory seems hacky.

What if we pushed the contents of the `drush` script into includes/startup.inc, as shown in the previous two commits?  (Change and correction.)

The alternative would be to leave the `drush` script as it is, and expect that people will  copy it to drush.finder.php if they want to set breakpoints in it.  That should be fairly rare, I imagine.

chx:
Why don't we use `popen` / `proc_open` ?

greg-1-anderson:
proc_open would be completely viable; in fact, that's what backend invoke uses.  We could use any of proc_open, popen, exec, system or passthru here, and I think that all of these would behave fairly similarly.  The one thing they all share in common is that they take a single `$command` parameter, so we would have to escape the other command arguments as described above, so we'd still need Windows-specific code. If we did go this route, I tend to think that passthru would be the most appropriate variant to use for this situation.

weitzman:
I thought we wanted only one process (the new forked one) and thus were
using pcntl_exec. If that doesn't matter, then we should launch using
Symfony Process component which should be cross-platform.

drush
```diff
@@ -1,131 +1,300 @@
...
+$env = $_ENV;
+
+// If PHP is not configured correctly, $_ENV will be
+// empty.  Drush counts on the fact that environment
+// variables will always be passed through, so we
+// need to repair this situation.
+if (empty($env)) {
+  exec('printenv', $env_items);
+  foreach ($env_items as $item) {
+    list($key, $value) = explode('=', $item, 2);
```
weitzman: One of the $item in my array has no = in it so gives a Notice.

greg-1-anderson:
Yes, I prefer using process replacement with pcntl_exec when it is available.  At the moment, I am trying to limit the Drush "finder" to only use PHP API functions, and things in includes/startup.inc.  However, at some point in the future I'd like to include the autoloader so that we can access alias files from the Drush finder.  Since that is the case, we might as well use the Symfony Process component in the Drush finder if we are going to support non-pcntl systems (which I am not super inclined to do at the moment, but we will see how important that turns out to be).

I pushed another commit to fix the assertion you mentioned in the empty-$_ENV case.

greg-1-anderson:
Gathering feedback about Windows support at #1612.

Does anyone think that non-Windows pcntl support is important?

examples/drush
```diff
@@ -1,13 +1,32 @@
 #!/usr/bin/env sh
```
weitzman: I'm inclined to rename this file to drush.wrapper for disambiguation with the root drush script. The docs can suggest a rename to `drush` when using it.

examples/drush
```diff
@@ -1,13 +1,32 @@
...
 #
-# A wrapper script which launches the Drush that is in your project's /vendor directory.
-# Copy it to the root of your project (or wherever you store handy scripts) and edit as desired.
+# DRUSH WRAPPER
 #
-# Below are options which you might want to add. More info at `drush topic core-global-options`:
+# A wrapper script which launches the Drush that is in your project's /vendor
+# directory.  Copy it to the root of your project (or wherever you store handy
+# scripts) and edit as desired.  If you already have a folder named 'drush',
```
weitzman: I think we need to omit the "or wherever you store handy scripts" bit. The finder only checks specific places.

examples/drush
```diff
@@ -1,13 +1,32 @@
...
-vendor/bin/drush --local $@
+# --local       Only discover commandfiles/site aliases/config that are
+#               inside your project dir.
+# --alias-path  A list of directories where Drush will search for site
+#               alias files.
+# --config      A list of paths to config files
+# --include     A list of directories to search for commandfiles.
+#
+# Note that it is recommended to always use --local when using a drush
+# wrapper script.
```
weitzman: This feels a bit strong. Maybe omit it.

examples/drush
greg-1-anderson: That's a good idea -- particularly since now, you can keep it named 'drush.wrapper' if using it with the drush finder.

includes/startup.inc
```diff
@@ -0,0 +1,260 @@
...
+          break;
+
+        case "-r":
+          $VAR = "ROOT";
+          break;
+
+        case "--verbose":
+        case "-v":
+          $VERBOSE = TRUE;
+          break;
```
weitzman: Might as well consider --debug and -d as verbose as well.

examples/drush
greg-1-anderson: I made the comment strong because without --local in the wrapper, it is possible to get an infinite loop in Drush -- but only if you set it up wrong, so we could take this recommendation out if you prefer.  Or maybe just take out the word "always"?

includes/startup.inc
greg-1-anderson: I've only included flags here that I use in the script.  I could put --debug in for future use, though.

examples/drush
greg-1-anderson: Yes, but if you put it in __ROOT__/vendor/bin, and you have added that to your $PATH, then it will work.  Maybe that's too complicated, and we should just remove this advice, though?

weitzman:
I've now tested this using just a global Drush and also with a site-local drush and its performing as designed. I also reviewed the code and it is looking good. Thanks for the excellent inline comments.

I added a checklist to the top post on this PR.

greg-1-anderson:
I added some documentation to install.md.  Feel free to un-check this task in the summary if you want more changes here.

weitzman:
For Windows, how about we tell them that the finder is unsupported and that they should always use drush.bat. Will redispatch/backend always call back into drush.bat? Incidentally, drush.bat looks a lot like a Windows version of drush.wrapper. Not sure if thats a useful observation or not.

greg-1-anderson:
Enhanced the installation instructions to give a better explanation of why you want to use a site-local Drush.  Note also that most of the common Drupal composer.json templates already require Drush, so you get site-local Drush support "for free" if you use one of these.

It is reasonable to expect Windows users to continue using drush.bat, as long as they are not using Composer with their Drupal site (e.g. any version of Drupal 8).  Backend invoke always dispatches directly to drush.php when making a call to another process on the same machine.

This brings up a good point, though.  If Drush ever uses backend invoke to dispatch to a different Drupal site on the same machine, but using different code, then Drush should really dispatch to the site-local drush.php, or use 'drush' instead of 'php drush.php'.  Otherwise, you could run into some "dependency hell" type problems by not running the correct site-local Drush.

This is rare, though; in most cases, local redispatches are to the same site, and when it is to a different site, it is usually to a different copy of the same site, which will also be okay.  Might be good to enhance this PR with some backend invoke support, though.

greg-1-anderson:
Backend invoke now respects site-local Drush. I think our Windows strategy is adequate. I'll add some simple tests of the Drush finder and wrapper a little later.

greg-1-anderson:
Added a simple test for the Drush finder.  The Drush wrapper is not tested yet.

greg-1-anderson:
Added tests for the drush.wrapper script.

greg-1-anderson:
We should probably check for --local=0 here; drush_startup() currently ignores values. I guess that means that --local=0 is invisible to drush_startup(), which is correct behavior, but --local=1 is missed.  Maybe this is a feature, though, as the Drush finder currently skips the wrapper if --local is specified.  Maybe that feature should be removed; then drush_startup() would not care about --local at all.  The reason we check --local today is to help prevent infinite loops that would be cause by folks who were using an old drush wrapper script that called "drush" instead of "drush.launcher".  We can't solve all of the problems that a bad wrapper could cause, though, so maybe we shouldn't care about this one.

greg-1-anderson:
As I mentioned above, handling --local in the Drush finder does not successfully handle all of the use-cases it was put in for (infinite loops still possible under misconfigured situations), and it does the wrong thing if someone adds --local on the commandline.  In this case, you'd end up running the global Drush instead of the site-local Drush, which is exactly the opposite of what you were asking for.

Because of this, I took --local out of the Drush finder script in the commit above.  --local is still processed as usual in Drush itself.

weitzman:
I just ticked the last checkbox in the OP. If I'm not mistaken, we can simply not support the Finder for windows users. They will have to find the right Drush themselves.

greg-1-anderson:
I think that's a workable strategy for Windows.  I think this is ready to merge, once you are satisfied that it has been exercised enough.  Once it is merged into master, I'd like to consider backporting to Drush 7, after it has had some time in an 8.x release.  In theory, it should not change the behavior of Drush commands, so should be backwards compatible.

weitzman:
Yeah, I agree that this is ready. Lets merge it on Monday. That way we don't slow down Barcelona folks if something goes wrong.

I dont think many D7 sites will be built via Composer, so I'm not seeing much benefit for backport. We're gonna release Drupal 8 and Drush 8 soon so Drush7 will be used by relatively few people I think. Happy to discuss.

greg-1-anderson:
The main reason to put this in D7 is for folks who want to use Drush 7 as their global Drush, and still have the convenience of the site-local Drush for their Drupal 8 sites.  We can see how this feature does in Drush 8 for a while before making a decision, though.

greg-1-anderson:
Rebased, tested, and squash-merged into master.

weitzman:
I am starting to get complaints about a missing pcntl extension. Do we have to always call pcntl.exe. If the "finder drush" never has to hand off to another drush, could we avoid pcntl call? /cc @greg-1-anderson

greg-1-anderson:
The "drush finder" script must chain to some other drush command (the launcher or the wrapper), regardless of whether said script is in the same Drush installation, or is living with the target site.

Right now, we use pcntl_exec() in the Drush Finder for efficiency.  In the issue on Windows support, we discussed having a fallback using some other form of exec (probably passthru) for platforms where pcntl_exec() does not exist.  I'll make a PR for this.

PR #1600, opened 2015-09-07
Add diff feature to make-update

basvdheijden:
The `drush make-update` command turns out to be really useful. However, I miss being able to see a diff between the old makefile and the newly (updated) one. Especially if you have a big makefile, it's not apparent which modules got updated.

By specifying the `--diff` flag, `drush make-update` gives you a diff between the new and old makefile.

### DIFF
--- a/commands/make/make.drush.inc
+++ b/commands/make/make.drush.inc
@@ -154,6 +154,8 @@ function make_drush_command() {
         'description' => 'Save to a file. If not provided, the updated makefile will be dump to stdout.',
         'example-value' => 'updated.make',
       ),
+      'security-only'  => 'Only update modules that have security updates available.',
+      'diff' => 'Return the diff instead of the whole new makefile.',
     ),
     'engines' => array('release_info', 'update_status'),
   );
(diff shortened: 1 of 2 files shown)


jonhattan:
I'd prefer that --diff (or --show-diff) always show the diff on screen, and that --result-file always save the updated makefile.

BasVanDerHeijden:
Good point. To be sure I know what you mean:
specifying ``--show-diff`` together with ``--result-file=bla.make`` shows the diff to the screen _and_ saves the resulting makefile to bla.make?
And if I only do a ``--show-diff``, it shows to the screen but doesn't save the updated makefile?

jonhattan:
> specifying --show-diff together with --result-file=bla.make shows the diff to the screen and saves the resulting makefile to bla.make?

yes

> And if I only do a --show-diff, it shows to the screen but doesn't save the updated makefile?

Not sure. At present, if result-file is not given, the makefile is printed to the screen. We could print the diff and then the makefile, or the diff only.

BasVanDerHeijden:
I would say the diff only, so you could potentially do something with the output (i.e. pipe to patch command). I will go ahead and implement this

jhedstrom:
Any possibility of adding a test for this? Also needs merge conflicts resolved.

weitzman:
make is deprecated

Issue #1453, opened 2015-06-22
Automated Testing of Updatedb


generalconsensus:
In order to make our infrastructure more agile we are investing more in automated tests -- via Behat and the https://github.com/jhedstrom/drupalextension repo.

In order to keep my sanity I am preferring that drush export only JSON so that I can do systematic matching against the responses I get from Drush.

While it was very easy to modify an existing drush contrib command (features) to make a command only export in JSON -- the same can not be said of drush updb. 

```php

    'outputformat' => array(
      'default' => 'key-value',
      'pipe-format' => 'json',
      'simplify-single' => TRUE,
      'table-metadata' => array(
        'list-separator' => ' ',
      ),
      'output-data-type' => 'format-list',
    ),
```

I do realize that the current direction of the project does a lot to support this -- and of which I am very grateful. 

Can you suggest anyway to make this functionality happen -- I would be glad to do the work but I believe that attention is focused on making this happen [first](https://github.com/drush-ops/drush/pull/673) before [this](https://github.com/drush-ops/drush/issues/1126)

weitzman:
If you want a list of pending updates, use the update-status command. Otherwise, I'm not sure what data you are trying to get out.

weitzman:
Closing until the request is clarified.

generalconsensus:
@weitzman Sorry for the delay in answering your original request: 

The point of this is take the following default output and make it just JSON

![image of normal updb]
(https://www.evernote.com/l/AFMS8yQ9zM9KUoJaozI82CZyTj8xilx-zT0B/image.png)

weitzman:
`drush updatedb-status --format=json`

generalconsensus:
Yes, but actionable -- the updatedb-status command only lists any pending database updates, but does not trigger the update itself. Indulge me for a moment to suggest the following as a backend behat test:

Given I run drush "updb" "--format=json"
Then I should see that the update works correctly

If this makes sense, I would be glad to send up a PR

weitzman:
Oh. The usual way you test to see if the update completed successfully is to examine the exit code from updatedb call. If it is zero, it was successful. Thats the unix standard.

If you really want more info, call `updatedb --backend` and you will get a JSON packet in return thats wrapped in some standard delimiters. Once you get JSON out of it you get log files, errors, etc.

PR #1603, opened 2015-09-08
Support hook_post_update_X().

### LINKED ISSUE #1602: Support hook_post_update_X implementations
Drush will need to update the `updb` command to take https://www.drupal.org/node/2538108 into account.

jhedstrom:
- Resolves #1602
- See https://www.drupal.org/node/2538108

### DIFF
--- a/commands/core/drupal/update.inc
+++ b/commands/core/drupal/update.inc
@@ -109,7 +109,11 @@ function update_main() {
   drupal_load_updates();
   update_fix_compatibility();
 
+  // Pending hook_update_N() implementations.
   $pending = update_get_update_list();
+
+  // Pending hook_post_update_X() implementations.
+  $post_updates = \Drupal::service('update.post_update_registry')->getPendingUpdateInformation();
 
   $start = array();
 
@@ -119,7 +123,7 @@ function update_main() {
   }
 
   // Print a list of pending updates for this module and get confirmation.
-  if (count($pending) || $change_summary) {
+  if (count($pending) || $change_summary || count($post_updates)) {
     drush_print(dt('The following updates are pending:'));
     drush_print();
 
@@ -130,16 +134,21 @@ function update_main() {
       }
     }
 
-    foreach ($pending as $module => $updates) {
-      if (isset($updates['start']))  {
-        drush_print($module . ' module : ');
+    foreach (array('update', 'post_update') as $update_type) {
+      $updates = $update_type == 'update' ? $pending : $post_updates;
+      foreach ($updates as $module => $updates) {
         if (isset($updates['start'])) {
-          $start[$module] = $updates['start'];
-          foreach ($updates['pending'] as $update) {
-            drush_print(strip_tags($update), 2);
+          drush_print($module . ' module : ');
+          if (!empty($updates['pending'])) {
+            $start += [$module => array()];
+
+            $start[$module] = array_merge($start[$module], $updates['pending']);
+            foreach ($updates['pending'] as $update) {
+              drush_print(strip_tags($update), 2);
+            }
           }
+          drush_print();
         }
-        drush_print();
       }
     }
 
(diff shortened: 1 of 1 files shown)


commands/core/drupal/update.inc
```diff
@@ -196,6 +205,15 @@ function drush_update_batch() {
     }
   }
 
+  // Apply post update hooks.
+  $post_updates = \Drupal::service('update.post_update_registry')->getPendingUpdateFunctions();
+  if ($post_updates) {
+    $operations[] = ['drupal_flush_all_caches', []];
```
weitzman: Hmmm.

1. I'm not sure why this needs to be added. I fear that we are going to clear caches twice. Does Drupal UI do that?
1. We always clear caches in a separate process now which can be skipped via an option. You can replace drupal_flush_all_caches() with drush_drupal_cache_clear_all().

weitzman:
It would be unicorn awesome if we had tests for updb. Just saying.

greg-1-anderson:
I have a question of the opposite bent:  is it now necessary for use to invariably clear the cache prior to a post-update operation?

Referring to  https://www.drupal.org/node/2538108 again, imagine that there is some hook 8001, that makes a schema change, and a follow-on hook 8002 that needs to run as a post-update operation, with all hook options.  If hook 8001 makes some change that invalidates caches, then it would be necessary to clear the caches before running hook 8002, or the results of the post-update hook would be unpredictable.

Also, the reason we allowed cache-clearing to be skipped via an option is to allow users to do multiple cache-clearing operations back-to-back, followed by one manual cache-clear at the end. I'm not sure that it is always safe to do that in the case of updatedb, though.

The bugaboo about updatedb functional tests and cache invalidation is that it's hard to know that you've really covered all contingencies; however, I agree that it would be super awesome to have at least one test that adequately covered one use case that was sufficiently complex (and no more so).

Something like:


1. Provide a 'demo' module in three versions:
    a. A base 8.x-1.0 version with no update hooks that initializes some data.
    b. An 8.x-1.1 version with an 8001 update hook, and an 8002 post-update hook
    c. An 8.x-1.2 version with an 8003 update hook.
2. Run two tests, starting with a fresh D8 base install for each one:
    a. Stepwise update:
        - Install base version, test data
        - Update to 8.x-1.1, run updatedb, test data
        - Update to 8.x-1.2, run updatedb, test data
    b. Skipped update:
        - Install base version, test data
        - Update to 8.x-1.2, run updatedb, test data

I think it would be best to just have the three "versions" in three different folders and "update" them via a filesystem copy, not via the pm code.  The module should provide an API function that could be called from the test to verify the data; the verify function can be different in each "version".

Not sure what the best update_N function should be; ideally, something that actually used Drupal APIs to set up and alter its schema would be best.

commands/core/drupal/update.inc
jhedstrom: Yeah, the patch that just landed calls that function between hook_update_N implementations and the new hook:

```php
    if ($post_updates) {
      // Now we rebuild all caches and after that execute the hook_post_update()
      // functions.
      $operations[] = ['drupal_flush_all_caches', []];
      foreach ($post_updates as $function) {
        $operations[] = ['update_invoke_post_update', [$function]];
      }
    }
```
 I will re-work to use `drush_drupal_cache_clear_all()`.

jhedstrom:
> It would be unicorn awesome if we had tests for updb. Just saying.

With the db dumps in D8 for testing, we may be able to extend core's `UpdatePathTestBase` just enough to test our own updates. We can pick that up in #1453.

jhedstrom:
Oh, just realized #1453 is closed.

weitzman:
@jhedstrom - Feel free to merge this when satisfied

jhedstrom:
Sorry, was on vacation--doing final review now.

Issue #1607, opened 2015-09-10
Remove calls to deprecated conf_path()


iant-tui:
conf_path() is due to be removed before Drupal 8.0.0 (see https://www.drupal.org/node/2457469 ), but doing so breaks drush. We need to remove calls to this function.

cilefen:
It seems like if many usages of conf_path were changed to drush_conf_path, and drush_conf_path made aware of anything it needs to know for 8, that may work.

weitzman:
Yeah, I suggest that drush_conf_path() be added to the three commands/core/drupal/environment*.inc files. Then when you use drush_conf_path(), you should add this line right above it - `drush_include_engine('drupal', 'environment');`

iant-tui:
How are you getting on cliefen? Is this ready for review?

cilefen:
@iant-tui See #1611

iant-tui:
Can I help with this in any way? It is a blocker to conf_path() being removed, and if we don't do that before RC1 (announced as 7th Oct) we'll need to support it until Drupal 9.

cilefen:
@iant-tui, in #1611, @weitzman suggested re-implementing conf_path in the bootstrap classes. You could try that.

weitzman:
Sorry I got delayed. You can help by making a PR.

iant-tui:
I take it you haven't written any code for this yet then. Are we basically suggesting copying D8's current conf_path() function into the bootstrap class?

weitzman:
Correct I have not started. I have not investigated the best way to implement this. There may be a better way than just copying the code - I dont yet know.

cilefen:
@weitzman Can you give me a hint on a way to test the changes, like a command that will definitely error or fail if the changes don't work?

weitzman:
Full bootstrap would fail if this code broke so I think we have sufficient implicit testing already.

weitzman:
Coded up a fix at #1607. Will merge when tests are green.

cilefen:
The PR is #1631

paul-m:
Actually removing conf_path() from Drupal is blocked by this issue.

https://www.drupal.org/node/2573443

PR #1611, opened 2015-09-11
Issue #1607: Removing calls to the deprecated conf_path() in Drupal 8.

cilefen:
This is my first time even looking at drush code, so, well, we'll see if this works.


weitzman:
drush_conf_path() does something slightly different than what conf_path() does so I dont think this is right. I'm going to work on this now. I will add a new conf_path() method to the bootstrap classes.

PR #1615, opened 2015-09-15
Allow local domain name to register as local host

danielnitschepeak:
This small change allows `drush_is_local_host()` to be a little more flexible. For example, this will prevent a "The source and destination cannot both be remote." error when doing a `drush sql-sync @alias.prd @alias.dev`, using this config (assuming dev.example.com has a host entry of 127.0.0.1):

````
$aliases['dev'] = array(
  '#check-local' => TRUE,
  'uri' => 'http://dev.example.com',
  ''remote-host' => 'dev.example.com',
);

$aliases['prd'] = array(
  'uri' => 'http://prd.example.com',
  ''remote-host' => 'prd.example.com',
);
````

### DIFF
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -618,7 +618,7 @@ function drush_is_local_host($host) {
   // In order for this to work right, you must use 'localhost' or '127.0.0.1'
   // or the machine returned by 'uname -n' for your 'remote-host' entry in
   // your site alias.
-  if (($host == 'localhost') || ($host == '127.0.0.1')) {
+  if (($host == 'localhost') || ($host == '127.0.0.1') || (gethostbyname($host) == '127.0.0.1')) {
     return TRUE;
   }
 


greg-1-anderson:
Not sure about the performance impact of adding gethostbyname() here.  Also, would Drush grind to a halt if you tried to use it without an internet connection?

We have discussed removing '#check-local' and drush_is_local_host() completely, and requiring folks who want this behavior to "local-ize" their aliases in a policy file hook.  Now that we have an alias alter hook, we should probably go ahead and do that.

danielnitschepeak:
Thanks @greg-1-anderson I had thought about localising aliases, so having a `dev.local` alias which doesn't include the remote-host option -- that would work as long as you remembered to use the right alias.

`drush_sitealias_alter` looks good. Here is what I went with which works with my setup:
````
/**
 * Allow drush to recognise a domain as a local destination.
 *
 * This prevents sql-sync from thinking the destination is remote, as it doesn't allow syncing
 * to a remote destination. Note: this requires the remote-host to point to 127.0.0.1 on the machine
 * you're running drush on.
 *
 * @param array $alias_record
 *   The alias record to modify.
 */
function policy_drush_sitealias_alter(array &$alias_record) {
  if (isset($alias_record['remote-host']) && '127.0.0.1' == gethostbyname($alias_record['remote-host'])) {
    unset($alias_record['remote-host']);
  }
}
````

For the googlers unfamiliar with policy files: put this in a `policy.drush.inc` file in one of the directories drush looks for alias files in (eg. ~/.drush, or drupal/drush).

greg-1-anderson:
Yes, that is a good solution.

PR #1617, opened 2015-09-15
Fixes #1614: Make module default project type for projects not hosted…

### LINKED ISSUE #1614: Don't check drupal.org for projects that specify a 'download' parameter in makefiles
Specifically, when a module isn't hosted on drupal.org, we need to specify ```projects[...][type] = "module"```. Otherwise, we'll get an error about not being able to find the project on d.o, when Drush tries to look up the project type. Since we filter out default or null values, when generating makefiles, ```drush make --lock``` is broken by this behaviour.

ergonlogic:
Let's give the Travis a crack at this...


ergonlogic:
Went with the simpler commit cherry-picked from the `make-fixes` branch.

PR #1619, opened 2015-09-19
Cast TranslationWrapper to string

webflo:
It is not possible to install latest D8 with Drush because PHP can´t compare to TranslationWrapper objects. I always hit the internal recursion limit.

```
drush site-install --verbose --yes --db-url=sqlite://tmp/site.sqlite                                                                                                    (8.x✱)
Initialized Drupal 8.0.0-dev root directory at /private/tmp/drupal-project/web                                                                                                                                            [notice]
You are about to DROP all tables in your 'tmp/site.sqlite' database. Do you want to continue? (y/n): y
Sites directory /private/tmp/drupal-project/web/sites/default already exists - proceeding.                                                                                                                                [notice]
Initialized Drupal site default at sites/default                                                                                                                                                                          [notice]
Executing: sqlite3  tmp/site.sqlite   < /private/tmp/drush_4f4AlU
Executing: sqlite3  tmp/site.sqlite   < /private/tmp/drush_XZViVo
Starting Drupal installation. This takes a while. Consider using the --notify global option.                                                                                                                           [ok]

Fatal error: Nesting level too deep - recursive dependency? in /private/tmp/drupal-project/web/core/lib/Drupal/Core/Form/FormBuilder.php on line 1310

Call Stack:
    0.0029     248744   1. {main}() /private/tmp/drupal-project/vendor/drush/drush/drush.php:0
    0.0050     248936   2. drush_main() /private/tmp/drupal-project/vendor/drush/drush/drush.php:11
    0.3496    8252232   3. Drush\Boot\BaseBoot->bootstrap_and_dispatch() /private/tmp/drupal-project/vendor/drush/drush/drush.php:70
    0.3680    6340544   4. drush_dispatch() /private/tmp/drupal-project/vendor/drush/drush/lib/Drush/Boot/BaseBoot.php:62
    0.5021    8624000   5. call_user_func_array:{/private/tmp/drupal-project/vendor/drush/drush/includes/command.inc:183}() /private/tmp/drupal-project/vendor/drush/drush/includes/command.inc:183
    0.5021    8624320   6. drush_command() /private/tmp/drupal-project/vendor/drush/drush/includes/command.inc:183
    0.5023    8629064   7. _drush_invoke_hooks() /private/tmp/drupal-project/vendor/drush/drush/includes/command.inc:215
    0.7887    7593784   8. call_user_func_array:{/private/tmp/drupal-project/vendor/drush/drush/includes/command.inc:364}() /private/tmp/drupal-project/vendor/drush/drush/includes/command.inc:364
    0.7887    7594024   9. drush_core_site_install() /private/tmp/drupal-project/vendor/drush/drush/includes/command.inc:364
    0.7899    7612264  10. drush_core_site_install_version() /private/tmp/drupal-project/vendor/drush/drush/commands/core/site_install.drush.inc:289
    1.9409   23562752  11. drush_op() /private/tmp/drupal-project/vendor/drush/drush/commands/core/drupal/site_install.inc:78
    1.9409   23564920  12. drush_call_user_func_array() /private/tmp/drupal-project/vendor/drush/drush/includes/drush.inc:705
    1.9409   23565136  13. install_drupal() /private/tmp/drupal-project/vendor/drush/drush/includes/drush.inc:720
    2.4069   25687344  14. install_run_tasks() /private/tmp/drupal-project/web/core/includes/install.core.inc:116
   51.2437   81991768  15. install_run_task() /private/tmp/drupal-project/web/core/includes/install.core.inc:527
   51.2437   81991952  16. install_get_form() /private/tmp/drupal-project/web/core/includes/install.core.inc:571
   51.2489   82571680  17. Drupal\Core\Form\FormBuilder->submitForm() /private/tmp/drupal-project/web/core/includes/install.core.inc:896
   51.4120   85561296  18. Drupal\Core\Form\FormBuilder->processForm() /private/tmp/drupal-project/web/core/lib/Drupal/Core/Form/FormBuilder.php:484
   51.4121   85565008  19. Drupal\Core\Form\FormBuilder->doBuildForm() /private/tmp/drupal-project/web/core/lib/Drupal/Core/Form/FormBuilder.php:560
   51.4190   85753408  20. Drupal\Core\Form\FormBuilder->doBuildForm() /private/tmp/drupal-project/web/core/lib/Drupal/Core/Form/FormBuilder.php:1003
   51.4192   85760896  21. Drupal\Core\Form\FormBuilder->doBuildForm() /private/tmp/drupal-project/web/core/lib/Drupal/Core/Form/FormBuilder.php:1003
   51.4193   85762592  22. Drupal\Core\Form\FormBuilder->handleInputElement() /private/tmp/drupal-project/web/core/lib/Drupal/Core/Form/FormBuilder.php:933
   51.4193   85768880  23. Drupal\Core\Form\FormBuilder->buttonWasClicked() /private/tmp/drupal-project/web/core/lib/Drupal/Core/Form/FormBuilder.php:1249

Drush command terminated abnormally due to an unrecoverable error.                                                                                                                                                     [error]
Error: Nesting level too deep - recursive dependency? in /private/tmp/drupal-project/web/core/lib/Drupal/Core/Form/FormBuilder.php, line 1310
```

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -165,7 +165,7 @@ function dt($string, $args = array()) {
       $output = $string;
     }
   }
-  return $output;
+  return (string) $output;
 }
 
 /**


tstoeckler:
See http://php.net/manual/en/language.oop5.object-comparison.php#71623 on why the recursion happens.

eiriksm:
Fixed the issue for me! :+1:

webflo:
New patch in https://www.drupal.org/node/2571375

webflo:
ping @greg-1-anderson @weitzman Can you take a look at it? Thanks!

greg-1-anderson:
How is Drush dt() being called in the call chain above?  It seems Drupal should be using t() directly here, so I don't see how the fix solves the problem (although empirically, it does).

If Drush dt() is in fact being called from Drupal code, then it seems like it might be undesirable to typecast the value to a string, since Drupal 8's translate() method does not work this way.  As far as Drush's internal use of dt() is concerned, though, the typecast to a string should be fine.

greg-1-anderson:
Looks like this was fixed in core; seems better to leave the typecast off here.  We can re-open if problems remain.

tstoeckler:
Well having it be a TranslatableString object makes sense only for HTML output. I don't think we have to fear XSS injection in Drush ;-), so I think the explicit string makes things more consistent in Drush

greg-1-anderson:
Yes, perhaps -- any place where the string is used by Drush, I agree.  But I am worried about the call chain above; since I do not understand how the t() from Drupal passed through Drush, I don't know if there would be any bad effects (e.g. in tests) if we stripped away the object. Since the purpose of the object is to provide data for output, maybe it never matters if we lose information like this.

webflo:
The object is created by ``dt() -> \Drupal::translation()->translate()``. TranslationManager::translate returns the TranslatableString.

greg-1-anderson:
Yeah, I understand how dt() gets the TranslatableString from Drupal; I just don't understand what call path through Drupal calls dt() in such a way that the Form handler gets the result.

neclimdul:
https://www.drupal.org/node/2571375#comment-10354155 should have fixed this.

greg-1-anderson:
Yes, it's fixed and closed; I'm just wondering how the output from dt() made it to the form handler. I think it's from the parameters such as the site name, but I haven't actually traced through it.

While this question is closed for this issue, it is still relevant in #1637.

tstoeckler:
Exactly, it's from `drush_core_site_install_version()` which uses `dt()` to translate the "op" values to simluate a form submit. Using the UI these values only ever reach the server as strings, that's why this only surfaces here.

greg-1-anderson:
OK, thanks. In that case, I think it is more appropriate to typecast here, or stop translating, as suggested in #1637.  We can continue discussions there.

webflo:
@greg-1-anderson I rebased the change. Its ready to merge.

greg-1-anderson:
Reversing my position; I think we might want to merge this after all.  GitHub won't let me re-open after a force-push (which is odd, since it recovers from force-pushes when an issue is open), but we will still be able to merge this in via the command line.  Discussion in #1637.

PR #1623, opened 2015-09-23
config-edit doesn't honor --editor on remote site

weitzman:
I'm getting vi when I run config-edit on a remote site. I don't like that much, and passing the --editor option help as it does not propagate to remote site. This PR lets the option propogate but I am seeing a "Error opening terminal: unknown." when passing --editor=nano. So this PR needs work.

An alternative to this PR is to use the local editor and then push back the edited file to the server. Do folks think thats a preferable solution?


weitzman:
--editor is deprecated. Setup $EDITOR as needed.

PR #1631, opened 2015-09-25
Fix #1607. Add conf_path() to boot classes.

### LINKED ISSUE #1607: Remove calls to deprecated conf_path()

weitzman:


cilefen:
Would \Drupal::service('site.path') work as well as reimplementing conf_path()?

commands/core/drupal/environment.inc
```diff
@@ -346,7 +347,7 @@ function _drush_theme_admin() {
 }
 
 function _drush_file_public_path() {
-  return Settings::get('file_public_path', conf_path() . '/files');
+  return PublicStream::basePath();
```
cilefen: +1 for PublicStream

cilefen:
I just did some manual testing with ``si``, ``cc``, ``cr``, ``pm-enable``, ``dl``, ``sa`` and ``status`` and there are no problems that I could see.

cilefen:
And @weitzman, that was with conf_path removed in D8.

weitzman:
Drupal::service('site.path') did not work unfortunately. Perhaps it could with some more research but I'm stopping here. The error is below for future reference ... Merged. Thanks for the review. 

Fatal error: Uncaught exception 'Drupal\Core\DependencyInjection\ContainerNotInitializedException' with message '\Drupal::$container is not initialized yet. \Drupal::setContainer() must be called with a real container.' in /Users/moshe.weitzman/htd/d8/core/lib/Drupal.php on line 129

Issue #1625, opened 2015-09-23
Add option to config-import to force installation with mismatched site UUID.


markdorison:
### Summary
It is not possible to import a site configuration "over" a freshly installed site. This would be very useful during development and for use with automated testing when the database would not be necessary as no content is required (or desired). 

### Steps to Reproduce
1. ```drush site-install minimal```
2. Log in and configure some pieces of the site. Set the site title, enable some modules, etc.
3. ```drush site-export```
4. ```drush site-install minimal```
5. ```drush site-import``` 

### Actual Results
The ```site-import``` command in step 5 fails with the following output: 

```
Drupal\Core\Config\ConfigImporterException: There were errors validating the config synchronization. in Drupal\Core\Config\ConfigImporter->validate() (line 730 of                    [error]
/var/www/development/chromatic.is/docroot/core/lib/Drupal/Core/Config/ConfigImporter.php).
The import failed due for the following reasons:                                                                                                                                      [error]
Site UUID in source storage does not match the target storage.
```

### Proposed Resolution
After discussing with [alexpott](https://www.drupal.org/u/alexpott), I tested the following manual work-around and saw a successful import:

1. ```drush site-install minimal```
2. Log in and configure some pieces of the site. Set the site title, enable some modules, etc.
3. ```drush site-export```
4. **Look up exported value of ```system.site uuid```.**
5. ```drush site-install minimal```
6. **drush cset system.site uuid [UUID-FOUND-IN-STEP-4]**
7. ```drush config-import``` 

It would be fantastic if ```config-import``` would provide an option (--force?) that would include the functionality in steps 4 and 6 so that a configuration could be imported over a clean site. Is there a philosophical reason not to do this? If not, I would be interested in adding it. Is there an existing mechanism to retrieve a specific config value from the configuration exported to disk?

weitzman:
I definately see the usefulness. Perhaps alexpott could chime in here and say if this is a horrible idea or not. It would have caveats in the description and confirmation.

markdorison:
Paging @alexpott. Can you weigh in on whether this is a horrible idea?

greg-1-anderson:
A --force on config-import, when importing everything, seems safe enough to me.

You could also codify your workaround into a short script, and use cget to look up the site UUID in the source site.  I think we can role this into Drush proper, although I'd still be interested to hear what @alexpott thinks.

weitzman:
Lets consolidate discussion in the PR #1635

Issue #1372, opened 2015-05-09
config_installer profile fails to install via Drush


andrewholgate:
The [config_installer](https://www.drupal.org/project/config_installer) profile is a special (and important) profile used to install an project using its existing yml config files.

Installing this profile from the UI works correctly but when running it via drush the following error occurs:

```
You are about to DROP all tables in your 'drupal8ddd2015' database. Do you want to continue? (y/n): y
Starting Drupal installation. This takes a while. Consider using the [ok]
--notify global option.
exception 'Drupal\Core\Installer\Exception\InstallerException' with  [error]
message 'Username field is required.
Email address field is required.' in
/var/www/project/web/core/includes/install.core.inc:895
Stack trace:
#0 /var/www/project/web/core/includes/install.core.inc(566):
install_get_form('Drupal\config_i...', Array)
#1 /var/www/project/web/core/includes/install.core.inc(522):
install_run_task(Array, Array)
#2 /var/www/project/web/core/includes/install.core.inc(117):
install_run_tasks(Array)
#3 /home/ubuntu/.composer/vendor/drush/drush/includes/drush.inc(709):
install_drupal(Object(Composer\Autoload\ClassLoader), Array)
#4 /home/ubuntu/.composer/vendor/drush/drush/includes/drush.inc(694):
drush_call_user_func_array('install_drupal', Array)
#5
/home/ubuntu/.composer/vendor/drush/drush/commands/core/drupal/site_install.inc(78):
drush_op('install_drupal', Object(Composer\Autoload\ClassLoader),
Array)
#6
/home/ubuntu/.composer/vendor/drush/drush/commands/core/site_install.drush.inc(289):
drush_core_site_install_version('config_installe...', Array)
#7 [internal function]: drush_core_site_install('config_installe...')
#8
/home/ubuntu/.composer/vendor/drush/drush/includes/command.inc(368):
call_user_func_array('drush_core_site...', Array)
#9
/home/ubuntu/.composer/vendor/drush/drush/includes/command.inc(219):
_drush_invoke_hooks(Array, Array)
#10 [internal function]: drush_command('config_installe...')
#11
/home/ubuntu/.composer/vendor/drush/drush/includes/command.inc(187):
call_user_func_array('drush_command', Array)
#12
/home/ubuntu/.composer/vendor/drush/drush/lib/Drush/Boot/BaseBoot.php(62):
drush_dispatch(Array)
[...]

PR #1635, opened 2015-09-26
Munge the uuids in sync dir before config-import

markdorison:
It was not previously possible to import a site configuration "over" a
freshly installed site. This is useful during development and in
automated testing environments when the database is not necessary as no
content is required (or desired).

Closes drush-ops/drush#1625

### To Test
1. ```drush site-install minimal```
2. Log in and configure some pieces of the site. Set the site title, enable some modules, etc.
3. ```drush site-export```
4. ```drush site-install minimal```
5. ```drush config-import --force```

You should be prompted as to whether you want to override the destination UUID. After this is complete, a config import will proceed.

screen shot 2015-09-26 at 14 21 28

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -125,10 +125,12 @@ function config_drush_command() {
       'source' => 'An arbitrary directory that holds the configuration files. An alternative to label argument',
       'partial' => 'Allows for partial config imports from the source directory. Only updates and new configs will be processed with this flag (missing configs will not be deleted).',
       'skip-modules' => 'A list of modules to ignore during import (e.g. to avoid disabling dev-only modules that are not enabled in the imported configuration).',
+      'force' => 'Override the existing site UUID to allow import of a configuration into an existing site (eg. to allow configuration of a site without the import of a database such as in an automated testing environment).',
     ),
     'core' => array('8+'),
     'examples' => array(
       'drush config-import --skip-modules=devel' => 'Import configuration; do not enable or disable the devel module, regardless of whether or not it appears in the imported list of enabled modules.',
+      'drush config-import --force' => 'Import configuration; force update the destination site UUID to allow importing of a configuration over an existing site.',
     ),
     'aliases' => array('cim'),
   );
@@ -591,6 +593,10 @@ function drush_config_import($source = NULL) {
     drush_print(implode("\n", $output));
   }
 
+  if (drush_get_option('force', FALSE) && drush_confirm(dt('Override the destination site UUID?'))) {
+    drush_op('_drush_config_import_site_uuid', $config_comparer);
+  }
+
   if (drush_confirm(dt('Import the listed configuration changes?'))) {
     if (drush_get_option('partial')) {
       // Partial imports require different processing.
@@ -645,6 +651,17 @@ function _drush_config_import_partial(FileStorage $source) {
   $active_storage = Drupal::service('config.storage');
   foreach ($source->listAll() as $name) {
     $active_storage->write($name, $source->read($name));
+  }
+}
+
+/**
+ * Replaces a destination site's UUID with the value in the source config.
+ */
+function _drush_config_import_site_uuid(StorageComparer $storage_comparer) {
+  if (!$storage_comparer->validateSiteUuid()) {
+    // Set the destination site UUID to match the source UUID.
+    drush_op('drush_config_set', 'system.site', 'uuid', $storage_comparer->getSourceStorage()->read('system.site')['uuid']);
+    $storage_comparer->reset();
   }
 }
 


weitzman:
Whats the correct behavior for incoming config entities with this option? Those carry UUIds that wont match.

greg-1-anderson:
I almost brought up entity UUIDs previously. I was writing some functional tests, and wanted to test config import. My first idea was to store exported block configuration files with the tests, and copy them in to the staging directory and import them.  The problem with this idea is that the exported configuration contains UUIDs that won't match with the UUIDs created during site install.  The reason I moved away from this idea is that I was afraid that there would be cross-references between different entities using UUID.  Seems that a force could make a mess of these, if only some of the files with cross-references were updated from a foreign site.

Perhaps this is expected behavior with --force ("I know what I am doing") it might be be helpful to log a warning every time a UUID is reset with the force option.

markdorison:
@weitzman @greg-1-anderson My intention was that it would replace *all* of the configuration entities so they in fact shouldn't match. This is reflected in the existing output which you can see part of in my screenshot that shows all of the existing configuration entities being deleted and new ones being created.

Please set me straight if I am missing something.

weitzman:
D8 has configuration dependencies so if you replace all the config entities, you wipe out all content types and fields and thus wipe all the content in the DB. We don't have a big enough warning for that! This is equivalent to doing a full site-install which means that --force would be a duplicate of https://www.drupal.org/node/1613424

markdorison:
@weitzman For the record, that *is* my intention with this functionality. I want to be able to spin up a site with the configuration files committed to version control *without* a database.

greg-1-anderson:
I ran into this use case myself when setting up CI for Drupal 8, and I have to say it is a little frustrating to not be able to commit configuration files and use them without the database that generated them.  I got around the limitation for my tests by exporting the entire configuration set after site-install, and then modifying configuration via Behat + Drush, and then testing imports and exports from there.  In my case, though, my purpose was only to test code; there was no specific site-under-test for me to worry about.  To do functional tests on the configuration of a specific site, this workaround would be ineffective.

Perhaps this PR is the solution, although perhaps the name of the option, the help text for the option, and the warning message should be altered to indicate that the option means "import the configuration and erase the site content".

For use cases that are not CI-related, I think that features should be used instead.

weitzman:
Blowing away your site is not something I would ever expect config-import to do. This could be an option to site-install or better yet, use the Contrib project that built for exactly this use case. I linked to it in my last message. If that profile does not work with Drush (we got a report of this), then lets fix the problem or find a site-install based solution.

greg-1-anderson:
Handling this in site-install sounds like the best option, since that is really the only valid time that you'd want to force the config import to rewrite the UUIDs.  The link above points to a postponed issue in core; is there a solution that we can use today? I did not see a reference to the contrib project, but I did not read everything in the referenced issue.

weitzman:
Lets close in favor of #1372

alexpott:
Another option is to update the staging config uuids. I do this like this:
```
/** @var \Drupal\Core\Config\FileStorage $sync */
$sync = \Drupal::service('config.storage.sync');
foreach ($sync->listAll() as $name) {
  $existing_config = \Drupal::configFactory()->get($name);
  if (!$existing_config->isNew() && ($uuid = $existing_config->get('uuid'))) {
    $data = $sync->read($name);
    $data['uuid'] = $uuid;
    $sync->write($name, $data);
  }
}

$core = $sync->read('core.extension');
unset($core['module']['standard']);
$core['module']['minimal'] = 1000;
$sync->write('core.extension', $core);
```

weitzman:
Oh, thats clever! Edit the files in the sync dir with so their uuids match those on the running site.

cweagans:
We're doing something pretty similar on our project. Excerpt from our makefile:

```
install-site:
	./vendor/bin/drush --root=$(DIR)/docroot site-install minimal --yes

config-init:
	cat ./config/system.site.yml | grep uuid | tail -c +7 | head -c 36 | ./vendor/bin/drush --root=$(DIR)/docroot config-set -y system.site uuid -

config-import:
	./vendor/bin/drush --root=$(DIR)/docroot config-import sync --yes

config-export:
	./vendor/bin/drush --root=$(DIR)/docroot config-export sync --yes

rebuild: install-site config-init config-import
```

I think it makes a lot more sense to put this in `site-install` than in `config-import`, because we're only ever doing this on the initial installation. I'd also point out that doing this with pretty much anything other than the minimal profile is likely going to fail. One example of a problem we ran into was Shortcuts. When we were installing with Standard, we'd get this on config-import:

```
Entities exist of type <em class="placeholder">Shortcut link</em> and
<em class="placeholder"></em> <em class="placeholder">Default</em>.
These entities need to be deleted before importing.
```

Given that, I'd suggest just making a separate command or flag or something, so that you can do: `drush si --from-config=/path/to/config/files`, which would install minimal, set the site UUID properly, and then import the config. That's not much different from what was suggested here earlier - the only difference is that Drush would just assume that it should install from Minimal if config is going to be imported

greg-1-anderson:
At BADCamp, @alexpott stated pretty clearly that he felt that the config installer was the way and the future; however, this won't be in until 8.1.0.
I agree that site-install should handle config import without requiring any installer that is not packaged with core.

weitzman:
Retitle based code from @alexpott above. This is exotic enough that it should be its own config-uuid command, not a flag on config-import.

greg-1-anderson:
I don't think this is so exotic. Apparently, the uuids are only in place to determine whether or not configuration was deleted and then re-created with the same name; all cross-references are done using data other than the uuids.  Because of this, I think that `config import --force` makes a lot of sense: "I want these two sites to be the same site now."  It would be much less discoverable (and ergo confusing) if it was split into two separate commands.

We could always warn and confirm when --force is used.

weitzman:
OK, makes sense. Anyone is welcome to submit a PR using @alexpott code

alexpott:
The one problem which does make it exotic are install profiles. If the incoming configuration has a different profile in core.extension (not so easy to figure out) then this command is going to need to deal with this.

greg-1-anderson:
@alexpott: Is this situation repairable (i.e., is the fixup limited to adjusting the installation profile recorded in settings.php), or are there deeper reasons why switching profiles won't work?

In any event, I think it would be a reasonable limitation if --force still failed with an error if the profile did not match.

cweagans:
Is it unreasonable to say "If you're going to use Drush for importing config, you need to plan on using Minimal for the default installation everywhere"? The group of people that are keeping a big pile of config in their Git repo probably doesn't overlap very much (if at all) with the group of people that are building install profiles.

Alternatively, Drush can just complain loudly if the target site and the incoming config have profiles that don't match. I think that would also be a reasonable path forward here.

alexpott:
Yep there is a bigger reason - install profiles change where we look for modules - if a module only exists in the profile and we do this we've got problems

greg-1-anderson:
@alexpott Yes, installation profiles change where Drupal looks for modules. Is updating the profile entry in settings.php sufficient to overcome this sort of change?

alexpott:
Yep it is. There are certain complications like if the starting profile has
modules. I'd suggest that you only allow swapping profiles if the profile
of the target site is minimal or standard before importing. That means you
know the site you are installing to has no custom modules in the profile.

simesy:
There's a bit of talk about using the minimal profile, which I'm not. I merged this branch into my drush and did:
```
drush config-export sync
drush site-install PROFILE
drush config-import sync
```
This gives me the described errors, but I'm also getting the errors if I try to replicate using the minimal profile, so either this branch is stale (it's 9 weeks old?) or I've screwed up something.

simesy:
**Use case:** We have 30 sites. We're using install profiles to neatly delineate custom code from shared code. The team has no shared server resources (yet). The sites have no content (yet). Trying to do initial builds from code with no databases.

simesy:
Oh wait. `--force` might help. This works with my custom profile.

simesy:
Update. I've been working successfully with --force for a few days, with a custom install profile. Not just core config, display suite for example.

simesy:
For those looking for similar solutions in this area... https://www.drupal.org/project/config_installer

weitzman:
The approach in https://github.com/drush-ops/drush/pull/1635#issuecomment-150643051 looks way more safe than the one in the PR. Edit the yaml files to match the site, not vice versa. Anyone is welcome to submit a new PR that does that.

cweagans:
What difference does it make? You're still going to end up importing a bunch of foreign config over the top of what's already in the site.

cweagans:
(Not arguing, btw. Genuinely curious as to why you see it that way)

weitzman:
See https://github.com/drush-ops/drush/pull/1635#issuecomment-144406706

simesy:
In my workflow this will result in noisy commits with unnecessary UUID changes in my config files, but I'm sort of re-appropriating the command for a different use case anyway...

cweagans:
Yeah, I saw that. I'm just not sure why changing the config resolves that issue. We still end up blowing away all the config in the site in favor of what's on disk, but going this route (as @simesy said), we end up modifying the config for every instance of the site, so the site UUID will never be in sync. As an example, if you have 3 developers working locally + dev, stage, prod, the UUID in the config checked into git will change basically every time the config is imported, won't it?

Kind of thinking it would be better to not do this on config-import, and have it be a flag on site-install instead (since that's the only use case I can think of where one would want to do this). Something like `drush si --from-config-dir="sync"`.

weitzman:
The problem isn't that config is overwritten. Thats a given, as you say. The problem is that the site's *content* will all get deleted as content depends on config entities for stuff like content types, fields, etc. Referential integrity is a bitch.

cweagans:
@weitzman Right. So what's the difference between changing the UUID in the config to match the site, or changing the UUID in the site to match the config, specifically for the site content problem?

What do you think about doing this in site-install instead?

weitzman:
In one scenario all content gets deleted and in the other it doesn't (because the config entity uuids are not changing and this the incomping config is treated as an update, not a delete+insert)

I already suggested doing this as part of site-install. Any now myself and @simesy have linked to config_installer project. This issue is going in circles.

cweagans:
@weitzman Oh, I see. I didn't read the code in https://github.com/drush-ops/drush/pull/1635#issuecomment-150643051 closely enough. It's changing *all* of the UUIDs, not just the one in system.site.yml. I was misunderstanding the direction here. Sorry for the noise.

Also, the config_installer project is nice, but I don't think it should be required to install from config. @greg-1-anderson said the same in https://github.com/drush-ops/drush/pull/1635#issuecomment-150665861. Maybe we can do both things? i.e. I can open another PR to add a `--from-config` option to `site-install`?

greg-1-anderson:
I think it would be great is the site-install command could handle a forced config import without a custom installer.  I think importing configuration during site-install is the ideal workflow for creating dev sites, where config changes might be committed back to the repository. A --force option on config-import would be handy for unit tests, but perhaps not necessary if site-install did the right thing, as subsequent re-imports would then have matching UUIDs.

weitzman:
Also, the config_installer project is nice, but I don't think it should be required to install from config.

if config_installer is headed toward core for 8.1.0, why don't folks just work on that? We are only 5 months from 8.1.0. We may or may not have Drush 8.1.0 before then

cweagans:
config_installer, IMO, is not very useful until we have install profile inheritance. Even if config_installer is in core, I'd probably still continue to `drush si minimal` and import my config on top of it.

greg-1-anderson:
I would tend to think that it would not take very long to release a version of Drush with an updated site-install command, iff someone went to the effort to make one.

simesy:
OK I looked at the patch, I will work on a PR for updating the site-install tonight. I won't use a `--force` argument because it's there's no target to destroy. But since `drush si` doesn't currently take a source directory/label, I'd like to know...

    # Do you want:
    --use-config-source=/path/to/file
    # Or...
    --use-config-dir=sync
    # Or both.

weitzman:
I'm thinking that a simple --config option would do. Thanks.

simesy:
I've opened a separate pull request at drush-ops/drush#1875

weitzman:
OK, lets close this in favor of #1875

weitzman:
Woops, meant to say - lets close this as #1875 went in. Hurray!

mriffault:
I managed to use @cweagans method [provided above](#issuecomment-150661519) (a big THANKS to him!). For the entity issue (which I encountered on language entities), I have moved them in the config/install folder of my installation profile. And I use [Drush CMI tools](https://github.com/previousnext/drush_cmi_tools) (which is awesome too!) to exclude this config from export and manage partial imports.

Now my build works fine, I can rebuild the website from scratch and import the configuration from code without using config_installer :)

Here is an example of my Makefile:
```
# @source https://github.com/previousnext/drush_cmi_tools/issues/5#issuecomment-250008389
APP_ROOT=$(CURDIR)/docroot
CONFIG_DIR=$(CURDIR)/config/sync
CONFIG_IGNORE=$(CURDIR)/config/config-ignore.yml
CONFIG_INSTALL=$(CURDIR)/config/install
CONFIG_SKIP_MODULES=devel

# Set DELFIL to 1 on make call to use the del file (on import)
ifeq ($(DEL),1)
	CONFIG_DELETE=$(CURDIR)/config/config-delete.yml
	DELFLAG=--delete-list=$(CONFIG_DELETE)
else
	DELFLAG=
endif

install:
# The dash before drush will ignore errors and allow "rebuild" recipe to continue
# even if some non-critical errors occur during install (like sending the
# confirmation e-mail to the administrator)
	-drush si -y my_profile \
		-r $(APP_ROOT) \
		--db-url=mysql://db_user:db_pass@mysql:3306/db_name \
		--db-su=root \
		--db-su-pw=rootpasswd \
		--account-name=admin \
		--account-pass=admin \
		--account-mail=admin@test.com \
		--locale=fr \
		--site-name="My website" \
		--site-mail=admin@test.com

# We have to override site uuid to be able to import the config from a previous
# install.
# @source https://github.com/drush-ops/drush/pull/1635#issuecomment-150661519
init:
	cat $(CURDIR)/config/sync/system.site.yml | grep uuid | tail -c +7 | head -c 36 | drush -r $(APP_ROOT) config-set -y system.site uuid -

# You may encounter errors on importing "entities" config (like
# language.entity...), this kind of config information have to be excluded from
# the export using config-ignore.yml, and could only been provided during site
# installation in the config/install directory of my_profile.
import:
	drush -r $(APP_ROOT) cimy -y --skip-modules=$(CONFIG_SKIP_MODULES) --source=$(CONFIG_DIR) --install=$(CONFIG_INSTALL) $(DELFLAG)

export:
	drush -r $(APP_ROOT) cexy -y --skip-modules=$(CONFIG_SKIP_MODULES) --destination=$(CONFIG_DIR) --ignore-list=$(CONFIG_IGNORE)

rebuild: install init import
```

PR #1636, opened 2015-09-26
Adding a check if the drush_get_username() and the remote-user in $si…

dasrecht:
…te_alias_settings differ. If so we use a ssh connection instead of the local to circumvent problems on shared environments (where rsync will fail because of restricted file permissions).

This was already brought up #1529 and #1530  but the fixes weren't entirely approved.

I now implemented it with the feedback of Johnattan (https://github.com/drush-ops/drush/issues/1528#issuecomment-132916754).

Please let me know if this is sufficient. I'd be happy to move this issue forward.

/bastian


greg-1-anderson:
This should be re-evaluated once #1650 is merged.  Please provide feedback if you think the checks in Drush still need to be adjusted.

Issue #760, opened 2014-08-15
Drush make - destination directory is not writable


ivanjaros:
Hi, I'm on Win7x64 with Drush 6.4-dev and I'm trying to use drush make.

The command I'm using is
drush make --no-core --no-gitinfofile --contrib-destination=. ../../theproject.make --projects=somemodule

When I run it(even as admin) I'll get this error:
Destination directory c:/Users/Ivan/.drush/cache/download/C-\www is not writable.      [error]
copy(c:/Users/Ivan/.drush/cache/download/C-\www\theproject-sites-all-patche
s-somemodule-metadata.patch):  [warning]
failed to open stream: No such file or directory drush.inc:752
Unable to download                                                       [error]

Seems to me the issue is with backslash in directory name "c:/Users/Ivan/.drush/cache/download/C-\www".

ergonlogic:
Please post the makefile with which you see this issue.

weitzman:
FYI we have [_drush_convert_path](http://api.drush.ws/api/drush/includes%21environment.inc/function/_drush_convert_path/master) if needed.

greg-1-anderson:
Drush make is maintenance only; closing.

PR #1640, opened 2015-09-29
Cache filename issues on Windows

matthijs-va:
On Windows, local patches fail to apply due to a backslashes in their cache file name. This pull request resolves that by replacing backslashes as well. See issue #760.


weitzman:
Merged into 8.x and 7.x Thanks.

PR #1643, opened 2015-09-29
Throw warning on variable set variables if set/overrridden in $conf

vijaycs85:


weitzman:
Variables are often overridden via settings.php and this does not prevent setting a value for those. It seems to prevent variables that are invented in setttings.php. Thats of limited usefulness IMO. Maybe I am missing something.

greg-1-anderson:
I can see folks getting frustrated if they forgot there was an override, and found variable-set ineffective.  Maybe this should just be a warning, though.

vijaycs85:
Exactly what @greg-1-anderson said. I agree that it can be warning to a doc page to explain: 
a) what happens when vget 
b) what happens when vdel

When having variable value in both file and DB.

greg-1-anderson:
Maybe you could push an update, with some docs added?

vijaycs85:
Yep, I can do that.

docs/usage.md
```diff
@@ -48,3 +48,10 @@ $ drush sql-sync --structure-tables-key=custom @live @dev
 
 See [example.aliases.drushrc.php](https://raw.githubusercontent.com/drush-ops/drush/master/examples/example.aliases.drushrc.php) for more information.
 
+Variable overrides
+------------------
+
+Variables can be overridden by global variable $conf in settings.php. When a variable is overridden, drush vset would set the value for the variable in database.
+However, the drush vget and vdel would behave as below:
+drush vget returns the value in $conf variable.
+drush vdel deletes the value from database, but the overridden value in $conf displayed for drush vget.
```
vijaycs85: We can add a link to settings.php file as it has a clear documentation under 'variable overrides'. But I leave it with doc experts.

weitzman:
I'll restate ... This patch doesn't help in the majority of cases. If I have overridden a google analytics variable or a search server varoiable in settings.php then this patch happily allows to vset it, and that vset accomplishes nothing.

greg-1-anderson:
@weitzman Do you prefer drush_set_error to drush_log, then?

This patch is completely worthless when you know you don't need it, but it could be quite valuable when you don't know what's going on, e.g. if it is someone else's settings.php file, or if you forgot that it was overridden, or simply are not thinking of variable overrides at the moment.  Seems useful to me.

I'm not sure about the placement of the docs in 'usage', though.  Not sure where it does go either, but I guess that only matters if you want the patch.

greg-1-anderson:
Just a sec, just read your original comment. Going to test this -- was presuming that it worked, but did not try it.

vijaycs85:
@weitzman Yep, vset has no effect if the variable is overridden in settings.php(or until you remove it from settings.php). The initial approach is to not to to allow to vset and the improved one is to allow vset, but throw warning that "there will be no effect".

greg-1-anderson:
Tested, behaves as advertised.  Seems useful to me.

weitzman:
I tested `dr vset site_name bar` and I got a warning despite the fact that site_name is not overridden.

$conf contains variables with overrides applied. It is non-trivial to get only the overrides AFAIK.

vijaycs85:
good catch. it works for custom variables, but site_name is indeed acting weird.

greg-1-anderson:
This could probably be checked by making a direct query against the `variables` table.

greg-1-anderson:
Alternately, it might be better to just call variable_get() at the end of variable-set, and log a warning if the values do not match.

vijaycs85:
@greg-1-anderson variable_get on the same run gives the value whatever just set. This is because of the way variable_set works :( it just override the value in $conf for that run.

greg-1-anderson:
```
$values = drush_invoke_process("@self", "variable-get", array($variable_name));
var_export($values);
```

vijaycs85:
Thanks @greg-1-anderson, it works. I checked https://github.com/drush-ops/drush/pull/1643#issuecomment-144429426 as well. it works now.

weitzman:
OK, code looks good now. The docs text needs to be clearer and moved to a topic attached to variable commands. Its too specific to be in usage.md. It is possible that the text could be abbreviated and put right into the command description or hook_help.

jonhattan:
/me draw upon the momentum to give attention to [Issue #1811224: vget for cron_safe_threshold always returns zero](https://www.drupal.org/node/1811224).

We are enfoncing the value of `cron_safe_threshold` in https://github.com/drush-ops/drush/blob/c129ea6ce51f6387d244f57debd04335b9c07dc1/lib/Drush/Boot/DrupalBoot.php#L437 and this has the undesired side effect that vget output is not accurate. This can be fixed in another PR but note that the final shape of this PR shouldn't forbid to vset cron_safe_threshold.

greg-1-anderson:
Speaking of [#1811224](https://www.drupal.org/node/1811224), this PR needs to be extended to the config-set command for Drupal 8.

vijaycs85:
This is ready for review.

sharique:
+1

greg-1-anderson:
Looks like it's my fault this never moved forward. At this point an update for Drush 9 would be needed. Closing.

PR #1645, opened 2015-09-29
Fix quoted variable

StryKaizer:


greg-1-anderson:
Thanks for this; fixed by #1648.

Issue #1637, opened 2015-09-28
After core's SafeMarkup changes, HTML tags are now output when using pm-enable


anavarre:
Per https://www.drupal.org/node/2571695#comment-10384893 we should evaluate if the below issue requires Drush to chase recent core changes with SafeMarkup.

```
$ drush en comment
      <em class="placeholder">comment</em> is already enabled.                                                                      [ok]
There were no extensions that could be enabled.  
```

Berdir says:

> That's because drush is still using !placeholders and because unknown placeholders fall through to [#2575703: Remove default fall-through from PlaceholderTrait::placeholderFormat()](https://www.drupal.org/node/2575703) would remove that and you'd get an error about an invalid placeholder instead, somehow.

jonhattan:
So we want to replace *all* usages of `!placeholder` with `@placeholder`.

neclimdul:
It isn't clear to me yet but yes that _seems_ to be the path suggested by the CR associated with that issue.

greg-1-anderson:
!placeholder used to mean no replacement.  @placeholder means escape, but do not add tags, whereas %placeholder means escape, and add em tags.

It seems that it would be a mistake to just change !var to @var everywhere in Drush. The big question I have is, what are the valid use cases for dt() calling through to t()?  Drush will use its own implementation of dt() when t() is not available.  Since dt() is really a completely different function than t() (the former being for cli output, the later for web browser output), shouldn't we just go ahead and make dt() use Drush's internal implementation all the time?  Are there any example code paths that would fail if we did this?

c.f. #1619. Seems that site-install passes some dt() output into the Drupal installer, the result of which is an object in the form field rather than a string. We talked about fixing that with a typecast; if the typecast is okay to put in, then perhaps avoiding Drupal's translation layer altogether would be preferable.

neclimdul:
Yeah, its a bit confused as well honestly. 

https://www.drupal.org/node/2506427 is the main issue for removing !placeholder.

greg-1-anderson:
Yes, so, if we keep calling Drupal's translation APIs from dt(), then in D8 we should not pass in any !var, because those are not allowed any longer.  However, there is no reason why Drush should not continue to use !var, as long as we do not call Drupal's translation API.  Really, all of Drush's placeholders should be !var, because there is never any need to cli terminal output to be html escaped.

[...]

PR #1647, opened 2015-09-29
#1637: Replace !var substitutions before calling D8 translate()

greg-1-anderson:
Drupal 8 may remove processing of !var substitutions. These are used all over Drush and also in Drush contrib modules; better to pre-process our strings before variable substitutions are removed for us by D8.

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -150,6 +150,10 @@ function dt($string, $args = array()) {
   // The language system requires a working container which has the string
   // translation service.
   else if (drush_drupal_major_version() >= 8 && \Drupal::hasService('string_translation')) {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Apparently, the Drupal-8.0.0-beta16 release due out tomorrow [will remove support for the ! placeholder](https://groups.drupal.org/node/483883).  We should therefore commit this, or some similar variant, ASAP so that Drush does not break when that change is committed.

weitzman:
I'm happy to delegate this one to @greg-1-anderson. Thanks!

greg-1-anderson:
Merging as-is, then.  We can always adjust later, if changes are desired, but I think this should do the trick, and keep our unit tests running tomorrow.  Thanks.

PR #1649, opened 2015-09-30
Remove `global` option in composer command for docs on `One Drush per Project`

rjsteinert:
The suggested composer command for `One Drush per Project` is the same as `One Drush for all Projects`. I think the intended command omits the `global` option. If that's correct, we could also use a note on how to path the `drush` command to your local instance.


greg-1-anderson:
Yes, that was a typo -- thanks for catching it.

PR #1650, opened 2015-09-30
Make drush_is_local_host() only match unambiguously local hostnames.

greg-1-anderson:
Also fixes a problem in drush_sitealias_is_remote_site(), which did not accept alias records, even though some callers used it as if it did.

With this change, users will have to implement an alias alter hook if they want "remote" aliases to act like local aliases when used on the system they point at.  Drush cannot really know the right way to test for local for every system, so it is best to not try, save for completely unambiguous options such as 'localhost' and '.local'.  This PR also eliminates the inconsistently-used "#check-local" item in site aliases -- an earlier, failed attempt to fix this problem.

### DIFF
--- a/drush.api.php
+++ b/drush.api.php
@@ -203,6 +203,22 @@ function hook_drush_command_alter(&$command) {
 }
 
 /**
+ * Adjust the contents of a site alias.
+ */
+function hook_drush_sitealias_alter(&$alias_record) {
+  // If the alias is "remote", but the remote site is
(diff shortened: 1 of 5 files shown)


greg-1-anderson:
Tests are green; recommend merging this now.

Schnitzel:
btw, this fixes https://github.com/drush-ops/drush/pull/1636 and https://github.com/drush-ops/drush/pull/1529 nice work!

PR #1652, opened 2015-10-01
Require ext-pcntl in composer.json

jonhattan:
Since #1595 we rely on pcntl_exec(). Some php binaries are not compiled with it (ex https://github.com/docker-library/php). Adding an explicit dependency in composer.json helps identifying the problem.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,7 @@
     "symfony/process": "2.4.5"
   },
   "suggest": {
+      "ext-pcntl": "*",
       "drush/config-extra": "Provides configuration workflow commands, such as config-merge."
   },
   "autoload": {


weitzman:
Well, we only require it if you actually use the finder. You are welcome to use the wrapper or launcher directly AFAIK. So maybe this should be 'suggest'? I'm concerned that fixing this will be a hassle for people and they really dont have to.

jonhattan:
Note that since finder is the default script, the first experience after updating to latest is a "function not found" error.

In this scenario, 'suggest' is helpful the same degree. I'll change the commit and merge.

skyred:
Just run into this problem as all my infrastructure is glued with Docker libraries. It looks like I should wait for the fall back solution mentioned in https://github.com/drush-ops/drush/pull/1595#issuecomment-144810398 ?

Currently, Docker official library doesn't support Composer https://github.com/docker-library/drupal/pull/7

Issue #833, opened 2014-09-23
vget for cron_safe_threshold always returns zero


hargobind:
Issue taken from https://www.drupal.org/node/1811224

It isn't possible to get the `cron_safe_threshold` variable via the command: `drush vget cron_safe_threshold` as it always returns 0.

Looking further into the code (7.0-dev - 2014-09-23), there's some code that overrides it:

```php
$override = array(
  'dev_query' => FALSE, // Force Drupal6 not to store queries since we are not outputting them.
  'cron_safe_threshold' => 0, // Don't run poormanscron during Drush request (D7+).
);
```

Setting the `cron_safe_threshold` variable still works fine. Please fix so that `vget` doesn't display internal drush overrides, but what is actually stored in the database.

vijaycs85:
https://github.com/drush-ops/drush/pull/1653

weitzman:
Lets continue in #1662

PR #1653, opened 2015-10-01
#833 - vget for cron_safe_threshold always returns zero

vijaycs85:


vijaycs85:
Lets follow at https://github.com/drush-ops/drush/pull/1654

PR #1654, opened 2015-10-01
fixes #833 vget for cron_safe_threshold always returns zero

### LINKED ISSUE #833: vget for cron_safe_threshold always returns zero

vijaycs85:


commands/core/variable.drush.inc
```diff
@@ -130,6 +130,11 @@ function drush_variable_get() {
 
   $keys = array_keys($conf);
   if ($args = func_get_args()) {
+    // Handle special variables case.
+    if (in_array($args[0], array('dev_query', 'cron_safe_threshold'))) {
+      return drush_set_error('DRUSH_VARIABLE_ERROR', dt('This variable is altered internally for drush. Use drush sql-query "select value from variable where name = \'cron_safe_threshold\'"'));
```
weitzman: Message is too specific as code above checks for dev_query as well.

vijaycs85:
got some changes not intended for this PR. Opened a new PR for this at https://github.com/drush-ops/drush/pull/1662

PR #1655, opened 2015-10-02
Added default option for php executable in drush runserver.

MartijnBraam:
The option 'php' in the cli context is not set when running drush.php from PhpStorm to debug. I think it is a good thing to at least try the php executable in PATH if no php executable is defined.


weitzman:
Seems reasonable to me.

weitzman:
Are you using the built in drush integration in phpstorm?

MartijnBraam:
I'm not using the drush integration but i'm running drush as composer project to debug commands in drush.

PR #1659, opened 2015-10-02
Fixes #1657 - Fatal error: Call to undefined function pcntl_exec()

### LINKED ISSUE #1657: Fatal error:  Call to undefined function pcntl_exec() in ncludes/startup.inc on line 256

vijaycs85:


greg-1-anderson:
Thanks for this; equivalent fix already committed.

PR #1660, opened 2015-10-03
#1657: Fallback for systems that have no pcntl_exec.

greg-1-anderson:
This PR uses `passthru` instread of pcntl_exec on systems where the later function is not available.

Note that this fallback **still** does not support Windows; Windows support for the drush "finder" could be provided if desired; at the moment, though, it is presumed that Windows users will be using drush.bat.

For the initial commit, I have completely disabled the pcntl_exec code path, so that the fallback can run through all of the tests.

### DIFF
--- a/includes/exec.inc
+++ b/includes/exec.inc
@@ -356,38 +356,6 @@ function _drush_escapeshellarg_windows($arg, $raw = FALSE) {
 }
 
 /**
- * Linux version of escapeshellarg().
- *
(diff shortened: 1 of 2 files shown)


weitzman:
Does drush output still show up on right stream (stderr, stdout, etc)?

greg-1-anderson:
Yep.

greg-1-anderson:
Merged this in to master; should make things better for folks. We can continue in another PR if anything additional is needed.

PR #1662, opened 2015-10-04
fixes #833 vget for cron_safe_threshold always returns zero

### LINKED ISSUE #833: vget for cron_safe_threshold always returns zero

vijaycs85:


greg-1-anderson:
1. If we are going to special-case Drush's vget override values, should we just call sql-query for the user, and return the appropriate value?

2. Are there error cases where that would not work (e.g. `drush vget cron_safe_`)?

3. It would be good if we got the list of "special" variable names from the same source that the override code gets them.  This would require some refactoring of the override code, probably.

vijaycs85:
Hi @greg-1-anderson, thanks for those very valuable questions/thoughts

1. Yeah, we can do that. We may not able to cover the $conf override cases though. 
2. Sure, I didn't think that one. May be we can do a filter on vget result as well?
3. Yeah, I will check that. 

Overall, it would be great, if we can do these "special" variables override only when it's necessary.

vijaycs85:
Just another approach to keep the original data before override and use it just before outputting the value to user.

jonhattan:
That looks more appropriate. In the end the usage of vget/vset/vdel should be transparent to the end user.

greg-1-anderson:
I did not test this, but (a) great idea on the approach here, and (b) looks like `$conf_original = $conf;` might be more appropriate here. Implementation of vget looks like it fails for non-overridden values in the current implementation. Maybe I'm missing something, since the tests are passing. Please add tests to vget overridden and non-overridden values, and correct code if failing. Thanks for your work here!

vijaycs85:
Thanks for your feedback @jonhattan and @greg-1-anderson.

> looks like $conf_original = $conf; might be more appropriate here.

thought, it would be nice to keep only the variables that we override internally to make drush work + explicitly added by passing as `--variables`. Though I don't see any problem using `$conf_original`  in `drush_variable_get` without touching `$conf`. Is that your suggestion?

I tested variables overridden by `$conf` in `settings.php` and in `DrupalBoot::bootstrap_drupal_configuration()` both looks working fine. I will add some test.

Finally, just noticed `DrupalBoot8::bootstrap_drupal_configuration()` calls `parent::bootstrap_drupal_configuration`, but D8 doesn't use `$conf`. Do we have separate issue for it?

vijaycs85:
After a closer look at code and @greg-1-anderson comment, realised his questions. Adding test case for variable in DB. As it needs `variable_init()`, test is failing (seems not bootstrapped to that level).

jonhattan:
@vijaycs85 in a fresh drupal installation, cron_safe_threshold is not set, so the correct output is:

```
drush @d7 vget cron_safe_threshold --pipe
No matching variable found.     [error]
```

I wonder if this is the failure you was seeing. To make a test accept [error] as a legitimate result, you want to pass `self::EXIT_ERROR` as the last parameter. The're several examples of this in drush tests.

OTOH, I think we don't need to call variable_init(), or not in the place you're calling it. Also, note that it is a Drupal 6 function. In drupal 7 it is variable_initialize().

vijaycs85:
@jonhattan, the error you mentioned is what is the correct result. but we are always getting '0' as it's overridden in drush. This is another action item of this issue.

> OTOH, I think we don't need to call variable_init(), or not in the place you're calling it. Also, note that it is a Drupal 6 function. In drupal 7 it is variable_initialize().

ah, I guess that makes lot of sense why the fail isn't all runs. Thanks. I will fix it.

vijaycs85:
This is ready for review.

PR #1666, opened 2015-10-05
Clarify that drush 8.x.x is needed for project level drush resolution

rjsteinert:
The docs currently states...

> When you run Drush, it will notice that you have a site-local Drush with the site you have selected, and will use that one instead.

That tripped a couple of us up over at Isovera as we did not realize we needed our global version of drush to be drush 8 in order for this to work. This pull request changes that sentence to...

>  When you run globally installed Drush 8.x.x, it will search for a project level drush install from the context of your current directory, and if found it will use that one instead.

I hope this rewording clarifies this for folks coming across this in the future.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -79,7 +79,7 @@ Composer - One Drush per Project
 -----------------
 Drush requires some of the same dependencies as Drupal 8.  If Drush and Drupal are installed separately, it is possible that different versions of these libraries will be loaded; this can lead to unpredictable results, usually resulting in difficult-to-diagnose crahses. It is therefore recommended that you use a separate copy of Drush for every Composer-managed Drupal site on your system.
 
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -79,7 +79,7 @@ Composer - One Drush per Project
...
 
-You should first follow the instructions "Composer - One Drush for all Projects", above, or "Git Clone (i.e. manual install)", so that you have a copy of Drush on your PATH.  When you run Drush, it will notice that you have a site-local Drush with the site you have selected, and will use that one instead.  This gives you the convenience of running Drush as "drush", without specifying the full path to the executable you want to use, without sacrificing the safety provided by a site-local Drush.
+You should first follow the instructions "Composer - One Drush for all Projects", above, or "Git Clone (i.e. manual install)", so that you have a copy of Drush on your PATH.  When you run globally installed Drush 8.x.x, it will search for a project level drush install from the context of your current directory, and if found it will use that one instead.  This gives you the convenience of running Drush as "drush", without specifying the full path to the executable you want to use, without sacrificing the safety provided by a site-local Drush.
```
weitzman: Thanks. 

* Missing an 'a': When you run *a* globally installed Drush ...
* "project level drush install from the context of your current directory". this seems awkward to me. it also misses the fact that a --root makes your current directory meaningless. perhaps revert to prior wording here.

rjsteinert:
@weitzman Agreed, seems awkward. I simplified this pull request to just add the Drush version required when mentioning Drush in the first sentence on the line.

PR #1673, opened 2015-10-07
Prevent endless loop on Windows.

mikeker:
This will handle the various flavors of root that `dirname()` on Windows gives us (eg: c:\, d:\, / or \)

### DIFF
--- a/includes/startup.inc
+++ b/includes/startup.inc
@@ -238,12 +238,17 @@ function drush_startup($argv) {
   //
   if (empty($found_script)) {
     $c = getcwd();
-    while (!empty($c) && ($c != "/")) {
+    // Windows can give us lots of different strings to represent the root
+    // directory as it often includes the drive letter. If we get the same
+    // result from dirname() twice in a row, then we know we're at the root.
+    $last = '';
+    while (!empty($c) && ($c != $last)) {
       $found_script = find_wrapper_or_launcher($c);
       if ($found_script) {
         chdir($c);
         break;
       }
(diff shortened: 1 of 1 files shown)


weitzman:
I didn't investigate, but we have [_drush_convert_path()](http://api.drush.ws/api/drush/includes%21environment.inc/function/_drush_convert_path/master) in case that would be useful here.

weitzman:
If not useful, lets just commit this as is.

mikeker:
I don't think we have access to _drush_convert_path at this point in the bootstrap...

weitzman:
That function is in includes/environment.inc which is included very early.
There is another environment.inc in the engine system which is usually
included later.

mikeker:
If I understand the new launcher code, it runs before `drush_main()` and before most stuff is included in `drush_preflight_prepare()`.

Regardless, `_drush_convert_path()` only normalizes the forward/backward slashes -- the issue here is that Windows includes the drive letter in `getcwd()` and that could vary from machine to machine. Also, `dirname()` considers `c:\` (or `c:/`) to be the filesystem root on Windows instead of `/`.

greg-1-anderson:
The Drush finder *only* includes startup.inc.  Long-term, I'd like to have the finder include Drush's autoloader, and use code in classes that it can autoload. I'd rather commit this PR as-is than move _drush_convert_path to startup.inc, or go down the dangerous path (or at least slippery slope) of including more and more files in the finder.

Issue #1612, opened 2015-09-11
Should drush continue to support the Windows platform?


greg-1-anderson:
Windows is not tested under CI, and it seems that Windows-related Drush questions do not get good support, either in this queue or on Stack Exchange.  This situation has gone on for years.

Roman Paska claims that [Drush on Windows is mostly functional](https://mobile.twitter.com/T2LPR/status/642041666208178177), although we still need a little more information about which shells work, what versions of Drush folks are using, etc.

I think that if we could set up CI for Windows Drush, and had a volunteer to do that, and if we had a number of people who could post here and say, "My name is ____, I use Drush on Windows regularly, and I have subscribed to the #drush tag on Drupal Stack Exchange, and will answer Drush on Windows questions there", well, then we'd have a supported platform, and there would be no question about that remaining.

I am somewhat dissatisfied with the current "unsupported but happens to work" state of affairs, though, and would prefer to firmly be in one camp (supported) or the other (support removed).

jibran:
How about this:

1. Find a good maintainer for windows.
2. Move the windows hacks out of drush.
3. Create a new project drush-win for windows only.

jibran:
Found https://github.com/travis-ci/travis-ci/issues/216 & https://github.com/travis-ci/travis-ci/issues/2104 after quick search.

weitzman:
I do agree that setting up CI for Windows is the crucial missing piece. Travis is never going to support Windows AFAICT. I've been subscribed to those issues for a long time.

Looks like http://www.appveyor.com/docs/build-configuration is free for open source projects and supports Windows build machines. We could run it in parallel with Travis.

Would be great if someone forked Drush and showed it running on appveyor. All you have to do is sign up on appveyor and add an appveyor.yml to your fork.

weitzman:
I got pretty far along with automated testing on Appveyor. Drush is now installed via Composer. Now we need to get Unish (our test suite) to complete a test run. See https://ci.appveyor.com/project/weitzman/drush/build/1.0.53/job/s70c2tla8fmoppu8 for the current trouble clearing the sandbox and finding drush.bat.

Help appreciated.

greg-1-anderson:
Great progress!  From the output, it looks as if the permission denied is caused by Unish trying to remove the sandbox on the first run, before it exists.  The code is wrapped in 'file_exists', though, so I don't know what to make of that.

My guess about the drush.bat is it seems that there are too many \'s in the PATH -- but I'm not sure about that.

[...]

PR #1674, opened 2015-10-07
drush.launcher fix for Windows

mikeker:
Followup to #1673: Quoting `$found_script` handles spaces in the path. Also does a poor-man's `_drush_convert_path()` so we don't have to include any other files in the finder.

I ran into an odd problem when using drush.launcher -- without specifying that this is to run in the shell, Windows would try to have me associate the .launcher extension with a program. (And would helpfully pull up the "Search the Windows Store" dialog...  Ha!) I don't know if the correct solution is to rename the drush.launcher to something else on Windows or to use the change in this PR. Feedback appreciated!


ivanjaros:
Yep, just updated drush so I can work with D8RC1 and am getting the same thing on Win10.

weitzman:
Hard coding sh like that is wrong for non Windows. It bypasses the shebang line at top of launcher.

mikeker:
I figured that was the case. I assume we don't want to drag
drush_is_windows into startup.inc.

I'll look into how best to deal with the association on the Windows side of
things.

mikeker:
I've tried a few ways of associating the .launcher file with `sh` on Windows 10 but all of them seem to lose the argument list in the process (so `drush status --verbose` ends up executing just `drush`). There doesn't seem to be a way to quote the entire command that doesn't confuse the system...

So, what about moving `_drush_get_os()`, `drush_is_windows()`, and maybe some similar commands into startup.inc? I know @greg-1-anderson voiced some concerns to that #1673 and this is starting to feel hacky.

weitzman:
how about we give up on this issue for the moment and ask folks to run drush.bat instead of the finder. i agree with keeping the finder simple.

greg-1-anderson:
Yes, we wouldn't want to see a lot of functions moving to startup.inc; we should definitely avoid testing for the platform here, and fall back on drush.bat if it is not possible to support Windows in a platform-independent way. 

Would it help Windows out if we renamed `drush.launcher` to `drush.sh`?  That name would be a little confusing for non-Windows users, as some folks might assume that `drush.sh` is what they should symlink to when setting up their $PATH, etc.; we could try to combat this with documentation if making that change helped Windows.

mikeker:
Changing `drush.launcher` to `drush.sh` works better on Windows (`.sh` is associated with the MinGW shell). But it opens Drush in a separate window and you lose the output. I'll poke around some more to see if there's a way around this.

mikeker:
> ask folks to run drush.bat instead of the finder

This does work, at least in my minimal testing of it. Adding something like:

```
alias drush=$HOME/Documents/dev/drush/drush.bat
```

to your `.bashrc` allows Drush to function on Windows, but I assume that prevents you from having separate one Drush per Drupal.

What about this change (971ef3a) instead?

kenorb:
Related: #1696

mikeker:
> So does one need Git bash in order to have sh?

I believe Cygwin has `sh` as well. I know it has `bash`. Cygwin users could use either shell option, though we probably need to update the docs around that.

ivanjaros:
Is this fixed in latest drush release(rc3)?

greg-1-anderson:
No.  Also, it seems that there hasn't been any testing of the drush.phar on Windows. It would be helpful if someone could investigate that, and see how it currently fares.

weitzman:
I tested as far as assuring help and core-status work. Obviously needs more
testing than that.

weitzman:
This seems OK to me. Still want it, @mikeker? Sorry for the delay.

When merging this, we need to squash to hide all the various unneeded commits along the way.

tstoeckler:
The PR doesn't actually fix things properly for me. Looking around the related code we already have a check for whether drush.launcher is executable. On my system drush.launcher is correctly reported as not-executable (as the Shebang is not respected), so that check would suffice. But we currently only perform that check in some cases, not in all. So for me this issue can be fixed, by just moving this check so that it is always performed. It _seems_ that is also generally sensible. I did not want to step on anyone's toes, so didn't create a PR, but you can look at it here: https://github.com/tstoeckler/drush/commit/a3682cb188e74052419c7dbc9b4d5a7d1ac0a646?diff=unified&w=1

e-r-n-i-e:
What about using `drupal-laucher` instead of `drupal.launcher` to avoid the problems on Windows?

greg-1-anderson:
Sounds reasonable to me. Has anyone tried it and confirmed that it works?

weitzman:
Makes sense.

However, Real Bash is now available on Windows. See #1612. Help there is much appreciated.

weitzman:
I debugged this a bit for a co-worker. We had success when we added an association for .launcher extension with `bash.exe` in Control Panel => Default Programs => Associate a file type or protocol with a program. `bash.exe` was found in our case inside of the [Cmder](http://cmder.net/) application. Alternatively it is installed by Acquia Dev Desktop and other programs. Install one of those if you don't have any bash.exe. One command you could use to find out if you have bash.exe is `where bash.exe.

Please report back if this helps.

weitzman:
In related news, Composer suggests that we should not have a drush.bat in our repo and instead [let it create .bat and proxy files](https://getcomposer.org/doc/articles/vendor-binaries.md#what-about-windows-and-bat-files-). I'm not sure Composer really passes along arguments and options properly but we could explore this.

weitzman:
Actually some code changes were needed. Those have now landed in master and 8.xz. See #2312

v-tsymbalistyi:
So is there some kind of solution for Windows prompting for a program to open drush.launcher?

PR #1682, opened 2015-10-12
Keep up with changes in Drupal 8's index.php.

greg-1-anderson:
Drupal 8's index.php is slightly different than it used to be.  Drush is still following the old pattern; while this works today, it would probably be prudent to keep up with changes in Drupal, just in case the old method is changed or removed someday.


greg-1-anderson:
This does not work; I guess the difference between Drush's bootstrap and Drupal's is deliberate.

PR #1684, opened 2015-10-13
Fixed #1627: Project type not set correctly.

### LINKED ISSUE #1627: Themes get the wrong project type now
Due to the changes in commits fcaf417c7a78c9619f949178fb93304b4fa0fe7c and e8125d083afe9087ced958a1356bb361116793e4, e.g. when the project type for themes is not explicitly defined, it is not autodetected anymore and the theme is placed in the modules folder. This seems like a breaking bug to me :)

danepowell:
Fixes #1627 

tl;dr: as of fcaf417c7a78c9619f949178fb93304b4fa0fe7c and e8125d083afe9087ced958a1356bb361116793e4, if you don't specify a project type in your makefile, it will be automatically set to `module`, even if it's actually theme or profile.

### DIFF
--- a/commands/make/make.drush.inc
+++ b/commands/make/make.drush.inc
@@ -387,7 +387,6 @@ function make_prepare_projects($recursion, $info, $contrib_destination = '', $bu
     // Merge the known data onto the project info.
     $project += array(
       'name'                => $key,
-      'type'                => 'module',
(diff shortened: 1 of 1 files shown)


jonhattan:
#1627 was already fixed.

Issue #1676, opened 2015-10-09
Update from drupal 8 beta 16 to Rc1 using drush gives error


jhmnieuwenhuis:
I use drush 8.0.0 rc1 to upgrade drupal ( drush up drupal)
The upgrade is performed, but at the end I get an error.
 Drush Version   :  8.0.0-rc1

Post updating views                                                  [ok]
Cache rebuild complete.                                              [ok]
Finished performing updates.                                         [ok]
PHP Fatal error:  Class 'Symfony\Component\HttpFoundation\Response' not found in /home/hn/.composer/vendor/drush/drush/lib/Drush/Boot/DrupalBoot8.php on line 133
PHP Stack trace:
PHP   1. {main}() /home/hn/.composer/vendor/drush/drush/drush.php:0
PHP   2. drush_main() /home/hn/.composer/vendor/drush/drush/drush.php:11
PHP   3. Drush\Boot\DrupalBoot8->terminate() /home/hn/.composer/vendor/drush/drush/drush.php:79
Drush command terminated abnormally due to an unrecoverable error.   [error]
Error: Class 'Symfony\Component\HttpFoundation\Response' not found in
/home/hn/.composer/vendor/drush/drush/lib/Drush/Boot/DrupalBoot8.php,
line 133
[hn@drupaldevel default]$ drush --version
 Drush Version   :  8.0.0-rc1

kordianbruck:
+1: Same thing happens to me. Any ideas?

greg-1-anderson:
Although I have not traced through this all, I think that what is happening is:

1. pm-updatecode bootstraps Drupal
2. The Composer autoloader caches all of the filesystem locations of all of the available classes
3. Drupal is updated, and the cached locations become unavailable
4. Drush tries to call the completion routines at the end of the command, and the failure occurs because the file containing Symfony\Component\HttpFoundation\Response no longer exists.

I think that the solution here is to make pm-updatecode be a DRUSH_BOOTSTRAP_NONE command, and have it use drush_backend_invoke() to fetch the information it needs about available updates via the pm-updatestatus command.  That way, the bootstrap + teardown will all happen before any Drupal code changes.

weitzman:
pm-updatestatus is already done within a drush_invoke_process(). I can see benefit by making updatecode into a NONE but eventually we do have to bootstrap Drupal and can run into errors. So I'm still a bit hazy on the real problem here and the proposed solution.

greg-1-anderson:
[...]

PR #1685, opened 2015-10-14
#1676: Run pm-updatecode without bootstrapping Drupal.

greg-1-anderson:
This works, but is unfinished in a couple of places. Here, we avoid bootstrapping Drupal during pm-updatecode, because you cannot bootstrap Symfony code, and then change the code that the autoloader was initialized in. That is a recipe for a crash.

I don't have time to write a test right now; I was testing like this:

```
drush dl drupal-8 --select --all --choice=3 --drupal-project-rename=drupal-8
cd drupal-8 
drush -y si --db-url=mysql://root@localhost/d8updatetestdb
drush status
drush pm-updatecode -y
drush status
```

That technique could be emulated to create a test case that upgrades from the previous version of Drupal to the current version.

### DIFF
--- a/commands/pm/pm.drush.inc
+++ b/commands/pm/pm.drush.inc
@@ -340,12 +340,14 @@ function pm_drush_command() {
       ),
       'update_status',
     ),
+    'bootstrap' => DRUSH_BOOTSTRAP_DRUPAL_ROOT,
   );
   // Merge all items from above.
   $items['pm-update'] = array(
     'description' => 'Update Drupal core and contrib projects and apply any pending database updates (Same as pm-updatecode + updatedb).',
     'aliases' => array('up'),
     'allow-additional-options' => array('pm-updatecode', 'updatedb'),
+    'bootstrap' => DRUSH_BOOTSTRAP_DRUPAL_ROOT,
   );
   $items['pm-updatecode-postupdate'] = array(
     'description' => 'Notify of pending db updates.',
--- a/commands/pm/updatecode.pm.inc
+++ b/commands/pm/updatecode.pm.inc
@@ -15,13 +15,10 @@
  * just to check on. In this case, say at the confirmation prompt.
  */
 function drush_pm_updatecode() {
-  // In --pipe mode, just run pm-updatestatus and exit.
-  if (drush_get_context('DRUSH_PIPE')) {
-    drush_set_option('strict', 0);
-    return drush_invoke('pm-updatestatus');
-  }
+  // We won't support --pipe mode with pm-updatecode any longer;
+  // folks should be using pm-updatestatus for this by now.
 
-  $update_status = drush_get_engine('update_status');
+  //$update_status = drush_get_engine('update_status');
 
   // Get specific requests.
   $requests = pm_parse_arguments(func_get_args(), FALSE);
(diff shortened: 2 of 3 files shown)


weitzman:
Well, folks probably should not use pm-updatecode in D8 to begin with. They should 

1. get recommendations with pm-updatestatus
2. edit the project's composer,json accordingly
3. composer update

Maybe we should write another command for this.

greg-1-anderson:
The steps above (update with `composer update`) is exactly what I recommended in my blog post a couple of days ago; unfortunately, it does not work for Drupal core.  This is because Drupal does not include drupal/core in its composer.json; instead, it uses wikimedia/composer-merge-plugin to merge in the composer.json for drupal/core, which is bundled as part of Drupal in the "core" directory.  As a result, `composer update` will never touch this part of the project.

In order to make `composer update` update your core files, you would have to modify your composer.json to remove the composer-merge-plugin, and add drupal/core as a dependency.  Rather than doing this by hand, though, it is probably a better idea to just use [drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project) to create a new project with a composer.json that is already set up to use the drupal/core split core.

It might be nifty, though, if pm-update could note whether dependent modules are stored in the composer.json, and, if so, use `composer update` to update contrib -- or at least be smart enough to skip this step.  At a minimum, though, it seems that pm-updatecode should continue to update the core directory.

greg-1-anderson:
As a general goal, though, I would like to get Drush out of the code-managing business.  I don't think this is feasible in the short-term, but maybe further down the road.

weitzman:
I recommend a new command that assumes you are using the best practice which is the drupal-project model, AFAIK.

greg-1-anderson:
As long as we are supporting drush dl, we should support updating code that was retrieved that way.  Drupal core is going out of their way to avoid requiring Composer. If we don't feel the same way, maybe we could convert composer.json automatically (at dl &/or pm-updatecode time) and then fall back on composer update.  We'd have to consider what to do about contrib modules too, though; if we convert, then we'd have to convert contrib modules to Composer-managed as well (a-la drush composer-generate).

I don't know if this will work well enough to just switch over, though.  It might.

mariagwyn:
Drush is an integral part of our update process. Removing this functionality would be a significant tooling loss.

greg-1-anderson:
I think it might be difficult to impose the Composer workflow on the user, even if we limited the action to Drupal 8.  Encouraging and facilitating the use of drupal-composer/drupal-project would be worthwhile, though, if there were some easy ways to do so.

mariagwyn:
I have no objections to the creation of a new and better workflow. However, I think that should likely be for a new version since it is removing functionality that I suspect many depend on. I use it personally on small, itty-bitty non-profit sites. And I use it as a tool on thousands of sites for a large company. Deprecating it without creating a separate project or warning would be ... very very frustrating.

weitzman:
OK, lets move Composer discussion to #1701 and focus this issue on a more robust pm-updatecode.

greg-1-anderson:
I cannot reproduce this (the problem in #1687) any longer. Maybe I mis-diagnosed the problem?

Anyone else seeing any problem with updating from Drupal 8.0.0-rc1 to rc2 with pm-updatecode?

PR #1688, opened 2015-10-15
Call drush_main() directly from the Drush finder if it is already loaded.

greg-1-anderson:
Just a quick prototype that I am putting here for now. Should be merged in with #675, and change the "main" from "drush.php" to "drush", and it would probably work.


weitzman:
Integrated into #1693

PR #1692, opened 2015-10-16
When calling user-password or user-login, blocked users should be prompted to be unblocked.

davereid:
I found it really odd that I can run `drush user-login` for a blocked account, but it does not warn me that the account is blocked. Hence the one-time login link will fail every time. I thought it would be nice if the user would be prompted to unblock the user when running these commands.

This also cleans up that the user-password command ran too many times (looped through each account, but then called each() on the user list each time).

### DIFF
--- a/commands/user/user.drush.inc
+++ b/commands/user/user.drush.inc
@@ -358,9 +358,8 @@ function drush_user_password($inputs) {
     if (empty($pass)) {
       $pass = drush_prompt(dt('Password'), NULL, TRUE, TRUE);
     }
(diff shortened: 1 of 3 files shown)


davereid:
I guess this probably needs some tests.

weitzman:
Code looks solid to me. I think it could go in after manually testing on D8-D8 if you dont have time to write automated tests.

greg-1-anderson:
Needs updating for Drush 9 now. Let's close for now, and re-open if there is renewed interest.

Issue #1641, opened 2015-09-29
Drush runserver HTTP server doesn't support callbacks with dot in the URL


kenorb:
I've the following testing URLs (which exists in Drupal 7):

    /taxonomy/autocomplete/foo_bar
    /taxonomy/autocomplete/foo.bar

These Drupal URLs normally works in Apache by giving the error:

> Taxonomy field foo.bar not found.

When testing using `drush rs`, the second URL doesn't work properly and it gives the following error:

> Not Found

> The requested resource /taxonomy/autocomplete/aa.00 was not found on this server.

Log entry:

> `[Tue Sep 29 15:56:11 2015] 127.0.0.1:54729 [404]: /pca/retrievebyid/25764183.00 - No such file or directory`

I've found this when implementing [PCA](https://github.com/mycognitive/postcodeanywhere) and [this callback](http://services.postcodeanywhere.co.uk/PostcodeAnywhere/Interactive/RetrieveById/1.3/json.ws?key=DJ95-ZP12-FW89-EN99&Id=23747212.00&PreferredLanguage=English&CallbackFunction=TestCallBack).

So my local menu callback (`pca/retrievebyid/1234.00`) doesn't accepts the Id argument, because it's never gets passed into Drupal, as it consist dot (`.`). So for some reason drush is giving 404 page.

The easiest way to reproduce the problem is by running: `drush qd`.

I'm using `drush rs` for testing purposes.

I believe the error is could be somewhere in [runserver_parse_uri()](https://github.com/drush-ops/drush/blob/b4c0683922ab08ef81c9f7553f58ab845ea6972e/commands/runserver/runserver.drush.inc#L186), but I'm not sure exactly how the URLs are handled.

MartijnBraam:
The issue that paths containing a dot aren't working is because the built-in php server assumes that all files containing a dot are an file in the filesystem. If a router script is passed to `php -S` then this behavior is bypassed and the all the urls start working.

I think this can be fixed by using the `d8-rs-router.php` for drupal 7

kenorb:
It also fails for Advanced Help module which scans for the help files and generates the appropriate callbacks.

So going into: `/help/commerce_paypoint_advanced/README.md`
generates `Not Found` page as well.

Related: #752

kenorb:
The solution is to start the server in docroot as:

    php -S localhost:8888 .ht.router.php

[...]

Issue #752, opened 2014-08-11
runserver: 404 when dot character in clean URL path


GarrettAlbright:
When using runserver, I noticed I was getting a PHP error from the server (not from Drupal) when attempting to go to a Memcache server report page with a path that looks like `http://127.0.0.1:8888/admin/reports/memcache/default/127.0.0.1%3A11211` (`%3A` is an encoded colon character).

However, if I manually adjust the path to make it "dirty," like `http://127.0.0.1:8888/index.php?q=admin/reports/memcache/default/127.0.0.1%3A11211`, it works just fine.

I thought this might be an issue in runserver-prepend.php, but after some consternation, as far as I can tell, that file doesn't even execute whenever there's a period in the clean path. I can only assume that PHP is assuming that there's a file involved whenever it sees a dot in the path, looking for that file, then failing when it can't find it without giving the prepend file a chance to have a say in the matter.

I notice that we don't seem to be using a router file. Would it be possible to avoid this problem if we had one? I'm not sure.

mdeltito:
I'm seeing the same, and it looks like this is the expected default behavior with PHP's built-in server:
https://bugs.php.net/bug.php?id=61286

Using a router file *does* mitigate the issue however, and the included D8 one seems to work just fine. I don't know enough about the history of the runserver-prepend file to determine if it's still necessary for D7, but the router file seems like the right approach moving forward.

kenorb:
It also fails for Advanced Help module which scans for the help files and generates the appropriate callbacks.

So going into: `/help/commerce_paypoint_advanced/README.md`
generates `Not Found` page as well.

weitzman:
Got fixed by #2033 AFAIK

weitzman:
For D7, see #1658

PR #1658, opened 2015-10-02
Use the d8-rs-router.php for drupal 7

MartijnBraam:
The d8-rs-router.php from the runserver module seems to work fine for drupal-7 and fixes the issue of paths containing a dot resolving to a 404 page. See issue #1641

PR #1699, opened 2015-10-22
Runserver prepend configure

typhonius:
This should help with #1641 & #752 when using Drush runserver on a D7 site.

### DIFF
--- a/commands/runserver/runserver.drush.inc
+++ b/commands/runserver/runserver.drush.inc
@@ -40,6 +40,7 @@ function runserver_drush_command() {
       'user' => 'If opening a web browser, automatically log in as this user (user ID or username). Default is to log in as uid 1.',
       'browser' => 'If opening a web browser, which browser to user (defaults to operating system default).',
       'dns' => 'Resolve hostnames/IPs using DNS/rDNS (if possible) to determine binding IPs and/or human friendly hostnames for URLs and browser.',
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
It's not easy creating a runserver prepend file that works right; seems that few folks would have the need or skill to use this.  Would it be possible to fix the problems in the referenced issues directly in the appropriate Drush prepend file?

typhonius:
Really it's only the bit which looks for actual files and doesn't cause PHP to be a sad :panda_face: when you want to send a path with a period in.

greg-1-anderson:
How does this PR fix the problem with paths with a period?

typhonius:
So my usecase for this was to use the d8 router file on a d7 site. I just relooked at the D7 prepend file and I'd probably have to do something like:

- Check if the path had a period in
- Check if the file existed on the server
- If yes to 1 and no to 2 call index.php directly with the path

I'm not really even sure if that'd work though.

greg-1-anderson:
Seems to me that I tried the D8 router file with D7, to see if we could have just one version, and it did not work.  It would be useful if someone could do the full matrix of testing -- Drupal verison vs. special characters in filename vs. different router files, etc., and report which combinations work and which do not. Ideally, Drush should always select a prepend file for the user that just worked.

kenorb:
Related PR: #1658

weitzman:
lets close in favor of #1658

PR #1706, opened 2015-10-24
Add current alias support in current command data.

IRuslan:
# Motivation
I struggled many times with commands with long and not-intuitive options.

# Solution
I've researched and figured out that impossible to know within invoked command, what alias was used for call actually.

My patch introduce new key for $command data, and intented to use in next way:
```php
$command = drush_get_command();
if (array_key_exists('current_alias', $command) && $command['current_alias'] == 'some-alias') {
  // Set up some specific stuff or default options values.
}
```

I've strongly motivated to remove 'is_alias' flag at all, because we just could check if 'current_alias'  is not empty. But probably it's not very compatibility-friendly...

### DIFF
--- a/includes/command.inc
+++ b/includes/command.inc
@@ -1037,7 +1037,10 @@ function drush_get_commands($reset = FALSE) {
         if (isset($command['aliases']) && count($command['aliases'])) {
           foreach ($command['aliases'] as $alias) {
             $commands[$alias] = $command;
+            // Actually is_alias flag could be replaced with usage of current_alias.
+            // But it's here for backwards compatibility.
             $commands[$alias]['is_alias'] = TRUE;
+            $commands[$alias]['current_alias'] = $alias;
           }
         }
       }


IRuslan:
See #1707 as a proof of concept for this functionality.

weitzman:
Whats the use case for this?

greg-1-anderson:
The example in the summary is poor; an alias should set the option values that it wants to. Commands should not test for specific aliases by name, ever.

If you want to get a reference to the current alias, it is available via `drush_get_context('DRUSH_TARGET_SITE_ALIAS')`.

It is definitely an aspiration to replace many/most drush_get_context uses with objects; if it was common for commands to reference the current alias, then I think this PR would be a good enhancement. However, since the usual use case is for aliases to set option values, and commands to act on option data, not alias data, perhaps this is not necessary. (Notable exceptions to the usual rule are sql-sync, rsync, but those are not able to use this anyway.)

IRuslan:
Greg, sorry if i was unclear.

I told not about site alias, but about the command alias being used to invoke a command. So it's not available via: 
```PHP
drush_get_context('DRUSH_TARGET_SITE_ALIAS')
```

As @weitzman pointed me here #1707, it's much better to use drushrc functionality to achieve my initial goal.

greg-1-anderson:
Oh, okay. If you define shell aliases, they can add options as well.

PR #1707, opened 2015-10-24
User login shortcut no browser

IRuslan:
# Motivation
From my practice it's a common case of user-login command usage is just to get link without being logged in browser.
In the same time it's quite painful to append each time '--browser=0'.

# Solution
This solution is based on this pull request — #1706 and serve as an example of usage in the same time.
I've didn't find any other convinient way to achieve this.

### DIFF
--- a/commands/user/user.drush.inc
+++ b/commands/user/user.drush.inc
@@ -189,7 +189,7 @@ function user_drush_command() {
   );
   $items['user-login'] = array(
     'description' => 'Display a one time login link for the given user account (defaults to uid 1).',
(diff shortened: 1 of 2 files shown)


weitzman:
if you get tired of typing options like --browser=0, the right fix is to add some configuration that sets this for you every time. See https://raw.githubusercontent.com/drush-ops/drush/master/examples/example.drushrc.php. Also run `drush topic` to read much more about Drush.

IRuslan:
Agree, your approach is much better. Didn't know about the drushrc functionality.

PR #1709, opened 2015-10-25
Allow Drush commands to be defined in YAML files rather than via php.

phenaproxima:
Implements the idea in http://blog.riff.org/2015_10_23_drupal_8_tip_of_the_day_replace_hook_drush_command_by_a_yaml_file#comment-4168

This allows command files to be static YAML files as well as PHP, similar to how permissions are implemented in Drupal 8.

### DIFF
--- a/commands/core/image.drush.inc
+++ b/commands/core/image.drush.inc
@@ -27,21 +27,6 @@ function image_drush_command() {
     ),
     'aliases' => array('if'),
   );
-  $items['image-derive'] = array(
-    'description' => 'Create an image derivative.',
-    'core' => array('7+'),
-    'drupal dependencies' => array('image'),
-    'arguments' => array(
-      'style' => 'An image style machine name.',
-      'source' => 'Path to a source image. Optionally prepend stream wrapper scheme.',
-    ),
-    'required arguments' => TRUE,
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
I like it.  @weitzman ?

weitzman:
Looks simple enough. However, I'd like to consider more thoroughly what ideal command definition looks like. Especially, I want to see what we can learn from Symfony Console. Better to change command definition once (even considering the nice backward compat proposed here). We already have supported multiple makefile formats for a year. Thats non-ideal.

greg-1-anderson:
Yes, good idea. Timing this with an object-oriented command definition would be good.

greg-1-anderson:
Not confirmed yet, but seems likely that I will be able to go (or at least possible). Session proposals are due at the end of the week, or thereabouts. I was planning on proposing two sessions: re-doing my "Drush in the Composer Era", because I think it contains important information that bears repeating, and also a beginner session. I got a lot of feedback at BADCamp that indicates that a lot of people never found out about a bunch of features, or have forgotten. I think there is a need for a review that would be of general interest.

Are you going to Mumbai? If so, are you interested in getting any Drush info out there?

 - Greg

PR #1711, opened 2015-10-26
Add config-pull command.

weitzman:
Add config-pull command - a simpler but less capable version of [config-merge](https://github.com/drush-ops/config-extra). This one does a config-export on a remote site (using a tmp dir) and rsync to the specified local config directory. From there, the developer must resolve conflicts (without the complex/pleasure of a 3 way merge) and do commit herself.

In order to support this, config-export gained the ability to pass --destination without a value. This saves the config files to a Drush backup directory (analogous to sql-dump).

Lets discuss how useful this is.


weitzman:
Fixed a bug where we passed all config dirs inside their backup dir. We want files without the subdir.

I think I've discovered a minor backend bug. This command sets `$backend_options['backend-simulate'] = TRUE;`. backend.inc should place the --simulate option in the global options section of the core-rsync command but it places it later, causing rsync to complain with `rsync: --simulate: unknown option`. The actual shell command is:

    rsync -e 'ssh ' -akzv --exclude=".git" --exclude=".gitignore" --exclude=".hg" --exclude=".hgignore" --exclude=".hgrags" --exclude=".bzr" --exclude=".bzrignore" --exclude=".bzrtags" --exclude=".svn" --stats --progress --remove-source-files --delete --simulate moshe.weitzman@bob:/Users/moshe.weitzman/drush-backups/config-export/20151027161329/ /Users/moshe.weitzman/htd/d8/sites/default/files/config_kcbWov0GMOzi0wJ5bNz-ZVq5gORUyOMPpLx-3Ox6M9qNodZPxWm22-sG5ygtC5qr9vbkGSwo-g/sync

greg-1-anderson:
Looks interesting.

So, the use case is that the user runs `drush sql-pull`, and this command copies the configuration from another site over the 'vcs' config store.  A 'git diff' will then show all of the differences made on both sites, and the user must figure out which to keep, and which to throw away.  If none of the config changes made on the two sites affect the same config files, then conflict resolution is as easy as 'git checkout -- file-to-keep.yml'.

I think that I could simplify config-merge quite a bit by: (a) prepare to pull (create a branch), (b) call config-pull, and (c) 3-way merge and/or just commit.

Of course, this is only the "rsync" mode of config-merge.  Need to contemplate whether the "git" mode can just be abandoned, or if there's an easy way to make config-pull pull from a branch.  Don't want to complicate config-pull too much, but the merge is more reliable if everything stays inside git.

I like the direction; I'd like to simplify / rewrite config-merge some.

weitzman:
@greg-1-anderson - when you get a chance could you look into the `rsync: --simulate: unknown option` bug mentioned above?

greg-1-anderson:
It is All-Hallow's eve, so I did not work hard on this -- I just committed a fix without testing it in context. Should be fine now, though. Will try again later if not.

weitzman:
That fixed backend-simulate. Thanks. I just merged this PR.

config-pull now takes two site aliases as arguments, just like sql-sync. I contemplated adding a config-import at the end and calling this config-sync but thats not really the use case I'm solving for. Input welcome.

Folks wishing to use this need update their Drush on both sides. If Drush is a dependency of your site (recommended), then you only need to update composer.json once, git commit, and then deploy new code two both environments.

greg-1-anderson:
I can see why you did not want to name this config-sync, or make it do a config-import on the end. It is probably best to keep this simple.

I could not think of a simple way to do either git or rsync transfers in this command. Probably better to leave this as implemented, and handle the "git" case in config-merge, where it is needed.

colans:
I'm wondering if my PR in config-extra (above) can be dropped now that `config-pull` exists.  Or would there still be a reason to keep that open?  I'm wondering if the extra complexity there adds something.

PR #1716, opened 2015-10-27
Pass shell ppid from Drush finder 

greg-1-anderson:
The Drush finder (and before it, the Drush wrapper, for those who used it) broke the site-set command, because the parent process ID for the Drush application is no longer set to the shell that called Drush.

This PR passes in the shell PID from the Drush Finder, so that it is available when needed.

### DIFF
--- a/includes/drush.inc
+++ b/includes/drush.inc
@@ -272,6 +272,7 @@ function drush_get_global_options($brief = FALSE) {
   $options['yes']                = array('short-form' => 'y', 'context' => 'DRUSH_AFFIRMATIVE', 'description' => "Assume 'yes' as answer to all prompts.");
   $options['no']                 = array('short-form' => 'n', 'context' => 'DRUSH_NEGATIVE', 'description' => "Assume 'no' as answer to all prompts.");
   $options['simulate']           = array('short-form' => 's', 'context' => 'DRUSH_SIMULATE', 'never-propagate' => TRUE, 'description' => "Simulate all relevant actions (don't actually change the system).");
+  $options['shell-pid']          = array('hidden' => TRUE, 'description' => 'Used by the Drush "Finder" script to pass in the PID of the shell for site-set.', 'context' => 'DRUSH_SHELL_PPID', 'never-propogate' => TRUE);
   $options['pipe']               = array('short-form' => 'p', 'hidden' => TRUE, 'description' => "Emit a compact representation of the command for scripting.");
   $options['help']               = array('short-form' => 'h', 'description' => "This help system.");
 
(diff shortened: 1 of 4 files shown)


weitzman:
The description and the code make sense. Oddly, I can't reproduce the problem. I'm sure I'm using pcntl and not passthru in the Finder. 

What are the steps you are using to reproduce this? Paste in alias definition too if possible.

greg-1-anderson:
0. Pull the master branch (and 7.x, if you want to compare behavior), as I pushed a debug log to site-set this morning.

1. Ensure that the Drush in your $PATH is a Drush finder, not the previous Drush launcher bash script.
```
$ head $(which drush)
#!/usr/bin/env php
<?php

/**
 * This is the Drush "finder" script, which is one part of the
 * Drush dispatching chain.  This is the script that
 * should appear in your global $PATH, or, if using Composer
 * (as usually the case), will be found in your vendor/bin directory.
```
2. Choose a site alias to test with
```
$ drush sa @badcamp.dev
$aliases["local.badcamp.dev"] = array (
  'root' => '/Users/ga/local/drupal/badcamp.dev',
  'uri' => 'localhost:9966',
);
```
3. Run site-set in --debug mode
```
$ drush site-set @badcamp.dev -d
Site set to @badcamp.dev
Site information stored in                                              [notice]
/var/folders/s4/_p47njzj3991f5m1450c9qkh0000gn/T//drush-env-ga/drush-drupal-site-79977
```
4. Compare with shell PID
```
$ echo $$
79476
```
5. Numbers are different?  Probably not working.
```
$ drush site-set -
Site set to @none
$ drush site-set -
Site set to @none
```
6. Apply this patch and try again:
```
$ drush site-set @badcamp.dev -d
Site set to @badcamp.dev
Site information stored in                                              [notice]
/var/folders/s4/_p47njzj3991f5m1450c9qkh0000gn/T//drush-env-ga/drush-drupal-site-79476
$ echo $$
79476
$ drush site-set -
Site set to @none
$ drush site-set -
Site set to @badcamp.dev
```
The tests are failing because --shell-pid is being included in backend invoke dispatches. I thought I set NEVER_PROPOGATE to prevent this; will go look and fix that now.

greg-1-anderson:
Had a typo in the 'never-propagate' flag. :P Working now.

weitzman:
Thanks for the clear instructions. I still could not reproduce the bug until I added a drush.wrapper. Now it all makes sense. Feel free to merge.

FYI, I just pushed a commit which hopefully clarifies the instructions for the wrapper.

greg-1-anderson:
Cool, thanks.

weitzman:
FYI, folks need to update their site-local Drush installs if they are going to update the global one. If you only update global Drush, you get `Unknown option: --shell-pid. `. An alternative to synchronized updating is adding --strict=0 to each Drush call.

greg-1-anderson:
Actually, we need to solve this problem, because the site-local Drush is going to have trouble with --shell-pid if it is a 7.x or 6.x version.

In the short term, we could backport --shell-pid to the 7.x and 6.x branches.  In the longer term, though, I wonder if we might have need to pass information from the Drush Finder to an arbitrary Drush that might not understand the new flag.

Should I perhaps rewrite this PR using an environment variable instead of a global option?

weitzman:
Changing to an env var seems like a durable, easy fix.

Issue #1712, opened 2015-10-27
Global drush 8.x causes sql-cli to hang at "proc_open(...)"


Shaun-Dychko:
`drush sql-cli` hangs on Mac OS X with MAMP (PHP 5.6.10). Other drush commands work fine, such as `drush sql-connect` and `drush sql-query`, so connection to the DB works. `drush sql-cli` also works with these same Drupal 7 sites hosted on different linux servers, one of which has the exact same version of drush, so I suspect the issue is with my Macbook, but I'm flummoxed. I ran `drush --debug sql-cli` and it gets to the line
```
Calling proc_open(mysql <various arguments>...
```
This is the case both with drush 7.0.0 and with drush 6.6.0. There aren't any error messages in the PHP log. Any pointers on how I could troubleshoot? I also moved ~/.drush in case something there was getting in the way.

Shaun-Dychko:
Oh, an interesting observation: if I remove ~/.composer/vendor/bin from my PATH and execute drush using ../vendor/drush/drush/drush from the site root, then `drush sql-cli` works! There's an issue with having drush 8.x installed globally with composer not handing off properly to the version installed locally with the site for this particular `sql-cli` command.

weitzman:
I can reproduce this if I force the drush_startup() to go through the passthru() code branch. Assigning to @greg-1-anderson

greg-1-anderson:
Maybe passthru doesn't do the right things with STDIN or tty setup or something.  Might need proc_open in startup.inc to fix.

weitzman:
fixed in 75aa74f

PR #1720, opened 2015-10-28
#1712: passthru does not play nicely with interactive commands

greg-1-anderson:
Use proc_open instead.


greg-1-anderson:
Tests pass, seems to work in casual testing.

weitzman:
cherry picked. thanks.

Issue #1618, opened 2015-09-15
Remove drush_is_local_host()


greg-1-anderson:
The drush_is_local_host(), and corresponding #check-local flag are unreliable, and should be removed.

Folks who want aliases that contain 'remote-host' entries that point at the local machine to behave like local aliases [should use an alias alter hook to remove the 'remote-host' entry](https://github.com/drush-ops/drush/pull/1615#issuecomment-140560370) using rules that are applicable for their system.

Should document this as well.

PR #1723, opened 2015-10-29
Fix '.local' being assumed as localhost

deviantintegral:
This fixes a bug from https://github.com/drush-ops/drush/commit/5fd73516d9d09c856fb4c70748bc0ca5dcd5ea54 / PR #1618 .

`drush_is_local_host()` added a test for the hostname ending in .local - but <a href="https://en.wikipedia.org/wiki/Multicast_DNS">.local is used for mDNS</a> and means machines on the local network, not the localhost itself. This breaks vagrant machines using mDNS for automatic publishing of records, but would also break using .local addresses for physical machines as well.

### DIFF
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -619,8 +619,7 @@ function drush_is_local_host($host) {
   // @see hook_drush_sitealias_alter() in drush.api.php.
   if (
     ($host == 'localhost') ||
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Looks good to me. Folks should just get used to using hook_alias_alter to identify their local sites.

Issue #1724, opened 2015-10-30
Make drush_drupal_version() pull from Boot classes


quicksketch:
With #342 and other enchancements, we have Boot classes that separate out the version-specific logic when bootstrapping D6/D7/D8 (and Backdrop).

One area where we're having trouble in Backdrop land is that `drush_drupal_version()` is hard-coded to look in particular areas for the `VERSION` constant. As Backdrop wants to identify itself as D7 (for now), but the logic is hard-coded, we are unable to make drush recocognize Backdrop without some pretty sad hacking (like making an empty file at `includes/bootstrap.inc` that contains a `VERSION` constant).

Separating out the logic for different versions of Drupal is what Boot classes were intended for, so could we move the logic of `drush_drupal_version()` into these classes?

quicksketch:
Hopefully this isn't too presumptious, but here are PRs for 7.x and master:

7.x: https://github.com/drush-ops/drush/pull/1725
master: https://github.com/drush-ops/drush/pull/1726

If they need work I can just turn into a single PR if that's preferred.

quicksketch:
Thanks @weitzman and @greg-1-anderson!

PR #1726, opened 2015-10-30
Issue #1724: Move logic for version information into Boot classes.

quicksketch:
Fixes https://github.com/drush-ops/drush/issues/1724 for master.

### DIFF
--- a/includes/drupal.inc
+++ b/includes/drupal.inc
@@ -36,29 +36,9 @@ function drush_drupal_version($drupal_root = NULL) {
 
   if (!$version) {
     if (($drupal_root != NULL) || ($drupal_root = drush_get_context('DRUSH_DRUPAL_ROOT'))) {
-      // Try and find D8.
-      if (file_exists($drupal_root . '/autoload.php')) {
-        // Load the autoloader so we can access the class constants.
-        drush_drupal_load_autoloader($drupal_root);
-        // Drush depends on bootstrap being loaded at this point.
-        require_once $drupal_root .'/core/includes/bootstrap.inc';
-        if (defined('Drupal::VERSION')) {
-          $version = Drupal::VERSION;
-        }
-      }
-      else {
-        // D7 stores VERSION in bootstrap.inc.
-        // D6 and below does it in system.module.
-        $version_constant_paths = array('/includes/bootstrap.inc', '/modules/system/system.module');
-        foreach ($version_constant_paths as $path) {
-          if (file_exists($drupal_root . $path)) {
-            require_once $drupal_root . $path;
-            if (defined('VERSION')) {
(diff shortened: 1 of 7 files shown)


greg-1-anderson:
This looks good to me.  We are still in the situation where we hope that no code will be calling "drush_drupal_version()", ultimately, but this should help us get along in compatibility mode until then.

This PR does shine a light on an existing problem in Drush:  drush_drupal_version needs to do bootstrappy-like stuff, like loading the autoloader. This could cause problems if some code at some point called drush_drupal_version on more than one site.  To avoid this danger, we should remove the $drupal_root parameter, and force commands that need this facility to instead call the "status" command via drush_invoke_process().  Currently, only the archive commands do this.

I think I'm fine merging this PR as-is, though, and handling the API adjustment in a follow-on issue.

includes/drupal.inc
```diff
@@ -36,29 +36,9 @@ function drush_drupal_version($drupal_root = NULL) {
...
-          if (file_exists($drupal_root . $path)) {
-            require_once $drupal_root . $path;
-            if (defined('VERSION')) {
-              $version = VERSION;
-              break;
-            }
-          }
-        }
+      $bootstrap = drush_bootstrap_class_for_root($drupal_root);
+      if ($bootstrap) {
```
weitzman: We have \Drush\Boot\EmptyBoot class. So should bootstrap class ever be FALSE?

weitzman:
I agree with removing $drupal_root param in a followup. The static caching in this function already makes it useless for calling on two different roots.

Code looks good to me. I added one line note for @greg-1-anderson to review.

quicksketch:
> We are still in the situation where we hope that no code will be calling "drush_drupal_version()", ultimately, but this should help us get along in compatibility mode until then.

Yeah, I thought about converting the existing code to use the Boot classes instead, but for starters I figured it'd be adequate to make it act as a compatibility layer. Maybe we could add a `@deprecated` indicator in the PHPdoc, but I'd defer to you guys on that one.

greg-1-anderson:
So, yes, drush_bootstrap_class_for_root might return NULL, because EmptyBoot always claims that it supports nothing at all.  EmptyBoot is inserted by drush_select_bootstrap_class(). I think the code should be refactored so that EmptyBoot is not added to the candidates, and drush_select_bootstrap_class instantiates one by name, rather than taking the last candidate.

The correct solution, though, is to remove the root parameter from the Drush version functions, and use drush_get_bootstrap_object() instead. Everyone should use that function to get the bootstrap object; only the bootstrap code should use drush_bootstrap_class_for_root() or drush_select_bootstrap_class().  We'd have to fix the archive commands to do this, though.

PR #1738, opened 2015-11-01
Just say "--no"

greg-1-anderson:
This PR supports --no- options as an alternate for =0 (e.g. --no-browser inste…ad of --browser=0).

I noticed that since forever ago (Drush 5, or whenever it was), Drush had a provision for --no options.  The function drush_append_negation_options() ensures that --no-foo does not raise an error if there is any command option --foo.  However, despite this assurance, nothing in Drush ever allowed these options to actually work.

This PR introduces a minor enhancement that does just that.  If this looks acceptable, I could update the documentation for options where --no makes sense.


weitzman:
Yes, this is much desired. Thanks.

We should document this. No idea where.

greg-1-anderson:
I scattered documentation around various places in Drush in the commit above. I'll merge this once the tests come back green.

PR #1740, opened 2015-11-02
Fix FormattableMarkup Invalid placeholder error

stefanruijsenaars:
See https://www.drupal.org/node/2575703#comment-10516850

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -154,6 +154,10 @@ function dt($string, $args = array()) {
     // so we'll pre-replace these before calling translate().
     $legacy_args = find_legacy_dt_args($args);
     $string = strtr($string, $legacy_args);
+    // Remove the legacy replacements.
+    foreach (array_keys($legacy_args) as $legacy_arg_key) {
+      unset($args[$legacy_arg_key]);
+    }
     $output = \Drupal::translation()->translate($string, $args);
   }
   else if (function_exists('t') && drush_drupal_major_version() <= 7 && function_exists('theme')) {
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -150,8 +150,8 @@ function dt($string, $args = array()) {
   // The language system requires a working container which has the string
   // translation service.
   else if (drush_drupal_major_version() >= 8 && \Drupal::hasService('string_translation')) {
-    // Drupal 8 deprecates (and will eventually remove) !var replacements,
-    // so we'll pre-replace these before calling translate().
+    // Drupal 8 removes !var replacements, creating a user-level error when
+    // these are used, so we'll pre-replace these before calling translate().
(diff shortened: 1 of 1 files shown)


wimleers:
This fixed it!

But I think this should also update

```
// Drupal 8 deprecates (and will eventually remove) !var replacements,
```

because now it's no longer just deprecated: it actually causes errors.

stefanruijsenaars:
Updated the comment

includes/output.inc
```diff
@@ -150,10 +150,14 @@ function dt($string, $args = array()) {
...
   // translation service.
   else if (drush_drupal_major_version() >= 8 && \Drupal::hasService('string_translation')) {
-    // Drupal 8 deprecates (and will eventually remove) !var replacements,
-    // so we'll pre-replace these before calling translate().
+    // Drupal 8 removes !var replacements, creating a user-level error when
+    // these are used, so we'll pre-replace these before calling translate().
     $legacy_args = find_legacy_dt_args($args);
     $string = strtr($string, $legacy_args);
+    // Remove the legacy replacements.
+    foreach (array_keys($legacy_args) as $legacy_arg_key) {
```
dawehner: Why do we not fix the usages of those entries or convert them?

stefanruijsenaars: Convert them into? strtr() inside of the dt() call in the 100+ instances where we use !var rather than doing it in find_legacy_dt_args()?

Not opposed to that but also wondering what we gain by doing so? In drush we shouldn't be having the usual sanitization concerns because the output is CLI

greg-1-anderson:
I think this PR is the right approach.

There are two questions that were raised her:

1) Why not just change all of the !var replacements in Drush, rather than fixing them up in dt()?

I am not sure that it is helpful to change !var to %var everywhere in Drush. Some instances may contain strings that the author intended to emit to the terminal without encoding. %var will encode.  Also, there are 500+ Drush extensions, many of which will also use !var.  It would be an extracted process to change all of these, even if it was a good idea to do so.

2) Why call through to t() at all?

One previously-stated reason was so that Drush translation could take advantage of translations done by Drupal. This will only be relevant for a very small number of very short strings, though, so I don't think it is very important. A more significant reason is that I believe that some code in Drupal, e.g. form submit button handlers, expect that their string parameters (e.g. the button label text) has been passed through t() (or the d8 translation class equivalent), and will not function correctly if a raw string is provided. I have to admit that I am not fully aware of the exact mechanisms involved here. It does seem to me that it would be possible to have dt() do its own substitutions, and provide a separate function for use in cases where we must use Drupal translation.  We could certainly go this route, but we'd have to get it right. The first step would be to stop calling through to Drupal in dt(), and see what (e.g. site-install?) breaks.

In the short term, though, I'd recommend just merging this PR.

PR #1741, opened 2015-11-02
Add a 'core-init' aka 'init' command.

greg-1-anderson:
A number of cli tools provide an 'init' command to help set up the tool's configuration files after installation.  This PR adds a basic init command to do the same thing for Drush.

There is more we could potentially do here, but this is a start.  It ensures that `drush` is in the global $PATH, and also sets up our example Bash configuration file, so that completion works, and various bash aliases and functions are offered.  Currently, this is done by modifying .bashrc, but we have the opportunity to add heuristics to find the best file for the platform here. I have heard anecdotally that .bash_profile is better for the Macintosh, but my recent experiments on Yosemite show that only .bashrc is sourced for newly-created accounts, so it seems to me that .bashrc is best for Mac and probably all flavors of Linux.  This is easily adjusted if anyone has more specific information about this.  `man bash` is not very helpful, as the behavior of Bash changes depending on how the host system chooses to call it.

### DIFF
--- a/commands/core/init.drush.inc
+++ b/commands/core/init.drush.inc
@@ -1,0 +1,115 @@
+<?php
+
+/**
+ * @file
+ *   Set up local Drush configuration.
+ */
+
+/**
+ * Implementation of hook_drush_command().
+ *
+ * @return
+ *   An associative array describing your command(s).
+ */
+function init_drush_command() {
+  $items['core-init'] = array(
+    'description' => 'Initialize local Drush configuration.',
+    'aliases' => array('init'),
+    'bootstrap' => DRUSH_BOOTSTRAP_NONE,
+  );
+  return $items;
+}
+
+/**
+ * Initialize local Drush configuration
+ */
+function drush_init_core_init() {
+  $home = drush_server_home();
+  $drush_config_dir = $home . "/.drush";
+  $drush_config_file = $drush_config_dir . "/drushrc.php";
+  $drush_bashrc = $drush_config_dir . "/drush.bashrc";
+  $examples_dir = DRUSH_BASE_PATH . "/examples";
+  $example_configuration = $examples_dir . "/example.drushrc.php";
+  $example_bashrc = $examples_dir . "/example.bashrc";
+  $bashrc_additions = array();
+
+  // Create a ~/.drush directory if it does not yet exist
+  if (!is_dir($drush_config_dir)) {
+    drush_mkdir($drush_config_dir);
+  }
+
+  // If there is no ~/.drush/drushrc.php, then copy the
+  // example Drush configuration file here
+  if (!is_file($drush_config_file)) {
+    @copy($example_configuration, $drush_config_file);
+    drush_log(dt("Copied example Drush configuration file to !path", array('!path' => $drush_config_file)), 'ok');
+  }
+
+  // If there is no ~/.drush/drush.bashrc file, then copy
+  // the example bashrc file here
+  if (!is_file($drush_bashrc)) {
+    @copy($example_bashrc, $drush_bashrc);
+    $bashrc_additions = array("%$drush_bashrc%" => "# Include Drush bash customizations\n. $drush_bashrc\n\n");
+    drush_log(dt("Copied example Drush bash configuration file to !path", array('!path' => $drush_bashrc)), 'ok');
+  }
+
+  // Decide whether we want to add our Bash commands to
+  // ~/.bashrc or ~/.bash_profile
+  $bashrc = drush_init_find_bashrc($home);
+
+  // If drush is not in the $PATH, then figure out which
+  // path to add so that Drush can be found globally.
+  if (!drush_which("drush")) {
+    $drush_path = drush_find_path_to_drush();
+    $drush_path = preg_replace("%^$home/%", '$HOME/', $drush_path);
+
+    $bashrc_additions = array("%$drush_path%" => "# Path to Drush, added by 'drush init'\n\$PATH=\$PATH:$drush_path\n\n");
+  }
+
+  // Modify the users bashrc file, adding our customizations.
+  $bashrc_contents = "";
+  if (file_exists($bashrc)) {
+    $bashrc_contents = file_get_contents($bashrc);
+  }
+  $new_bashrc_contents = $bashrc_contents;
+  foreach ($bashrc_additions as $pattern => $addition) {
+    // Only put in the addition if the pattern does not already
+    // exist in the bashrc file.
+    if (!preg_match($pattern, $new_bashrc_contents)) {
+      $new_bashrc_contents = $addition . $new_bashrc_contents;
+    }
+  }
+  if ($new_bashrc_contents != $bashrc_contents) {
+    file_put_contents($bashrc, $new_bashrc_contents);
+    drush_log(dt("Updated bash configuration file !path", array('!path' => $bashrc)), 'ok');
+  }
+}
+
+/**
+ * Determine which .bashrc file is best to use on this platform.
+ */
+function drush_init_find_bashrc($home) {
+  return $home . "/.bashrc";
+}
+
+/**
+ * Determine where Drush is located, so that we can add
+ * that location to the $PATH
+ */
+function drush_find_path_to_drush($home) {
+  // First test: is Drush inside a vendor directory?
+  // Does vendor/bin exist?  If so, use that.  We do
+  // not have a good way to locate the 'bin' directory
+  // if it has been relocated in the composer.json config
+  // section.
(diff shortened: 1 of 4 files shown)


commands/core/init.drush.inc
```diff
@@ -0,0 +1,115 @@
...
+ * Implementation of hook_drush_command().
+ *
+ * @return
+ *   An associative array describing your command(s).
+ */
+function init_drush_command() {
+  $items['core-init'] = array(
+    'description' => 'Initialize local Drush configuration.',
+    'aliases' => array('init'),
+    'bootstrap' => DRUSH_BOOTSTRAP_NONE,
```
weitzman: Perhaps provide an example of running this on a remote alias (it would work OK?)

commands/core/init.drush.inc
```diff
@@ -0,0 +1,115 @@
...
+    @copy($example_bashrc, $drush_bashrc);
+    $bashrc_additions = array("%$drush_bashrc%" => "# Include Drush bash customizations\n. $drush_bashrc\n\n");
+    drush_log(dt("Copied example Drush bash configuration file to !path", array('!path' => $drush_bashrc)), 'ok');
+  }
+
+  // Decide whether we want to add our Bash commands to
+  // ~/.bashrc or ~/.bash_profile
+  $bashrc = drush_init_find_bashrc($home);
+
+  // If drush is not in the $PATH, then figure out which
```
weitzman: Lets capitalize Drush in this case.

commands/core/init.drush.inc
```diff
@@ -0,0 +1,115 @@
...
+  // path to add so that Drush can be found globally.
+  if (!drush_which("drush")) {
+    $drush_path = drush_find_path_to_drush();
+    $drush_path = preg_replace("%^$home/%", '$HOME/', $drush_path);
+
+    $bashrc_additions = array("%$drush_path%" => "# Path to Drush, added by 'drush init'\n\$PATH=\$PATH:$drush_path\n\n");
+  }
+
+  // Modify the users bashrc file, adding our customizations.
```
weitzman: user's

commands/core/init.drush.inc
```diff
@@ -0,0 +1,115 @@
...
+function drush_init_find_bashrc($home) {
+  return $home . "/.bashrc";
+}
+
+/**
+ * Determine where Drush is located, so that we can add
+ * that location to the $PATH
+ */
+function drush_find_path_to_drush($home) {
+  // First test: is Drush inside a vendor directory?
```
weitzman: I know that you did not use dirname(DRUSH_COMMAND) for a reason. I guess that finds drush.php and not the Finder? Maybe the finder should pass itself to to the launcher via Env Variable?

greg-1-anderson: dirname(DRUSH_COMMAND) == DRUSH_BASE_PATH; I could use either, as they both indicate the same thing. Switching to dirname(DRUSH_COMMAND) would be fine if you think that is a more documented / more official constant than DRUSH_BASE_PATH.

In the use-case where we are looking for the location of the Drush executable, the Drush Finder and Launcher that were used in this execution of Drush will always be in the same directory as drush.php. If a different Drush Finder was in use, then presumably it is already on the $PATH.

One use case we might want to consider improving here is where the user creates a new Drupal 8 site via composer create-project, and then runs `./vendor/bin/drush init`. This is a fairly compelling use-case, but in the current implementation, you end up with the current site-local Drush in your $PATH, which might not be what you want. It might seem odd for Drush to install and/or copy itself to a "global" location, but perhaps a `composer global install drush` is what is needed here.  Was uncertain about that, so I just let it select the site-local Drush for now.

commands/core/init.drush.inc
greg-1-anderson: Yes, it would work okay, but only if Drush was already installed remotely. It seems that a local call to `drush init` after installation is going to be the typical use-case. Even if the user is trying to "catch up" and init a bunch of remote Drush installations, `drush @remote ssh` followed by `drush init` is a more reasonable approach.

Another issue about `drush @remote init` that might be strange is that the remote version of Drush might be different than the local version of Drush.  Right now, the operations performed by `drush init` should work all the way back to Drush 5; I don't know how likely it is that we would make `drush init` set something up that only worked on the most recent version of Drush, but I am all the same reluctant to promote remote execution of init.

weitzman: If you think remote execution is a bad idea, then we could trap it and recommend the `drush ssh` approach in the error message. I'm ambivalent.

commands/core/init.drush.inc
weitzman: Yeah, I'd go with selecting site-local Drush as well. This does mean that wiping a random codebase can unknowingly ruin your drush command. If would be good if init were idempotent so that folks could rerun in a different codebase and get up to speed. Is it idempotent already? I'm OK with extra junk getting appended to bash files in this unusual case.

greg-1-anderson: Yes, `drush init` is idempotent; in the case of wiping a radom codebase, then `drush` would not be in the $PATH any longer.  If they can find any copy of Drush and run `drush init` on it, that would fix the problem.

The other issue is that the example files are copied over to the user's $HOME/.drush directory, so they can be customized. This means that they do not automatically benefit from enhancements that are made to these in new versions of Drush.  Maybe there could be options that control "freshening" these files, although that could be dangerous.

Perhaps the example drushrc.php file could be converted into a usable-as-is file, so that 'unsuck' & c. become available immediately after `drush init`

weitzman: I'm not too concerned about the freshness issue.

I :heart: unsuck, but thats not much of a reason to switch to usable-as-is. If we can find a compelling reason, I'm up for uncommenting a few lines.

commands/core/init.drush.inc
greg-1-anderson: I think if someone decides on their own to use a remote alias, they probably know what they are doing. I see this as an edge case, not as something to actively prevent.

greg-1-anderson:
Is this ready to merge as-is now?  We can work on improving the drushrc example, if there are things that can be done there, in a separate issue.

weitzman:
Made a tiny improvement. I think the command is ready. It seems like this command is testable in terms of checking that the files go where they are expected, have correct $PATH, the regex works, etc.

weitzman:
Perhaps you can write some tests, and I will update the [configure page](http://docs.drush.org/en/master/configure/) which is looking crufty given the finder/launcher change and this init command,

greg-1-anderson:
Sure, I will write some tests. Would be good to also start publishing .phar builds; that would ease installation quite a bit.

greg-1-anderson:
I tried to write a test that got bash to source the bashrc file, and confirm that an alias was defined. I finally came up with something that worked locally, but it does not work on Travis.  :(

Do you think the tests are good enough as they are, without the failing test?

tests/Unish/CommandUnishTestCase.php
```diff
@@ -369,6 +369,31 @@ function parse_backend_output($string) {
...
+   */
+  function assertLogHasMessage($log, $message, $logType = FALSE) {
+    foreach ($log as $entry) {
+      if (!$logType || ($entry['type'] == $logType)) {
+        if (strpos($entry['message'], $message) !== FALSE) {
+          return TRUE;
+        }
+      }
+    }
+    $this->assertTrue(FALSE, "Could not find expected message in log: " . $message);
```
weitzman: Slightly simpler: $this->fail() method.

tests/initCommandTest.php
```diff
@@ -0,0 +1,34 @@
...
+    $this->drush('core-init', array(), array('backend' => NULL));
+    $parsed = $this->parse_backend_output($this->getOutput());
+    // First test to ensure that the command claimed to have made the expected progress
+    $this->assertLogHasMessage($parsed['log'], "Copied example Drush configuration file", 'ok');
+    $this->assertLogHasMessage($parsed['log'], "Copied example Drush bash configuration file", 'ok');
+    $this->assertLogHasMessage($parsed['log'], "Updated bash configuration file", 'ok');
+    // Next we will test to see if there is evidence that those
+    // operations worked.
+    $home = getenv("HOME");
+    $this->assertTrue(is_file("$home/.drush/drushrc.php"), "The expected ~/.drush/drushrc.php file does not exist");
```
weitzman: $this->assertFileExists(). With this, you can omit the $text

weitzman:
Can we test that our $bashrc_additions happenned as expected?

greg-1-anderson:
The tests do not add Drush to the $PATH, since it is already on the path from the unish configuration.  I can add a test to see if our Drush bashrc is sourced in .bashrc, though.

greg-1-anderson:
Tests are passing now.

PR #1743, opened 2015-11-03
Issue #1742: Fix array_flip error in find_legacy_dt_args() function when the dt() function was chained many times

### LINKED ISSUE #1742: Method find_legacy_dt_args() fails when I give the result of a dt() function to another dt() function
With Drupal 8, when I give a dt() function result to another dt() function as arg, like:

```
dt('test translate with @arg', ['@arg' => dt('An argument')])
```

I have this error:

```
array_flip(): Can only flip STRING and INTEGER values! output.inc:195   [warning]
array_flip(): Can only flip STRING and INTEGER values! output.inc:195   [warning]
```

The problem is:

```
function find_legacy_dt_args($args) {
  return array_flip(array_filter(array_flip($args), function ($v) { return $v[0] == '!'; } ));
}
```

You try to flip an array with a `\Drupal\Core\StringTranslation\TranslatableMarkup` as value
(It's impossible to an array_flip to call the __toString() method of this object)

I think it would be better to change your `find_legacy_dt_args($args)` like this to be sure the array_value is transformed to a string:

```
function find_legacy_dt_args($args) {
  $legecy_args = [];
  foreach ($args as $arg => $replacement) {
    if ($arg[0] == '!') {
      $legecy_args[$arg] = $replacement;
    }
  }
  return $legecy_args;
}
```

What do you think about that ?

vpeltot:
Fixes #1742 for master.

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -185,7 +185,13 @@ function dt($string, $args = array()) {
  * will flip the array twice to filter on the key.
  */
 function find_legacy_dt_args($args) {
-  return array_flip(array_filter(array_flip($args), function ($v) { return $v[0] == '!'; } ));
+  $legacy_args = [];
+  foreach ($args as $arg => $replacement) {
+    if ($arg[0] == '!') {
+      $legacy_args[$arg] = $replacement;
+    }
+  }
+  return $legacy_args;
 }
 
 /**


greg-1-anderson:
Looks good to me. Thanks.

Issue #589, opened 2014-04-16
php-options in aliases ignored


techmunk:
I have an alias in aliases.drushrc.php that sets the 'php-options' setting.
This is ignored totally. Is this a mistake?

Removing the --php-options part from the final exec line in drush executable makes my php-options get respected again.

Am I missing something?

Steps to test:
1. Create a simple php file in `/tmp/test.php`

``` php
<?php
echo "I was run\n";
```
1. Setup a drush alias.

``` php
<?php
$aliases['test'] = array(
  'php-options' => '-d auto_prepend_file=/tmp/test.php'
);
```
1. Run `drush @test | head -n 1`
   - Expected output - `"I was run"`.
   - Actual output - `"Execute a drush command......."`
2. In main drush file, change line 129 from
   `exec "$php" $php_options "$SCRIPT_PATH" --php="$php" --php-options="$php_options" "$@"`
   to
   `exec "$php" $php_options "$SCRIPT_PATH" --php="$php" "$@"`
   (Removes the php-options settings).
3. run `drush @test | head -n 1` again.
4. Output - We see the first line is `"I was run"`

PR #1752, opened 2015-11-06
Vastly simplify install documentation

weitzman:
I removed a lot of cruft, including the whole configure.md file. The interesting bits from there are preserved as bullets in the global install section. 

- [x] I put the PS1 code into example.bashrc file. Also do a $bash_addition in core-init? @greg-1-anderson 
- [ ] Add some pointers/docs for legacy users of DRUSH_INI, DRUSH_PHP, etc. @weitzman
- [x] drush_build_drush_command() sets env variables like DRUSH_PHP and PHP_OPTIONS. drush.phar users dont use launcher so these are useless to them, right? We should set --php and --php-options, no? Thats what the branch above it does. @greg-1-anderson


greg-1-anderson:
This looks good.
There is no need to do a $bash_additions for the prompt if it's already in the example bashrc file, as we source the example bashrc file already.
I think --php and --php-options can still be potentially useful to a phar user.  Consider the user who has installed the Drush phar as the global Drush on their laptop, and have also installed the Drush phar as the global Drush on their remote server.  If this user than installs a composer-managed Drupal site, the execution path should be:

(1) laptop drush.phar -> (2) remote drush.phar -> (3) remote site-local drush launcher

If (1) applies --php or --php-options, those will be ignored when launching the phar at (2), but will still be passed along to (3), which will use them.  So, I think that drush_build_drush_command() should continue to work the way that it does.  More testing for this sort of code path would be helpful, though.

I was looking at the Drush redispatch code, for `drush @remote cmd`.  I wanted to refactor it a bit, so that it used backend invoke for all cases except interactive commands. Maybe this can happen later, when backend invoke is factored into its own library, and altered to use the Symfony process component.

greg-1-anderson:
I think that in the past, backend invoke did not send DRUSH_PHP or PHP_OPTIONS to remote sites, although it sort of looks like it does do that now. That might be a mistake, as local and remote paths / options might vary from the remote and local sites.

The change was made in adf70771e7f39ae1bde27e47b8ea6db3cb03e0d8 to fix  #617 and #589. The intent of those issues was to allow users to set php and php-options in a site alias, which is extremely useful. I am wondering if at some point we might need to differentiate between DRUSH_PHP and PHP_OPTIONS that are set locally, vs. those that are set via an alias. I didn't consider that at the time we were making the change.

It's been like that for more than a year; without complaint, so maybe it's okay. I never set DRUSH_PHP or PHP_OPTIONS myself, though.

weitzman:
I was thinking about $additions for PS1 just to make it more discoverable. But yeah, not that important.

    (1) laptop drush.phar -> (2) remote drush.phar -> (3) remote site-local drush launcher

I'd like to consider what we can do in the case where there is no 3), because we are going to get bug reports about these alias options not working.

1. I think could deprecate DRUSH_PHP and tell people to just use the right $PATH. In the case of a remote alias with 'php' specified, we could pass PATH=/path/to/php:$PATH (assuming that works).
1. In the case of a remote alias with php-options, I think our only hope is to explicitly specify a php executable so we can add the options like so: `php $php-options drush.phar`. Sound feasible? If so, we could actually solve 1. above with this approach.

weitzman:
I went ahead and merged this. We can keep discussing here.

greg-1-anderson:
In the case of a site-local Drush, this should never be a .phar for Drupal 8 or any Composer-managed Drupal site, because in those instances, Drush's dependencies would not be correctly shared with the Drupal site's dependencies, as is the intention for a site-local Drush.  I don't know if we should try to detect this situation (Drush is a phar, and it loads an autoload.php file that is part of a Drupal site == trouble).

For Drupal 7 and earlier, it is perfectly reasonable to have a site-local Drush that is a phar. One use-case here would be to use an older version of Drush with older versions of Drupal (e.g. if they have some Drush extensions that have not been updated to work with the latest versions of Drush). So, it seems reasonable that we should try to support a site-local .phar.  It is also equivalent to want to support a remote global Drush that is a .phar.

`PATH=/path/to/php:$PATH` won't work directly, because the local system won't know what the current value of `$PATH` is, so it won't be able to append it. It feels wrong to require the local system to have recorded the full $PATH of the remote system. If might be possible to modify the PATH on the remote system if we used `ssh user@remote bash -c "env PATH=/path/to/php:$PATH drush.phar ..."`, but that seems a little roundabout. Also, I am not sure whether there would be additional commandline escaping (e.g. once for ssh, and once for bash).

The question, though, is how do we know that the remote Drush is a phar?  You suggestion to run `php $php-options drush.phar --php-options=$php-options` works great if Drush is a .phar, or if Drush is a "finder" script (which is also php), but it would fail if Drush was an older version, where the `drush` script is the Bash 'launcher'.  Maybe we only support selection of PHP and the passing of remote PHP options to a remote .phar if the user has specified a %drush-script that end `drush.phar`, and the remote script is named equivalently.  That much could work.

weitzman:
How would one use a drush.phar as a site-local drush? My only vision for a site-local drush is one that is listed in composer.json. 

For the record, `ssh user@remote bash -c "env PATH=/path/to/php:$PATH drush.phar ...` looks OK to me.

    Maybe we only support selection of PHP and the passing of remote PHP options to a remote .phar if the user has specified a %drush-script that end drush.phar, and the remote script is named equivalently.

I would vote for an explicit #drush-script-is-phar key on the alias, or something like that. We are encouraging folks to rename drush.phar to drush so relying on name won't work.

PR #1753, opened 2015-11-06
Make 'drush cache-clear' not break when not bootstrapped high enough

rmuit:
If drush is not bootstrapped to DRUSH_BOOTSTRAP_DRUPAL_FULL, 'drush cc' on D8 throws a ContainerNotInitializedException.

Below fixes this. Unless you care about 'views_invalidate_cache' always running (for some cases of 'cc'?) -- then the fix is more involved.

### DIFF
--- a/commands/core/views.d8.drush.inc
+++ b/commands/core/views.d8.drush.inc
@@ -455,7 +455,8 @@ function _views_drush_view_is_disabled(View $view) {
  * Adds a cache clear option for views.
  */
 function views_drush_cache_clear(&$types) {
-  if (\Drupal::moduleHandler()->moduleExists('views')) {
+  if (drush_has_boostrapped(DRUSH_BOOTSTRAP_DRUPAL_FULL)
+      && \Drupal::moduleHandler()->moduleExists('views')) {
     $types['views'] = 'views_invalidate_cache';
   }
 }


weitzman:
Can you elaborate on a use case where this error occurs?

rmuit:
Right; should have been a bit clearer.
* cd into a D8 docroot where the site cannot be immediately derived from that docroot. (e.g. some kind of multisite install that does not have DB credentials defined in sites/default/settings.php)
* run 'drush cc drush'
* ContainerNotInitializedException.

weitzman:
In  that case we are supposed to throw an error in drush_cache_clear_pre_validate(). Would be great if someone could figure out why we aren't doing that.

rmuit:
Actually, the error is thrown before drush_cache_clear_pre_validate() is doing actual checks. It goes drush_cache_clear_pre_validate() -> drush_cache_clear_types() -> drush_command_invoke_all_ref('drush_cache_clear', $types) -> views_drush_cache_clear($types) -> BOOM.

Here's another example, which makes clearer why drush cc drush should actually be able to run without booting to database level, than my previous one: suppose you're an automated system and you're provisioning a new site. You
* copy D8 skeleton
* do whatever stuff
* copy your own fancy additions (like, say, acsf module)
* want to run 'drush cc drush' to have your own fancy site init command recognized
* so you can run 'drush cc your-fancy-site-init-command'. All before having a fully bootstrappable live site.

weitzman:
I pushed an equivalent fix. Thanks.

PR #1756, opened 2015-11-09
Do not add --local flag when redispatching to a site-local Drush

greg-1-anderson:
Drush will automatically redispatch to a site-local Drush in `drush_preflight_command_dispatch()` if it finds that the current Drupal site has a site-local Drush, and a different (presumably global) Drush is being used.  This use case happens whenever you run `drush @localsite somecommnad` for a local site that has a site-local Drush.  The Drush finder does not handle this case, as the finder does not parse site alias files.  In instances where the Drush finder **does** redispatch, it does not add the --local flag.  `drush_preflight_command_dispatch()` adds --local only because it was implemented before we decided that --local should only be added if the user deliberately adds it via a Drush wrapper script. It was an oversight that it was not removed earlier.

### DIFF
--- a/includes/preflight.inc
+++ b/includes/preflight.inc
@@ -567,17 +567,15 @@ function drush_preflight_command_dispatch() {
     $this_drush = drush_find_drush();
     // If there is a local drush selected, and it is not the
     // same drush that is currently running, redispatch to it.
-    // We compare the dirname instead of the full dir + filename
-    // so that the finder / launcher will compare as the "same"
-    // Note that the "wrapper" will be in a different directory
-    // than the launcher; however, the wrapper should set --local,
-    // and therefore will not be considered here due to the
-    // 'empty($is_local)' test above.
-    if (file_exists($local_drush) && (dirname($this_drush) != dirname($local_drush))) {
+    // We assume that if the current Drush is nested inside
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
It looks like there's an infinite loop with the wrapper here; we're also seeing #1758.

greg-1-anderson:
Tests are all green, except for the unrelated Drupal 8 make test, which is failing. Ready for review.

includes/preflight.inc
```diff
@@ -567,17 +567,15 @@ function drush_preflight_command_dispatch() {
     $this_drush = drush_find_drush();
     // If there is a local drush selected, and it is not the
```
weitzman: please capitalize Drush

weitzman:
Looks good. Ready to merge - my capitalization request is debatable.

greg-1-anderson:
Sometimes, it's ambiguous whether a comment should refer to "Drush", the name of the program, or "drush", the thing you type into your shell to run Drush.  I'll try to standardize on only using "drush" when we literally mean "type this snippet into the shell to do X."  That should reduce comment capitalization arbitrariness.

I capitalized Drush in the comments here, and squash-merged to master.

Issue #1766, opened 2015-11-10
Replace Drupal version number pattern-based classes with named services/engines


quicksketch:
This issue grows out of https://github.com/drush-ops/drush/issues/1750. Where we were discussing move constants into Boot classes.

Right now we have a number of version-based names for our classes such as:
- Queue6/7/8
- Role6/7/8
- Sql6/7/8
- StatusInfo6/7/8
- User6/7/8
- UserSingle6/7/8

Rather than using number-based magic class names, it was suggested that the boot classes provide the names of these classes instead.

quicksketch:
If we pursue this, it starts looking in some ways a lot like these numbered classes because "services" in the D8 usage of the word. That is, they are swappable classes that return defined, consistent results regardless of the underlying system.

So what I'd like to suggest is that Drush adopt this concept of loading these existing classes as though they were "services" in the D8 sense. We already have these existing classes, and I'm not suggesting we modify them extensively (or at all), just that we change the way they are loaded away from being magically loaded with `drush_get_class()` and `drush_drupal_version()` to consolidating their loading into the boot classes. I'll try to work up a PR as a proof of concept.

quicksketch:
I created a PoC PR at https://github.com/drush-ops/drush/pull/1767.

I haven't figured out how Drush "engines" fit into all of this. I can't quite figure out how they work and what they accomplish. It looks like in a9e63b956ac5cafc2af0fbf0a83b60ea65094072, the ability for engines classes to live in `/lib` was added. If we pursued this further, engines could use service names instead of class names, reducing further the use of `drush_get_class()` and `drush_drupal_version()`. However, I don't think that engines and services are the same thing.

One thing I like about this concept of services is that commands could also list dependencies in an version-agnostic way. Right now commands can specify a version of Drupal core on which they depend, but with these services they could instead specify what services they need. These services are entirely version and CMS independent. It's just up to the boot class to implement (or not implement) each service. If the service isn't implemented, then the command wouldn't be allowed.

For example the SQL commands could easily be allowed within any system that provided the SQL service:
```php
  $items['sql-cli'] = array(
    'description' => "Open a SQL command-line interface using Drupal's credentials.",
    'bootstrap' => DRUSH_BOOTSTRAP_NONE,
    'service-requirements' => array('sql'), // <- New line.
    // ...
  );
```

[...]

PR #1767, opened 2015-11-11
Issue #1766: Replace Drupal version number pattern-based classes with named services.

quicksketch:
Proof of concept for https://github.com/drush-ops/drush/issues/1766.

This mostly untested PR just attempts to show how `drush_get_class()` could be replaced with a named-services approach, where each boot class would keep track of the services it supported.

### DIFF
--- a/commands/core/queue.drush.inc
+++ b/commands/core/queue.drush.inc
@@ -62,7 +62,7 @@ function drush_queue_run_validate($queue_name) {
  * Return the appropriate queue class.
  */
 function drush_queue_get_class() {
-  return drush_get_class('Drush\Queue\Queue');
+  return drush_get_service('queue');
 }
 
 /**
--- a/commands/core/role.drush.inc
+++ b/commands/core/role.drush.inc
@@ -267,10 +267,10 @@ function drush_role_perm($action, $rid, $permission = NULL) {
  * @param string $role_name
  * @return RoleBase
  *
- * @see drush_get_class().
+ * @see drush_get_service().
  */
 function drush_role_get_class($role_name = DRUPAL_ANONYMOUS_RID) {
-  return drush_get_class('Drush\Role\Role', func_get_args());
+  return drush_get_service('role', func_get_args());
 }
 
 /**
--- a/commands/sql/sql.drush.inc
+++ b/commands/sql/sql.drush.inc
@@ -594,7 +594,7 @@ function drush_sql_get_class($db_spec = NULL) {
  * @return Drush\Sql\SqlVersion
  */
 function drush_sql_get_version() {
-  return drush_get_class('Drush\Sql\Sql', array(), array(drush_drupal_major_version())) ?: NULL;
+  return drush_get_service('sql');
 }
 
 /**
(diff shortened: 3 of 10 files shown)


greg-1-anderson:
I think this looks pretty good as far as basic structure and concept is concerned.

One thing of note. The old "drush_get_class" method took a string parameter that contained a partial class name.  The new method takes a simple string identifier.  It would be good if there was a strong relationship between the object being requested, and the interface that the object is expected to return. For example, maybe the parameter should be an interface name rather than a simple identifier.

My other thought is that this is starting to look a lot like [Symfony Dependency Injection](http://symfony.com/doc/current/components/dependency_injection/index.html). Could we perhaps give each boot object a chance to provide a service container (yaml file and/or dynamic object) that knew how to inject the appropriate framework-specific dependencies for each subsystem (sql, pm, role, etc.) that needed them?

This is in conflict with my previous suggestion. I'm not sure it's a good idea. It would work for "core" subsystems easily enough, but what if you wanted to write a Drush extension that supported Drupal 7 and Drupal 8 implementation for a certain command it provided, and what if someone later wanted to submit a PR to that extension that added Backdrop support? What would the relationship between the boot object and the various extension framework support classes be?  

That same question also applies to the design in this PR.  For example, we currently have:
```
 class DrupalBoot6 extends DrupalBoot {
 
  function __construct() {
    // Add services supported by Drupal 6.
    $this->add_service('queue', 'Drush\Queue\Queue6');
    $this->add_service('role', 'Drush\Role\Role6');
    $this->add_service('sql', 'Drush\Sql\Sql6');
    $this->add_service('user', 'Drush\User\User6');
    $this->add_service('user_single', 'Drush\User\UserSingle6');

    parent::__construct();
}
```
I guess if a "foo" service wanted to provide Drupal 6 and Drupal 7 implementations, then it would provide classes Foo\Foo6 and Foo\Foo7, and would know how to get references to the DrupalBoot6 and DrupalBoot7 classes (is this currently available?), and would then call `add_service()` to each boot object.  This satisfies my question above, presuming we have a way for an extension to ask for a specific boot object, and determine if it is available.  For example:
```
function foo_drush_init() {
  $backdrop_boot = drush_magic_function_to_get_boot_object("BACKDROP");
  if ($backdrop_boot) {
    $backdrop_boot->add_service('foo', 'Foo\FooBackdrop');
  }
  // repeat for all supported frameworks
}
```
This is eminently doable, with appropriate accessors.

Could we also do the same thing using Symfony DI?  Could we do this even better with some similar mechanism?  e.g. Drupal 8 has a [plugin API](https://www.drupal.org/developing/api/8/plugins) that fills similar needs, but I don't know if we could adapt and/or re-use that code. (Could be interesting to maintain separability from D8 without conflicting with existing classes when using D8.)

greg-1-anderson:
c.f. #1884 - dependency injection for Drush.  Does not address engines.

weitzman:
Drush 9 now has dependency injection via a services container.

weitzman:
This system is gone from drush9

Issue #1754, opened 2015-11-07
More PsySH hotness


bobthecow:
You should check out @grota's [drush-psysh](https://github.com/grota/drush-psysh). It proxies drush commands inside the REPL, which is super rad :)

weitzman:
Psysh is in core drush already

bobthecow:
Yep. But I'm talking about incorporating features of drush-psysh that core doesn't yet have :)

weitzman:
I don't see what you refer to. Please be specific about what you think core
is missing.

bobthecow:
> It proxies drush commands inside the REPL, which is super rad.

bobthecow:
So, for example, you can run `cache-clear`, or `role-list`, or really any other drush command, while inside a REPL session.

weitzman:
thats kinda neat. @damiankloip is our de facto psysh maintainer so lets get his opinion.

damiankloip:
I did see that project, indeed has some awesome stuff in! We could easily bring this into drush core - the creation of the commands for the shell.

We would not want to use that project in it's entirety as it has it's own command file, and install command. We already have commands and dependencies.

So, in a nutshell, yes!

damiankloip:
OK, added a modified version of drush-psysh as that was not working with this version of drush, but a few fixes later and it's working pretty well! https://github.com/drush-ops/drush/pull/1768

PR #1768, opened 2015-11-11
Add a modified version of drush-psysh commands

damiankloip:
Modified version of the commands from https://github.com/grota/drush-psysh

This needed a bit of work to get working properly with drush, I guess just the newer version.

The help command needed a couple of fixes but mainly the DrushCommand execution needed some work to run correctly. This initial attempt is working pretty well though!

Based on discussion in https://github.com/drush-ops/drush/issues/1754
 
### Problems

- ~~After running a command, you need to use the psysh `exit` command twice to exit the shell session~~
- ~~Can we stop drush printing the output? I would prefer to pass it to the output object (@todo in the code)~~
- ~~Doesn't have support for site aliases yet (command aliases are fine), could be some sort of follow up?~~

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -1,5 +1,9 @@
 <?php
 
+use Drupal\Component\Assertion\Handle;
+use Drush\Psysh\DrushHelpCommand;
+use Drush\Psysh\DrushCommand;
+
 /**
  * Implements hook_drush_command().
  */
@@ -22,9 +26,27 @@ function drush_cli_core_cli() {
   if (drush_drupal_major_version() >= 8) {
     // Register the assertion handler so exceptions are thrown instead of errors
     // being triggered. This plays nicer with PsySH.
-    \Drupal\Component\Assertion\Handle::register();
+    Handle::register();
     $shell->setScopeVariables(['container' => \Drupal::getContainer()]);
   }
+
+  // Add drush commands in the shell.
+  $commands = [new DrushHelpCommand()];
+
+  foreach (drush_get_commands() as $name => $config) {
+    // Ignore some commands that don't make sense inside PsySH.
+    if (in_array($name, ['help', 'drush-psysh', 'php-eval', 'core-cli', 'php'])) {
+      continue;
+    }
+    // Don't add hidden commands or aliases.
+    if ($config['hidden'] || $name !== $config['command']) {
+      continue;
+    }
+
+    $commands[] = new DrushCommand($config);
+  }
+
+  $shell->addCommands($commands);
 
   // PsySH will never return control to us, but our shutdown handler will still
   // run after the user presses ^D.  Mark this command as completed to avoid a
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -3,6 +3,7 @@
 use Drupal\Component\Assertion\Handle;
 use Drush\Psysh\DrushHelpCommand;
 use Drush\Psysh\DrushCommand;
+use Drush\Psysh\Shell;
 
 /**
  * Implements hook_drush_command().
@@ -21,7 +22,7 @@ function cli_drush_command() {
  * Command callback.
  */
 function drush_cli_core_cli() {
-  $shell = new Psy\Shell();
+  $shell = new Shell();
 
   if (drush_drupal_major_version() >= 8) {
     // Register the assertion handler so exceptions are thrown instead of errors
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -33,16 +33,7 @@ function drush_cli_core_cli() {
   // Add drush commands in the shell.
   $commands = [new DrushHelpCommand()];
 
-  foreach (drush_get_commands() as $name => $config) {
-    // Ignore some commands that don't make sense inside PsySH.
-    if (in_array($name, ['help', 'drush-psysh', 'php-eval', 'core-cli', 'php'])) {
-      continue;
-    }
-    // Don't add hidden commands or aliases.
-    if ($config['hidden'] || $name !== $config['command']) {
-      continue;
-    }
-
+  foreach (_drush_core_cli_get_commands() as $name => $config) {
     $commands[] = new DrushCommand($config);
   }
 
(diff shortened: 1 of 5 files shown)


jonhattan:
> Can we stop drush printing the output? I would prefer to pass it to the output object (@todo in the code)

Yes, add `backend => TRUE` to the options passed to `drush_invoke_process()`, and pick the results with `drush_backend_get_result()`.

damiankloip:
Thanks @jonhattan! See https://github.com/damiankloip/drush/commit/611bf3574fb193147a235ef85000f3c5f373c794. I tested that this works great. Output is indeed only printed the once now.

damiankloip:
@jonhattan Also, do you know if there is a way I can reliable tell from the command definition that an option is meant to just be a flag? E.g. `--core` and `--no-core` vs `--package` from the pml command.

damiankloip:
Otherwise I am considering just adding a static list for flag options for now that I can check.

commands/core/cli.drush.inc
```diff
@@ -17,18 +22,47 @@ function cli_drush_command() {
...
+
   // PsySH will never return control to us, but our shutdown handler will still
   // run after the user presses ^D.  Mark this command as completed to avoid a
   // spurious error message.
   drush_set_context('DRUSH_EXECUTION_COMPLETED', TRUE);
   $shell->run();
 }
+
+/**
+ * Returns a filtered list of drush commands used for CLI commands.
```
weitzman: Please use capitalized Drush unless referring specifically to the finder script.

lib/Drush/Psysh/DrushCommand.php
```diff
@@ -0,0 +1,227 @@
+<?php
+/**
+ * @file
+ * Drush Command class.
+ *
+ * Original author: Justin Hileman
+ *
+ * DrushCommand is a PsySH proxy command which accepts a drush command config
```
weitzman: Same - Capitalize D for Drush

lib/Drush/Psysh/DrushCommand.php
```diff
@@ -0,0 +1,227 @@
+<?php
+/**
+ * @file
+ * Drush Command class.
+ *
+ * Original author: Justin Hileman
```
weitzman: We dont use author attribution

lib/Drush/Psysh/DrushCommand.php
```diff
@@ -0,0 +1,227 @@
...
+      foreach ($this->config['options'] as $name => $option) {
+        if (!is_array($option)) {
+          $option = ['description' => $option];
+        }
+
+        if (!empty($option['hidden'])) {
+          continue;
+        }
+
+        // @todo: Figure out if there's a way to detect InputOption::VALUE_NONE
```
damiankloip: Spoke to @weitzman about this one, there is indeed no reliable way to do this currently. Let's punt this for a follow up.

commands/core/cli.drush.inc
```diff
@@ -17,18 +22,52 @@ function cli_drush_command() {
...
 
   if (drush_drupal_major_version() >= 8) {
     // Register the assertion handler so exceptions are thrown instead of errors
     // being triggered. This plays nicer with PsySH.
-    \Drupal\Component\Assertion\Handle::register();
+    Handle::register();
     $shell->setScopeVariables(['container' => \Drupal::getContainer()]);
   }
 
+  // Add drush commands in the shell.
```
weitzman: Cap D

lib/Drush/Psysh/DrushCommand.php
```diff
@@ -0,0 +1,218 @@
...
+      // Add a newline after so the shell returns on a new line.
+      $output->writeln('');
+    }
+    else {
+      $output->page(drush_backend_get_result());
+    }
+  }
+
+  /**
+   * Extract drush command aliases from config array.
```
weitzman: Cap D

lib/Drush/Psysh/DrushCommand.php
```diff
@@ -0,0 +1,218 @@
...
+   *
+   * @return array
+   *   The command aliases.
+   */
+  protected function buildAliasesFromConfig() {
+    return !empty($this->config['aliases']) ? $this->config['aliases'] : [];
+  }
+
+  /**
+   * Build a command definition from drush command configuration array.
```
weitzman: Cap D

lib/Drush/Psysh/DrushCommand.php
```diff
@@ -0,0 +1,227 @@
...
+      $command = array_shift($args);
+    }
+    // Otherwise, default the alias to '@self' and use the first argument as the
+    // command.
+    else {
+      $alias = '@self';
+      $command = $first;
+    }
+
+    $return = drush_invoke_process($alias, $command, array_values($args), $input->getOptions(), ['backend' => TRUE]);
```
weitzman: 'backend' should be sent in 3rd param. Also, please try to `drush dl page_manager` or something like that twice. You should get prompted about overwrite the 2nd time. If no such prompt happens, try to set `array('interactive' => TRUE)` as $backend_options value (5th param).

damiankloip:
@weitzman added support for the global options now too

dawehner:
In general it works, some observations:

* Startup time isn't that fast anymore
* Autocompletion for the command names itself

weitzman:
I would think that the startup time would be unchanged all we did is register commands. The bigger delay is the Drupal bootstrap and we've been doing that from the beginning. Any thoughts on why we are slower?

damiankloip:
The start up time will be slightly longer, there are a lot of commands to register, all of which need every option registered too. We could improve this slightly by re-using the global commands - create once and merge into every definition.

TBH, I think being slightly slower to start the shell is inevitable. It's also not a speed critical operation at all :)

dawehner:
Yeah right, its not that horrible and well, we can still try to cache stuff
later. Just saying +1 for the general functionality!

damiankloip:
We can always make things better. Thanks @dawehner, you're my favourite.

@weitzman what do you think about trying to tackle more efficient shell initialization in a follow up? or would you rather try and get something here? IMO, this is not a deal breaker at all.

We will have to look into the command completion, I would think it should work, as psysh has a completion matcher for commands.

damiankloip:
The completion problem is an upstream bug with PsySH itself. I will open a PR to fix it there.

Issue #1769, opened 2015-11-11
[7.x] Modules in included makefiles do not override the "parent" makefile


kienanstewart:
In a case similar to [#1268](https://github.com/drush-ops/drush/issues/1268), specifications to modules included in later makefiles do not override the ones from previous makefiles.

Makefile1

>  projects[webform][version] = 3
>  include Makefile2

Makefile2

> projects[webform][version] = 4

The makefile generation will end up with webform = 3.x

I would expect later includes to override previous entries.

ergonlogic:
Actually, no, this works as expected. The makefile that contains includes should be thought of as a wrapper around the includes. The order they're defined in the makefile is irrelevant.

Issue #1268, opened 2015-03-18
Makefiles using multiple includes don't end up with all elements parsed


kienanstewart:
I believe a change introduced in 0f103bee607a411e8a9c58042477e31a40752cfd causes makefiles with multiple includes to have output that doesn't include all the projects (or any keys) when the default keying is used.

The following is a simplified example that I've tested; for a more complex example see drupal.org/project/kplatforms

test.make

> api = 2
> core = 7.x
> projects[drupal][version] = 7.35
> includes[] = a.make
> includes[] = b.make

b.make

> api = 2
> core = 7.x
> projects[] = addressfield
> projects[] = views
> projects[] = search_api

c.make

> api = 2
> core = 7.x
> projects[] = openlayers
> projects[] = geophp
> projects[] = features

Result w/ revision e28a45f996c9dc89f84af6863de03a44b74a2f94

> ; This file was auto-generated by drush make
> core = 7.x
> api = 2
> ; Core
> projects[drupal][version] = "7.35"
> projects[drupal][type] = "core"
> ; Modules
> projects[features][version] = "2.4"
> projects[geophp][version] = "1.7"
> projects[openlayers][version] = "3.0-beta1"
> projects[addressfield][version] = "1.0"
> projects[views][version] = "3.10"
> projects[search_api][version] = "1.14"

Results with revision 0f103bee607a411e8a9c58042477e31a40752cfd

> ; This file was auto-generated by drush make
> core = 7.x
> api = 2
> ; Core
> projects[drupal][version] = "7.35"
> projects[drupal][type] = "core"
> ; Modules
> projects[features][version] = "2.4"
> projects[geophp][version] = "1.7"
> projects[openlayers][version] = "3.0-beta1"

In commit e28a45f996c9dc89f84af6863de03a44b74a2f94 (just prior), the output is as expected.

ergonlogic:
I'll take a look at this later tonight, since it was my commit that appears to have introduced this bug.

weitzman:
ping @ergonlogic

ergonlogic:
Sorry for the delay. Looking into this now.

PR #1770, opened 2015-11-11
Issue #1769: Reversed order of array replacement when merging include…

kienanstewart:
Reversed order of array replacement when merging included makefiles - later includes override previous values.

See issues #1769 and #1268


weitzman:
Failed tests

ergonlogic:
See #1769.

Issue #1667, opened 2015-10-05
array_flip(): Can only flip STRING and INTEGER values! output.inc:184


bradjones1:
```
bradj@home:~/www/brjllc$ ddrush uli
array_flip(): Can only flip STRING and INTEGER values! output.inc:184
```
Call trace from Xdebug:
![screenshot from 2015-10-05 15 51 52](https://cloud.githubusercontent.com/assets/981966/10294689/26db7bca-6b79-11e5-8b04-a95dee2c12f2.png)

Documenting for the moment since I don't see a related ticket on search... will try to debug later.

weitzman:
we'll need steps to reproduce.

which xdebug client is that?

jonhattan:
I'm seeing this when `drush @site uli`. 

```
Array
(
    [!phase] => 5
    [!max] => 
)
array_flip(): Can only flip STRING and INTEGER values! output.inc:185     [warning]
```

It comes from `drush_log(dt("Find command files for phase !phase (max=!max)", array('!phase' => $phase, '!max' => $phase_max)), 'debug');` when `$phase_max` is FALSE.

jonhattan:
Fixed in master and 7.x

FGM:
Same error on master Drush nad master D8 (just pulled both right now) when creating a command with required-arguments truthy, and not passing the required argument. The error appears to occur because the $args passed to find_legacy_dt_args() is like : ['@missing' => ..., '@required' => ..., '@command' => ...] while that function expected !-prefixed strings, not @-prefixed ones.

weitzman:
Just tried on master - works fine for me.

FGM:
Indeed, with today's HEAD on drush and core, the symptom changes in that case : it is now `Illegal offset type drush.inc:1584` which is a different issue, happening because in `drush_set_error()`, `$error` is not a string but a TranslatableMarkup instance, causing `$error_log[$error][] = $message;` to emit this warning, which can be removed by wrapping $error in double quotes to trigger its `__toString()`.

weitzman:
The doxygen for drush_set_error() needs clarity on what types $error can be. Mind to send a PR? Thanks for retesting.

FGM:
I can do this, but AIUI the error is still in Drush code, which calls drush_set_error() incorrectly. Here is a drush plugin to reproduce the problem:

```php
function bug_drush_command() {
  return [
    'foo' => [
      'arguments' => ['arg' => 'bar'],
[...]

PR #1771, opened 2015-11-12
Allow drush_set_error() to handle TranslatableMarkup.

FGM:
Issue #1667: Allows passing the result of dt() on Drupal 8 to drush_set_error().

### DIFF
--- a/includes/drush.inc
+++ b/includes/drush.inc
@@ -1551,19 +1551,17 @@ function drush_format_size($size) {
 /**
  * Set an error code for the error handling system.
  *
- * @param error
+ * @param \Drupal\Component\Render\MarkupInterface|string $error
  *   A text string identifying the type of error.
- *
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks.

The Doxygen states that $error must be a string. Can we push this casting to the caller? If not, lets change Doxygen.

FGM:
Since we want to support Drupal 8, I think it will be much more convenient to allow any instance of TranslatableMarkup, which is likely to appear often now that most strings are no longer plain PHP strings. Probably using a MarkupInterface doc hint to specify an interface rather than a class. This will fix all such cases in one change instead of having to change all callers in core and contrib. 

Which solution do you prefer ?

weitzman:
Yeah, I'm OK with accepting MarkupInterface. Thanks.

PR #1774, opened 2015-11-15
Fixed some relative links in install.md

hansfn:


hansfn:
Try the link "alternative ways to install" on http://docs.drush.org/en/master/install/ to reproduce the issue.

weitzman:
I would rather these be relative links but those are proving very hard to make work on docs.drush.org and on Github. Thanks.

PR #1775, opened 2015-11-16
Filter command aliases for PHP keywords

damiankloip:
I was testing out something for one of the PsySH PRs and we have a problem with aliases. There is a high chance one of them could conflict with a PHP keyword. E.g. `if` is an alias of `image-flush` this means we can't eval any php code starting with if :)

Maybe we should check commands too to be safe?

Needs a bit of tidy, just trying to post ASAP before my laptop dies (on 2% atm).

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -70,14 +70,98 @@ function drush_cli_core_cli() {
 function _drush_core_cli_get_commands() {
   $commands = drush_get_commands();
   $ignored_commands = ['help', 'drush-psysh', 'php-eval', 'core-cli', 'php'];
+  $php_keywords = _drush_core_cli_get_php_keywords();
 
   foreach ($commands as $name => $config) {
-    // Ignore some commands that don't make sense inside PsySH, are hidden, or
-    // are aliases.
-    if (in_array($name, $ignored_commands) || !empty($config['hidden']) || ($name !== $config['command'])) {
+    // Ignore some commands that don't make sense inside PsySH, are PHP keywords
+    // are hidden, or are aliases.
+    if (in_array($name, $ignored_commands) || in_array($name, $php_keywords) || !empty($config['hidden']) || ($name !== $config['command'])) {
       unset($commands[$name]);
+    }
+    else {
+      // Make sure the command aliases don't contain any PHP keywords.
+      if (!empty($config['aliases'])) {
+        $commands[$name]['aliases'] = array_diff($commands[$name]['aliases'], $php_keywords);
+      }
     }
   }
 
   return $commands;
+}
+
+/**
+ * Returns a list of PHP keywords.
+ *
+ * This will act as a blacklist for command and alias names.
+ *
+ * @return array
+ */
+function _drush_core_cli_get_php_keywords() {
+  return [
+    '__halt_compiler',
+    'abstract',
+    'and',
+    'array',
+    'as',
+    'break',
+    'callable',
+    'case',
+    'catch',
+    'class',
+    'clone',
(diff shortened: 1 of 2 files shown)


lib/Drush/Psysh/DrushCommand.php
```diff
@@ -118,7 +125,11 @@ protected function execute(InputInterface $input, OutputInterface $output) {
    *   The command aliases.
    */
   protected function buildAliasesFromConfig() {
-    return !empty($this->config['aliases']) ? $this->config['aliases'] : [];
+    $aliases = !empty($this->config['aliases']) ? $this->config['aliases'] : [];
+
+    return array_filter($aliases, function($alias){
```
weitzman: Could array_diff() simplify this a little?

damiankloip: Totally. When I plug my laptop in and have some power i'll carry on with tidying this up.

commands/core/cli.drush.inc
```diff
@@ -70,14 +70,98 @@ function drush_cli_core_cli() {
...
   $ignored_commands = ['help', 'drush-psysh', 'php-eval', 'core-cli', 'php'];
+  $php_keywords = _drush_core_cli_get_php_keywords();
 
   foreach ($commands as $name => $config) {
-    // Ignore some commands that don't make sense inside PsySH, are hidden, or
-    // are aliases.
-    if (in_array($name, $ignored_commands) || !empty($config['hidden']) || ($name !== $config['command'])) {
+    // Ignore some commands that don't make sense inside PsySH, are PHP keywords
+    // are hidden, or are aliases.
+    if (in_array($name, $ignored_commands) || in_array($name, $php_keywords) || !empty($config['hidden']) || ($name !== $config['command'])) {
```
damiankloip: @weitzman this just removes the command if it's conflicts, we could potentially prefix it with drush or something? This is very edge-casey, not sure if it's worth doing that or not?

commands/core/cli.drush.inc
```diff
@@ -70,14 +70,98 @@ function drush_cli_core_cli() {
...
       unset($commands[$name]);
     }
+    else {
+      // Make sure the command aliases don't contain any PHP keywords.
+      if (!empty($config['aliases'])) {
+        $commands[$name]['aliases'] = array_diff($commands[$name]['aliases'], $php_keywords);
```
damiankloip: filters out any aliases like `if`.

damiankloip:
I prefer this approach, filtering the data before creating the commands, rather than having this logic in the actual command class.

weitzman:
Looks good.

commands/core/cli.drush.inc
dawehner: which commands are that?

commands/core/cli.drush.inc
dawehner: seems reasonable, but should afaik be done ideally by pysh itself

commands/core/cli.drush.inc
damiankloip: A good example is the alias for image-flush, if. Some of those keywords could easily clash with an alias or custom command.

commands/core/cli.drush.inc
damiankloip: Yes, we could try to get this upstream. Let's keep it here for now though, as we can have this working before the 8.0.0 release, if we wait for upstream, we will likely not.

dawehner:
for now +1

damiankloip:
Thanks both, merging as-is so we have this fix in before 8.0.0.

PR #1782, opened 2015-11-18
Add VarDumper casters for Entity and Config classes to the shell

damiankloip:
we can add a few casters to make inspecting some objects a bit nicer. As almost everything is now a protected property, you would have to `dump -a $object` to see everything. So save doing that all the time we can add some casters, these will show properties of entities, config entities, config objects (for now).

So instead of not much:

1____forks_drush_drush_____sites_d8__php_

We can have properties shown by default like this:

1____forks_drush_drush_____sites_d8__php_

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -31,13 +31,17 @@ function cli_drush_command() {
  * Command callback.
  */
 function drush_cli_core_cli() {
-  $shell = new Shell();
+  $configuration = new \Psy\Configuration();
+  $shell = new Shell($configuration);
 
   if (drush_drupal_major_version() >= 8) {
     // Register the assertion handler so exceptions are thrown instead of errors
     // being triggered. This plays nicer with PsySH.
     Handle::register();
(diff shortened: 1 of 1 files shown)


commands/core/cli.drush.inc
```diff
@@ -90,6 +94,64 @@ function _drush_core_cli_get_commands() {
...
+ * Returns a mapped array of casters for use in the shell.
+ *
+ * These are Symfony VarDumper casters.
+ * See http://symfony.com/doc/current/components/var_dumper/advanced.html#casters
+ * for more information.
+ *
+ * @return array.
+ *   An array of caster callbacks keyed by class or interface.
+ */
+function _drush_core_cli_get_casters() {
```
dawehner: Ideally we would have added typed data

damiankloip: Follow ups are more than welcome @dawehner :)

damiankloip: OR maybe the `Drupal\Core\Field\FieldItemInterface` caster could just use `Drupal\Core\TypedData\TypedDataInterface` instead?

PR #1785, opened 2015-11-19
Run terminate before the shell runs to avoid database errors when the core-cli command finishes

damiankloip:
See comment for explanation. This is not ideal but it does work, not sure there is a better way at the moment to be honest. This will actually run terminate() twice, but because we already have the subscriber instances created before the shell is run, when the shell exits they still work.

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -75,14 +75,14 @@ function drush_cli_core_cli() {
     $bootstrap->terminate();
   }
 
-  $shell->run();
-
   // To fix the above problem in Drupal 7, the connection can be closed manually.
-  // This will make sure a new connection is created again if/when needed in
-  // registered shutdown functions.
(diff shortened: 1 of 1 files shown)


dawehner:
Nice!

```
.... SUCCESS!
~/w/d8 (2616816) $ drush php
Psy Shell v0.5.2 (PHP 5.5.28 — cli) by Justin Hileman
>>> exit
Exit:  Goodbye.
~/w/d8 (2616816) $
```

dawehner:
RTBC

damiankloip:
This fixes it for Drupal 8 only. I can still see this in D7 when stuff uses the database in a shutdown function.

damiankloip:
OK, going to merge this now. For me it fixes the errors when exiting the shell and doesn't cause any side effects for D7 or D8 usage. If this does have issues later on, we will only really know from more widespread testing :)

PR #1789, opened 2015-11-19
Bring pm-projectinfo in drush

### LINKED ISSUE #1757: Add pm-projectinfo into Drush core
I would like to add [pm-projectinfo from drush extras](http://cgit.drupalcode.org/drush_extras/tree/pm_extras.drush.inc) into Drush core.

This was discussed before, and pm-projectinfo was relegated to Drush extras due to lack of demand. However, I'd like to be able to rely on it being available in Drush core in the composer-generate command. This is more convenient in the remote case.

jonhattan:
Resolves #1757

### DIFF
--- a/commands/pm/pm.drush.inc
+++ b/commands/pm/pm.drush.inc
@@ -248,6 +248,39 @@ function pm_drush_command() {
       'output-data-type' => 'format-table',
     ),
   );
+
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
Thanks!

Issue #456, opened 2014-02-12
Ability to use variables in .make files


kenorb:
It would be good to use e.g. system variables in .make files, so make process could be more dynamic.
Or any other ways to pass some variables into .make files.

Example scenario:
I've got repo on drupal.org with mirror on GitHub and few other forks. Code in repos are not always in sync, so sometimes you've to use different repo urls (projects[foo][download][url]). You could use separate .make files for each one, but as the project become more complex, it'll require more and more separate .make files to meet the build criteria.

ergonlogic:
How would you envision setting such variables, provide defaults and so forth? I think an easier and more maintainable solution might be to provide an option to load one or more include makefiles. This would allow you to override entries in the main makefile.

So this might look like:

    drush make example.make --includes=github.make,sandboxes.make,devel.make

While `example.make` might specify core and a bunch of contrib, including some of your own modules hosted on d.o. `github.make` could specify download urls for your projects, `sandboxes.make`could specify some sandboxes to use instead of the released projects, and `devel.make`could add devel and other projects that you might only want to run in dev environments.

ergonlogic:
I believe the solution in #1790 would provide a mechanism to accomplish something along these lines. Feel free to open another feature request if there are requirements that wouldn't be satisfied by that solution.

PR #1790, opened 2015-11-19
Fixes #1151: Add option to include makefiles on the fly.

### LINKED ISSUE #1151: Option to include makefiles on the fly
It'd be nice to be able to add includes on the command-line when calling make. Specifically, this'd be useful in development environments. It'd avoid having to keep an extra stub makefile for development. I envision this just tacking on the included makefiles like a regular `includes[]` statement, but at the very end of the makefile being built.

ergonlogic:
Pretty simple implementation of #1151.

### DIFF
--- a/commands/make/make.drush.inc
+++ b/commands/make/make.drush.inc
@@ -119,6 +119,7 @@ function make_drush_command() {
         'description' => 'Return project info for use by core-quick-drupal.',
         'hidden' => TRUE,
       ),
+      'includes' => 'A list of makefiles to include at build-time.',
     ),
     'engines' => array('release_info'),
     'topics' => array('docs-make', 'docs-make-example'),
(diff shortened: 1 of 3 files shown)


ergonlogic:
@jhedstrom Thoughts?

Pretty straight-forward, imo. Also, a nice way to accomplish #456.

ergonlogic:
Actually, after reviewing #1769 and #1770 again, this does *not* help with #456. We could look at using a similar approach with an `--overrides` option to merge over values in the makefile itself.

jhedstrom:
This makes sense to me. Let's add an example to the docs though.

ergonlogic:
Added docs for `--includes` and `--overrides`: https://github.com/drush-ops/drush/blob/dev/1151/docs/make.md#includes

jhedstrom:
:+1:

PR #1794, opened 2015-11-20
Allow site and Drupal version specific CLI history

damiankloip:
The shell history is currently using the PsySH global history. So we get regular psysh and any Drupal sites/versions you are using in the same history. This gets slightly annoying when switching between them.

Let's set the history file path ourselves, and allow this to be per site or per Drupal version! The default will be per site, with an option, `version-history` which can be used to use drupal version specific history instead.

I added an example in the `example.drushrc.php` file as this is a good one for command specific defaults.

It also prints out the history file path when verbose is enabled.

### Directory structure
```
★ tree ~/.drush/cli/
$HOME/.drush/cli/
├── drupal-7
├── drupal-8
├── drupal-site-7a914924f49f52862af4fdc8b1813f52
├── drupal-site-7df5e3082987556585d903657b4cdeec
└── drupal-site-testalias.local
```

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -15,6 +15,9 @@ function cli_drush_command() {
     'aliases' => array('php'),
     'bootstrap' => DRUSH_BOOTSTRAP_MAX,
     'topics' => array('docs-repl'),
+    'options' => array(
+      'version-history' => 'Use command history based on Drupal version (Default is per site).',
+    ),
   );
   $items['docs-repl'] = array(
     'description' => 'repl.md',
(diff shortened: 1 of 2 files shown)


commands/core/cli.drush.inc
```diff
@@ -133,6 +141,61 @@ function _drush_core_cli_get_casters() {
...
+ * This can either be site specific (default) or Drupal version specific.
+ *
+ * @return string.
+ */
+function drush_history_path_cli() {
+  $drupal_major_version = drush_drupal_major_version();
+  $drush_core_directory = _drush_core_directory('root', 'path', TRUE);
+  $cli_directory = drush_directory_cli();
+
+  $file_name = drush_get_option('version-history', FALSE) ? "drupal-$drupal_major_version" : 'drupal-site-' . md5($drush_core_directory);
```
dawehner: I wonder whether we could use a site alias, if available?

damiankloip:
@dawehner there you go, alias support added!

dawehner:
Nice! RTBC

weitzman:
The drush_clear_error() looks heavy handed to me. Can we avoid that?

Can we use DRUPAL_ROOT instead of _drush_core_directory()?

damiankloip:
@weitzman This was just mostly copied from `drush_directory_cache`. See the comment there:

```php
        // If the base directory is writable, but the cache directory
        // is not, then we will get an error. The error will be displayed,
        // but we will still call drush_clear_error so that we can go
        // on and try the next location to see if we can find a cache
        // directory somewhere.
```

If you think it's OK to remove, let's remove it.

damiankloip:
I guess using DRUPAL_ROOT is fine now we have the alias code, as DRUPAL_ROOT will just be a fallback when there is no alias, which is totally fine.

damiankloip:
Changed to use `DRUPAL_ROOT` and remove `drush_clear_error`.

PR #1795, opened 2015-11-20
Add a caster for the container

damiankloip:
Quick attempt for a container caster. This is not ideal, as we have to get every service... Better than nothing, and probably better than just showing the service IDs.

If you just want the service IDs you can run the devel `devel-container-services` (dcs) command.


weitzman:
Feel free to merge when you are satisfied.

PR #1802, opened 2015-11-23
Don't clear cache when no database updates, fixes #755

### LINKED ISSUE #755: Don't clear cache when no database updates
Hello,
when `drush updb` is run but not database updates are there it will print "No database updates required". Ok, but why it clears caches in this case? I think caches should be run only if some database updates were run.

helmo:


weitzman:
updb now has --cache-clear=0 if you want to manually control the cache step. we also have updatedb-status. 

neither of those does what this PR does. someone needs to look at this closely. i probably wont do that myself given the available workarounds.

helmo:
Both workarounds are just that ... :(

It just bugs me that caches are cleared when nothing has been done.
If I set `--cache-clear=0` then I have to check myself if caches need to be flushed... and with the updatedb-status command I'm adding extra overhead and complexity.

ergonlogic:
This looks pretty reasonable to me. It also matches Drupal's behaviour more closely. Drupal doesn't clear caches if there weren't any updates to apply.

PR #1805, opened 2015-11-23
Fix drush tests

webflo:


webflo:
Closed in favour of #1806

PR #1808, opened 2015-11-25
Use Psr\Log for all logging.

greg-1-anderson:
I/O is the number one obstacle we face to (a) break out some of our functionality into libraries that might be usable independently of Drush, and (b) adopt external libraries that may perform some of the operations we require, so that we can reduce our maintenance requirements and become more standard.

This is just the first pass; need to move more code into the actual Logger, etc.


greg-1-anderson:
[Symfony\Components\Console\Output\ConsoleOutput](http://api.symfony.com/2.0/Symfony/Component/Console/Output/ConsoleOutput.html) is a little frustrating.

You can write to stderr from symfony using [StreamOutput](http://api.symfony.com/2.0/Symfony/Component/Console/Output/StreamOutput.html); however, there does not seem to be any standard concept in Symfony for writing log messages to stderr, and normal output to stdout.

Robo, DrupalConsole et. al. use the Symfony console output class, so this is a big shortcoming in terms of standardization across cli tools.

I am also frustrated that there is no easy mapping between Psr\Log LogLevels and Symfony output levels (which are just “nomal”, “verbose” and “quiet”). Seems like this should already be solved in a standard way.

Still considering using ConsoleOutput, to take advantage of their formatting, rather than continuing to use our own.

lib/Drush/Log/DrushLog.php
```diff
@@ -3,6 +3,8 @@
 /**
  * @file
  * Contains \Drush\Log\DrushLog.
+ *
+ * This
```
jhedstrom: Unfinished comment here?

greg-1-anderson:
@jmolivas: Trying to standardize I/O in Drush for better interoperability of components. Would be good if DrupalConsole could do likewise. Please review.

lib/Drush/Log/DrushLog.php
greg-1-anderson: Lots still to come here!  Lots!  Not just comments.  :)

greg-1-anderson:
Tests are all green again; I think this PR is at a good spot to be reviewed and merged, as at this point we can now:

(a) Start adding \Psr\Log\LoggerInterface to the constructor of new classes that we may wish to break out into external libraries (e.g. backend invoke and site aliases).

(b) Fetch a \Psr\Log\LoggerInterface via `$logger = drush_get_context('DRUSH_LOG_CALLBACK');`. This logger can then be passed in to the constructor of any external library that we may wish to use that logs to a \Psr\Log\LoggerInterface. Any log messages emitted by that library will be handled correctly by Drush (works with backend invoke, etc.)

There is more we can do to improve and separate logging, but I think we can continue in follow-on issues. This PR is kind of a merge conflict waiting to happen. :)

commands/core/browse.drush.inc
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use \Drush\Log\LogLevel;
```
weitzman: Unused (pun intended).

commands/core/config.drush.inc
```diff
@@ -5,12 +5,13 @@
...
 
+use \Drush\Log\LogLevel;
 use Drupal\Core\Config\StorageComparer;
 use Drupal\Core\Config\ConfigImporter;
 use Drupal\Core\Config\ConfigException;
 use Drupal\Core\Config\FileStorage;
-use Drush\Config\StorageWrapper;
-use Drush\Config\CoreExtensionFilter;
+use \Drush\Config\StorageWrapper;
+use \Drush\Config\CoreExtensionFilter;
```
weitzman: whats the meaning of the leading slash?

weitzman:
Looks good to me. Looks OK for 8.1 since there is no API change?

greg-1-anderson:
I think this is fine for 8.1, as the code will continue to work if folks are still using 'warning' instead of LogLevel::WARNING.

I don't think that the leading slash that I put into the new 'use' statements (which forces a global scope) are necessary.  I just added another commit that takes them out; we'll see if the tests come out green. (Oh, I need to fix the merge conflicts first -- one sec...)

greg-1-anderson:
Okay, removing the extraneous leading slash in the 'use' statements worked fine (all tests passed). I made another commit to remove the 'use' statements in files where they were unused; these tests should also pass (*).

Shall I go ahead and squash-commit these to master once everything comes back green?

(*) Presuming I did not remove too much! :)

weitzman:
Yes ok to merge to master. 

So should drush stop using types preflight, bootstrap, etc? Do that in a follow up ? I don't think that's a significant Api change.

greg-1-anderson:
It is okay for an app to extend the number of log levels if it has extra that it wants to use. I eliminated the 'rollback' and 'sitealias' types here, because I think they are not significant. Any library-type / potentially-reusable code should log directly to a LoggerInterface, and should only log using the standard log levels. Preflight and bootstrap are in very Drush-specific parts of the codebase, so I don't think there is any harm n leaving them as they are. I think we can wait for Drush 9 to decide to what extent we do or do not want to standardize our log levels.

greg-1-anderson:
Squashed and merged -- thanks!

greg-1-anderson:
What do you think about backporting this to Drush 7.x? Might make it less annoying to backport other bugfixes, if anyone is planning on doing that.

I'll roll a 7.x PR if this is desired.

weitzman:
I think it would be OK to back port.
>
>

Issue #1817, opened 2015-11-30
Add commands for working with interface translations (locale)


tstoeckler:
Drush currently does not have any way to work with translations. It is often an important part of the deployment process to update translations, i.e. just like doing a `drush config-import` as part of the deployment to have the latest config on the site you also want to have the latest translations.

tstoeckler:
So this initial version basically replicates the _Available translation updates_ screen in Drupal 8 core. Interface Translation module sadly does not provide a proper API, so I basically re-implemented what happens in the respective form submission handlers. See the respective `@see` entries for the actual code I used. The `locale-check` and `locale-update` commands do work, however this does not yet contain commands for exporting and importing single PO files, which would also be useful. That feature is also available in Drupal 7 so we could provide that there, as well.

The translation update functionality is handled by the contrib [Localization update](https://www.drupal.org/project/l10n_update) module, so probably is not eligible for Drush core.

tstoeckler:
Here's a screenshot:
![drush-locale](https://cloud.githubusercontent.com/assets/3929504/11466735/79de5630-9743-11e5-91c9-ceddeeec8050.png)

Also pinging @penyaskito because we talked about this yesterday...

weitzman:
@tstoeckler will merge the PR (#1818) when he has a chance. no need to keep this issue open.

PR #1818, opened 2015-11-30
Add locale-check and locale-update commands (#1817)

tstoeckler:
See #1817


damiankloip:
This is looking pretty good. The fact that there is not something at the API level in D8 is kind of crazy...

tstoeckler:
Thanks!

Yes, I agree completely. It's totally crazy. But IMHO all the more reason to have this in Drush, because you can't just do
```bash
drush ev "\Drupal::service('locale.updater')->update();"
```
or similar as a temporary shiv, because something like that simply doesn't exist.

weitzman:
If these are for D8+, we should rename the file with an 8 as suffix like the views8.drush.inc file.

Is there an issue to move this logic to an API function. I'd like for a Drupal patch to exist so I have a hope of not maintaining this forever.

tstoeckler:
Opened https://www.drupal.org/node/2631584 on Drupal.org for that, added @todo's and renamed the file.

weitzman:
Looks good. Feel free to merge this.

tstoeckler:
Totally failed at Git as I just realized. Now actually added @todo's et al.

Thanks for the endorsement, it will be great to have this for multilingual workflows. Looking forward to it

tstoeckler:
From #1817:
> @tstoeckler will merge the PR (#1818) when he has a chance. no need to keep this issue open.
I don't have write access...

weitzman:
Woops - my bad. I thought I had done that. Now invited.

tstoeckler:
Yay, thanks!

PR #1823, opened 2015-11-30
Remove Drupal 8 specific code, since D8 is not supported in Drush 7.

greg-1-anderson:
Specific checks for Drupal 8 embedded in files shared between Drush 7 and Drush 8 remain unchanged, to facilitate backporting; files that are never used in Drush 7 are removed, to reduce confusion caused when folks try to use Drush 7 on Drupal 8 and do not believe the 'not supported' message because of the large number of Drupal 8 support files present.


greg-1-anderson:
Tests pass here; does this look like a good idea, @weitzman?

weitzman:
"do not believe the 'not supported' message"

Even if they don't believe it, they will get an error message upon any
usage with Drupal 8. I think we can skip this PR to be honest. It could
cause the occasional hassle when backporting a fix so we I dont think its
worth it.

PR #1825, opened 2015-11-30
Convert DrushLog class into an AbstractLogger adapter

greg-1-anderson:
Call through to the core Drush logger.

Once we are using a service container, we can set up these relationships with a yml loader and get rid of the global.

Let's see if these tests pass, though.

### DIFF
--- a/lib/Drush/Boot/DrupalBoot8.php
+++ b/lib/Drush/Boot/DrupalBoot8.php
@@ -70,7 +70,8 @@ class DrupalBoot8 extends DrupalBoot {
     // channel.
     $container = \Drupal::getContainer();
     $parser = $container->get('logger.log_message_parser');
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Why is Travis running both "PR" and "push" tests for every PR?  These seem to be equivalent -- that is, it looks like the same tests are being run twice.

weitzman:
Yeah, I noticed the double testing recently. Travis seems to think it is not a bug - https://github.com/travis-ci/travis-ci/issues/3241 and https://github.com/travis-ci/travis-ci/issues/1147

greg-1-anderson:
Tests are green. This PR brings us one step closer to using a service container to manage our logger and other dependencies.

weitzman:
Looks good.

PR #1828, opened 2015-12-01
Fixes #1827: Download Console_Table from GitHub rather than Pear.

### LINKED ISSUE #1827: Drush version 6.* not working due offline pear.php.net
Hi,
Just to let you know, Drush version 6.* isn't working at this moment because pear.php.net is offline.
Drush tries (internally) to download a dependency on Console_Table directly from http://pear.php.net/ (wich is offline at the moment).
See https://github.com/drush-ops/drush/blob/6.x/includes/environment.inc#L150

Hopefully this helps other people to fix failing builds on Travis CI for example.

greg-1-anderson:
This will make the installation of the Drush 6.x branch dependent on GitHub rather than Pear.  This will hopefully be an improvement.

Note that the oldest version of the Console Table available on GitHub is 1.1.5, so I used that. Previously, we were using 1.1.3.

### DIFF
--- a/commands/pm/release_info/updatexml.inc
+++ b/commands/pm/release_info/updatexml.inc
@@ -211,7 +211,7 @@ function release_info_print_releasenotes($requests, $print_status = TRUE, $tmpfi
       $xml = simplexml_import_dom($dom);
 
       // Extract last update time and the notes.
-      $last_updated = $xml->xpath('//div[@class="last-updated"]');
+      $last_updated = $xml->xpath('//div[contains(@class,"views-field-changed")]');
       $last_updated = $last_updated[0]->asXML();
       $notes = $xml->xpath('//div[contains(@class,"field-name-body")]');
       $notes = $notes[0]->asXML();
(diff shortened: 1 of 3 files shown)


jhedstrom:
Makes sense to me. I ran into this issue earlier today (my fix was actually just to upgrade to 8.0.1, but this is good for folks that can't do that for whatever reason).

greg-1-anderson:
This is going to need the pm-releasenotes commit also.

greg-1-anderson:
Merging, so that the 6.x dev branch, at least, will be usable.

dmromanov:
Are you planning to revert this line back to download.pear.net when github will be down?

greg-1-anderson:
No. Pear is no longer used in the more recent versions of Drush; this is a permanent change.

weitzman:
[Drush 6.7.0 is now released, without PEAR depedency](https://github.com/drush-ops/drush/releases/tag/6.7.0).

@dmromanov - yes, we are trading dependencies here. modern software is built from a collection of disparate dependencies. those dependencies are usually fetched at install/update time. do you think drush should be doing something different? i find your comment unhelpful.

Issue #1835, opened 2015-12-03
Add more defensive code to .bashrc


drupol:
If for some reason the file .bashrc is modified by Drush at a certain point by adding: 

````
# Include Drush bash customizations
. /home/USER/.drush/drush.bashrc
````

If you remove the drush bin, then, when the user will login, it will be automatically logged out. It's frustrating.

I think we should add a check for the existence of the drush binary before including the drush.bashrc file.

philsward:
Ran into the same issue.

drush init is adding the include lines at the very top of the .bashrc file.  "If" a person adds all of their exports into the .bashrc file, the exports are usually much lower in the file.  So, if the exports haven't run to generate the path locations, then the scripts for drush are trying run before it even knows where to look for drush.  Theoretically, if drush is installed in /usr/local/drush/bin, it should be available before the .bashrc profile loads.  However, if you're on a shared server (like me), you can't add drush to /usr/local/bin/drush

[/usr/local/bin]# touch test.txt
touch: cannot touch `test.txt': Read-only file system

In this case, you have to first tell .bashrc which paths to look in, then you can tell it to look in .drush.

Long story short, drush init needs to add to the end of the file, not the beginning.

weitzman:
It already runs at the end of the file in 8.0.1

drupol:
Wow.

Hello @weitzman,

This thread is about two problems.
The first is, if drush is not found, it automatically logs out the user uppon login, it's very dangerous and can lock you out of your own machine.
The second, reported by @philsward, is for adding stuff at the end of the .bashrc file.

Does the code in 8.0.x fixes those two or only one ?

Thanks.

weitzman:
8.0.1 fixes 2nd bug. The first issue awaits a fix by anyone who is interested. Note that a missing source statement should never log a user out of their system. Its a simple bash error and the shell continues to initialize AFAIK.

tornewuff:
Actually, 8.0.1 still adds the extra code to the beginning of .bashrc, not the end. The code in master adds it to the end, but the change isn't there in the 8.0.1 version of core-init.

PR #1836, opened 2015-12-03
Safer .bashrc additions.

drupol:
Check if the drush executable exist before making those changes.

This might fix issue #1835


weitzman:
Fixed in a different way.

PR #1838, opened 2015-12-03
Fix for PHP 5.3.3.

AndreBaumeier:
fix for "Exception 'ReflectionException' with message 'Class Drush\Sql\Sql7 does not have a constructor, so you cannot pass any constructor arguments"

### DIFF
--- a/includes/drush.inc
+++ b/includes/drush.inc
@@ -114,7 +114,7 @@ function drush_get_class($class_name, $constructor_args = array(), $variations =
       $variant_class_name = $class_name . implode('', array_slice($variations, 0, $i));
       if (class_exists($variant_class_name)) {
         $reflectionClass = new ReflectionClass($variant_class_name);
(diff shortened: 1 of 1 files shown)


weitzman:
#544

greg-1-anderson:
Some folks are stuck on php 5.3.3 for their LTS releases (Centos, Ubuntu).  Let's at least let the tests run; seems like it would be a good idea to get this in to Drush 7.  We'll probably release Drush 9 and make Drush 7 unsupported pretty soon; this should be available as a fallback.

greg-1-anderson:
All of the tests passed; it shouldn't hurt anything to put this in, and will help folks on old LTS be a little more up-to-date.

JorisVanEijden:
This was never merged into the 7.x branch :-1:

greg-1-anderson:
Opened #2136.

PR #1840, opened 2015-12-04
Disable zend assertions in the shell for PHP7

damiankloip:
The warning about serializing the container is back when exiting the shell when using PHP7. This happens if zend assertions are enabled, which is the default setting. The Handle::register() enables assertion exceptions. Core might want to make some tweaks to how this is handled in the register() method.

I think it does not matter if we set this ini setting regardless. Lower versions (PHP5) will just not care about it. Otherwise, we could check `PHP_MAJOR_VERSION`.

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -44,6 +44,10 @@ function drush_cli_core_cli() {
   $shell = new Shell($configuration);
 
   if ($drupal_major_version >= 8) {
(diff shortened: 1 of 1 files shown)


damiankloip:
Hmm, I don't think this is right actually. This just disables everything.

damiankloip:
Closing this in favour of an approach that actually works in https://github.com/drush-ops/drush/pull/2635

PR #1190, opened 2015-02-17
Fix datestamp to time of last commit.

danepowell:
see issue: #1059

PR #1846, opened 2015-12-07
Fix datestamp to time of last commit (.info.yml files).

danepowell:
See #1190 for original issue description. In #1190, we fixed the datestamp for traditional .info files. This PR ports that fix to work with .info.yml files as well.


jhedstrom:
Makes sense to me. I didn't realize we had such different code paths for ini vs yaml.

Thanks!

Issue #1842, opened 2015-12-04
make-update updates core from 7.x to 8.x.x


kazdegroot:
Given the following .make.yml file
```yml
core: 7.x
api: 2
defaults:
  projects:
    subdir: contrib
    translations:
      - nl
projects:
  drupal:
    version: '7.41'
    type: core
  features:
    version: '2.7'
  strongarm:
    version: '2.0'
```

running `drush make-update makefile.make.yml` produces:

```yml
core: 7.x
api: 2
defaults:
  projects:
    subdir: contrib
    translations:
      - nl
projects:
  drupal:
    version: 8.0.1
    type: core
  features:
    version: 8.x-3.0-alpha5
  strongarm:
    version: '2.0'
```

While I guess there is a point to be made that it really does attempt to update everything to the latest version, I imagine that this function shouldn't suggest core updates.

pjcdawkins:
It's a new command and it looks like it just doesn't support D7 [yet]. I guess, at least for now, it should fail with a "Not supported" error on D7 projects.

kazdegroot:
From a naive understanding of the functionality, it seems that it shouldn't differ that much from the way the current lock functionality works (just with some additional version comparisons). But reality tends to be more complex.
I'm currently working on a project where this would be pretty handy, so if I find the time, and no one has beaten me to the punch, I might take a stab at it in the next few weeks.

basvdheijden:
It seems that @matthijs-va already fixed it, but his code is still unmerged.

weitzman:
Lets close in favor of #1876.

PR #1848, opened 2015-12-08
Minor fixes to make commands

pjcdawkins:
I was hunting for a bug in `make-update` (#1842), unfortunately I couldn't find how to fix it (it seems more like it just hasn't been implemented for D7, yet).

But meanwhile my IDE has identified some probable bugs (variable names reused within loops), and some unused variables which can at least be confusing.

### DIFF
--- a/commands/make/make.download.inc
+++ b/commands/make/make.download.inc
@@ -370,7 +370,7 @@ function make_download_git($name, $type, $download, $download_location) {
     if (!empty($download['submodule'])) {
       $command = 'git submodule update';
       foreach ($download['submodule'] as $option) {
-        $command .= ' --%s';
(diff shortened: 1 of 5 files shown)


commands/make/make.download.inc
```diff
@@ -370,7 +370,7 @@ function make_download_git($name, $type, $download, $download_location) {
     if (!empty($download['submodule'])) {
       $command = 'git submodule update';
       foreach ($download['submodule'] as $option) {
-        $command .= ' --%s';
+        $command .= sprintf(' --%s', $option);
```
pjcdawkins: This one is possibly not the right fix, but it seems no worse than before

greg-1-anderson:
Make is maintenance only; closing.

PR #1862, opened 2015-12-15
Refactored make_move_build

sherakama:
This pull request is to allow for a recursive build or merge and takes on the:  `TODO: This only protects one level of directories from being removed.`

Refactored make move build to allow for a better --no-core and --override merge of builds. Now compares directories and only replaces those that actually were built in to place. This way builds with --no-core and --overwrite.


jhedstrom:
Test failures seem to be PHP 5.4-specific.

sherakama:
Thanks for the tip, I'll switch back to php 5.4 and see what is causing the meltdown.

weitzman:
Also, code needs into the master branch. Thanks for the contribution.

sherakama:
Going to close this one and open up one for the master branch. Back porting should be easy enough as it is just one function and is pretty similar between major versions.

PR #1873, opened 2015-12-17
8.x - Fixes create and restore database with drush arr

sherakama:
Drush arr fails when database parameter is provided on create command.


greg-1-anderson:
Tests are failing, no activity: closing

PR #1874, opened 2015-12-17
Speed up pm-download a lot by trying to rename tmp dir into place

weitzman:
Still uses copy as a fallback plan, since drush_move_dir already does that.

### DIFF
--- a/commands/pm/download.pm.inc
+++ b/commands/pm/download.pm.inc
@@ -215,7 +215,7 @@ function drush_pm_download() {
     }
 
     // Copy the project to the install location.
-    if (drush_op('_drush_recursive_copy', $request['full_project_path'], $request['project_install_location'])) {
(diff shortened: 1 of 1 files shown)


weitzman:
We lose [the code at end of _drush_recursive_copy()](https://github.com/drush-ops/drush/blob/8.0.1/includes/filesystem.inc#L271-L284) but in the case of pm-download it is not needed, I think.

jonhattan:
@weitzman in https://www.drupal.org/node/1168812 we discussed about respecting ACLs, group and s-bit ownership inheritance. We need to copy instead of move for group inheritance to work. Also the snippet you highlighted is responsible of the s-bit thing.

The commit in 1168812 also introduces a test that is not run by default. It should be enabled in Travis, if feasible. I'll run it locally and report back.

jonhattan:
Checked locally that filesystemTest.php fails with this PR. In #1890 I'm configuring Travis to run that test.

PR #1875, opened 2015-12-18
Add a --config parameter to import config after install.

simesy:
Passing  a --config parameter (eg. `drush site-install --config=path/to/config`) triggers the equivalent of a `drush config-import --source=path/to/config` after the installation is complete.

This is all-or-nothing, any UUIDs generated during the site installation will be overwritten, any config created during the installation that is not represented in the --config directory will be deleted.

A use case for feature is CI/testing where the source directory contains a known set of configuration to be tested.

For history of this feature see drush-ops/drush#1635

### DIFF
--- a/commands/core/site_install.drush.inc
+++ b/commands/core/site_install.drush.inc
@@ -85,6 +85,7 @@ function site_install_drush_help_alter(&$command) {
     else {
       unset($command['options']['writable']);
       unset($command['options']['keep-config']);
+      unset($command['options']['config']);
     }
   }
 }
--- a/commands/core/site_install.drush.inc
+++ b/commands/core/site_install.drush.inc
@@ -57,7 +57,7 @@ function site_install_drush_command() {
       ),
       'writable' => 'Make CMI and other dirs writable by both web and CLI users. Suitable for non Prod environments.',
       'keep-config' => 'Keep CMI directories untouched. This preserves existing configuration.',
-      'config' => 'After installation, override all configuration and UUIDs that were generated during installation with config from this file path.'
+      'config' => 'Update site UUID and then import configuration from this path.'
     ),
     'examples' => array(
       'drush site-install expert --locale=uk' => '(Re)install using the expert install profile. Set default language to Ukrainian.',
@@ -104,11 +104,11 @@ function drush_core_site_install_validate() {
     drush_set_context('DRUSH_SELECTED_URI', 'http://' . $sites_subdir);
   }
 
-  if ($source = drush_get_option('config')) {
-    if (!file_exists($source)) {
+  if ($config = drush_get_option('config')) {
+    if (!file_exists($config)) {
       return drush_set_error('config_import_target', 'The config source directory does not exist.');
     }
-    if (!is_dir($source)) {
+    if (!is_dir($config)) {
       return drush_set_error('config_import_target', 'The config source is not a directory.');
     }
   }
--- a/commands/core/site_install.drush.inc
+++ b/commands/core/site_install.drush.inc
@@ -1,6 +1,7 @@
 <?php
 
 use Drush\Log\LogLevel;
+use Drupal\Core\Config\FileStorage;
 
 function site_install_drush_command() {
   $items['site-install'] = array(
@@ -55,7 +56,8 @@ function site_install_drush_command() {
         'example-value' => 'directory_name',
       ),
       'writable' => 'Make CMI and other dirs writable by both web and CLI users. Suitable for non Prod environments.',
-      'keep-config' => 'Keep CMI directories untouched. This preserves existing configuration.'
+      'keep-config' => 'Keep CMI directories untouched. This preserves existing configuration.',
+      'config' => 'After installation, override all configuration and UUIDs that were generated during installation with config from this file path.'
     ),
     'examples' => array(
       'drush site-install expert --locale=uk' => '(Re)install using the expert install profile. Set default language to Ukrainian.',
@@ -99,6 +101,15 @@ function drush_core_site_install_validate() {
     }
     // Make sure that we will bootstrap to the 'sites-subdir' site.
     drush_set_context('DRUSH_SELECTED_URI', 'http://' . $sites_subdir);
+  }
+
+  if ($source = drush_get_option('config')) {
+    if (!file_exists($source)) {
+      return drush_set_error('config_import_target', 'The config source directory does not exist.');
+    }
+    if (!is_dir($source)) {
+      return drush_set_error('config_import_target', 'The config source is not a directory.');
+    }
   }
 }
 
@@ -289,4 +300,13 @@ function drush_core_site_install($profile = NULL) {
   }
   drush_include_engine('drupal', 'site_install');
   drush_core_site_install_version($profile, $form_options);
+
+  // Post installation run the configuration import.
+  if ($config = drush_get_option('config')) {
+    // Set the destination site UUID to match the source UUID, to bypass a core fail-safe.
+    $source_storage = new FileStorage($config);
+    drush_op('drush_config_set', 'system.site', 'uuid', $source_storage->read('system.site')['uuid']);
+    // Run a full configuration import.
+    drush_invoke_process('@self', 'config-import', array(), array('source' => $config));
+  }
 }
--- a/commands/core/site_install.drush.inc
+++ b/commands/core/site_install.drush.inc
@@ -298,6 +298,12 @@ function drush_core_site_install($profile = NULL) {
       $form_options[$key] = $value;
     }
   }
+
+  // If the profile is not explicitly set, default to the 'minimal' for an issue-free config import.
+  if (empty($profile) && drush_get_option('config')) {
+    $profile = 'minimal';
+  }
+
   drush_include_engine('drupal', 'site_install');
   drush_core_site_install_version($profile, $form_options);
 


commands/core/site_install.drush.inc
```diff
@@ -289,4 +300,13 @@ function drush_core_site_install($profile = NULL) {
   }
   drush_include_engine('drupal', 'site_install');
   drush_core_site_install_version($profile, $form_options);
+
+  // Post installation run the configuration import.
+  if ($config = drush_get_option('config')) {
+    // Set the destination site UUID to match the source UUID, to bypass a core fail-safe.
+    $source_storage = new FileStorage($config);
+    drush_op('drush_config_set', 'system.site', 'uuid', $source_storage->read('system.site')['uuid']);
```
cweagans: Doesn't this modify the UUID from the source config (i.e. what's on disk)? That is not a desirable behavior if the config is in version control - that config on disk should never change IMO. Since this option is only available when installing a fresh site, I think it's safe enough to change what's in the site instead of in the config directory.

cweagans:
I think this should also change the default profile choice to Minimal if --config is passed. Minimal doesn't create any config entities by default, whereas Standard does (shortcut links is what bit us in the past). Minimal provides the most likely path to success when installing from config, so it makes sense to me to switch that default.

commands/core/site_install.drush.inc
simesy: No, it modifies the UUID in the active config of the newly installed site. I've been using @markdorison's variation on this and it has never modified the source yml file on disk.

simesy:
That seems like a sound idea. It also makes the whole process faster.

simesy:
Commit added to default to 'minimal' per @cweagans suggestion.

simesy:
Hmm. Regardless of what you default the installation to, the imported configuration (assuming it includes `core.extensions`) has to match the name of the install profile or you get:

    Unable to uninstall the <em class="placeholder">Minimal</em> profile since it is the install profile.

cweagans:
@simesy I think it's okay to default to Minimal. If they're using some other profile, they can change it by adding the profile argument.

commands/core/site_install.drush.inc
```diff
@@ -55,7 +56,8 @@ function site_install_drush_command() {
         'example-value' => 'directory_name',
       ),
       'writable' => 'Make CMI and other dirs writable by both web and CLI users. Suitable for non Prod environments.',
-      'keep-config' => 'Keep CMI directories untouched. This preserves existing configuration.'
+      'keep-config' => 'Keep CMI directories untouched. This preserves existing configuration.',
```
weitzman: options that are version specific should get altered in so they aren't valid or shown for other versions. see user_drush_help_alter() for example.

commands/core/site_install.drush.inc
```diff
@@ -100,6 +102,15 @@ function drush_core_site_install_validate() {
     // Make sure that we will bootstrap to the 'sites-subdir' site.
     drush_set_context('DRUSH_SELECTED_URI', 'http://' . $sites_subdir);
   }
+
```
weitzman: Should move to drush_core_site_install_validate()

commands/core/site_install.drush.inc
```diff
@@ -287,6 +298,21 @@ function drush_core_site_install($profile = NULL) {
       $form_options[$key] = $value;
     }
   }
+
+  // If the profile is not explicitly set, default to the 'minimal' for an issue-free config import.
+  if (empty($profile) && drush_get_option('config')) {
+    $profile = 'minimal';
```
weitzman: Im still not clear why we do this? Can folks still use a custom profile with --config? If so then this is fine.

weitzman:
Left some minor comments. Would be good to hear that this actually works for folks.

commands/core/site_install.drush.inc
simesy: It works for me either way, but the whole process is faster with 'minimal' - there is no time wasted installing things that the config import will override anyway.

commands/core/site_install.drush.inc
simesy: It's already in drush_core_site_install_validate().

commands/core/site_install.drush.inc
simesy: Commit just pushed removes --config for versions less than 8.

simesy:
To recap: All comments addressed. The call needs to be made about removing the 'minimal' profile part, I would suggest removing it unless we can replicate the problem it's mitigating.

commands/core/site_install.drush.inc
```diff
@@ -55,7 +56,8 @@ function site_install_drush_command() {
         'example-value' => 'directory_name',
       ),
       'writable' => 'Make CMI and other dirs writable by both web and CLI users. Suitable for non Prod environments.',
-      'keep-config' => 'Keep CMI directories untouched. This preserves existing configuration.'
+      'keep-config' => 'Keep CMI directories untouched. This preserves existing configuration.',
+      'config' => 'After installation, override all configuration and UUIDs that were generated during installation with config from this file path.'
```
weitzman: "After a Minimal installation, import configuration from this path. Update site UUID accordingly."

commands/core/site_install.drush.inc
```diff
@@ -100,6 +103,15 @@ function drush_core_site_install_validate() {
     // Make sure that we will bootstrap to the 'sites-subdir' site.
     drush_set_context('DRUSH_SELECTED_URI', 'http://' . $sites_subdir);
   }
+
+  if ($source = drush_get_option('config')) {
+    if (!file_exists($source)) {
```
weitzman: Do we mean $config instead of $source? Another instance below.

simesy:
Resolved comments in last commit.

weitzman:
Squashed and merged. Thanks much.

greg-1-anderson:
This looks like it will be a good and useful feature, but we are not done yet; the new site-install option --config conflicts with the existing Drush global --config option, so folks like me who set --config cannot run site-install.

This local option needs to be renamed to something else.

simesy:
Thanks greg!

commands/core/site_install.drush.inc
alexpott: If the profile contains sub modules this is not going to work because the modules will not be found... actually what needs to happen here for correctness is the we should read the core.extension file in the incoming config and determine which is the install profile from the list of modules.

alexpott:
There are quite a few gotcha with this approach. If the install profiles on the source config and the drush install don't match then there are issues. Also the install language is interesting - I think it should match the site default language from system.site. To be honest I really wish we all got behind https://www.drupal.org/project/config_installer since this is doing exactly what people want, does it without installing incorrect config first, and has tests.

greg-1-anderson:
I believe that the [config installer works with drush site-install](https://github.com/drush-ops/drush/issues/1372).

It sounds like the implementation of config_installer is currently more robust than what we have here, but some folks prefer the DX of the --config. Maybe some work could be done here to ensure consistency between installation variables and their corresponding configuration values. I don't know that it's a problem to install minimal profile first.

I think that the ideal solution, for the future would be to take the operation of the config_installer and make it part of the core installation code rather than having it be a separate installation profile. Until then, what we have is just a choice between different workarounds.

simesy:
@alexpott "To be honest I really wish we all got behind config_installer" - I've passed over project solution because the projects I'm working on already use custom install profiles.

dgtlmoon:
Some help for those just catching up here.

So I've got a site profile which installs fine with `../vendor/bin/drush site-install my_profile` and I've already exported the configs with `../vendor/bin/drush config-export`. 

however when I run `../vendor/bin/drush site-install my_profile --config-dir=../config`

```
Installation complete.  User name: admin  User password: YdkMThXM9B                                                                                                                             [ok]
No config value specified.                                                                                                                                                                   [error]
Invalid argument supplied for foreach() DrupalBoot.php:240                                                                                                                                 [warning]
 Collection  Config                                    Operation 
             views.view.taxonomy_term                  delete    
             views.view.frontpage                      delete    
...
             core.date_format.html_datetime            delete    
             core.date_format.html_date                delete    
             core.date_format.fallback                 delete    
 sync        core.date_format.fallback                 create    
 sync        core.date_format.html_date                create    
 sync        core.date_format.html_datetime            create    
 sync        core.date_format.html_month               create    
...
Import the listed configuration changes? (y/n): y
Drupal\Core\Config\ConfigImporterException: There were errors validating the config synchronization. in Drupal\Core\Config\ConfigImporter->validate() (line 725 of                           [error]
/var/www/site/web/core/lib/Drupal/Core/Config/ConfigImporter.php).
The import failed due for the following reasons:                                                                                                                                             [error]
This import is empty and if applied would delete all of your configuration, so has been rejected.
```

This error does not really mean anything to me, where am I going wrong?

dgtlmoon:
@simesy have you seen https://www.drupal.org/node/2642398 ?

fluxsauce:
@dgtlmoon The practical solution is to use https://www.drupal.org/project/config_installer

For a `drush.make.yml`:

```yml
projects:
  config_installer:
    type: "profile"
    version: "1.3"
```

Installation:

```bash
drush site-install config_installer --config-dir=PATH/TO/config/base --yes
```

marcelovani:
I need to enable my own profile and import config from the repo to build a fresh site from code.

`i.e. drush site-install my_profile --config-dir=PATH/TO/config/base --yes`

How can I use config_installer to import the confi? It is not possible to have two profiles enabled at the same time.

marcelovani:
I think I figured out why it wasn't working and how config_installer knows that I want **my_profile** enabled.

Basically **core.extension.yml** is what defines which profile will be enabled.

I had an entry for **standard: 0** which I changed to **my_profile: 0**

Now it works.

I can only make it work using the UI. When using drush, I get an error saying that the folder config/base is empty.

PR #1879, opened 2015-12-21
Fix #1878. Consider drush commandfiles located in all enabled themes, not just default+admin themes.

### LINKED ISSUE #1878: Drush regression with themes' commands
Some themes (eg Omega) defines drush commands to help themers to do their job.

With Drush 7, these commands were runnable even when the theme was not installed.
With Drush 8, they are runnable only if the theme is installed and set by default. Even if it's one of its subthemes that's set by default, the commands won't run.

Steps to reproduce:
1. Download and install Drupal 7
2. $ composer require drush/drush:8.*
3. $ drush cc drush
4. $ drush omega-guard
Command omega-guard needs the following module(s) enabled to run : omega
5. $ rm -rf composer.* vendor
6. $ composer require drush/drush:7.*
7. $ drush cc drush
8. $ drush omega-guard
Work as expected

weitzman:


DuaelFr:
It works like a charm :)
Thanks a lot.
Do you plan to make a new release soon?

weitzman:
Thanks - cherry picked onto master.

weitzman:
Release expected every month, coinciding with Drupal. See https://www.drupal.org/core/release-cycle-overview

Issue #1493, opened 2015-07-14
user-login to a remote host does not pass along a cli override of --uri option


ian-rose:
I have an `example.aliases.drushrc.php` file with contents of:

```php
<?php
$aliases['prod'] = array(
  'parent' => '@parent',
  'site' => 'example',
  'env' => 'prod',
  'root' => '/path/to/drupal',
  'remote-host' => 'server.tld',
  'remote-user' => 'example',
  'uri' => 'www.example.com',
);
```

When I run `drush @example.prod uli` the command returns `http://www.example.com/user/reset/1/[token]/login`.

When I run `drush @example.prod uli -l edit.example.com` I expect to get `http://edit.example.com/user/reset/1/[token]/login`, but instead the command returns `http://www.example.com/user/reset/1/[token]/login`

I believe this is an issue, as command-line arguments should override any defaults defined in an alias.

weitzman:
This is an edge case when you are calling uli on a remote host and have specified a custom uri. We dont pass aong that uri in our redispatch. See https://github.com/drush-ops/drush/blob/8.0.0-beta12/commands/user/user.drush.inc#L378. Ping @greg-1-anderson for ideas on how to fix. Not urgent.

ian-rose:
I tested with the following alias configuration on a local environment and am still seeing drush not prioritize the `--uri` flag over the alias's `uri` definition:

```php
<?php
$aliases['local'] = array(
  'parent' => '@parent',
  'site' => 'example',
  'env' => 'local',
  'root' => '/path/to/drupal',
  'uri' => 'local.example.com',
);
```

This issue appears to apply to both remote hosts and local sites.

weitzman:
@ian-rose - what version of drupal and drush are you using?

weitzman:
I tried to reproduce the problem and was not able. Here is my alias and command. I got back an URL that 127.0.0.1 domain, not mm domain:

```
~/c/h/drush (master<)$ dr sa @d7
$aliases["d7"] = array (
  'root' => '/Users/moshe.weitzman/htd/d7',
  'uri' => 'http://mm/d7',
  'path-aliases' => 
  array (
    '%dump-dir' => '/Users/moshe.weitzman/sqldump',
  ),
);
```

`dr @d7 --uri=127.0.0.1/d7 uli`

ian-rose:
I was using a (outdated) 6.x version. I pulled the latest 6.x and tested on that and 7.x, and am no longer seeing the issue on local environments.

[...]

PR #1883, opened 2015-12-22
#1493 - override alias records with cli context options

blazindrop:
First stab at fixing the issue where command line arguments (especially --uri) can't override site aliases.

### DIFF
--- a/includes/command.inc
+++ b/includes/command.inc
@@ -700,7 +700,12 @@ function drush_redispatch_get_options() {
   $global_option_list = drush_get_global_options(FALSE);
   foreach (drush_get_context('alias') as $key => $value) {
     if (array_key_exists($key, $global_option_list)) {
-      $options[$key] = $value;
+      if (isset($cli_context[$key])) {
+        $options[$key] = $cli_context[$key];
+      }
+      else {
+        $options[$key] = $value;
+      }
     }
   }
 


weitzman:
ping @greg-1-anderson.

This one is biting some folks. I did some tests with this command `drush @d7 --uri=http://foo pm-updatecode`. Below is the site alias @d7:

```
$aliases['foo'] = array(
 'root' => '/Users/moshe.weitzman/htd/d7',
 'uri' => 'http://mm/d7',
);
```

My expectation is that the --uri=http://foo on the CLI is propagated to backend calls but that's not happenning on drush7 or drush8. It *is* happenning on drush6. Is this by design? The non-overridden --uri is visible in log excerpt below:

```
Backend invoke: /usr/local/bin/php /Users/moshe.weitzman/c/h/drush/drush.php --php=/usr/local/bin/php  --backend=2 --verbose --debug --no  --root=/Users/moshe.weitzman/htd/d7 --uri=http://mm/d7  pm-updatestatus 2>&1
```
This PR fixes the bug for some commands like `drush @foo --uri=http://buytaert.net browse` but not the pm-updatecode example above.

greg-1-anderson:
Yeah, I looked at that PR a few days ago, and my initial impression was, "why is the existing code doing it that way?"  Adding explicit cli overrides on top of an explicit alias override (band-aid on the band-aid) looks like too much doctoring.  I was wondering if it would hurt too much to just pull off the band-aid, and see if we can heal the owie some other way. The existing Drush context priority logic should be right most of the time; I'm not sure why it is not in this case.  (n.b. cli overrides alias if we just take off the band-aid).

weitzman:
The PR changes drush_redispatch_get_options() but thats not even called for `drush @d7 --uri=http://foo pm-updatecode`

greg-1-anderson:
So, I think the fix for `drush_redispatch_get_options()` should be:

`$options = $cli_context + drush_get_context('specific') + $alias_options;`

I think alias options should just be all of the options.  Then, below, in the part where we get the command in order to filter out options that are not defined in the command definition, we should union in the global options, and treat all options uniformly.

It does not make any sense that we should filter options based on where they come from.  What matters is: (a) which contexts are relevant, (b) what their relative priority is, and (c) which options should be sent on to the remote command.  We should just do that.

I don't know about the code path that does not go through `drush_redispatch_get_options()`, though.

weitzman:
I think the other code path has to do with the alias 'self'. In drush6 that had a --uri value from cli but in 7/8 it has alias' value.

ElijahLynn:
This one is biting us too, Acquia just did a drush default switch from drush6 to drush8. Previously working Cloud Hooks were working fine and now they are broken for our particular setup. We have a multisite setup and the Cloud Hooks run from $HOME. The scripts are calling drush --uri with the old 'foldername' for drush 6 but when I try the recommend full URI that should map to the sites.php it shows it initiates the sites in the right folder but then as you say it doesn't pass that to the backend and I get the original URI in the alias file.

I know you guys know this but just posting this here for the search engines.

weitzman:
Just posted a proposed fix for both the pm-updatecode code path and the drush_redispatch_get_options() code path at #1924. Please test and report back. let's closer this and discuss there.

Also could use some unit tests.

PR #1884, opened 2015-12-24
Add dependency injection container to Drush.

greg-1-anderson:
**Motivation:**

We'd like to be able to factor out key parts of Drush into libraries; key among these are the alias / backend invoke handling. We also want to move away from `drush_get_context()` for fetching important Drush service object references.

**First step:**

This PR initializes a DI container that has only the logger in it.  This will allow us to also initialize Alias and Backend Invoke classes from an external library, and use DI to give them a logger.

**Next steps:**

We could use DI to construct all of the bootstrap objects. If some external plugin provided a bootstrap object, it could also initialize it with its own foo-services.yml, and make references to 'logger' et. al. as needed to construct its objects, since the DI container will already be initialized with all of the standard Drush instances.  It would then need to make one more explicit call to add its bootstrap object to our bootstrap object manager.

We could use this technique for our bootstrap objects as well, so each DrushBootN had its own set of services, which instantiated the correct Rules8 vs Rules7 object, and gave them the same tag (e.g. 'rules').  This would make it way easier to provide platform-independent helper classes, or even use DI to provide the implementation for a group of Drush commands associated with a platform.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "5483a9ed6ab914076b7aa8305d7b1903",
-    "content-hash": "32fceb4a785595b4e40cf4e602703132",
+    "hash": "8435c9eb245e5b5ad0e26c7d7ef6e8f9",
+    "content-hash": "7092f159ce3306251aa9f5440b18d233",
     "packages": [
         {
             "name": "dnoegel/php-xdg-base-dir",
@@ -455,16 +455,16 @@
         },
         {
             "name": "symfony/dependency-injection",
-            "version": "v2.7.6",
+            "version": "v2.7.7",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/dependency-injection.git",
-                "reference": "af284e795ec8a08c80d1fc47518fd23004b89847"
+                "reference": "7a201bf08c29e47075047cbb378724ad46dfe217"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/dependency-injection/zipball/af284e795ec8a08c80d1fc47518fd23004b89847",
-                "reference": "af284e795ec8a08c80d1fc47518fd23004b89847",
+                "url": "https://api.github.com/repos/symfony/dependency-injection/zipball/7a201bf08c29e47075047cbb378724ad46dfe217",
+                "reference": "7a201bf08c29e47075047cbb378724ad46dfe217",
                 "shasum": ""
             },
             "require": {
@@ -492,7 +492,10 @@
             "autoload": {
                 "psr-4": {
                     "Symfony\\Component\\DependencyInjection\\": ""
-                }
+                },
+                "exclude-from-classmap": [
+                    "/Tests/"
+                ]
             },
             "notification-url": "https://packagist.org/downloads/",
             "license": [
@@ -510,7 +513,7 @@
             ],
             "description": "Symfony DependencyInjection Component",
             "homepage": "https://symfony.com",
-            "time": "2015-10-27 15:38:06"
+            "time": "2015-11-23 10:17:36"
         },
         {
             "name": "symfony/filesystem",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "160e694a860dd8b012104bae69cd2d75",
-    "content-hash": "345215a79dc35e20c5858b0a05b0aba6",
+    "hash": "afef9aa9d988aad79812fb0e143ea9ca",
+    "content-hash": "01111258a3e0b2304ddc2d01e11c7082",
     "packages": [
         {
             "name": "dnoegel/php-xdg-base-dir",
@@ -129,32 +129,38 @@
         },
         {
             "name": "nikic/php-parser",
-            "version": "v1.4.1",
+            "version": "v2.0.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/nikic/PHP-Parser.git",
-                "reference": "f78af2c9c86107aa1a34cd1dbb5bbe9eeb0d9f51"
+                "reference": "c542e5d86a9775abd1021618eb2430278bfc1e01"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/nikic/PHP-Parser/zipball/f78af2c9c86107aa1a34cd1dbb5bbe9eeb0d9f51",
-                "reference": "f78af2c9c86107aa1a34cd1dbb5bbe9eeb0d9f51",
+                "url": "https://api.github.com/repos/nikic/PHP-Parser/zipball/c542e5d86a9775abd1021618eb2430278bfc1e01",
+                "reference": "c542e5d86a9775abd1021618eb2430278bfc1e01",
                 "shasum": ""
             },
             "require": {
                 "ext-tokenizer": "*",
-                "php": ">=5.3"
+                "php": ">=5.4"
             },
+            "require-dev": {
+                "phpunit/phpunit": "~4.0"
+            },
+            "bin": [
+                "bin/php-parse"
+            ],
             "type": "library",
             "extra": {
                 "branch-alias": {
-                    "dev-master": "1.4-dev"
+                    "dev-master": "2.0-dev"
                 }
             },
             "autoload": {
-                "files": [
-                    "lib/bootstrap.php"
-                ]
+                "psr-4": {
+                    "PhpParser\\": "lib/PhpParser"
+                }
             },
             "notification-url": "https://packagist.org/downloads/",
             "license": [
@@ -170,7 +176,7 @@
                 "parser",
                 "php"
             ],
-            "time": "2015-09-19 14:15:08"
+            "time": "2015-12-04 15:28:43"
         },
         {
             "name": "pear/console_table",
@@ -338,26 +344,77 @@
             "time": "2015-11-12 16:18:56"
         },
         {
-            "name": "symfony/console",
-            "version": "v2.7.7",
+            "name": "symfony/config",
+            "version": "v3.0.0",
             "source": {
                 "type": "git",
-                "url": "https://github.com/symfony/console.git",
-                "reference": "16bb1cb86df43c90931df65f529e7ebd79636750"
+                "url": "https://github.com/symfony/config.git",
+                "reference": "40bae8658dbbb500ebc19aa9fde22dc4295fc290"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/console/zipball/16bb1cb86df43c90931df65f529e7ebd79636750",
-                "reference": "16bb1cb86df43c90931df65f529e7ebd79636750",
+                "url": "https://api.github.com/repos/symfony/config/zipball/40bae8658dbbb500ebc19aa9fde22dc4295fc290",
+                "reference": "40bae8658dbbb500ebc19aa9fde22dc4295fc290",
                 "shasum": ""
             },
             "require": {
-                "php": ">=5.3.9"
+                "php": ">=5.5.9",
+                "symfony/filesystem": "~2.8|~3.0"
+            },
+            "type": "library",
+            "extra": {
+                "branch-alias": {
+                    "dev-master": "3.0-dev"
+                }
+            },
+            "autoload": {
+                "psr-4": {
+                    "Symfony\\Component\\Config\\": ""
+                },
+                "exclude-from-classmap": [
+                    "/Tests/"
+                ]
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "authors": [
+                {
+                    "name": "Fabien Potencier",
+                    "email": "fabien@symfony.com"
+                },
+                {
+                    "name": "Symfony Community",
+                    "homepage": "https://symfony.com/contributors"
+                }
+            ],
+            "description": "Symfony Config Component",
+            "homepage": "https://symfony.com",
+            "time": "2015-11-02 20:34:04"
+        },
+        {
+            "name": "symfony/console",
+            "version": "v3.0.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/symfony/console.git",
+                "reference": "175871ca8d1ef16ff8d8cac395a1c73afa8d0e63"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/symfony/console/zipball/175871ca8d1ef16ff8d8cac395a1c73afa8d0e63",
+                "reference": "175871ca8d1ef16ff8d8cac395a1c73afa8d0e63",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=5.5.9",
+                "symfony/polyfill-mbstring": "~1.0"
             },
             "require-dev": {
                 "psr/log": "~1.0",
-                "symfony/event-dispatcher": "~2.1",
-                "symfony/process": "~2.1"
+                "symfony/event-dispatcher": "~2.8|~3.0",
+                "symfony/process": "~2.8|~3.0"
             },
             "suggest": {
                 "psr/log": "For using the console logger",
@@ -367,7 +424,7 @@
             "type": "library",
             "extra": {
                 "branch-alias": {
-                    "dev-master": "2.7-dev"
+                    "dev-master": "3.0-dev"
                 }
             },
             "autoload": {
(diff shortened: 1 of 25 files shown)


composer.lock
```diff
@@ -338,26 +344,77 @@
...
+                "reference": "40bae8658dbbb500ebc19aa9fde22dc4295fc290"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/symfony/config/zipball/40bae8658dbbb500ebc19aa9fde22dc4295fc290",
+                "reference": "40bae8658dbbb500ebc19aa9fde22dc4295fc290",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=5.5.9",
```
weitzman: I don think we upped our requirements to 5.5 yet. I'm fine doing so for Drush9. I'd be OK with this patch being our starting point for Drush9. No strong opinion.

includes/drush.inc
```diff
@@ -1261,11 +1265,46 @@ function drush_log($message, $type = LogLevel::NOTICE, $error = null) {
 }
 
 /**
+ * Set up our dependency injection container, and load drush-services.yml.
+ */
+function drush_init_dependency_injection_container() {
```
weitzman: Perhaps all this code belongs in preflight.inc?

greg-1-anderson:
Yes, this is definitely a Drush 9 feature. The biggest problem we have right now is that Drupal 8 has its own version of the Symfony DependencyInjection classes, which is causing the ContainerBuilder to blow up due to some dependency hell problems. Since Drupal has its own copy, these incompatibilities can't be resolved simply by including Drush in the same composer.json as Drupal.

I'm still considering how to best solve this problem. It might be best if Drush also had its own DI container. It depends on exactly why the ContainerBuilder is failing; I haven't had the time to really look into this yet.

weitzman:
http://pimple.sensiolabs.org/ is a popular alternative. It looked nice and simple when I briefly looked a while back. I have no useful input beyond that.

greg-1-anderson:
I looked at the Drupal ContainerBuilder class, and it merely extends the symfony container builder; it does not replace it. It is possible that always using a site-local Drush would be a solution here. I might try setting up the Drush Drupal 8 tests to always use a site-local Drush.

Pimple looks fine if you want to initialize your DI in code. I saw that some folks use the Symfony YamlFileLoader with Pimple. That might be a solution; I have not looked into whether it would be an equivalent amount of revlocking as we already have with the existing ContainerBuilder. There is also [td7650/yml2pimple](https://github.com/td7650/yml2pimple), but it is a fork of an unmaintained project, and has no issue queue (the parent project has the issue queue). It also uses Pimple 1.x instead of Pimple 3.x, and an older yaml parser than we use.

Maybe I'll try the YamlFileLoader with Pimple, and see how far that gets us.

greg-1-anderson:
The tests pass when Drush is using exactly the same version of symfony/dependency-injection as Drupal 8; perhaps this is sufficient, since we already use components that must be compatible with Drupal 8.

weitzman:
I'm still slightly hesitant to give up on the feature where a single Drush
can operate on multiple versions of Drupal.

greg-1-anderson:
We are already missing that feature--the global Drush can break against future versions of Drupal 8 at any time, and the current version of Drush doesn't work against older Drupal 8 betas and alphas.  We don't notice the problem so much, because we happen to work with both 8.0.0 and 8.0.1, but this will be harder to do as time goes on, when there are more Drupal 8 versions using different sets of dependencies.

Still, I'll try out Pimple with the YamlFileLoader. If it worked out, it may delay the time until we run into problems.

greg-1-anderson:
I considered this a bit more, and I am a little reluctant to use pimple. It is similar to the Symfony dependency injection component, yet different; it seems undesirable and confusing to have different containers in Drush and Drupal.

I think it may be possible to make Drush slightly less vulnerable to some types of dependency hell, but it would take a little refactoring. If it does not work out, we can try switching to pimple later. For now, though, I think I'll continue from 64f617b, and refactore a small amount of Drush preflight logic into classes.

greg-1-anderson:
My idea above was a dead-end; it was my thought that we could reduce some types of dependency hell by including the Drupal autoloader immediately after we included our autoloader, before we attempted to use any symfony classes. This could possibly work for `drush --root=...`, but in the case of `drush @alias ...`, we want to use DI to manage our logger and alias classes, so we would have to use symfony before we knew where the Drupal root is.

Fortunately, though, this whole problem seems to be a false alarm. I failed to check in composer.lock on one of my tests, leading to spurious failures that I thought was a dependency hell problem, but was actually simply an incompatible library versions problem.

**Summary**

- We will still have problems if Drupal upgrades to symfony/dependency-injection 3.x, as, at that time, some versions of Drupal will be using 2.x (as it currently is), and we will have to pick one of 2.x or 3.x to use, and therefore will not support all versions of Drupal 8 with one version of Drush.
- We already have this risk from other components we share with Drupal 8.
- Using pimple will not significantly improve the situation, as we will still be revlocked to the YamlFileLoader. There are no good alternatives for this right now, as mentioned above, unless we want to maintain our own DI yaml loader.
- Minor updates of symfony/dependency-injection are okay; so, we can use symfony/dependency-injection 2.7.7 while Drupal is using 2.7.6, and this does not cause a problem.

Given this, are we okay with moving forward with symfony/dependency-injection 2.7* in Drush 9?

greg-1-anderson:
I'm starting to create a \Drush class modeled off of the \Drupal class from Drupal 8. This insulates the dependency injection code with static wrapper methods. If everyone uses these methods for any needed direct access to the DI container for legacy code, then we could switch to pimple later with minimal disruption, should it become necessary to do so.

I'll commit the code once it's ready, but I think it's safe to move forward with symfony/dependency-injection 2.7.* for now.

weitzman:
I'm generally +1 for this. I'll have a more considered opinion once the PR is more complete. Thanks for working on this.

greg-1-anderson:
I meant "add the code to this PR", not commit. Thanks for the interim feedback; will add more here when it is ready. I am also considering a non-DI version of the "Drush" class for Drush 7 & 8 to make it easier to backport code changes from Drush 9. Might be handy.

greg-1-anderson:
Here is the DI code using a \Drush class to manage the container, just as Drupal 8 does.

Next steps:

- The various 'boot' objects, esp DrupalBoot8, could be created via dependency injection, and we could insert an instance of the logger into these instances via constructor injection. This would eliminate the call to \Drush::logger() in DrupalBoot8.

- We could add an Alias service class, and use that for loading and resolving alias files.

- We could create a wrapper object for Drush backend invoke.

greg-1-anderson:
Another thing to consider is how to handle Drush version. In Drupal, this is just \Drupal::VERSION, which has the advantage of being able to be accessed easily after including the autoload file. In the current implementation of this PR, though, this is not implemented, because the current version of Drush stores the version number in a "drush.info" file that could potentially be read by some other tool or process. PHP constants cannot be initialized with code; as the name implies, they must be initialized with a constant. We therefore use a slightly different implementation that can still read the drush_info file for backwards compatibility. If there are no external programs that read this file, we could consider just using a \Drush::VERSION constant, like Drupal 8 does.

greg-1-anderson:
I moved our global includes into the autoloader, because that is where they belong.

greg-1-anderson:
I am going to back out 6c84c76. Placing all of our global includes in the autoloader is very convenient and useful, but there are two problems with it:

1. All of the Drush global includes would be parsed by Drupal on every request, even though they were not needed.

2. It would make it perhaps even **too** convenient to have all of the legacy global functions available at all times, which could potentially slow down our efforts to migrate these to lazy-loadable classes.

greg-1-anderson:
This PR adds dependency injection for the bootstrap objects; some additional code cleanup still needed, but the basic structure is in place.

Note that this PR also removes the mechanism that Backdrop is using to hook into Drush to provide its own bootstrap object. To bring that back, we would need to implement a [compiler pass to find and add bootstrap extensions](http://symfony.com/doc/current/components/dependency_injection/tags.html).

We should decide on coding conventions to use here. For functions I moved over from bootstrap.inc, I retained underscore_naming, but these should probably all be converted to camelCase.

greg-1-anderson:
Another good thing to do here would be to move all of bootstrap.inc into the BootstrapManager class; however, there are references in bootstrap.inc to both site alias and backend invoke code; any such references in the Drush\Boot classes should also be done via dependency injection. I'm not sure if these code paths should even exist here; I'll probably work on separating out site aliases before refactoring any more functions in bootstrap.inc.  I think we could just remove the word "DRUPAL" from the bootstrap levels, though, and make these standard for use by all frameworks, since the concepts are general enough. I think that the `drush_bootstrap_value()` function could be removed, though, and other storage could be used instead here. In general, I think it is a good strategy to remove any use of `drush_set_context()` that is used for global variable storage rather than context (option) storage.

Note that Drush-specific code can use custom log levels, like `LogLevel::BOOTSTRAP`, but any set of classes that might be separated out into a standalone library should only use the standard PSR log levels. I think that the bootstrap log level is still useful enough to keep around.

lib/Drush.php
```diff
@@ -0,0 +1,209 @@
...
+ *   $object = drush_get_context('DRUSH_CLASS_LABEL');
+ *
+ * Better:
+ *   $object = \Drush::service('label');
+ *
+ *   function hook_do_stuff() {
+ *     $lock = lock()->acquire('stuff_lock');
+ *     // ...
+ *   }
+ * @endcode
```
weitzman: Are there two examples here? I don't get the Lock example

lib/Drush.php
```diff
@@ -0,0 +1,209 @@
...
+   * to follow the style used in Drupal 8.  For now, we read
+   * our version information from the drush.info file.
+   *
+   * Do any external clients try to read this file?  With
+   * a class constant, any code that includes autoload.php
+   * may reference \Drush::VERSION.  The way we have implemented
+   * it here allows \Drush::getVersion(), or direct parsing
+   * of drush.info, which is better for backwards compatibility.
+   */
+  //const VERSION = '8.0.0-dev';
```
weitzman: Let's leave this info in the .info file for now.

lib/Drush/Boot/BootstrapManager.php
```diff
@@ -0,0 +1,147 @@
...
+   * Return the bootstrap object in use.  This will
+   * be the latched bootstrap object if we have started
+   * bootstrapping; otherwise, it will be whichever bootstrap
+   * object is best for the selected root.
+   */
+  public function getBootstrap() {
+    if ($this->bootstrap) {
+      return $this->bootstrap;
+    }
+    return $this->select_bootstrap_class();
```
weitzman: I agree that switching to camel case makes sense. Can be a follow up if desired.

drush-services.yml
```diff
@@ -0,0 +1,25 @@
+parameters:
```
weitzman: Are users supposed to edit this file? Code comments at top would help. If the do edit this file we need to consider phar users who have read only copy.

weitzman:
I'm traveling so can only give brief comments. 

this looks terrific. KEEP GOING!

lib/Drush.php
greg-1-anderson: Cut-copy-pasta error!  I based my comment on the advice given in Drupal 8, edited it for Drush, and did not remove that part. It should not be there.

drush-services.yml
greg-1-anderson: No, this file is for Drush core only. I agree that comments here would be helpful anyway.

My idea is that Drush could also look for a services file at the document root; this would allow Backdrop & c. to add their own dependency injection easily. We'd load the Drush services first, which would allow others to refer to labels like 'logger' and 'bootstrap.manager' without needing to know exactly what their implementation was, so long as they fulfilled the contract. As I mentioned in a comment, I am going to re-implement the way Boot classes are added to the BootstrapManager for better extensiblity.

It might also make sense for Drush to look for service files anywhere that it currently looks for Drush extensions. This will need a little consideration and discussion, but I think that ultimately this could result in the ability to write extensions that are very separable / testable.

greg-1-anderson:
I will keep going here. I kept getting blocked on other attempts to refactor for various reasons; I think this will break the log-jam, and allow us to move forward in a number of areas we have wanted to improve for a long time.

greg-1-anderson:
Now we use a compiler pass to identify tagged services, so we could potentially add more 'boot' classes simply by loading more service files (e.g. a Backdrop Drush service file, found at the document root, or maybe docroot/drush).

greg-1-anderson:
So, it is kind of a problem that we cannot alter the container after it has been compiled. Re-initializing the container like this is undesirable both in terms of performance and behavior, as you risk dumping dynamic parameters, if we end up using that feature. Even if we do not use that feature, it feels wrong to throw out the container and re-initialize it multiple times.

I think we need to do something more like this:

- Initialize the container and load our services
- Preflight far enough to identify the site root. Load additional services defined by the framework.
- Ask for all of the services tagged 'bootstrap.boot'
- Manually add those services to the Bootstrap manager
- Bootstrap far enough to identify modules (this pretty much means a full bootstrap for the command)
- Add services / commands defined by the modules via the [Symfony extension mechanism](http://symfony.com/doc/current/components/dependency_injection/compilation.html#managing-configuration-with-extensions)
- Compile the container

The thing that is dissatisfying about this is that [parent services](http://symfony.com/doc/current/components/dependency_injection/parentservices.html) and potentially others--and certainly any compiler passes--cannot be used until the container is compiled.

I am unsure whether or not it is important to use parent services (a simple optimization) or compiler passes. I sort of presumed that the reason Symfony has a compiler pass feature is to allow objects to be fully initialized before they are given to any other object (e.g. via constructor or setter injection).  However, it seems that dependency injection happens at `load` time, and compiler passes happen at `compile` time. If this is in fact the case, then perhaps the proposal above is sufficient.

greg-1-anderson:
Drupal 8 modules can [define their own services](https://www.drupal.org/node/2133171) (scroll down). These are part of the Drupal 8 container rather than the Drush container.  We could use the Drupal 8 container to find tagged classes that are Drush extensions, just as we find tagged `Boot` classes in this PR.  Doing this would be compelling insofar as we would not need to spend any effort on the container, or finding modules, as Drupal has already done that for us. The down-side of this approach, though, is that the dependency injection done in the module's services file could only reference Drupal services; the Drush services would all be invisible.

The alternative would be for us to load our own example.drush-services.yml file (for the "example" module), in which case it would be possible to use DI to inject Drush services into the module's cli classes.  However, this alternative has the inverse problem, which is to say that the module's cli classes then could not use DI to reference the services defined in Drupal. This might be even more disadvantageous, as the purpose for writing a Drush extension in a module is to make it easy to call the Module / Drupal code.

I am not sure whether we might be able to just merge the two containers together somehow; that would be cool, but might not be possible.

greg-1-anderson:
Combining the two containers together is pointless; this presumes that you'd have both Drush and Drupal services in the same module services.yml file, which would blow up when it was loaded by Drupal w/out Drush.

We'll have to find our own drush-services.yml file in each module, and load it separately if it exists. The module's Drush extension will then need to use `\Drupal::service()` to obtain objects constructed via Drupal's DI loader.

includes/preflight.inc
```diff
@@ -160,9 +171,11 @@ function drush_preflight_prepare() {
...
   }
 
-  $drush_info = drush_read_drush_info();
-  define('DRUSH_VERSION', $drush_info['drush_version']);
-  $version_parts = explode('.', DRUSH_VERSION);
+  // Set up for backwards-compatibility. New code should call
+  // \Drush::getVersion() and \Drush::getMajorVersion() directly
+  $drush_version = \Drush::getVersion();
+  $version_parts = explode('.', $drush_version);
+  define('DRUSH_VERSION', $drush_version);
```
weitzman: Let's use the @deprecated annotation on things like this,. That way PHPStorm lets you know that you are being naughty.

lib/Drush.php
```diff
@@ -0,0 +1,216 @@
...
+ * accessor to arbitrary services for use by legacy Drush code.
+ *
+ * Advice from Drupal 8's 'Drupal' class:
+ *
+ * This class exists only to support legacy code that cannot be dependency
+ * injected. If your code needs it, consider refactoring it to be object
+ * oriented, if possible. When this is not possible, and your code is more
+ * than a few non-reusable lines, it is recommended to instantiate an object
+ * implementing the actual logic.
+ *
```
weitzman: Based on your most recent comment, it sounds like Drush extensions are going to have to call Drupal::service() instead of getting Drupal dependencies injected. So this paragraph preaches an ideal which can't be met for Drush extensions. Let's remove or rephrase?

lib/Drush/Symfony/BootstrapCompilerPass.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
weitzman: This file can be removed?

weitzman:
This looks terrific. When you are satisfied, feel free to create 8.x branch and then merge this into master.

A few non-urgent questions:
1. Where do Drush properties living like DRUSH_VENDOR_PATH,  DRUSH_COMMAND, etc. 
1. Where should we do validation like minimum PHP check, is_cli() check, etc.

lib/Drush.php
greg-1-anderson: Actually, I hope to provide dependency injection for Drush extensions in a future commit here.

lib/Drush/Symfony/BootstrapCompilerPass.php
greg-1-anderson: Possibly; I am still uncertain about our best strategy vis-a-vis freezing the container.

greg-1-anderson:
I hope to provide a configuration class to replace our context system. Ideally, new commands will gain access to `drush_get_option()` and `drush_get_context()` via an object that they may obtain via dependency injection.  Still working out how this should best work; ideally, all of the "work" should be done by library functions (classes) that can be called with **parameters** instead of global state.

I could potentially commit here before solving that problem; I want to take one more go at determining our container strategy first. We can always adjust the implementation later if I don't get it exactly right the first time.

greg-1-anderson:
Continued at #2018

PR #1885, opened 2015-12-25
Suggested fix for #1875: rename --config to --config-dir

greg-1-anderson:
site-install local option --config conflict with existing Drush global --config option. Rename the new site-install option to --config-dir.

I am not too partial to one name over another, as long as it is free from conflict.


weitzman:
That name looks good. Thanks.

PR #1886, opened 2015-12-25
Strange problem with double-include of make.utilities.inc

greg-1-anderson:
Does not seem to be happening on Travis, but let's see if this at least does no harm. Just creating this to force the tests to run.


greg-1-anderson:
Not entirely sure this is necessary; just reducing the entropy in my life.

PR #1888, opened 2015-12-29
Trimming value read from .drush-use.

grasmash:


grasmash:
Some IDEs will add extra lines to files, particularly when configured to follow Drupal coding standards. This whitespace prevents the value of .drush-use from working as expected. Trimming the value fixes this issue.

PR #1889, opened 2015-12-29
Removing check for index.php when scanning for drush wrapper or launcher.

grasmash:


grasmash:
`drush_startup()` only calls `find_wrapper_or_launcher()` in the following situations:
* --root is specified
* a .drush-use file exists
* index.php is in cwd

It would be great if drush scanned for ./drush-wrapper even when index.php is not in the cwd. E.g., when we're one level above the docroot. This change causes that to happen.

greg-1-anderson:
The existing code is broken.  It is too restrictive; the loop exits immediately if there isn't an index.php in the directory being checked. This means that the algorithm will seldom (never) iterate upwards.

The suggested change is necessary, but insufficient; it is a little too unrestrictive. `find_wrapper_or_launcher()` should be broken up; two more functions should be created, one that does the iteration over directories and then calls the second function, which iterates over the kinds of drush executables.  `find_wrapper_or_launcher()` itself should call the first function if index.php exists, or just call the second function directly if there is no index.php.

Then I think it would be right.

greg-1-anderson:
I'm testing my proposal now.

greg-1-anderson:
Continuing at #1901 because GitHub and git permissions.

PR #1890, opened 2015-12-29
Configure Travis env to allow filesystem tests

jonhattan:
To allow filesystemTest, we need to give UNISH_USERGROUP the name of a group the user belongs to.

Ref #1874


jonhattan:
Travis provides `travis` user/group, with no extra groups. So we need `sudo` to add the user to a group.

We also want this change takes effect immediately. In an interactive session you use `newgrp` command for that, but this doesn't work in Travis. Here's the instructions I've followed https://docs.travis-ci.com/user/ci-environment/#Group-membership

---

This PR enables sudo. The drawback is that we loose faster builds with containers -- https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

jonhattan:
Updated to "Trusty Beta" environment. It is on 20-52s boot time, faster than the standard method (50-145s) - https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

jonhattan:
Disabled xdebug.

jonhattan:
The two first commits are enough. Subsequent ones are an attempt to speed up the build. Sadly the legacy environment required for `sudo` operations is slow as hell.

weitzman:
I'd prefer not to slow down all tests for a seldom needed test. Any suggestion how to proceed here?

Issue #1887, opened 2015-12-28
Drush make fails sometimes due to Backend invoke limitation


mavimo:
### Env:

PHP 5.6.16 (cli) (built: Dec  5 2015 18:05:12)
### Issue

When I try to build a make file with `drush make myfile.make.yml mydest` I receive the error _No projects passed to drush_make_process_ and the build fail, but only when some modules are in the make file.
### Description

I'm try to build the makefile (`test_ko.make.yml`):

```

---
core: 6.x
api: 2

projects:
  drupal:
    version: '6.37'
  advuser:
    version: '3.0-alpha2'
  autoload:
    version: '2.1'
  content_taxonomy:
    version: '1.0-rc2'
  cck:
    version: '2.10'
  date:
    version: '2.10'
  unique_field:
    version: '1.4'
  computed_field:
    version: '1.0'
  filefield:
    version: '3.13'
  force_password_change:
    version: '3.4'
  google_analytics:
    version: '4.1'
  imageapi:
    version: '1.10'
  imagecache:
    version: '2.0-rc1'
  imagefield:
    version: '3.11'
  jquery_update:
    version: '2.0-alpha1'
  libraries:
    version: '1.0'
  link:
    version: '2.11'
  noreqnewpass:
    version: '2.0'
  path_redirect:
    version: '1.0-rc2'
  pathauto:
    version: '1.6'
  poormanscron:
    version: '2.2'
  print:
    version: '1.19'
  smtp:
    version: '1.1'
    do_recursion: 0
  token:
    version: '1.19'
  token_filter:
    version: '1.0'
  translation_helpers:
    version: '1.0'
  userplus:
    version: '2.5'
  uuid:
    version: '1.0-rc2'
  views:
    version: '3.2'
  views_customfield:
    version: '1.0'
  views_nivo_slider:
    version: '2.0'

defaults:
  projects:
    subdir: 'contrib'
```

that fail with the error:

```
Beginning to build test_ko.make.yml.                                                                                                                                                         [ok]
drupal-6.37 downloaded.                                                                                                                                                                      [ok]
No projects passed to drush_make_process                                                                                                                                                    [error]
Drush Make failed to download all projects. See the log above for the specific errors.                                                                                                       [error]
```

after some debug i found that a similar make file (`test_ok.make.yml`, changed `views_nivo_slider`with `devel` module) complete without issue:

```

---
core: 6.x
api: 2

projects:
  drupal:
    version: '6.37'
  advuser:
[...]

PR #1896, opened 2016-01-04
refs #1887 - Allow long CLI string

mavimo:
Allow receive all stream before decode JSON, see #1887


weitzman:
Lets close this and discuss/experiment in #1887 until we have a solid plan.

PR #1901, opened 2016-01-07
Allow drush finder to find scripts at the cwd.

greg-1-anderson:
Let's see how this fares with the tests.  Might want to hold off on this until 8.0.3.


greg-1-anderson:
This seems pretty good; however, I think it might be a good idea to tag 8.0.2 before merging, so this can live in `master` for a while before going into a tagged release.

weitzman:
Cherry picked after tagging 8.0.2

Issue #1532, opened 2015-08-04
Drush site-set fails for aliases in site specific locations (e.g. sites/all/drush)


alex-mccabe:
#1241 may or may not be related, I'm not sure. In that issue aliases stopped working entirely, in here they only stop working with site-set, but in both cases it breaks at 7.0.0-alpha4 - alpha3 works fine.

Running a command with an explicit alias works:

`$ drush @mysite.local en -y devel`
`devel is already enabled.                                                                                                                                                                                                  [ok]`
`There were no extensions that could be enabled.                                                                                                                                                                            [ok]`

Running site-set (or the alias, use) and then the command without the alias fails:

`$ drush use @mysite.local`
`$ drush en -y devel`
`Command pm-enable needs a higher bootstrap level to run - you will need to invoke drush from a more functional Drupal environment to run this command.                                                              [error]`
`The drush command 'en devel' could not be executed.                                                                                                                                                                 [error]`
`Could not find a Drupal settings.php file at sites/default/settings.php.                                                                                                                                            [error]`

The alias is for a local development site that has its own settings.php in its own directory, `mysite.local/settings.php`. I'm on Ubuntu 14.04 LTS, Drupal 7.38, running PHP 5.5.9-1ubuntu4.11.

weitzman:
Please report bugs using master branch or at least a released version of 7.x

alex-mccabe:
@weitzman I'm not sure I follow. This happens in 7.0.0-alpha4 and higher, *including* 7.0.0, and dev-master.

Or is there something that you're asking me to do here in GitHub to attach it to one of those branches?

weitzman:
Thanks for clarifying.

weitzman:
Can you please provide the output of `drush sa @mysite.local`.

joachim-n:
I am seeing the same problem. Aliases work with commands such as 'status' and 'en' and 'fl', but while site-set reports success, it has no lasting effect.

Here's the requested output for my alias:

```
drush sa @goh
$aliases["goh"] = array (
  'uri' => 'http://localhost:8888/mysite/subsite/',
  'path-aliases' =>
  array (
  ),
);
```

ohthehugemanatee:
I'm having the same problem. Aliases file is in ```sites/all/drush/foo.aliases.drushrc.php``` :

```php
<?php

// Figure out where the vagrant private key is stored (it moved in a recent version)
[...]

PR #1902, opened 2016-01-07
Treat site-set aliases the same way that cli-specified aliases are handled.

greg-1-anderson:
Add an automatic un-set on errors, so that users do not get stuck if a site-local alias is set, and later cannot be found (e.g. if the user cd's out of the site root.

This is my preferred solution to #1532.


weitzman:
This looks pretty sane to me. Travis is happy. Needs some manual testing.

ohthehugemanatee:
Tested working for me.

greg-1-anderson:
Merged. Thanks.

PR #1903, opened 2016-01-07
Try to redo the drush_sitealias_site_get() during bootstrap.

greg-1-anderson:
in case someone is using 'use' with a site-local alias. Note that 'use'-ing site-local aliases breaks horribly as soon as you change your working directory.

This is my alternate proposal for #1532.


greg-1-anderson:
Since all of the tests passed in #1902, I'd say we should go with that & just close this alternative.

PR #1906, opened 2016-01-09
Try to turn off double-testing of Drush PRs on Travis.

greg-1-anderson:
Let's see if it works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,11 @@ php:
 # http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
 sudo: false
 
+branches:
+  only:
+    - master
+    - 7.x
+
 env:
   matrix:
 #D7


greg-1-anderson:
@weitzman: Seems to work. Do we want to still test EOL branches, like 6.x and 5.x?  What about 8.0.x?  Any others?

weitzman:
We had this before (its still in 7.x branch). I think I removed it when messing around with drush.phar deployment. Perhaps the branches directive stopped testing of tags and thus stopped auto-upload of the phar. I'm a bit wary of putting this back.

greg-1-anderson:
Hm. The docs on deploy say that deploy tasks run on tags only, so the branches config should not affect that. The [branches configuration allows regular expressions](https://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches); don't know that that helps us.

Seems like it should work, but up to you whether or not you want to merge it.

weitzman:
Lets leave as is. I ran into issues like travis-ci/travis-ci#2111.

greg-1-anderson:
Okay, so it looks like `branches` inadvertently applies to tags as well.  We could use a regex that matched `N.N.N`, and it would probably work.

Maybe I'll reopen later with an updated patch if I test this on some other repo first.

PR #1921, opened 2016-01-14
Conflicting UUIDs in shortcut.set.default when using --config-dir

jimafisk:
When using the config-dir option on site-install I was receiving the following error:

`Import the listed configuration changes? (y/n): y
Drupal\Core\Config\ConfigImporterException: There were errors validating the config synchronization. in 
Drupal\Core\Config\ConfigImporter->validate() (line 730 of /var/www/SITENAME/core/lib/Drupal/Core/Config/ConfigImporter.php).
The import failed due for the following reasons:
Entities exist of type <em class="placeholder">Shortcut link</em> and <em class="placeholder"></em> <em class="placeholder">Default</em>. These entities need to be deleted before importing.`

Setting the destination site shortcut.set.default UUID to that of the imported config resolved the issue for me.

### DIFF
--- a/commands/core/site_install.drush.inc
+++ b/commands/core/site_install.drush.inc
@@ -251,6 +251,7 @@ function drush_core_site_install($profile = NULL) {
     // Set the destination site UUID to match the source UUID, to bypass a core fail-safe.
     $source_storage = new FileStorage($config);
     drush_op('drush_config_set', 'system.site', 'uuid', $source_storage->read('system.site')['uuid']);
+    drush_op('drush_config_set', 'shortcut.set.default', 'uuid', $source_storage->read('shortcut.set.default')['uuid']);
     // Run a full configuration import.
     drush_invoke_process('@self', 'config-import', array(), array('source' => $config));
   }


weitzman:
Were you installing minimal profile or something else. Please install with minimal and then let the --config-dir overwrite everything.

jimafisk:
I was passing "standard" as the profile argument since the source site was installed that way. When I leave as default (minimal), I receive the following error:

Import the listed configuration changes? (y/n): y
Drupal\Core\Config\ConfigImporterException: There were errors validating the config synchronization. in Drupal\Core\Config\ConfigImporter->validate() (line 730 of /var/www/test/core/lib/Drupal/Core/Config/ConfigImporter.php).
The import failed due for the following reasons:
Unable to install the <em class="placeholder">standard</em> module since it does not exist.
Unable to uninstall the <em class="placeholder">Minimal</em> profile since it is the install profile.

It doesn't seem like the profile listed in core.extension.yml in the config-dir overrides the default site-install profile. Maybe I'm missing something obvious.

dsdeiz:
Same. I was getting this as well. Not sure if it's worth some information but the config I exported was from a fresh install of the standard profile.

> Import the listed configuration changes? (y/n): y
Drupal\Core\Config\ConfigImporterException: There were errors      [error]
validating the config synchronization. in
Drupal\Core\Config\ConfigImporter->validate() (line 728 of
/var/www/web/core/lib/Drupal/Core/Config/ConfigImporter.php).
The import failed due for the following reasons:                   [error]
Unable to install the <em class="placeholder">standard</em> module
since it does not exist.
Unable to uninstall the <em class="placeholder">Minimal</em>
profile since it is the install profile.

weitzman:
I actually think Drush should forcibly set the initial install to minimal when using --config-dir. This feature will not work otherwise

jimafisk:
Agreed, that seems best. Although, is that already happening around line 241?
```
    // If the profile is not explicitly set, default to the 'minimal' for an issue-free config import.
    if (empty($profile) && drush_get_option('config-dir')) {
      $profile = 'minimal';
    }
```
I assume you're referring to overriding the incoming config?

Also want to mention that the config_installer (https://www.drupal.org/project/config_installer) profile seems to help in this situation, although being able to install a new site from existing config directly through drush would be ideal.

Thanks Moshe!

weitzman:
The existing code _defaults_ profile to minimal, but it won't override a profile if supplied. I'm suggesting that we override and provide an 'ok' message stating that minimal was used and that the desired config will still be applied.

danepowell:
This seems similar to https://github.com/acquia/reservoir/issues/47 and https://github.com/acquia/lightning/issues/387

If you are using a custom profile, make sure the config in your profile has UUIDs that match those in your config export

PR #1923, opened 2016-01-15
Console_Table throws deprecated warnings on PHP 7.

pfrenssen:
Whenever I execute a drush command that returns tabular data I get the following notice on PHP 7:

> Methods with the same name as their class will not be constructors in a future version of PHP; Console_Table has a deprecated constructor in Table.php on line 58

Here is the PR that fixes this: [PR #13: Old style constructors are deprecated in PHP 7](https://github.com/pear/Console_Table/pull/13).

We can use this version for the time being until the PR is accepted.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "psy/psysh": "~0.6",
     "symfony/yaml": "2.7.*",
     "symfony/var-dumper": "^2.6.3",
-    "pear/console_table": "dev-master#95607d141c"
+    "pear/console_table": "~1.3.0"
   },
   "require-dev": {
     "phpunit/phpunit": "4.*",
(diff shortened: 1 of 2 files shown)


weitzman:
No commits to tjat repo in 2015 so yeah, a fork probably makes sense. Could anyone verify the bug and its fix?

pfrenssen:
The pull request has been accepted upstream, so there's no need to maintain a fork.

Currently Packagist considers the fork of @RobLoach as the canonical source of Console Table, probably because he initially added composer support to it. So until Packagist is updated we need to specify the URL.

RobLoach:
We need a Console_Table `1.2.2` tag?

yunosh:
You even got a 1.3.0 tag. :-)

composer.json
```diff
@@ -34,7 +34,7 @@
     "psy/psysh": "~0.6",
     "symfony/yaml": "2.7.*",
     "symfony/var-dumper": "^2.6.3",
-    "pear/console_table": "~1.2.0"
```
RobLoach: Can likely target `~1.3.0`, or `^1.2.0` now. Thanks @yunosh! :+1:

pfrenssen:
Awesome, thanks @yunosh!

PR #1924, opened 2016-01-15
Fix #1883. Allow cli overrides in drush_redispatch_get_options() and in @self.

weitzman:


includes/sitealias.inc
```diff
@@ -1705,8 +1707,8 @@ function _drush_sitealias_set_context_by_name($alias, $prefix = '') {
       drush_sitealias_cache_alias_by_path($site_alias_settings);
       if (empty($prefix)) {
         // Create an alias '@self'
-        _drush_sitealias_cache_alias('@self', $site_alias_settings);
-        // change the selected site to match the new --root and --uri, if any were set
+        drush_sitealias_create_self_alias();
```
weitzman: drush_sitealias_create_self_alias() is preferred here since it considers the --uri override provided via CLI as opposed to  _drush_sitealias_cache_alias('@self', $site_alias_settings) which appears not to.

greg-1-anderson: Looks right to me.

includes/command.inc
```diff
@@ -696,9 +696,9 @@ function drush_redispatch_get_options() {
   unset($cli_context['php']);
   unset($cli_context['php-options']);
   // cli overrides and command specific
-  $options = $cli_context + drush_get_context('specific');
+  $options_soup = $cli_context + drush_get_context('specific') + drush_get_context('alias');
   $global_option_list = drush_get_global_options(FALSE);
-  foreach (drush_get_context('alias') as $key => $value) {
+  foreach ($options_soup as $key => $value) {
```
greg-1-anderson: This introduces a minor change in behavior in addition to the bugfix.

Bugfix: corrects priority of options, so cli options override alias options, as they should. This part of the change is good.
Behavior change: Non-global cli options that are NOT defined as part of the command were formerly passed along to the remote side, but now are silently dropped.  This is probably a bad thing.

I'll try pushing another commit here to fix this.

greg-1-anderson:
1c55c17 is causing shellAliasTest.php to fail.  I think the rest of the changes here are okay.

I don't have an alternate solution for the existing problem yet.

greg-1-anderson:
You backed out 8509f33; did you intend to change the behavior here to not include unrecognized options?

Consider:

`drush -s user@server/path/to/drupal#sitename core-topic core-global-options  --uri=othersite --foo=bar`

In the previous implementation, --foo=bar is passed on (perhaps to cause an error), whereas in the current PR is it silently ignored.  There are two potential bad effects here:

1) Option key typos are no longer reported; user runs the remote command without an option they thought they passed in, probably with incorrect results.

2) It is hard to send options to commands that exist on the remote side, but do not exist on the local machine. Not even --strcit=0 will get Drush to pass --foo=bar.

Other than that, getting closer here.

weitzman:
I had to revert that change in order to the new tests to pass. I don't like the side effects you mention though. Not sure when I will have time to get back to this one. Any patch which is green and passes your code review is good enough for me.

weitzman:
I had to start a new PR since I committed a couple parts of this PR that were logically separable. See #1952. Please review there.

PR #1926, opened 2016-01-16
Support color tags in drush_print

Chi-teck:
Since #88 does not seem to happen, I propose we leverage some _Symfony/Console_ subsystems individually. Colorful output may improve developer experience significantly.

Usage: `drush_print('<info>This will be green.</info>');`


weitzman:
We rarely use drush_print(). Use drush_log() instead, which does color code based on severity.

And we do intend to complete #88.

PR #1937, opened 2016-01-25
Allow drush rsync to operate on two remote aliases.

greg-1-anderson:
An oft-requested feature. Handy for migrations.

### DIFF
--- a/commands/sql/sqlsync.drush.inc
+++ b/commands/sql/sqlsync.drush.inc
@@ -263,7 +263,7 @@ function drush_sqlsync_sql_sync($source, $destination) {
       // Cleanup if this command created the dump file.
       $rsync_options['remove-source-files'] = TRUE;
     }
-    $runner = drush_get_runner($source, $destination);
+    $runner = drush_get_runner($source_settings, $destination_settings);
     // Since core-rsync is a strict-handling command and drush_invoke_process() puts options at end, we can't send along cli options to rsync.
     // Alternatively, add options like --ssh-options to a site alias (usually on the machine that initiates the sql-sync).
     $return = drush_invoke_process($runner, 'core-rsync', array("$source:$source_dump_path", "$destination:$destination_dump_path"), $rsync_options);
--- a/commands/sql/sqlsync.drush.inc
+++ b/commands/sql/sqlsync.drush.inc
@@ -44,6 +44,7 @@ function sqlsync_drush_command() {
       // 'no-cache' => 'Do not cache the sql-dump file.',
       'no-dump' => 'Do not dump the sql database; always use an existing dump file.',
       'no-sync' => 'Do not rsync the database dump file from source to target.',
+      'runner' => 'Where to run the rsync command; defaults to the local site. Can also be "source" or "destination".',
       'source-db-url' => 'Database specification for source system to dump from.',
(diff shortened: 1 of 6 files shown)


greg-1-anderson:
Also including here some improvements to make it easier to run sql-sync on Pantheon. Using `core-status` instead of `php-eval` is preferable in any event, because we do not want to send executable php code in our payload. Requires Drush 7 or later on the target system. Could include php-eval as a fallback if you thought it important to maintain compatibility with older versions of the remote Drush.

commands/sql/sqlsync.drush.inc
```diff
@@ -250,9 +250,9 @@ function drush_sqlsync_sql_sync($source, $destination) {
   else {
     $tmp = '/tmp'; // Our fallback plan.
     drush_log(dt('Starting to discover temporary files directory on Destination.'), LogLevel::OK);
-    $return = drush_invoke_process($destination, 'php-eval', array('return drush_find_tmp();'), array(), array('integrate' => FALSE, 'override-simulated' => TRUE));
-    if (!$return['error_status']) {
-      $tmp = $return['object'];
+    $return = drush_invoke_process($destination, 'core-status', array(), array(), array('integrate' => FALSE, 'override-simulated' => TRUE));
```
weitzman: This change looks good. Feel free to cherry pick it ahead of the rest of the PR. If you want, you can force-push to this branch after this change is in.

greg-1-anderson: Cool. I moved the 'core-status' commit to master and rebased this PR as suggested.

weitzman:
Bah - I lost a longer comment on this PR.

1. Today we support both source and destination being remote. We just require that one have a working Drush (the runner). This PR lets neither side have a working Drush. We pretty much abandoned this goal a few years ago. Not sure how valuable it is to add it now. I'm not opposed, just noting that it is not a current goal for sql commands and other parts of Drush
1. Any rsync docs need updating?
1. #remote-rsync needs documenting. also its a bit confusing as an api. it looks like the code arbitrarily looks at the destination and prefers running there in the case when both source and dest have #remote-rsync

greg-1-anderson:
1. The goal here is actually to make it more convenient to sql-sync from remote-to-remote. Today, this requires that the runner (the destination) have an alias for the source. More importantly, this also means that you need to add credentials for the source to the destination machine. This is not always as convenient as just using the local machine (already authenticated to both servers) to act as the intermediary.

2. Yep, some docs will need updating, once this is in the right form.

3. Today, the runner is always on the destination, and there is no way to select the source at all, so this is a bit of an improvement. We could improve things further by adding a --runner option (or some other name) to sql-sync that could be passed through as a parameter to drush_get_runner, to allow folks to select a remote running via an option, without editing the alias records.

weitzman:
1. OK, that makes sense now.
3. --runner sounds good to me Could we then omit #remote-rsync?

greg-1-anderson:
I added --runner, removed #remote-sync, and updated the docs a bit.

weitzman:
Looks good. Please take a look at config-pull to make sure no changes are needed there (also calls into core-rsync). Then ready to merge.

greg-1-anderson:
Added doxygen and merged. config-merge calls drush_get_runner only via config-pull, which was already updated.

PR #1955, opened 2016-01-31
Improve help texts on config-export

Schnitzel:
I'm working on documenting and improving the Drupal 8 configuration management experience, it spoke to some developers and they are sometimes confused with the helper-texts that drush shows, so I took some time and checked how we could improve the experience.

Added some inline comments why I think my changes are better

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -444,10 +444,10 @@ function _drush_config_export($destination, $destination_dir, $branch) {
 
     $config_comparer = new StorageComparer($comparison_source, $target_storage, Drupal::service('config.manager'));
     if (!$config_comparer->createChangelist()->hasChanges()) {
-      return drush_log(dt('There are no changes to export.'), LogLevel::OK);
+      return drush_log(dt('The active config does not differ to the export directory (!target).', array('!target' => $destination_dir)), LogLevel::OK);
     }
 
-    drush_print("The following configuration changes have been made since the last export:\n");
+    drush_print("Differences of the active config to the export directory:\n");
     $change_list = array();
     foreach ($config_comparer->getAllCollectionNames() as $collection) {
       $change_list[$collection] = $config_comparer->getChangelist(NULL, $collection);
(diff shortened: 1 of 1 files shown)


commands/core/config.drush.inc
```diff
@@ -444,10 +444,10 @@ function _drush_config_export($destination, $destination_dir, $branch) {
 
     $config_comparer = new StorageComparer($comparison_source, $target_storage, Drupal::service('config.manager'));
     if (!$config_comparer->createChangelist()->hasChanges()) {
-      return drush_log(dt('There are no changes to export.'), LogLevel::OK);
+      return drush_log(dt('The active config does not differ to the export directory (!target).', array('!target' => $destination_dir)), LogLevel::OK);
     }
 
-    drush_print("The following configuration changes have been made since the last export:\n");
+    drush_print("Differences of the active config to the export directory:\n");
```
weitzman: I'm not sure that this is an improvement. Anyone else have an opinion or suugestion?

commands/core/config.drush.inc
```diff
@@ -462,7 +462,7 @@ function _drush_config_export($destination, $destination_dir, $branch) {
     }
     $comment .= "\n\n$output";
 
-    if (!$commit && !drush_confirm(dt('The .yml files in your export directory (!target) will be deleted.', array('!target' => $destination_dir)))) {
+    if (!$commit && !drush_confirm(dt('The .yml files in your export directory (!target) will be deleted and overwridden with the active config.', array('!target' => $destination_dir)))) {
```
weitzman: Perhaps we dont need this prompt. Its already not shown when --commit is passed. It should always be safe to re-export. Thoughts?

commands/core/config.drush.inc
Schnitzel: See: https://github.com/drush-ops/drush/commit/f7eccff3cac6fe1ef643a496d1ec0f06070f8a4d#commitcomment-15784936
the big Problem I see that the changes are not between the last export and the current export (would imply that we're using the snapchot system), but it is just between the active config and the export directory.

commands/core/config.drush.inc
Schnitzel: I think we need the prompt:
I you're using GIT, yes it is save to reexport, but if you don't track the config files at all, it is not save to export and not prompt the user, as he would loose a config. 
Also we use `drush cex` to shortly have a preview of what would be exported and if this is not exactly what we expect (let's say because there is some configuration that should not be exported), you can still abort right there.

weitzman:
Improved a couple words and committed in 28f3d52

PR #1956, opened 2016-02-01
Fixes #1895: Check for autoload.php rather than index.php 

### LINKED ISSUE #1895: Remove file_exists for index.php in Drupal 8 installations
I'am using the Drupal 8 core package with a Symfony Project, so a don't need the index.php file as front controller, because I use the app.php of Symfony. But drush (\Drush\Boot\DrupalBoot8) exec a file_exists at index.php to validate a D8 installation.

This file is essential to the server command, but not to others, and this file don't define a D8 installation.

I suggest remove this verification and add an option to pass alternative front controller to server command.

The drupalconsole check for autoload.php, and it's the right.

greg-1-anderson:
Apparently, some folks use Drupal 8 with a Symfony front-controller.  Not sure why, but autoload.php is a file that definitely must exist in Drupal 8, so this will do as well as index.php for our purposes.

Seems legitimate to me, anyway; let's see what the tests say.

### DIFF
--- a/lib/Drush/Boot/DrupalBoot8.php
+++ b/lib/Drush/Boot/DrupalBoot8.php
@@ -19,7 +19,7 @@ class DrupalBoot8 extends DrupalBoot {
   protected $request;
 
   function valid_root($path) {
-    if (!empty($path) && is_dir($path) && file_exists($path . '/index.php')) {
+    if (!empty($path) && is_dir($path) && file_exists($path . '/autoload.php')) {
       // Additional check for the presence of core/composer.json to
       // grant it is not a Drupal 7 site with a base folder named "core".
       $candidate = 'core/includes/common.inc';


greg-1-anderson:
Tests are green. @weitzman: does this change look okay to you?

weitzman:
https://www.drupal.org/node/1475510 is very close to commit. Once thats in, there will be no autoload.php in a git download of Drupal. And if folks use drupal-project, then the autoload.php will be one dir up from Drupal root. So we should use a different file or different heuristic IMO.

greg-1-anderson:
https://www.drupal.org/node/1475510 does remove vendor/autoload.php, but it does not remove the autoload.php from the Drupal root, which is the one we are testing here.

Similarly, drupal-composer/drupal-project relocates vendor/autoload.php to ../vendor/autoload.php, but it does not remove the autoload.php from the Drupal root -- it rewrites it to point at the new location.

As near as I can tell, the Drupal 8 convention of including the autoload.php at the project root as a way to find the "real" autoload.php is not changing.

weitzman:
OK, makes sense.

PR #1959, opened 2016-02-01
Add a new log level, DEBUG_NOTIFY. 

greg-1-anderson:
Messages logged at this level are only displayed when --debug and --verbose are used together.

--verbose:  Prints verbose log messages (e.g. 'notice'), but not debug.

--debug:  Prints verbose AND debug log messages (e.g. 'notice' and 'debug'), but does not print LogLevel::DEBUG_NOTIFY messages.

--debug --verbose:  Like --debug, but also prints new LogLevel::DEBUG_NOTIFY messages.

With this PR, only 'preflight' and the site-alias 'Scanning into' messages are moved to LogLevel::DEBUG_NOTIFY. Even with this small change, I like --debug output much better, as a lot of noise that is usually not interesting is removed.

We will have a minimal impact of occasionally having to ask folks to run with `-dv` for extra debugging info, but this should be very rare. It's unlikely that we'll ever need the 'preflight' messages, and the alias-scanning stuff is only useful when debugging site alias usage.  In general, I think it will be very helpful to have this extra logging level, as we can start to add in info that is useful only in edge cases, without making ordinary debug logs arduously long.

Note that I also remove DEBUG_NOTIFY (but not PREFLIGHT) log message from inclusion in the backend invoke packet log.

### DIFF
--- a/includes/preflight.inc
+++ b/includes/preflight.inc
@@ -355,9 +355,12 @@ function drush_preflight_site() {
 
 function _drush_preflight_global_options() {
   // Debug implies verbose
-  drush_set_context('DRUSH_VERBOSE',     drush_get_option(array('verbose', LogLevel::DEBUG), FALSE));
-  drush_set_context('DRUSH_DEBUG', drush_get_option(LogLevel::DEBUG));
-  drush_set_context('DRUSH_SIMULATE',    drush_get_option('simulate', FALSE));
+  $verbose = drush_get_option('verbose', FALSE);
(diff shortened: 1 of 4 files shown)


weitzman:
Looks good to me. These changes are all backwards compatible? If so, feel free to merge if green.

greg-1-anderson:
Yes, the changes are backwards compatible, except for the ~couple of log messages that are dropped, which shouldn't affect anything. We'll see if the tests reveal any edge-cases, though.

PR #1961, opened 2016-02-02
Fixes #1845: Report bootstrap errors that may arise in bootstrap_max()

### LINKED ISSUE #1845: Consistent behavior/error if missing settings.php
When trying to run most Drush commands and a `settings.php` file does not exist, Drush prints out a helpful error message:

`"Could not find a Drupal settings.php file at !file."`

https://github.com/drush-ops/drush/blob/3657fa99a9c47e4684786ccac71270b2db27615e/lib/Drush/Boot/DrupalBoot.php#L395

But when performing `drush cc`, a different (and arguably less helpful) error message is printed:

`"No Drupal site found, only 'drush' cache was cleared."`

https://github.com/drush-ops/drush/blob/3657fa99a9c47e4684786ccac71270b2db27615e/commands/core/cache.drush.inc#L146

I think it would improve the Drush experience to have consistent behavior in these circumstances.

greg-1-anderson:
This PR delays the reporting of bootstrap error until Drush is terminating, instead of reporting bootstrap errors during initialization. This makes it possible to show bootstrap errors if any arise during command execution in a call to `drush_bootstrap_max()`.

It also includes a change to drush cc to ensure that `drush_bootstrap_max()` is called if the user is trying to clear any cache other than just the `drush` cache.

### DIFF
--- a/commands/core/cache.drush.inc
+++ b/commands/core/cache.drush.inc
@@ -140,14 +140,17 @@ function drush_cache_command_clear($type = NULL) {
       return drush_user_abort();
     }
   }
+  // Require a full bootstrap unless we are just clearing the Drush cache.
+  // This will give more consistent error messages when a site cannot be found.
+  if ($type != 'drush') {
+    drush_bootstrap_max(); // bootstrap quietly - drush_bootstrap may fail ungracefully
+    if (!drush_has_boostrapped(DRUSH_BOOTSTRAP_DRUPAL_FULL)) {
+      return drush_set_error('DRUSH_CC_COULD_NOT_BOOTSTRAP', dt("The cache could not be cleared, because the Drupal site could not be bootstrapped."));
+    }
+  }
   // Do it.
   drush_op($types[$type]);
-  if ($type == 'all' && !drush_has_boostrapped(DRUSH_BOOTSTRAP_DRUPAL_FULL)) {
-    drush_log(dt("No Drupal site found, only 'drush' cache was cleared."), LogLevel::WARNING);
-  }
-  else {
-    drush_log(dt("'!name' cache was cleared.", array('!name' => $type)), LogLevel::SUCCESS);
-  }
+  drush_log(dt("'!name' cache was cleared.", array('!name' => $type)), LogLevel::SUCCESS);
 }
 
 /**
--- a/commands/core/cache.drush.inc
+++ b/commands/core/cache.drush.inc
@@ -143,7 +143,7 @@ function drush_cache_command_clear($type = NULL) {
   // Require a full bootstrap unless we are just clearing the Drush cache.
   // This will give more consistent error messages when a site cannot be found.
   if ($type != 'drush') {
-    if (!drush_has_boostrapped(DRUSH_BOOTSTRAP_DRUPAL_FULL)) {
+    if (!drush_require_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_FULL)) {
       return drush_set_error('DRUSH_CC_COULD_NOT_BOOTSTRAP', dt("The cache could not be cleared, because the Drupal site could not be bootstrapped."));
     }
   }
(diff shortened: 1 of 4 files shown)


weitzman:
The cache change makes sense.

My hesitation about the rest is that user expects log messages to be printed in chrono order. This seems relatively safe, but I'll pause and get your opinion on the potential for out of order messages.

greg-1-anderson:
In the case of existing failures that are handled correctly, the log message order does not change. The message that was normally printed during bootstrap will also cause command execution to be skipped, so the log message will be printed out at about the same time, right before exit.

In the case of new failures, where the messages previously were not displayed at all, commands will typically (read: should always) exit as soon as drush_bootstrap_max() fails. So, the only messages that will be out of order will be the final drush_set_error() message that the command prints before bailing, which comes out before the bootstrap errors.

There are two possible solutions to the ordering problem:

1) Have commands `return FALSE` without calling `drush_set_error()` at all. This would work brilliantly, except in cases where `drush_bootstrap_max()` fails without ever setting any bootstrap errors, in which case the result is a dissatisfying silent failure. Probably rare, but definitely not good.

2) Have drush_set_error no longer call through to drush_log(), and instead, queue up the message to be printed during terminate.  We'd also have to adjust the way errors are handled with backend invoke. Deferring the printing of error messages seems easy enough to me, but I'm a little concerned about the changes to backend invoke.

3) Expose the set bootstrap error function to commands (a policy, not technology change), and expect commands to call that instead of drush_set_error if they fail after drush_bootstrap_max().

I think that (3) works.  I didn't think of it until I typed up (1) and (2).

weitzman:
I'm not sure its an improvement to have command authors use different error functions depending on bootstrap completion. Is that your 3)? I know that few commands use MAX but still a complexity concern.

greg-1-anderson:
You are right, (1) - (3) are all wrong.  The thing is, this function should be using a regular bootstrap_full here, not a max. It's only using max because full terminated improperly. I need to go back and fix full rather than work around it.

greg-1-anderson:
I added another commit that put the reporting of bootstrap errors in BaseBoot back to the point in the code they used to be (but still factored into a separate function). Also, I kept the code in terminate() that displays unreported bootstrap errors on exit, with a warning message alerting the user to the fact that these errors are out-of-order.  Next I will work on fixing `cc all` so that the bootstrap errors will be reported in order (via bootstrap full), so that the out-of-order message does not appear. However, I think we should keep the behavior of terminate() like this, in case some other code fails after a failed bootstrap_max.  I think it's better to see these messages out of order than never see them at all. Let me know what you think about this, though -- the negative aspect here is that if someone is using a bootstrap-max command in a context where it is expected that it should not be possible to bootstrap the site (e.g. when working with an uninstalled site using a command that does not require database access), then you might occasionally see spurious bootstrap errors when a command fails for an unrelated reason.

Hypothetical example:

Could not write to foo.yml: permission denied [error]
Unreported bootstrap errors: [warning]
Could not bootstrap for whatever reason [error]

This should be rare, as it will only happen for bootstrap max commands. We can think about it a bit, anyway; I'm going to try to fix the bootstrap full error in cache-clear now.

greg-1-anderson:
Actually, it gets interesting in that drush_bootstrap_max() was a no-op here, because the cache-clear command is already a bootstrap-max command.  Drush has already tried to bootstrap as high as it can, so there is no point in calling bootstrap max or bootstrap full here.

We need a different API to allow commands to exit cleanly with appropriate error message if bootstrap max does not go far enough for the command's needs.

greg-1-anderson:
Introduced `drush_require_bootstrap()` to straighten this all out.  This should work fine.

greg-1-anderson:
Perhaps I should have named this drush_require_**boo**strap(), though.

greg-1-anderson:
@weitzman - believe this is working cleanly now.

weitzman:
This still doesn't feel especially clean to me. 

I wonder if we shouldn't move `cc drush` to its own new `flush` command. `flush` could have various levels of clearing for things like cached tarballs, release xml, etc.

All existing cache clear types stay in cache.drush.inc. which would be solely owned by Drupal and can require FULL bootstrap for all commands. As a deprecated feature, this file could still implement `cc drush` but that might not work in all situations since FULL would be required.

greg-1-anderson:
That sounds sort of okay for `drush cc drush`, but I'd like to take a moment to survey the other bootstrap max commands, and see if they are also going to be okay.

Issue #1820, opened 2015-11-30
Dependencies are not skipped when config-export --skip-modules is used


gaelg:
If one skips some module (let's say Devel) when exporting configuration, the dependent config entities won't be skipped (you'll still get a devel.settings.yml file in the export). And this can break the import in another instance of the same site, because of dependency validation. The error will come from https://api.drupal.org/api/drupal/core!lib!Drupal!Core!EventSubscriber!ConfigImportSubscriber.php/function/ConfigImportSubscriber%3A%3AvalidateDependencies/8

nevergone:
Confirmed. :(

nevergone:
drush config-import also affected.

thenetexperts:
still a problem in 8.0.2, for now one could add the dependent config files to a .gitignore until solved

nevergone:
This issue is independent from git.

thenetexperts:
yes of course. I was merely pointing out that, in the meantime, you could add the created dependent configuration .yml files (for example devel.settings.yml) to a .gitignore until this issue is resolved.

gaelg:
I tried to work on this, but it's harder than I expected. The storage filters are called with source config objects only, so that there is no way to filter out a destination config object using storage filters. I'll try to find another way.

gaelg:
Here's what I did today: https://github.com/gaelg/drush/commit/44887b11bada471f0fe90a6591abbe7d54c47053
Any help would be very appreciated, specially to improve dependency detection and to make tests.

gaelg:
I improved dependency detection and fixed a bug.
system.menu.devel is still not ignored, but this is because Devel should enforce the dependency for this config (https://www.drupal.org/node/2404447).
It seems to work well. Should I make a pull request, or should this still be improved before any pull?

greg-1-anderson:
Thanks for working on this. I haven't had a chance to review the referenced branch yet, but it is always okay to make a work-in-progress PR.

gaelg:
It would be nice if someone in this issue queue could write up tests for this. I'll do it when I get time otherwise.
The tests could be:
For export :
* install devel on a website, drush cex --skip-modules=devel, ensure that devel.settings.yml is not created
* do it a second time, now that sync dir is not empty, ensure the same
* website without devel, sync dir with devel.settings.yml, drush cex --skip-modules=devel, ensure devel.settings.yml still exists
[...]

PR #1964, opened 2016-02-02
Skip dependencies on config import/export when modules are skipped (--skip-modules)

gaelg:
See https://github.com/drush-ops/drush/issues/1820


bircher:
The approach here is very much centred around adding extra functions.
I think it would be possible to take care of the dependency filtering in the `\Drush\Config\CoreExtensionFilter` with a little help from `\Drupal\Core\Config\ConfigManagerInterface::findConfigEntityDependents`

See https://github.com/nuvoleweb/config_split/blob/8.x-1.x/src/Config/SplitFilter.php#L62 (though maybe not the perfect example, then the `CoreExtensionFilter` just needs to ignore the blacklisted configuration.

gaelg:
I rewrote this from scratch using bircher's advice. The code should be much better now, as it reuses existing code when possible (dependency detection, modules sorting).
I'd love to have feedback on this so that I can work on it during the DrupalCon Dublin sprint tomorrow, for it to be hopefully committed.

weitzman:
@greg-1-anderson would be the person to review this. I will say that I have been resistant to this feature existing in Drush core. I'd rather not see it grow in complexity within core. Having said that, I defer to Greg's judgement on this PR.

gaelg:
Allright, thank you for the quick feedback! It would have been better if this issue had been closed sooner if it's a "won't fix", but anyway I learnt much by trying to contribute to this.
IMHO the feature is "skip-modules" (already in drush core) and this PR is more a bugfix of the feature.
If someone tries to just use skip-modules right now, its config will never import. It needs a good .gitignore file referencing any dependency. If it's only for Devel, it's easy, but when you're used to have many dev modules enabled, it can be harder.
For various reasons, it might indeed be relevant to have this out of core. But then maybe the whole skip-modules feature should be out, no? At least if would mean some changes to core to make it extensible.

robertoneshoe:
@gaelg, I've tested your pull request, but there's still some unwanted config being exported. For example, when Devel is enabled, system.menu.devel is still being exported.

gaelg:
@robert-os Yes, it's known problem from the Devel module that did not correctly declare the dependency. But it's solved in the latest dev version of Devel.

weitzman:
This looks like a useful improvement. Would be great if someone found time to review it.

bircher:
I looked over it, and I like the direction this is going in.
It is a good idea to make the source storage available to more than just one filter operation. I found out that it is also useful to have the wrapped storage available. So instead of passing it as an argument to all filter options why not give it to the filter at the start.

You could make your life a lot easier if you would base this PR on top of #2313. My estimate would be around 25 lines of code changes to filter dependent configuration.

In addition it would be good to have some automated tests.

weitzman:
Please work with config_split project. --skip-modules is moving out of Drush and into there.

PR #1974, opened 2016-02-04
Document install alternative for all users via composer

christopher-hopper:
This change explains how to do a single, common drush install for all users and all projects on a system using composer. I believe this documentation better explains how that works and gives a repeatable, simple set of steps that are robust and could be followed by anyone.

Also included is a fix for the Windows dot-points which had become munged due to end-of-line characters being lost at some point.

### DIFF
--- a/docs/install-alternative.md
+++ b/docs/install-alternative.md
@@ -75,8 +75,12 @@ Windows
 ------------
 Drush on Windows is not recommended, since Drush's test suite is not running there ([help wanted](https://github.com/drush-ops/drush/issues/1612)).
 
-* [Acquia Dev Desktop](https://www.acquia.com/downloads) is excellent, and includes Drush. See the terminal icon after setting up a web site.
-* Or consider running Linux/OSX via Virtualbox. [Drupal VM](http://www.drupalvm.com/) and [Vlad](https://github.com/hashbangcode/vlad) are popular.* These Windows packages include Drush and its dependencies (including MSys).     * [7.0.0 (stable)](https://github.com/drush-ops/drush/releases/download/7.0.0/windows-7.0.0.zip).    * [6.6.0](https://github.com/drush-ops/drush/releases/download/6.6.0/windows-6.6.0.zip).    * [6.0](https://github.com/drush-ops/drush/releases/download/6.0.0/Drush-6.0-2013-08-28-Installer-v1.0.21.msi).
-* Or install LAMP on your own, and run Drush via [Git's shell](https://git-for-windows.github.io/), in order to insure that [all depedencies](https://github.com/acquia/DevDesktopCommon/tree/master/bintools-win/msys/bin) are available.   
-* Whenever the documentation or the help text refers to `drush [option] <command>` or something similar, 'drush' may need to be replaced by 'drush.bat'.
-* When creating site aliases for Windows remote machines, pay particular attention to information presented in the example.aliases.drushrc.php file, especially when setting values for 'remote-host' and 'os', as these are very important when running Drush rsync and Drush sql-sync commands.
+- [Acquia Dev Desktop](https://www.acquia.com/downloads) is excellent, and includes Drush. See the terminal icon after setting up a web site.
+- Or consider running Linux/OSX via Virtualbox. [Drupal VM](http://www.drupalvm.com/) and [Vlad](https://github.com/hashbangcode/vlad) are popular.
+- These Windows packages include Drush and its dependencies (including MSys).
+    - [7.0.0 (stable)](https://github.com/drush-ops/drush/releases/download/7.0.0/windows-7.0.0.zip).
+    - [6.6.0](https://github.com/drush-ops/drush/releases/download/6.6.0/windows-6.6.0.zip).
+    - [6.0](https://github.com/drush-ops/drush/releases/download/6.0.0/Drush-6.0-2013-08-28-Installer-v1.0.21.msi).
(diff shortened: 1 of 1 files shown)


christopher-hopper:
I've reviewed the tests failed above and the seem unrelated to the changes. It looks like a connectivity problem during the test run where it was unable to download a library or something. I'll look into how to re-run the tests, unless someone else needs to do something to kick it off again.

weitzman:
Assigning to @jonhattan or @greg-1-anderson since they last touched these instructions. 

Note that following the .phar install instructions gives you a (slightly limited) copy of drush for all users.

christopher-hopper:
@jonhattan, @greg-1-anderson, 
Review of these documentation changes would be appreciated.

christopher-hopper:
Re-rolled against master. Still awaiting comment.

I had thought it might be worth including a summarised version of the instructions I've added, without the explanation, as it's really only three commands that someone might want to copy and paste into a terminal all at once. 

Thoughts?

docs/install-alternative.md
```diff
@@ -17,19 +17,81 @@ Follow the instructions below, or [watch a video by Drupalize.me](https://youtu.
...
+        sudo composer install
+
+    The `composer.json` file tells composer to install (or update) drush at the [package version][composer package version] we specified and to put the [vendor binaries][composer vendor binaries] where all users can access them.
+
+1. Lastly, if you are using bash with command completion you can enable drush completion for all users by simply symlinking the `drush.complete.sh` shell script into the correct location.
+
+        sudo ln -s /usr/local/bin/drush.complete.sh \
+          /etc/bash_completion.d/drush
+
+**Important Tip:** When installing *drush 6.x* there are *dependencies not managed by composer* that will require download on first run. You
```
greg-1-anderson: Drush 6 is no longer supported. Should we still document the installation procedure for it?

docs/install-alternative.md
```diff
@@ -17,19 +17,81 @@ Follow the instructions below, or [watch a video by Drupalize.me](https://youtu.
...
-
 * [Documentation for composer's require command.](http://getcomposer.org/doc/03-cli.md#require)
 * Uninstall with : `composer global remove drush/drush`
 
+Install Drush for all users via Composer
+------------
+If you need a common Drush install available for all users on a system, follow the composer install steps below.
+
+1. [Install Composer globally][composer install global].
+1. Create and/or navigate to a directory path for the single composer drush install. In this example we are going to install drush version 7.x
```
greg-1-anderson: Which version should we recommend here?  Drush 9 may go through periods of instability in the near future.  Drush 7 is two versions back now, and will probably get fewer backported bugfixes.  Maybe Drush 8 is best.

docs/install-alternative.md
```diff
@@ -17,19 +17,81 @@ Follow the instructions below, or [watch a video by Drupalize.me](https://youtu.
...
+
+After composer update completes, the binaries and library will be updated for all users.
+
+### Major Version Upgrade
+
+If upgrading to a new major version the steps are the same, with one addition:
+
+1. Remove the existing symlinks to drush [vendor binaries][composer vendor binaries]. From our example above these would be located in the `bin-dir` path `/usr/local/bin`:
+
+        sudo rm -i /usr/local/bin/drush*
```
greg-1-anderson: I wish Composer had a command to uninstall the items it added to the bin directory. Beyond this one shortcoming, I am overall positively inclined towards these instructions.

Some folks advise against `sudo drush`, and `sudo composer` has similar characteristics vis-a-vis custom installers. However, the basic procedure here involves installing code to a directory that is the the global PATH for all users, so I suppose that adding `sudo` does not add a lot of extra exposure over that. It would complicate the instructions a lot recommend creating a `composer` user, making a separate Composer bin directory, adding that to the PATH, etc. etc., so it's probably best to stay with the current form.

Is `sudo composer` and the global bin technique described being used elsewhere now? The comments below imply that perhaps this is becoming a general use pattern.

weitzman:
1. Would be nice to make these directions shorter. I think install option this is an edge case.
1. In almost all cases we should begin the project name with a capital letter - Drush not drush

docs/install-alternative.md
christopher-hopper: While these steps require a privileged user to write to the paths `/opt` and `/usr/local/bin` during install, the resulting `drush` command can be run by any user without the need for elevated privileges via `sudo`.

I've not experienced issues when using `sudo composer` commands. I don't recommend the use of `sudo drush` though. I can't imagine that being good or necessary.

christopher-hopper:
@greg-1-anderson, @weitzman 
I've updated this pull to address your comments and include suggested improvements. I think this is now simpler and I've summarized to just the commands at the start.

weitzman:
I really want a few lines of code and comments.

christopher-hopper:
I've collapsed this down now into brief, commented code snippets as requested.

christopher-hopper:
@weitzman 
I've removed all the optional stuff, stripped it down to just the commands, with comments above each to explain their purpose. 

It reads a lot cleaner. I see what you were getting at.

christopher-hopper:
Thanks for helping me to get the documentation right. 

I'd like to clean-up the Windows Install Alternatives docs now as Drush on Windows will be easier now Microsoft have started shipping with a Linux Bash command-line. Will explain in #1612 

A little sad about the commit attribution, but hey, I know you're busy.

weitzman:
I squashed the commits and then missed --author. Thanks for the reminder, and thanks for the attention to docs.

christopher-hopper:
No problem. Thanks for being a supportive maintainer who encourages participation.

PR #1975, opened 2016-02-04
Respect rsync-version on sql-sync

### LINKED ISSUE #758: rsync-version not picket up from alias file
The example.drushrc.php file says this:

```php
/**
 * Drush requires at least rsync version 2.6.4 for some functions to work
 * correctly.  rsync version 2.6.8 or earlier may give the following error
 * message: "--remove-source-files: unknown option".  To fix this, set
 * $options['rsync-version'] = '2.6.8'; (replace with the lowest version of
 * rsync installed on any system you are using with Drush).  Note that this
 * option can also be set in a site alias, which is the preferred solution if
 * newer versions of rsync are available on some of the systems you use.
 * See: http://drupal.org/node/955092
 */
```

I have a live server with rsync 2.6.8 and I tried unsuccessfully to add it to the live alias. I tried adding it to dev (the other alias) as well and nothing.

I added it to the $alias['live'] directory, then to specific-comments key, then also just as plain $options['rsync-version'] and nothing. The only way this worked for me was adding it to a drushrc.php file, but that makes it global to drush.

Any ideas?

hanoii:
Attempt to fix #758.

### DIFF
--- a/commands/sql/sqlsync.drush.inc
+++ b/commands/sql/sqlsync.drush.inc
@@ -80,6 +80,7 @@ function sqlsync_drush_command() {
       ),
       // 'no-ordered-dump' => 'Do not pass --ordered-dump to sql-dump.  sql-sync orders the dumpfile by default in order to increase the efficiency of rsync.',
       'sanitize' => 'Obscure email addresses and reset passwords in the user table post-sync.',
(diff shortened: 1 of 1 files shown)


weitzman:
We need to solve this in general, not just for one options. See #842

hanoii:
OK, although this pull request had a more meaningful contribution that was not related to commands propagation, but rather to the fact that rsync-version was not taken into account when doing sql-sync.

hanoii:
I  attempted another pull request with only the fix for sql-sync.

PR #1976, opened 2016-02-04
Make sure rsync-version handle --remove-source-files all the time

hanoii:
This is another try for #1975. It was related to #758 and closed thought as not addressing #842.

I believe this is a separate fix that sorts something important on #758 which is not command propagation. I have reverted the commit that adds a command while #842 is being fixed, but the fix for sql-sync is still valid I believe.

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -286,7 +286,7 @@ function core_drush_command() {
       'include-paths' => 'List of paths to include, seperated by : (Unix-based systems) or ; (Windows).',
       '{rsync-option-name}' => "Replace {rsync-option-name} with the rsync option (or option='value') that you would like to pass through to rsync. Examples include --delete, --exclude=*.sql, --filter='merge /etc/rsync/default.rules', etc. See the rsync documentation for a complete explanation of all the rsync options and values.",
       'rsync-version' => 'Set to the version of rsync you are using to signal Drush to go into backwards-compatibility mode when using very old versions of rsync. For example, --rsync-version=2.6.8 or earlier will cause Drush to avoid the --remove-source-files flag.',
-
+      'remove-source-files' => 'If possible, rsync removes synchronized files (non-dir) at sender.',
     ),
     'strict-option-handling' => TRUE,
     'examples' => array(
--- a/commands/core/rsync.core.inc
+++ b/commands/core/rsync.core.inc
@@ -184,21 +184,25 @@ function _drush_build_rsync_options($additional_options, $include_settings_is_de
     $options .= ' --stats --progress';
   }
 
-  // Check if the user has set $options['rsync-version'] to enable rsync legacy version support.
-  // Drush was written for rsync 2.6.9 or later, so assume that version if nothing was explicitly set.
-  $rsync_version = drush_get_option(array('rsync-version','source-rsync-version','target-rsync-version'), '2.6.9');
+  if (_drush_rsync_option_exists('remove-source-files', $additional_options)) {
+    // Check if the user has set $options['rsync-version'] to enable rsync legacy version support.
+    // Drush was written for rsync 2.6.9 or later, so assume that version if nothing was explicitly set.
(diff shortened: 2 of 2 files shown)


weitzman:
Even this part should not be necessary. What you should be doing is adding rsync-version to the command-specific['rsync'] section of your site alias. Then #842 could potentially solve it. I'm not willing to complicate sql-sync for special handling of rsync-version.

commands/sql/sqlsync.drush.inc
```diff
@@ -262,7 +262,19 @@ function drush_sqlsync_sql_sync($source, $destination) {
   if ($do_rsync) {
     if (!drush_get_option('no-dump')) {
       // Cleanup if this command created the dump file.
-      $rsync_options['remove-source-files'] = TRUE;
```
hanoii: sql-sync adds this regardless or any rsync-version option, it doesn't really matter where it is set, it's simply not taken into account, this code is simply adding this option and passed along to drush_invoke_process() slightly below.

Then when you get to the rsync part of the command, because of:

```php
  $options = _drush_build_rsync_options($additional_options, $include_settings_is_default);

  // Get all of the args and options that appear after the command name.
  $original_args = drush_get_original_cli_args_and_options();
  foreach ($original_args as $original_option) {
    if ($original_option{0} == '-') {
      $options .= ' ' . $original_option;
    }
  }
```

This is adding --remove-source-files again to the rsync command which fails.

The only thing this patch does, is an attempts to emulate what happens on `_drush_build_rsync_options` in regard to this specific rsync setting.

hanoii:
If this is not an acceptable fix, that's fine, but I am pretty sure #842 will not fix this unless the way `remove-source-files` is being added is reworked as well. This is a command-specific issue, not a matter of alias settings.

weitzman:
Thanks for clarifying. I missed that sql-sync was "misbehaving"  by unconditionally adding --remove-source-files. 

Any chance we can refactor to that this shares some code with rsync command and keeps the ugliness contained in a new function there?

hanoii:
I am looking into refactoring this a bit better, but I am confused on a few things, because some things does't look right, so maybe you can help me in working this the best way possible.

If we look at rsync on its own, you have:

```php
function drush_core_rsync($source, $destination, $additional_options = array()) {
```

Then inside that function you have 

```php
  $options = _drush_build_rsync_options($additional_options, $include_settings_is_default);
```

And that function is the one that holds, for now, the logic that builds the option arguments and it expects $additional_options to override things, and there's where it expects remove-source-files and have to logic to change it.

However, I can't seem to understand where $additional_options could be filled is coming from. I don't see it anywhere in the command flow and not be able to build up regardless of what I do. I project-wide search for a call to it, as it seems it might be just an argument with the possibility of executing this command through an internal function call, but I found none.

I will still attempt to provide at least a better patch, but this smells something carried over from old code and not properly use somehow.

```php
  $options = _drush_build_rsync_options($additional_options, $include_settings_is_default);
```

Works fine because it defaults to things if a certain settings are not on $additional_options.

hanoii:
Well, after a few hours this am I kind of happy with the proposed solution. I reworked this two times in different ways.

I am adding remove-source-files as an actual drush command. I think it makes more sense as it has its own logic. By doing that, It's handled by drush and not pass-through. I kept the logic in the same place, but move it above and treat as as any other drush option that could be overridden in the same (odd) way with `$additional_options`.

Hope you like it!

weitzman:
I'm not sure when I will have time to review this. One complication is that people sometimes use different drush versions on each side of a sql-sync and such. Ideally stuff keeps working in this case.

hanoii:
It doesn't seem that this will affect that. I personally am using this on my local dev against a rather old drush version on the remove server with no issue.

hanoii:
Because I spotted a small issue when using -v due to spaces not being added, and I rebase the codee against latest changes just checking in to see if there's any time available to review this.

greg-1-anderson:
This has been idle for a long time, and old rsync versions just keep getting older. Closing.

PR #1987, opened 2016-02-10
startup.inc printenv: Fixed drush_env obtaining the environment variables through `exec('printenv', $env_items)`

R-eenignE:
Fixed `drush_env` obtaining the environment variables through `exec('printenv', $env_items)` so that `$env_items` properly contains bash functions. Each newline in a function creates another index in the `$env_items` array which deviates from the expected key=value format. There is _always_ at least one newline in printenv function output, even if the function exported was all on one line, printenv will place the closing `}` on another line.

Coreutils `printenv` technically contains a `-0` or `--null` option which would separate keys by ASCII NUL characters instead of newlines to address this, though not all printenv implementations (ex: Mac OS) have this option available. Below is the options excerpt from `info printenv`

```
‘-0’
‘--null’
     Output a zero byte (ASCII NUL) at the end of each line, rather than
     a newline.  This option enables other programs to parse the output
     even when that output would contain data with embedded newlines.
```

Note there is still a possibility that variable values in addition could contain embedded newlines. Also the parsing looks for a line matching the regex `/^}$/` which may be possible to see before the actual end of a function depending on the function code formatting. I have not been able to cause a closing bracket to be on a line of its own before the function closure in the printenv output with a number of attempts I made; in each of those tests this update worked as expected.

### DIFF
--- a/includes/startup.inc
+++ b/includes/startup.inc
@@ -30,6 +30,21 @@ function drush_env() {
   // variable values using 'printenv'.
   if (empty($env)) {
     exec('printenv', $env_items);
+
+    // Exported bash functions contain newlines in the
+    // printenv output. If any are present this causes
+    // the array items to no longer contain key=value
+    // with new indices for every line in the function.
+    // Linux printenv has a flag -0 which delimits vars
+    // by an ASCII NUL instead of newlines to account
+    // for this problem, though other systems (OS X)
+    // printenv do not have this option. We filter
+    // the resulting array for any functions, concat
+    // each into a contiguous string and merge the
+    // resulting array; the end result having bash
+    // functions take up one array index each.
+    $env_items = drush_env_fix_bash_functions($env_items);
+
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
It's been a long time with no activity here. Sorry I didn't see the assignment a few years ago.

This is only needed for users who have not set `variables-order` to include environment variables. At this point in time, I am almost more inclined to say that such users simply should not get environment variables. That would be a breaking change, though (and in fact this feature is gone in Drush 9).

If this is still needed, we could revisit. I'm not sure why `printenv` should be printing bash functions. If we were to revisit, what I would suggest doing instead is run `printenv`, look at the first "word" of each line, if it is `var=` then use getenv to get the value of `var`. If there is an environment variable with that name, then store its value in  our `$env` array; otherwise, ignore it as junk.

Please open a new PR against the 8.x branch if interested in continuing.

PR #1988, opened 2016-02-10
--uri override fix broke Pantheon aliases.

greg-1-anderson:
Roll back part of #1883 / 70752cf: breaks remote aliases that do not set 'root'.

I did one simple test with --uri on the commandline, and found it was still overriding the 'uri' element set in the alias record.  Might need more testing, to insure that my fix does not break something else.  First step: see how the travis tests fare.

### DIFF
--- a/includes/sitealias.inc
+++ b/includes/sitealias.inc
@@ -88,15 +88,15 @@ function drush_sitealias_check_site_env() {
  * Check to see if a '@self' record was created during bootstrap.
  * If not, make one now.
  */
-function drush_sitealias_create_self_alias($fallback_site_record = array()) {
-  $drupal_root = drush_get_context('DRUSH_SELECTED_DRUPAL_ROOT');
-  $uri = drush_get_context('DRUSH_SELECTED_URI');
-  // Create an alias '@self'
-  if (!empty($drupal_root) && !empty($uri)) {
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Yeah, the unit test for this is showing --uri not being passed on. Need to fix that.

greg-1-anderson:
@weitzman: Please review. I don't really like ed68920 very much, but I prefer it to the behavior of  70752cf, where is is unclear how the selected root and uri were set.

I'll try to generate a test to catch the situation where a remote alias has no 'root', and the cwd is inside a Drupal site. Then, I'd recommend we commit this (if you don't see any problems with it), and fix it better in the alias rewrite.

greg-1-anderson:
The tests that I added succeed with this PR, fail on master (remote command becomes a local command) and fail on 8.0.2 (remote command is still a remote command, but --uri override not fixed yet).

I'm going out of town for a few days, and don't want to leave Pantheon aliases broken on master, so I'm going to go ahead and merge this in. We can continue to discuss and adjust in follow-on issues as necessary.

mhumeSF:
Is this saying that if a uri value is set in a drush alias, is the behavior supposed to that one cannot override that hardcoded value with `--uri=site2.com`?

greg-1-anderson:
This is an old PR. I believe that overriding the URI set in a site alias with --uri is presently working on 8.x and master, but didn't test. Please open a new issue if you have problems.

Issue #1759, opened 2015-11-09
make with custom themes in info file fails


fuzzy76:
At some point ```projects[zen][version] = "5.4"``` was enough for downloading the zen theme (as it should according to the example.make file). But with the latest drush, the theme is downloaded to sites/all/modules instead of sites/all/themes unless you add a type key.

daften:
See #1627 
Didn't commit #1684 fix this?

daften:
Apparently, part of the revert commit was reverted again in this commit : https://github.com/danepowell/drush/commit/4066da0a020faeb3224fd658de1d41b26f4552e5

It should be noted exactly why this is done, because the commit message doesn't explain enough IMO.

And this is a breaking change that has happened twice now, can this at least be documented properly? :)

ergonlogic:
Can someone provide a sample makefile that illustrates this behaviour on Drush's master branch?

fuzzy76:
I don't have a server with a newer Drush available right now, but dropping the line in my original post into any make-file (like https://github.com/drush-ops/drush/blob/master/examples/example.make ) should show the behaviour.

ergonlogic:
example.make doesn't build properly because of an invalid patch and git branch. But removing those so that it does build, then adding `projects[zen][version] = "5.4"` does indeed show that it was downloaded to `sites/all/modules/contrib/`. Here's a simpler makefile illustrating the behaviour:

    core = 7.x
    api = 2

    projects[] = drupal
    projects[zen][version] = "5.4"
    defaults[projects][subdir] = "contrib"

ergonlogic:
Adding `projects[zen][type] = "theme"` puts it in the correct place. The defaults must not be getting overridden by the data retrieved from update.drupal.org.

PR #1992, opened 2016-02-11
Backport fix for #1759 to Drush 7

PJnes:
Ran into the issue discussed in #1759 while using drush 7. The same fix works without any changes, so have just cherrypicked the same commit over.


weitzman:
Thanks ... In general, there isn't much reason to use Drush7 as Drush8 supports the same drupal versions.

Issue #1435, opened 2015-06-10
Drush 7 sql-sync fails miserably.


JorisVanEijden:
These all work fine with drush 6.5.0:
```
1. drush6 --target-dump=/dumpdir/db.sql    --source-dump=/dumpdir/db.sql    sql-sync @remote @local
2. drush6 --target-dump=/dumpdir/db.sql.gz --source-dump=/dumpdir/db.sql    sql-sync @remote @local
3. drush6 --target-dump=/dumpdir/db.sql    --source-dump=/dumpdir/db.sql.gz sql-sync @remote @local
4. drush6 --target-dump=/dumpdir/db.sql.gz --source-dump=/dumpdir/db.sql.gz sql-sync @remote @local
```
And all fail with drush 7.0.0:
```
1. drush7 --target-dump=/dumpdir/db.sql    --source-dump=/dumpdir/db.sql    sql-sync @remote @local
  => import failed   -- "Failed to gunzip input file."

2. drush7 --target-dump=/dumpdir/db.sql.gz --source-dump=/dumpdir/db.sql    sql-sync @remote @local
  => import OK, but no dumpfiles present afterward

3. drush7 --target-dump=/dumpdir/db.sql    --source-dump=/dumpdir/db.sql.gz sql-sync @remote @local
  => dump failed -- "already has .gz suffix"

4. drush7 --target-dump=/dumpdir/db.sql.gz --source-dump=/dumpdir/db.sql.gz sql-sync @remote @local
  => dump failed -- "already has .gz suffix"
```

How is this supposed to work?

weitzman:
sql-sync was rewritten for Drush7. We gained some functionality and lost some. It is far more maintainable. 

After import, Drush7 always deletes the dumpfiles on both source and target. So 2. looks like a success to me. Feel free to suggest better help text for the --*-dump options which are more clear about this.

PR #1996, opened 2016-02-12
Update drush sql-sync --target-dump and --source-dump documentation.

jbickar:
See https://github.com/drush-ops/drush/issues/1435


weitzman:
options can declare an example-value. would be better here.

jbickar:
Thanks, updated.

weitzman:
Merged into 7/8/master branches.

Issue #1977, opened 2016-02-04
site-install with no sites/default leads to endlessly recursing drush_mkdir calls


leeomara:
This may be an edge case too esoteric to care about, but I recently ran into the problem trying to site-install a copy of a site that (accidentally) had not `sites/default` directory.

It seems that inside of `drush_core_pre_site_install`, there's no provision for `$conf_path` being empty (or false), and then `drush_mkdir` keeps trying to call itself with the `dirname` of "".

Here is an abbreviated backtrace of the error:

```
PHP Fatal error:  Maximum function nesting level of '256' reached, aborting! in /usr/local/Cellar/drush/8.0.2/libexec/includes/filesystem.inc on line 360
PHP Stack trace:
PHP   1. {main}() /usr/local/Cellar/drush/8.0.2/libexec/drush.php:0
PHP   2. drush_main() /usr/local/Cellar/drush/8.0.2/libexec/drush.php:12
PHP   3. Drush\Boot\BaseBoot->bootstrap_and_dispatch() /usr/local/Cellar/drush/8.0.2/libexec/includes/preflight.inc:66
PHP   4. drush_dispatch() /usr/local/Cellar/drush/8.0.2/libexec/lib/Drush/Boot/BaseBoot.php:67
PHP   5. call_user_func_array:{/usr/local/Cellar/drush/8.0.2/libexec/includes/command.inc:185}() /usr/local/Cellar/drush/8.0.2/libexec/includes/command.inc:185
PHP   6. drush_command() /usr/local/Cellar/drush/8.0.2/libexec/includes/command.inc:185
PHP   7. _drush_invoke_hooks() /usr/local/Cellar/drush/8.0.2/libexec/includes/command.inc:217
PHP   8. call_user_func_array:{/usr/local/Cellar/drush/8.0.2/libexec/includes/command.inc:366}() /usr/local/Cellar/drush/8.0.2/libexec/includes/command.inc:366
PHP   9. drush_core_pre_site_install() /usr/local/Cellar/drush/8.0.2/libexec/includes/command.inc:366
PHP  10. drush_mkdir() /usr/local/Cellar/drush/8.0.2/libexec/commands/core/site_install.drush.inc:159
PHP  11. drush_mkdir() /usr/local/Cellar/drush/8.0.2/libexec/includes/filesystem.inc:361
PHP  12. drush_mkdir() /usr/local/Cellar/drush/8.0.2/libexec/includes/filesystem.inc:361
...
[...]

PR #1998, opened 2016-02-14
Issue #1977: If there is no default.settings.php then fail gracefully…

serundeputy:
…; no seg fault.

This fails gracefully with the failed to copy `default.settings.php` file error rather than unhelpful `Segmentation fault`.

### DIFF
--- a/commands/core/site_install.drush.inc
+++ b/commands/core/site_install.drush.inc
@@ -129,7 +129,13 @@ function drush_core_pre_site_install($profile = NULL) {
   if ($dir = drush_get_option('sites-subdir')) {
     $sites_subdir = "sites/$dir";
   }
-  $conf_path = $sites_subdir;
+  if ($sites_subdir) {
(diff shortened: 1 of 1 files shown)


commands/core/site_install.drush.inc
```diff
@@ -137,6 +143,9 @@ function drush_core_pre_site_install($profile = NULL) {
   $default = realpath($alias_record['root'] . '/sites/default');
   $sitesfile_write = drush_drupal_major_version() >= 8 && $conf_path != $default && !file_exists($sitesfile);
 
+  if (!file_exists($conf_path)) {
+    $msg[] = dt('create a @default directory', array('@default' => 'sites/default'));
```
greg-1-anderson: This should be `'@default' => $conf_path`, maybe?

serundeputy: indeed; good point!

greg-1-anderson:

PR #2002, opened 2016-02-17
Fixed sneaky fallback on mysql when mysqli is enabled and configured

eXistenZNL:
## What

This removes the sneaky switching from the mysqli driver to the [deprecated](http://php.net/manual/en/migration55.deprecated.php) mysql driver in Drush, even though the mysqli extension is configured and loaded, and working and the Drupal project we use Drush on uses the mysqli driver.

## Why

We are currently relocating a legacy Drupal 6 project from an old server which runs PHP 5.3, to a newer infrastructure that runs PHP 5.6. We run on Ubuntu 14.04 TLS, but since the default PHP version in Ubuntu 14.04 is 5.5, we make use of the widely used [PPA of Ondřej Surý](https://launchpad.net/~ondrej/+archive/ubuntu/php5-5.6).

In these builds of PHP 5.6, the legacy mysql driver / extension has not been compiled and packaged, and is thus not available. Debating whether this is wise and should be included by Ondřej Surý is beyond the point, since it the extension is deprecated as of PHP 5.5 and will be removed completely in PHP7, so if we do not deal with it now, we have to later.

Since Drupal 6 does not support support PDO but relies only on mysql and mysqli, it is not an option to switch to PDO instead. It is however important that Drush follows what has been configured in Drupal and should not wander of on it's own, deciding it is probably a good idea to instead opt for the legacy mysql extension even though the Drupal project is using mysqli.

## Drush output before my patch

```
$ drush up admin_menu
Command pm-update needs a higher bootstrap level to run - you will need to invoke drush[error]
from a more functional Drupal environment to run this command.
The drush command 'up admin_menu' could not be executed.                        [error]
Drush was not able to start (bootstrap) the Drupal database.                           [error]
Hint: This may occur when Drush is trying to:
 * bootstrap a site that has not been installed or does not have a configured database.
In this case you can select another site with a working database setup by specifying
the URI to use with the --uri parameter on the command line. See `drush topic
docs-aliases` for details.
 * connect the database through a socket. The socket file may be wrong or the php-cli
may have no access to it in a jailed shell. See http://drupal.org/node/1428638 for
details.

Drush was attempting to connect to: 
 Drupal version         :  6.37                                          
 Site URI               :  http://default                                
 Database driver        :  mysql                                         
 Database hostname      :  127.0.0.1                                     
 Database port          :                                                
 Database username      :  dev
 Database name          :  dev
 PHP executable         :  /usr/bin/php                                  
 PHP configuration      :  /etc/php/5.6/cli/php.ini                      
 PHP OS                 :  Linux                                         
 Drush script           :  /vagrant/drush-8/vendor/drush/drush/drush.php 
 Drush version          :  8.0.3                                         
 Drush temp directory   :  /tmp                                          
 Drush configuration    :                                                
 Drush alias files      :                                                
 Drupal root            :  /vagrant/test-project
 Drupal Settings File   :  sites/default/settings.php                    
 Site path              :  sites/default
```

## Output after my fix has been applied

```
$ drush up admin_menu
Update information last refreshed: wo, 02/17/2016 - 14:38
 Name                              Installed Version  Proposed version  Message                   
 Administration menu (admin_menu)  6.x-1.8            6.x-1.9           Nieuwe versie beschikbaar 

Code updates will be made to the following projects: Administration menu [admin_menu-6.x-1.9]

Note: A backup of your project will be stored to backups directory if it is not managed by a supported version control system.
Note: If you have made any modifications to any file that belongs to one of these projects, you will have to migrate those modifications after updating.
Do you really want to continue with the update process? (y/n): y
Project admin_menu was updated successfully. Installed version is now 6.x-1.9.
Backups were saved into the directory                                                  [ok]
/home/vagrant/drush-backups/test-project/20160217133810/modules/admin_menu.
No database updates required                                                           [success]
'all' cache was cleared.                                                               [success]
Finished performing updates.                                                           [ok]
```
Also, when checking `drush status`, the Database driver has been switched from `mysql` to `mysqli` after applying my patch.

## How to test

1. Grab the latest Drush, and on a PHP installation **without the legacy mysql extension**, configure a Drupal project to use mysqli as the driver, and see that Drupal will work but Drush will fail because it decides to use legacy mysql driver instead.
1. Now grab my patch, and see that Drush will work because it no longer switches from the mysqli driver to the mysql driver on it's own.
1. Rejoice, for all was good again.

### DIFF
--- a/includes/sitealias.inc
+++ b/includes/sitealias.inc
@@ -1641,7 +1641,7 @@ function drush_convert_db_from_db_url($db_url) {
       );
       $url = (object)array_map('urldecode', $url);
       $db_spec = array(
-        'driver'   => $url->scheme == 'mysqli' ? 'mysql' : $url->scheme,
+        'driver'   => $url->scheme,
         'username' => $url->user,
         'password' => $url->pass,
         'host' => $url->host,
--- a/lib/Drush/Sql/Sqlmysqli.php
+++ b/lib/Drush/Sql/Sqlmysqli.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Drush\Sql;
+
+class Sqlmysqli extends Sqlmysql {
+
+}


weitzman:
Thanks ... I think this change is too risky at this time in the Drupal 6 lifecycle (past EOL). Won't Fix.

eXistenZNL:
Lol! So people using PHP 5.5 and higher will have to start forking Drush and applying my patch manually so Drush actually doesn't break?!

It doesn't break Drupal 6 since the mysql driver still works if that's configured in the project. It only allows using the mysqli driver if that's configured in the project. Did you even read at all?

You really inspire people to contribute to open source projects this way.

EugenMayer:
Good work @eXistenZNL , was working on the same issue.

EugenMayer:
@weitzman interested in a patch where i test if the ext/mysql extension exists?

I would use the legacy (old) strategy if:
- mysql:// is used ( obvious )
- if mysqli:// is used AND ext/mysql is present (to not break this legacy hack, we kee pit

i would use the new mysqli code if:
- mysqli:// is used
- ext/mysql is **not** installed
- ext/mysqli is installed

This way, not old behavior is touched, its as broken as it has been.

Just in the case were even with the old drush, drush would become disfunctional otherwise (not ext/mysql but used mysqli) we would have a better fallback.

Beware, if this fix does not happen, drush is unusable under php7 (mysql has been removed)

weitzman:
Does this only affect Drupal 6 sites, or are more modern versions of Drupal affected as well? Drush is currently tested on PHP7 against Drupal 7 and Drupal 8 so I'm pretty confident that those Drupal versions work with PHP7+Drush9.

If this is just a Drupal 6 problem, I'm not inclined to fix it on any stable branch. Drush9 does not support Drupal 6 so there would be nothing more to do in this case.

EugenMayer:
i would not consider this stable, if drush does not offer any functionality at all - i would argue that this a critical fix and we could make it work, without breaking the old "broken behavior"

For now, i cannot understand how this can work in D7 and D8. What i can tell you, drush 7/8 does not work under php7 with D6 (tested)

weitzman:
[Drush6 does not claim to run under PHP7](http://cgit.drupalcode.org/drupal/tree/INSTALL.txt?h=6.x). Drush isn't going to try to make it do so.

EugenMayer:
You got me wrong. Drush7 and 8 do not work under PHP7 for drupal6, not drush6

weitzman:
Woops - I updated my link text. My point stands. You are using a combination that is not supported by Drupal project.

EugenMayer:
I see you linked drupal6 install.txt. I can get Drupal6 running fairly easy - it runs stable on php5.5 for more then 1 year now, stepping up to PHP7 is very trivial.

Even though i understand you argument, it does not make the above code working under php5.2, php5.3 or php5.

Drush 7 and Druah 8 will **also not** work for drupal6 under php5.x with deactivated ext/mysql which is suggested by all PHP docs like http://php.net/manual/en/intro.mysql.php

`This extension is deprecated as of PHP 5.5.0, and has been removed as of PHP 7.0.0. Instead, either the mysqli or PDO_MySQL extension should be used. See also the MySQL API Overview for further help while choosing a MySQL API.`

People do follow this advice and will discover a disfunctional drush.

eXistenZNL:
@EugenMayer Thanks! You wonder why it works for Drupal 7 and Drupal 8: simply because they support PDO and Drupal6 doesn't.

I think that running Drupal 7 or 8 with the MySQLi extension on a system without the legacy mysql extension , and then trying to run Drush, will break Drush as well, no matter the version (6,7,8,9).

cisso:
@weitzman I think the conditions proposed by @EugenMayer are sound. Could you elaborate how you would still consider them "too risky", as they clearly keep the old behavior where appropriate while also allowing to use the mysqli extension as intended?

eXistenZNL:
I will update my PR with extra backwards compatibility, if the intention to merge that is here. If @weitzman keeps saying "Meh wont merge" without valid reason, I won't waste my time on it anymore.

aft-christophe:
Because for some reason people dont want to merge this fix in the official branch, I believe this is the best workaround:
download the patch 
- https://github.com/drush-ops/drush/commit/0109de7fc814094c0bc388380fd9dce1377a70f5.patch
put it in your project .
- then in your composer file
```
  "require": {
    "cweagans/composer-patches": "~1.0",
    "drush/drush": ""
  },
  "extra": {
    "patches": {
      "drush/drush": {
        "fix mysqli driver": "path/to/the/file.patch"
      }
    }
  }
```

This will automatically apply the patch for you. on install and update.

segi:
+1

weitzman:
I'm happy to accept this patch if someone can run the test suite with just mysqli and drupal 6 and say that it passes

cptX:
So it looks like I have encountered the same problem. This patch looks like it will solve it. So I suppose my assumption regarding wrong DB driver was correct...

cptX:
So I manually applied the aft-christophe instructions (https://github.com/drush-ops/drush/pull/2002#issuecomment-373762885) and it works!!! DB is now connected!
I tried to set the patch in the composer json but after reinstalling drush/drush it didn't apply any changes, but I have zero experience with composer so it doesn't matter...
Thanks guys, I wouldn't expect to solve this issue so quickly. Please consider merging it in the official release...

greg-1-anderson:
So, per the [Mysqli overview](http://php.net/manual/en/mysqli.overview.php), the mysqli driver was introduced in PHP 5.0 and is strongly recommended in favor of the older mysql driver. There is no reason I can see that anyone running PHP 5.0 or later with Drupal 6 or later should require the fallback here.

I think that the reason this has stayed the way it is (that is, taking exactly the opposite advice of the documentation) is that the current implementation happens to work, and it's hard to know how many edge cases we might break until we put this into a stable release and a lot of people try it.

In the balance, I think that so much time has gone by that I'm 👍 on deploying this -- and making an equivalent change in Drush 9, which is also currently exhibiting this ancient behavior.  We might want to be a touch more conservative, though; something like:
```
    'driver'   => $url->scheme == 'mysqli' ? 'mysql' : $url->scheme,
...
function mysqlFallbackDriver($scheme)
{
    if (($scheme != 'mysqli') || !getenv('DRUSH_MYSQL_FALLBACK')) {
        return $scheme;
    }
    return 'mysql';
}
```
That is, the 'mysql' fallback would only happen if the user set the DRUSH_MYSQL_FALLBACK environment variable. Maybe Drush 9 only does the fallback if the environment variable is set to 1, and Drush 8 does the fallback **unless** the environment variable is set to 0.

dsnopek:
Another way of being conservative is to check if the 'mysql' extension is even enabled before swapping 'mysqli' to 'mysql'.

As it is now, under PHP 7, Drush is falling back on an extension that isn't there and, so, broken. For people without the 'mysql' extension, not falling back couldn't break anything (as it's aready broken) and only improve things!

greg-1-anderson:
I'm confused about the failure condition. I'm running PHP 7.2 and the hacky mysql fallback works fine.
```
$ php -a
Interactive shell

php > var_export(extension_loaded('mysqli'));
true
php > var_export(extension_loaded('mysql'));
false
```
Why isn't Drush blowing up on my system?

cptX:
On which drupal version are you running Drush?

greg-1-anderson:
@cptX Just Drupal 7 and Drupal 8, but it seems that the mysql fallback should also be failing in that environment, and I don't know why it is working.

EugenMayer:
You should consider reading the comments above. This would save you a lot of time.

greg-1-anderson:
Sure enough, @eXistenZNL [answered this above](https://github.com/drush-ops/drush/pull/2002#issuecomment-197214309).

dsnopek:
Here's a PR https://github.com/drush-ops/drush/pull/3706 which does what I attempted to describe in my comment above https://github.com/dsnopek

More details on the PR, but my hope is that this is conservative enough to merge without fear of breaking anyone :-)

PR #2003, opened 2016-02-17
Flexible versions for Symfony components

DarioSwain:
Version of ```symfony/var-dumper``` is increased, because at now I receive conflict with ```psy/psysh``` package, [they use](https://packagist.org/packages/psy/psysh#v0.6.0) ```symfony/var-dumper``` version ```~2.7|~3.0```.

Also for symfony/yaml I made version more flexible.

This PR can solve the next problems:

Use case 1:
We already have application with Symfony components and we try to add drush like dependency into our application.
In this case we receive conflict when our symfony yaml or var-dumper component ```>= 2.8``` or ```< 2.7```.

Use case 2:
Some dependency package increase they version of any symfony component. For example ```psy/psysh:0.7``` start using version ```2.8``` of yaml.
In this case we never receive version ```0.7``` of ```psy/psysh```, because we have hard dependency on ```symfony/yaml``` version ```2.7.*``` instead of ```~```.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,8 +32,8 @@
     "php": ">=5.4.5",
     "psr/log": "~1.0",
     "psy/psysh": "~0.6",
-    "symfony/yaml": "2.7.*",
-    "symfony/var-dumper": "^2.6.3",
+    "symfony/yaml": "~2.3|~3.0",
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Thanks for this.

PR #2004, opened 2016-02-17
Fixes #2000: Resolve git branches and revisions in make-lock.

### LINKED ISSUE #2000: Allow make lock commands to specify a git revision
When we specify a git repo as the source for a project, `make-lock` and its cousins won't lock down the revision that would be loaded. As a result, builds are not reliably reproducible unless we specify the revision ourselves in the source makefile. It looks like this ought to be fairly straight-forward:

    $ git ls-remote http://github.com/drush-ops/drush.git --branch 7.x
    66bff105a6161f9e3e0df7775db22f636477ce2f        refs/heads/7.x
    $ git ls-remote http://github.com/drush-ops/drush.git --branch master
    28f3d52cdc3ea2bba54f55e3acbfb598e301f978        refs/heads/master

ergonlogic:
Here's a pretty straight-forward solution to #2000.

### DIFF
--- a/commands/make/generate.contents.make.inc
+++ b/commands/make/generate.contents.make.inc
@@ -110,6 +110,10 @@ function make_generate_from_makefile($file, $makefile) {
     }
     $projects[$name]['_type'] = $project['type'];
 
+    if ($project['download']['type'] == 'git') {
+      drush_make_resolve_git_refs($projects[$name]);
+    }
+
     // Don't clutter the makefile with defaults
     if (is_array($defaults)) {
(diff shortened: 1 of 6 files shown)


weitzman:
Code looks good. Any chance e can get test coverage? Can be a todo if needed.

ergonlogic:
Because this locks down a generated makefile to a specific commit, we'd probably want a project that isn't going to have any further commits...

ergonlogic:
Maybe just add some old 5.x module as a library...

PR #2010, opened 2016-02-19
Fix undesirable redispatch

greg-1-anderson:
drush_preflight_command_dispatch would redispatch even if the DRUSH_FINDER_SCRIPT environment variable was set; that is not desirable, so check for that explicitly.


greg-1-anderson:
This is incorrect.

Issue #2020, opened 2016-02-22
Tidy up: VCS software no longer uses nested control directories


AdamPS:
CVS and early versions of SVN put a control directory at every level of the code tree.  Modern systems seem to only have them in the root.

Since this commit https://github.com/drush-ops/drush/commit/e999ecc38df431a6c5e2c6e286aa99e407cebb4f whenever Drush runs pm-update, the old module is deleted and a new one installed.  So any nested control directory would be lost.  This commit is a while ago now, so it seems reasonable to assume that nobody is still using the older VCS software.

Therefore we might as well delete the similar code in pm-download.  There is a significant motivation for this tidy up because, version control mechanisms other that backup are not widely tested.  At the moment there unnecessary code paths with an if test on version control - and hence bugs can come in the not widely tested code path.

AdamPS:
I also propose removing some delete_dir code specific to backup.inc that isn't necessary any more because subsequent non-specific code will overwrite the dir anyway.  Again this helps avoid a bug where someone starts relying on the code in backup.inc but that code is missing for other version control mechanisms.

AdamPS:
@weitzman This is removing some ancient CVS related code so I think it could be helpful, and seems aligned with your goal to simplify and remove SVN support also.

AdamPS:
OK, I'm not that interested and don't think maintainers are either so let's close it.

Issue #2015, opened 2016-02-21
Rollback fails with version-control=svn


AdamPS:
If rollback is needed, changes are not actually rolled back automatically.  However it's pretty easy to roll them back manually.

Originally noted as a bug as part of #1056, but it's separate from the main issue there, so opening a new one.  Fix is easy, pull request coming up soon.

AdamPS:
@weitzman so this one is definitely SVN-specific and I understand if you want to reject it because SVN support is scheduled to be removed.  (It is just one line mind you:-))

AdamPS:
I learnt that maintainer no longer wants to accept changes to SVN because it will be removed in drush 9.

Issue #1056, opened 2014-12-17
Getting SVN errors when updating Drupal core with pm-updatecode and Drush 7


mortenson:
I'm starting to use Drush 7 with SVN, and noticed an error when updating Drupal core that I never got with Drush 6:

```
$ drush pm-updatecode drupal --version-control=svn --svnsync
Update information last refreshed: Never
 Name    Installed Version  Proposed version  Message
 Drupal  7.26               7.34              SECURITY UPDATE available


Code updates will be made to drupal core.
WARNING:  Updating core will discard any modifications made to Drupal core files, most noteworthy among these are .htaccess and robots.txt.  If you have made any modifications to these files, please back them up before updating so that you can re-create your modifications in the updated version of the file.
Note: Updating core can potentially break your site. It is NOT recommended to update production sites without prior testing.

Do you really want to continue? (y/n): y
Destination directory docroot/drupal-7.34 already exists.                                                                                          [error]
Updating project drupal failed. Attempting to roll back to previously installed version.                                                                                       [error]
Drush was unable to get the svn status on docroot/drupal-7.34. Check that you have Subversion                                                      [error]
installed and that this directory is a subversion working copy.
The specific errors are below:
svn: 'docroot' has no ancestry information
```

As I'm not that familiar with how Drush 7 works with SVN, does anyone have any ideas regarding this?

jonhattan:
This could be a regression introduced with e999ec. I'll check that.

I'd prefer to not lose code that is already isolated. Perhaps keeping version-control feature, but moving svn and bzr implementations to contrib is an option. I'd be happy to do such extraction --and provide a basic git vcs implementation that I'll use myself.

weitzman:
I like that plan, jonhattan.

whittaker007:
Excuse me for butting in, but I'm seeing similar issues which I'm not sure is SVN-specific or not. We do have our project version control under SVN, and after updating Drush we are unable to update modules with `svn up`.

What happens is `svn up` lists the available updates and asks for confirmation as usual, but once confirmed it fails on the first module with the message:

````
Destination directory [path to module] already exists.
Updating project [module name] failed. Attempting to roll back to previously installed version.
````
Except it doesn't roll back the module, it deletes it instead.

Updating the modules using `drush dl` works as expected.

kgaut:
I'm experiencing the same issue as @whittaker007 when trying to do a pm-update on my drupal 7 project.

[...]

PR #2021, opened 2016-02-22
Get SVN back working again

AdamPS:
OK, I'm not an expert on drush, nor github.  However not many people are using SVN, so I guess if I want it to work, I'm going to have to give it a try.  Sorry for any mistakes.  I would appreciate some sort of comment even if it's "thanks I'll look in a few weeks when I have time" or "thanks, but you've done it all wrong because..."

1) Main SVN fix in #1056:
- I noted the new wget package update code simply overwrites the whole old package with the new, with no fancy recursive preserving of control files, orphan dir handling etc.
- Actually that's not a big problem as in modern SVN (and I think BZR as best I could tell from a search) the control files are only at the root.
- So firstly a quick extra call to skip control files when updating drupal core.
- Secondly, set the option to drupal_move_dir to make sure we actually do overwrite.

2) The rollback part #2015 should be simple and clear.

3) #2020 is not essential to make SVN work, but will help stop it breaking in future.  I've removed various hard-coded if tests on VCS system, so there is much greater commonality of code.  I can take this one out of the pull request if needs be.

4) In passing I removed some old code/comments relating to CVS.

### DIFF
--- a/commands/pm/download.pm.inc
+++ b/commands/pm/download.pm.inc
@@ -198,37 +198,18 @@ function drush_pm_download() {
     if ($version_control->engine == 'backup') {
       // Check if install location already exists.
       if (is_dir($request['project_install_location'])) {
-        if (drush_confirm(dt('Install location !location already exists. Do you want to overwrite it?', array('!location' => $request['project_install_location'])))) {
-          drush_delete_dir($request['project_install_location'], TRUE);
-        }
-        else {
+        if (!drush_confirm(dt('Install location !location already exists. Do you want to overwrite it?', array('!location' => $request['project_install_location'])))) {
           drush_log(dt("Skip installation of !project to !dest.", array('!project' => $request['name'], '!dest' => $request['project_install_location'])), LogLevel::WARNING);
           continue;
         }
       }
     }
-    else {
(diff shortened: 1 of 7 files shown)


weitzman:
Code looks OK Needs review and merge by @jonhattan

AdamPS:
10 months later, unfortunately this seems to have stalled out!

I would really appreciate if this could be merged.  I feel that I was encouraged to contribute because no one else was really interested in the pm area.  So I took the time to learn drush and try and help out.  At the moment I have to keep merging my patches into the latest drush - which for someone who doesn't really use git gets increasingly complicated the longer it goes on:-)

I have a second pull request #2083 based on this one that solves two more long-standing drush bugs.  I feel it would be even better to commit that one.  However, the maintainers have not commented on that PR yet.

weitzman:
alas, it seems that jonhattan never found time to review this. 

svn and bzr engines are scheduled for removal in drush9. sorry, but this will not be merged.

AdamPS:
@weitzman OK we can lose the SVN part, but 99% of the code in this PR and the referenced one is not specific to SVN.

In fact 3 of my commits are to remove outdated code, so that should actually help in removing SVN.

PR #2037, opened 2016-02-27
Fixes #2033: Force script name for run-server router

### LINKED ISSUE #2033: run-server command fails with paths with dots (config entities) in URL
run-server command works really well for local development, but it currently fails with paths like http://localhost:8888/admin/structure/types/manage/event/fields/node.event.body

The HTML for the page is returned correctly, but for some reason no assets will load, and I see lots of 404s in the log for paths like http://localhost:8888/admin/structure/types/manage/event/fields/core/assets/vendor/normalize-css/normalize.css?o34nnk

mikeyp:


weitzman:
Do people use runserver with other scripts like update.php?

mikeyp:
The file_exists check at the top of the router handles update.php and core/install.php just fine. This also makes sense at the end since we're requiring index.php directly anyway, so there's no reason the SCRIPT_NAME should be anything other than index.php.

weitzman:
Thanks. Any chance you can add a code comment linking to this issue?

mikeyp:
Updated with an explanation and link to the issue

weitzman:
Fixed a typo and committed in 2404559. Thanks.

Issue #178, opened 2013-10-03
Allow extra arguments to mysqldump in sql-dump command


kenorb:
--tab=path, -T path

     Produce tab-separated text-format data files. For each dumped table, mysqldump creates a tbl_name.sql file that contains the CREATE TABLE statement that creates the table, and the server writes a tbl_name.txt file that contains its data. The option value is the directory in which to write the files.

See:
http://dev.mysql.com/doc/refman/5.1/en/mysqldump.html#option_mysqldump_tab
http://stackoverflow.com/a/6300583/55075

com2:
I would like this too, but it would be nice when some arguments / switches could be added by default. I think of the the missing --routines switch. When you work frequently with CiviCRM you need to include stored functions / routines in your dump. With mysqldump you would think that everything is safely backed up until you discover that without this switch the stored functions are missing!

PR #2039, opened 2016-02-28
Add option 'no-data' to sql-dump command

AdamPS:
The --skip-tables-key option is handy to exclude temporary tables such as caches when doing sql-dump.

However when you come to restore a site from such a dump, you need to create the schema for these skipped tables.  mysql has such a feature, and this PR exposes it through drush with a new "--no-data" option.

Typical use case:
```
drush sql-dump --data-only --skip-tables-key=XX
drush sql-dump --no-data
```

I marked this as mysql only because that is the only platform I can test or know about.


weitzman:
I would rather see #178

PR #2040, opened 2016-02-28
In watchdog-show, allow filtering by a range of severity levels

AdamPS:
When filtering log messages, arguably the most common and useful option is to display all messages of a given severity or higher.

The watchdog-show command currently support filtering by severity, but only for exact match to a specific level.  This PR adds support to filter on a range of severities with a new option.  The default behaviour is unchanged for back-compatibility.

### DIFF
--- a/commands/core/watchdog.drush.inc
+++ b/commands/core/watchdog.drush.inc
@@ -48,6 +48,7 @@ function watchdog_drush_command() {
     'options' => array(
       'count' => 'The number of messages to show. Defaults to 10.',
       'severity' => 'Restrict to messages of a given severity level.',
+      'severity-range' => 'When matching by severity, include messages of the given level or any more severe level (by default, matches exact level).',
       'type' => 'Restrict to messages of a given type.',
       'tail' => 'Continuously show new watchdog messages until interrupted.',
       'sleep-delay' => 'To be used in conjunction with --tail. This is the number of seconds to wait between each poll to the database. Delay is 1 second by default.',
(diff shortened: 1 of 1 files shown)


jonhattan:
I'm not a fan of adding a new option for this case.

I'd prefer for `--severity` and `--type` to accept multiple values, comma separated (based on [_convert_csv_to_array()](http://api.drush.org/api/drush/includes!drush.inc/function/_convert_csv_to_array/master)). On that, you can create for example a shell alias for `wd-show --severity=notice,warning,error`.

Other nice addition would be to negate severities: `wd-show --severity=!info`.

AdamPS:
@jonhattan Thanks for the reply.  I can see it could be neater to avoid an extra option.  On the other hand, it's potentially a bit clumsy and not user-friendly to have to type:
`drush wd-show --severity=notice,warning,error,critical,alert,emergency`

So perhaps some mechanism to make it easier to specify a range of values?  E.g.
`drush wd-show --severity=notice+`

I admit I'm struggling to understand the use-case for cherry-picking individual severities.  Do you have any example why someone might want notice, warning, error, but definitely not want critical, alert, emergency?

I think what you say for filter by type makes a lot of sense.  Users may want multiple values of type, specifically excluded a type and so on.  It's just the severities where I feel it is more of a linear range with a threshold.

greg-1-anderson:
This can now be done with the `--filter` option

AdamPS:
@greg-1-anderson That's great thanks for the information.

PR #2043, opened 2016-02-29
#178 Allow extra arguments in sql-dump command

AdamPS:
The --extra option is already present on other SQL commands, so all it needs is to reuse that on sql-dump.  As far as I can see, --extra should work for any back-end, so I've changed some description references from mysql to SQL.

Note the the original https://www.drupal.org/node/518184 additionally allowed removing of the default options (--no-autocommit --single-transaction --opt -Q) by removing them from the code and adding them back in to example.drushrc.php.  However I'm not clear how existing users would get their actual drushrc.php modified.  Anyway I've _not_ done this extra bit as I don't need it, and it's not part of the title of #178.  I suggest that if anyone cares about it they can create a new task.

### DIFF
--- a/commands/sql/sql.drush.inc
+++ b/commands/sql/sql.drush.inc
@@ -71,11 +71,11 @@ function sql_drush_command() {
     'description' => 'A string for connecting to the DB.',
     'bootstrap' => DRUSH_BOOTSTRAP_NONE,
     'options' => $options + $db_url + array(
-        'extra' => array(
-          'description' => 'Add custom options to the mysql command.',
-          'example-value' => '--skip-column-names',
-        ),
+      'extra' => array(
+        'description' => 'Add custom options to the SQL command.',
+        'example-value' => '--skip-column-names',
       ),
+    ),
     'examples' => array(
       '`drush sql-connect` < example.sql' => 'Bash: Import SQL statements from a file into the current database.',
       'eval (drush sql-connect) < example.sql' => 'Fish: Import SQL statements from a file into the current database.',
@@ -118,6 +118,7 @@ function sql_drush_command() {
       'data-only' => 'Dump data without statements to create any of the schema.',
       'ordered-dump' => 'Order by primary key and add line breaks for efficient diff in revision control. Slows down the dump. Mysql only.',
       'gzip' => 'Compress the dump using the gzip program which must be in your $PATH.',
+      'extra' => 'Add custom options to the SQL command.',
     ) + $options + $db_url,
   );
   $items['sql-query'] = array(
@@ -139,7 +140,7 @@ function sql_drush_command() {
       ),
       'file' => 'Path to a file containing the SQL to be run. Gzip files are accepted.',
       'extra' => array(
-        'description' => 'Add custom options to the mysql command.',
+        'description' => 'Add custom options to the SQL command.',
         'example-value' => '--skip-column-names',
       ),
       'db-prefix' => 'Enable replacement of braces in your query.',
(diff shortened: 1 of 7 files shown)


greg-1-anderson:
Should we use strict option handling here, perhaps, like rsync?

weitzman:
Thanks for taking on this long time todo.

1. A bit too simplisitic. See \Drush\Sql\Sqlmysql::dumpCmd. There we run mysqldump twice if needed. Each would need the --extra I think. IMO we need to add this to dumpCmd in the supported Db backends.
1. Please add an entry to Examples section in command help so folks get a clue that this is available.
1. I don't see much value right now for making this a strict command so nothing to do there.

AdamPS:
Thanks for the feedback - done.  I have tested on mysql, but don't have access to any of the other SQL servers.

commands/sql/sql.drush.inc
```diff
@@ -71,11 +71,11 @@ function sql_drush_command() {
     'description' => 'A string for connecting to the DB.',
     'bootstrap' => DRUSH_BOOTSTRAP_NONE,
     'options' => $options + $db_url + array(
-        'extra' => array(
-          'description' => 'Add custom options to the mysql command.',
-          'example-value' => '--skip-column-names',
-        ),
+      'extra' => array(
+        'description' => 'Add custom options to the SQL command.',
```
weitzman: The extra bits are added to the *dump* command, not any SQL statement.

commands/sql/sql.drush.inc
```diff
@@ -101,6 +101,7 @@ function sql_drush_command() {
     'examples' => array(
       'drush sql-dump --result-file=../18.sql' => 'Save SQL dump to the directory above Drupal root.',
       'drush sql-dump --skip-tables-key=common' => 'Skip standard tables. @see example.drushrc.php',
+      'drush sql-dump --extra=--no-data' => 'Pass extra option to SQL dump command.',
```
weitzman: Omit SQL here

commands/sql/sql.drush.inc
```diff
@@ -118,6 +119,7 @@ function sql_drush_command() {
       'data-only' => 'Dump data without statements to create any of the schema.',
       'ordered-dump' => 'Order by primary key and add line breaks for efficient diff in revision control. Slows down the dump. Mysql only.',
       'gzip' => 'Compress the dump using the gzip program which must be in your $PATH.',
+      'extra' => 'Add custom options to the SQL command.',
```
weitzman: The extra bits are added to the *dump* command, not any SQL statement.

commands/sql/sql.drush.inc
```diff
@@ -139,7 +141,7 @@ function sql_drush_command() {
       ),
       'file' => 'Path to a file containing the SQL to be run. Gzip files are accepted.',
       'extra' => array(
-        'description' => 'Add custom options to the mysql command.',
+        'description' => 'Add custom options to the SQL command.',
```
weitzman: Maybe "Add extra arguments to the database connection. Perhaps add an item to the Examples section of the command definition that illustrates.

AdamPS:
Done, thanks, with minor tweaks.
- This PR includes some fixes to the description for other commands referring to the specific engine 'mysql'.  So in the first comment where you suggested "dump command" I swapped for "connect string"; hopefully I got it right.
- In the last comment you suggested to add an example, but in fact there is example-value already, so I didn't duplicate it.

weitzman:
OK, looks great. Thanks.

Do you see anywhere in the test suite where we could add a --extra and add an assert to make sure it worked. It is OK if its mysql only.

weitzman:
Thanks ... I added a test and squash merged.

AdamPS:
Many thanks.  Sorry I've not learnt about running Drush tests yet.

PR #2044, opened 2016-02-29
Tolerate recoverable errors in batch operation callbacks

### LINKED ISSUE #1930: DrushBatchContext brakes batch processes if batch operation callback uses typehint for $context
From Drupal 8.0.2 [`locale_translation_batch_status_check`](
http://cgit.drupalcode.org/drupal/tree/core/modules/locale/locale.batch.inc?h=8.0.2#n39) uses typehint for `$context` param.

Drush 8.0.2 passes a [`DrushBatchContext`](https://github.com/drush-ops/drush/blob/8.0.2/commands/core/drupal/batch.inc#L150) instance as `$context` param.

Even if `DrushBatchContext` extends `ArrayObject` the `array` typehint causes a fatal error:
`Argument 4 passed to locale_translation_batch_status_check() must be of the type array, object given locale.batch.inc:39`

jonhattan:
Attempts to fix #1930, based on #1465. Untested.

### DIFF
--- a/commands/core/drupal/batch.inc
+++ b/commands/core/drupal/batch.inc
@@ -149,7 +149,13 @@ function _drush_batch_worker() {
       // Magic wrap to catch changes to 'message' key.
       $batch_context = new DrushBatchContext($batch_context);
 
+      // Tolerate recoverable errors.
+      // See https://github.com/drush-ops/drush/issues/1930
+      $halt_on_error = drush_get_option('halt-on-error', TRUE);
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This had one test that was failing for no rational reason; I re-ran that one test, and now everything is green.

I don't like the modal nature of the fix; however, I can't think of a cleaner option (other than fixing Drupal -- not possible in the short term -- or turning off recoverable errors altogether -- not desirable), though, so I think we should go with this.

PR #2045, opened 2016-02-29
Try testing just one testing group with Drupal 6.

greg-1-anderson:
Let's find out how many of the unit tests actually pass on Drupal 6.  Just running one test group to start out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,8 +52,11 @@ matrix:
   exclude:
     - php: 7.0
       env: UNISH_DRUPAL_MAJOR_VERSION=6 PHPUNIT_ARGS=--group=base
+    - php: 7.0
       env: UNISH_DRUPAL_MAJOR_VERSION=6 PHPUNIT_ARGS=--group=commands
+    - php: 7.0
       env: UNISH_DRUPAL_MAJOR_VERSION=6 PHPUNIT_ARGS=--group=pm
+    - php: 7.0
       env: UNISH_DRUPAL_MAJOR_VERSION=6 PHPUNIT_ARGS=--exclude-group=base,make,commands,pm,quick-drupal
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ sudo: false
 
 env:
   matrix:
+#D6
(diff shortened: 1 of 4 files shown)


weitzman:
The test matrix on master branch already too huge. 

So my strategy for D6 and for alternate DB engines is to periodically pull the latest code into a  "fulltest" branch (e.g. [6.x-fulltest](https://github.com/drush-ops/drush/tree/6.x-fulltest), [master-fulltest](https://github.com/drush-ops/drush/tree/master-fulltest) and then more complete Travis tests get run automatically. So, we do test drush6 "every once in a while". drush6-fulltest was passing except for some known postgres issues at the time we released Drush 8.0.0.

We could use a more rigor around testing these edge case configurations. I dont know that Drush6 is really the place to start though. [We have marked in as unsupported in our matrix](http://docs.drush.org/en/master/install/). I think that's a realistic position given our current manpower.

greg-1-anderson:
Okay, the test case fixes in this PR could be merged into master-fulltest.  The changes here are for Drupal 6 in the master branch; some changes are needed to make Drupal 6 tests pass under the latest Drupal.org infrastructure.  I think it would be good to run at least one quick Drupal 6 test on each master (soon, 8.x) branch test, but whatever you want to do about this is fine.

I don't plan on doing anything in the 6.x or 6.x-fulltest branch. Drush 6 is EOL, no need to work on it any longer.

I see that the fulltest branch does not test with php 7. You can keep it like that, or use the matrix exclude expressions from this PR to add it in.

greg-1-anderson:
I've come down with a cold and was not in top form last night; this probably should not have been merged.  I thought I was testing on HEAD of master, but in fact I had not pulled in Jonathan's fixes. As a result, some of the changes here are workarounds for a bug that is now fixed. It's possible that all of the changes except the .travis.yml update is unnecessary.

I'll work on backing out what is not needed here later.

Regarding the number of tests in the queue on the 8.x branch, while it is certainly true that we have a lot of tests here, soon we will have master ==> Drush 9, and at that point we can remove Drupal 6 from master at that time.  I think it's probably fine to have extra tests on 8.x, since we'll only be backporting commits there.

PR #2055, opened 2016-03-04
Fix #1965 site drushrc

### LINKED ISSUE #1965: Site not recognized in advanced setup
How to reproduce (with today's 8.0.x-dev):
* Have a drupal 7.x multisite containing a site named "foo", publicly accessible as http://www.foo.com
(...and some "default" site)
* $sites.php maps www.foo.com to foo
* some modules need the correct $SERVER['host'] also in drush calls, so sites/foo/drushrc.php contains _uri = www.foo.com_

Now do
`drush @sites sa @self`
which internally does
`drush --uri=foo sa @self`

Expected: uri=http://www.foo.com
Experienced: uri=default

(At least i think these are the relevant things to reproduce...)

The real-world trouple is a multisite, that caches links with wrong base_url in _drush cc all_.

axel-rutz:


Cottser:
I can't speak to the DrupalBoot.php change but the one in includes/context.inc definitely solved my issue:

sites/default/drushrc.php exists.
sites/example.com/drushrc.php also exists.

`cd` to sites/example.com/drushrc.php and run `drush status` and the information from the default site is presented, not the example.com site.

With the change to context.inc the behaviour works as I would expect it to, running `drush status` (or any other commands) from sites/example.com targets that site instead of the default.

Cottser:
Sorry. Nevermind, the change to context.inc makes it so the default drushrc.php is not picked up so I think this needs to be fixed another way.

greg-1-anderson:
Neither of the changes here seem appropriate.

Rather than providing a default for DRUSH_SELECTED_DRUPAL_SITE_CONF_PATH, try setting it correctly at selection time.

Avoid making out-of-order calls to bootstrap operations. Every bootstrap phase should happen once, and in order. There's currently one spot in preflight.inc where stuff is re-done if the site selection changes during bootstrap.  Any adjustments necessary for similar reasons should be done in the same place.

It might be easier to work on fixing `drush sa @sites` rather than `drush @sites sa @self`.  That is, if `@sites` is set up correctly, then it should pass the correct URI in to each call to backend invoke. Trying to fix up the URI after the fact seems messier.  Of course, for the more straightforward case of `drush @onsiteofmany`, just make sure that the uri is correct in your aliases file.

PR #2062, opened 2016-03-08
Exclude node_modules and bower_components from directory scans

jhedstrom:
Quick workaround to #1347


jhedstrom:
I backported this to `8.0.x` as well.

PR #2071, opened 2016-03-16
Remove unrecognized options used by postgresql in archive-dump.

blauerberg:
archive-dump on postgresql does not works correctly because it refer unrecognized options.

```
$ drush ard
pg_dump: unrecognized option '--no-align'
Try "pg_dump --help" for more information.
Database dump failed
...
```

### DIFF
--- a/lib/Drush/Sql/Sqlpgsql.php
+++ b/lib/Drush/Sql/Sqlpgsql.php
@@ -121,7 +121,7 @@ class Sqlpgsql extends SqlBase {
     if (isset($data_only)) {
       $extra .= ' --data-only';
     }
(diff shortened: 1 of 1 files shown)


kasperg:
This is the same issue as in #2072 but with another solution.

This PR removes the options entirely - even for [`psql` where they are valid](http://www.postgresql.org/docs/9.4/static/app-psql.html). #2072 removes them when running `pg_dump`.

weitzman:
Fixed in df6fba0d4a9024aa4deecbac2c36c657d4d9d806 for Drush9. Postgres is fully tested every night for Drush9. Alas archive-dump is no longer available.

PR #2072, opened 2016-03-16
Do not add default extra options to pg_dump

kasperg:
Current options which are valid for psql do not work for pg_dump:
- no-align
- field-separator
- pset
- tuples_only

See http://www.postgresql.org/docs/9.5/static/app-pgdump.html

Adding them causes the command to fail.


weitzman:
Fixed in df6fba0d4a9024aa4deecbac2c36c657d4d9d806

PR #2073, opened 2016-03-16
Set password automatically for pg_dump

kasperg:
Without this users are prompted for database password when running
sql-dump for Postgres.

This change utilizes the same PGPASSFILE environment variable approach
as is used for running normal psql commands.


weitzman:
Fixed in #3480

PR #2076, opened 2016-03-17
core-init: Fix PATH export

vieresjoki:
`export $PATH` causes variable interpolation which breaks .bashrc.


vieresjoki:
Same issue is present in 8.x branch which is where I originally meant to fix this.

weitzman:
cherry picked into master and 8.x. thanks.

Issue #2069, opened 2016-03-15
drush config-import --partial doesn't create field tables in DB


siolfyr:
Sorry if this isn't an issue, but it feels like one.

In D8, if using Drupal's config sync from the UI, new fields create their tables in the DB.  If I do the following:

1. Make a DB dump
2. Make a new field
3. Config-export with Drush so the new field configs will be imported next time
4. Delete DB and restore from dbdump
5. Drush config-import

...the tables in the DB aren't created, but the config for this field exists in the DB.  What's worse, if I also changed the config of a view to rely on the new field I made, the site will crash trying to load the field from the tables that don't exist.

I greatly appreciate things that automate what could be done in the UI.  If drush cim is not made for this, is there a way to tell Drupal, "Hey, you know those fields in your config? I want those."

Tried with D8 8.0.5 and Drush 8.0.5.  Thanks for reading.

Berdir:
drush just calls core API's. This should work and we have tests in core that it does. I've also imported fields like this.

siolfyr:
Any idea why it's not working in my case, then?  I'm using the --source and --partial options.  I'm very certain that it isn't working, I've tried the test a number of times.

weitzman:
Can you try without --partial? That is, do a config-export, add/update the
new files, and config-import from there.

siolfyr:
When I didn't use the partial option, the node__field and node__revision tables are created.  It seems like it must only be the --partial option that this is happening for, then.

weitzman:
Thanks for testing that.

--partial uses a very low level API and apparently bypasses a lot of the config event system. We either need to put warnings on --partial, remove it, change it, or something else. I have long thought that--partial should 

1. config-export into a temp dir
1. copy the contents of the --partial dir into the temp dir
1. config-import from the temp dir

You can see us bypassing the API by comparing these:

- [_drush_config_import_partial](http://api.drush.org/api/drush/commands%21core%21config.drush.inc/function/_drush_config_import_partial/master)
- [_drush_config_import](http://api.drush.org/api/drush/commands%21core%21config.drush.inc/function/_drush_config_import/master)

@jhedstrom since he added it in #1170.

podarok:
The thing is that config-import removing configs from active storage if they aren't included to importing one.
So here is a use-case why --partial should be kept
User creating Contact Form (Config Entity) and working with it
DEV team trying to push their own changes to the user's environment and drush config-import removes created Contact Forms...

[...]

PR #2082, opened 2016-03-18
Remove --file option form config-edit

weitzman:
It duplicates config-import --partial, and is buggy. A fix for --partial is coming in #2069


jhedstrom:
:+1: 

This looks good to me.

Issue #2022, opened 2016-02-23
Follow correct update procedure for drupal core


AdamPS:
Correct procedure is [here](https://www.drupal.org/node/1223018).

Principal problem relates to step 3), "Remove all old core files and directories".
- Latest master preserves all files (_pm_update_core search for 'backup target') from old version.  Consequence is that if the new version of Drupal removes a file it doesn't get removed locally.  Although unlikely, worst case the file could have been removed to fix a security issue.
- The preserving code only works if there was a backup (depends on command line options).  Without that, it's like old drush (see #614), and any custom files in the root or profile directory are removed.

Why don't we instead do exactly what the instructions say?
- After installing core, save a list of the files installed.  
- Download the new core to a working directory.
- Delete all files on the list from the old core.  If the list is missing, take it from the new core which normally will be right.
- Copy from the working directory to the root.  (Added bonus is that files like /sites/README.txt are updated which the instructions say should be.)

Secondary problem: drush doesn't set maintenance mode.  Does anyone know if this is deliberate?  Looks like it does in D8, but only during database update.  Of course whilst we are copying files, it will be pure luck if any page works at all.  But it seems like it would be safest in case the process has a long delay at any stage.  I guess the downside is that a minor error could leave the site in maintenance mode which might be unwanted.

weitzman:
1. Hopefully someone will chime in about file preservation.
2. We don't set maint mode because thats rarely needed for code only updates. Furthermore, pm-updatecode already has enough responsibilties. pm-update adds the complexity of updatedb on top of it. If you want maint mode, script it into your process as per the Example in vset docs.

AdamPS:
Thanks.  In terms of point 1), I believe this proposal is the more or less the same as proposed by @greg-1-anderson in #614.  Minor differences:
- I am proposing it only for core (the use-case for adding site-local files within a module directory is much less clear)
- I am not suggesting we actual fetch the old release again at upgrade time, but instead keep a record of the top-level files in it.

If there is consensus, I will have a go at coding it, although it will be a few weeks before I have time.

I'm happy to ignore maint mode.

greg-1-anderson:
I am wary about the `save the list of files downloaded` part. Maybe this should be an example pm-updatecode hook rather than a part of the core Drush pm code.

AdamPS:
Thanks.  OK, so if we don't save the list, I guess the other options I can see are
- re-fetch the old core to check what files are in it (and figure out what to do if we can't)
[...]

Issue #614, opened 2014-04-28
Provide a way to not delete files or folders during pm-update drupal


kgaut:
When running 

```
  drush pm-update drupal 
```

All the files and folder within the rootpath are deleted, drupal's, but also, for example IDE's settings.

It would be nice to provide a way to not delete these files / folders, maybe by a drush.ignore file or by an argument to pm-update command.

What do you think about this idea? Do you have any suggestion ?

superbiche:
This only applies to the .htaccess issue (that was one of the main topic of https://drupal.org/node/1018190): couldn't we write some of the directives we want to keep when upgrading core between .htaccess comment tags, something like #<start drush-keep> and #<end drush-keep>, and add a Drush behavior to catch this block and its contents, the rewrite it in the upgraded .htaccess?
This would make the upgrade process safer, avoid extra manipulation/hooks, and prevent not upgrading .htaccess, which isn't safe either.

Can try to make a PR for this.

AdamPS:
@superbiche Your idea sounds fine, but it is not really part of this issue (maybe more the one you reference?).  This issue is talking about avoiding deleting files that weren't put there by Drupal in the first place.  Drupal does provide .htaccess.

Maybe best if you create a new issue for your idea, or use another one that is is more closely related to continue your discussion?

superbiche:
In the issue I reference, @greg-1-anderson asked to move the debate about this potential feature here on Github. So I'll open a new one here.
Sorry for the inconvenience :)

philsward:
Would it be worth having a .drushignore file within the root of Drupal and anything within that file is ignored (or copied out and copied back)

This would allow the .htaccess and robots.txt issue to be handled a bit better and if there are duplicates, the new one could be re-named.  I realize the existing Drupal files are on another issue, but they do all kind of come together.  Personally, I think one issue should cover all of the use cases until a clear understanding of how to approach each use case is determined, then they can be split off to separate issues.

Right now we have:
Drupal files vs Non-Drupal files
Is it possible for Drush to ignore files or would they have to be copied out of root and copied back?
If Drupal files, how do we deal with existing files?  Merge? Overwrite? Rename New, Rename Old?

I can see Drush asking the user what they want to do with each existing file.  Since Drush does a backup by default, if the user fat-fingers their choice, they can always retrieve out of the backup, right?

I've got some other thoughts on the existing Drupal files, so I'll go find that issue to post in.

AdamPS:
[...]

PR #2083, opened 2016-03-18
Follow correct update procedure for drupal core #2022

AdamPS:
Also solves
- #614 completely (currently it is fixed in some common cases, but not always).
- #4 Update Drupal core without moving core files out of the way.

See a more detailed list of benefits in the comments for #2022.  This PR solves several long open issues and simplifies code so it would be great to get it committed.

Note this PR is cumulative on top of #2021 because both affect the same areas of code, and it would be a mess of merging to consider them separately.  It would be sufficient to pull this PR then reject the other one.

### DIFF
--- a/commands/pm/download.pm.inc
+++ b/commands/pm/download.pm.inc
@@ -198,37 +198,18 @@ function drush_pm_download() {
     if ($version_control->engine == 'backup') {
       // Check if install location already exists.
       if (is_dir($request['project_install_location'])) {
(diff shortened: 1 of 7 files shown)


AdamPS:
This PR currently includes some SVN-specific changes.  However I could remove them if needs be.

Please can you let me know if you have any interest in taking these fixes?  If not, then please can we close the issues to save anyone else wasting time creating PRs?

AdamPS:
OK, looks like maintainers don't want to merge this PR because drush pm is deprecated.

PR #2089, opened 2016-03-21
Improved the compatibility table

damienmckenna:
Simplified some of the existing columns of the compatibility table and added a new "Main changes" column.


weitzman:
Thanks. I dont see how we can fit enough info in Main changes to be useful. Text wrapping would hurt legibility.  I think a [changelog](http://keepachangelog.com/) would be better.

damienmckenna:
There goes my dream of Drush being rewritten in COBOL ;-)

Issue #2093, opened 2016-03-23
rsync between two remote directories fails


paulmer:
drush rsync @site1:%files/ @site2:%files should copy the contents of the files directory in site 1 to the files directory in site 2, however, this command fails with an error similar to the following:

rsync: link_stat "/tmp/drush_tmp_1458747200_56f2b740a52ef/files" failed: No such file or directory (2)

This is due to a flaw in _drush_core_rsync_both_remote.  The comments correctly describe 3 possible cases and the sequence of rsync commands that must be executed.  The above command matches case 3, and the sequence of rsync commands that should be executed are

rsync @site1:dir/ /tmp/tmpdir
rsync /tmp/tmpdir/ @site2:location

However, the final paragraph of the comments incorrectly states "...and the second rsync always copies from $tmpDir/basename($source_path)."  Clearly case 3 does not include basename($source_path) as part of the path since the contents of tmpdir are the contents of $source_path, not the directory $source_path.  The code implements this erroneous comment and generates the incorrect rsync command:

rsync /tmp/tmpdir/files @site2:location

which as expected is falling with a "No such file or directory" error.  In reality the source path for the second rsync should include basename($source_path) only if $source_path does not end with a "/".

weitzman:
See #2094

PR #2094, opened 2016-03-23
PR: rsync between two remote directories fails

paulmer:
Modification to correct comments and append the basename of the source path to the intermediate directory only when the source is a specific file or directory.


weitzman:
Looks good to me. I'll leave for a couple days in case @greg-1-anderson can review. Background info is in #2093.

greg-1-anderson:
Yes, this looks good. The only test failures are for Drupal 6, and those tests have been removed from the master branch, so I think this is good to merge.

Issue #1582, opened 2015-08-29
drush_sql_register_post_sync_op() query with braces enclosing table does not work


AohRveTPV:
Works when using `drush sql-sanitize`:
`drush_sql_register_post_sync_op('password_policy_history', dt('Delete all Password Policy history table entries'), "DELETE FROM password_policy_history;");`

Does not work:
`drush_sql_register_post_sync_op('password_policy_history', dt('Delete all Password Policy history table entries'), "DELETE FROM {password_policy_history};");`

The former, though, presumably does not work for sites using table prefixes.  So there seems to be no single way to write the query that works for all configurations.

Using Drush Git master latest.

This issue arose from this issue:
https://www.drupal.org/node/2555053

mstrelan:
sql_drush_sql_sync_sanitize() includes checks to determine if the table name should be wrapped in braces. For me this seems to wrap the table names regardless because my db prefix is apparently an array and therefore not empty. Regardless of whether or not it should be wrapped, ultimately the query is written to a temporary file and read directly by mysql, so if there are braces they will throw an error. Braces should be replaced by the actual prefix.

weitzman:
Drush sql-sanitize replaces braces when the --db-prefix option is provided.

AohRveTPV:
> Drush  sql-sanitize replaces braces when the --db-prefix option is provided.

But it errors when --db-prefix is not provided and braces are used in the query.  So it seems to be impossible to write a query for sql_drush_sql_sync_sanitize() that works for both sites using and not using database prefixes.

weitzman:
Oh crap. That does look like awkward. Today, you have to write your sanitize hook like Drush core does it in sql_drush_sql_sync_sanitize():
```
if (drush_get_option('db-prefix') || !empty($databases['default']['default']['prefix'])) {
    $wrap_table_name = TRUE;
  }
  else {
    $wrap_table_name = FALSE;
  }
```
Or more tersely:
```
$wrap_table_name = drush_get_option('db-prefix') || !empty($databases['default']['default']['prefix'])`. 
```

Then you add braces based on $wrap_table_name.

Ping @bjeavons since this went in with fdffc6519c9d6e3f02c7f85ba733a8c635680cbf. 

I'm not sure why we don't unconditionally to the prefix replacement. This is such a useless part of Drupal that I hesitate to spend much time making it better. It could easily cause more harm than help.

weitzman:
[...]

PR #2100, opened 2016-03-28
Fix #1869 and #1582. Better support for db prefixes during sql-sanitize

### LINKED ISSUE #1869: Drupal 8: sql-conf returns an array instead of a string
Drush is reading the database configuration prefix as an array, not a string.

This is causing several Drush operations to fail because the array index is set (as an array) when it shouldn't be.  So Drush assumes that there is a prefix, and we end up with extra "{}" around table names in SQL.  This causes a syntax error.

Known operations affected:
* drush sql-conf
* drush sql-sanitize
* drush [sql-sync-pipe](https://www.drupal.org/project/drush_sql_sync_pipe) --sanitize (see [Fix --sanitize option](https://www.drupal.org/node/2537146))

For example, with the following in settings.(local.)?php:

```php
$databases['default']['default'] = array(
  ...
  'prefix' => '',
  ...
);
```

... or leaving out "prefix" entirely, `drush sql-conf` returns:

```php
Array
(
    [driver] => mysql
    [database] => DATABASE
    [username] => USER
    [host] => localhost
    [port] => 
    [prefix] => Array
        (
            [default] => 
        )

)
```

If "prefix" is set, you get:

```php
Array
(
    [driver] => mysql
    [database] => DATABASE
    [username] => USER
    [host] => localhost
    [port] => 
    [prefix] => Array
        (
            [default] => foo
        )

)
```

... instead of a string.

This is not a problem with Drush 8 on sites running Drupal 7.  It only happens on D8 sites.

weitzman:
Database prefixes are still discouraged by Drush maintainers.

### DIFF
--- a/commands/sql/sql.drush.inc
+++ b/commands/sql/sql.drush.inc
@@ -477,21 +477,10 @@ function drush_sql_sanitize() {
     return drush_user_abort();
   }
 
-  $sanitize_query = '';
-  foreach($options as $id => $data) {
-    // Enable prefix processing when db-prefix option is used.
-    if (drush_get_option('db-prefix')) {
-      if (drush_drupal_major_version() >= 7) {
-        $data['query'] = Database::getConnection()->prefixTables($data['query']);
-      }
-      else {
-        $data['query'] = db_prefix_tables($data['query']);
-      }
-    }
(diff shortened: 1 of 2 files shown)


weitzman:
Merged in 61056c4

PR #2102, opened 2016-03-28
End of life for Drupal 6.

weitzman:
These users are encouraged to continue using Drush 8.


greg-1-anderson:
I'd love to see Drush 9 continue to support Drupal 6; however, I don't imagine that there is anyone who has the time and resources to make that happen. Drush 9 will start moving forward quickly, and that would be a lot of effort to put into an EOL release.

+1 for removal.

Issue #502, opened 2014-03-05
pmi --format=csv skips missing fields


jcheek:
Hello,

I am parsing the output of **drush pmi --format=csv** with certain fields enabled: **--fields=extension,version,path,status**.  All is fine unless one of the fields is missing in the  module, e.g. version.

When displaying output in another format, it may be acceptable to simply delete the information if it doesn't exist, but in CSV, that's not ok.  A blank entry must be output or else all of the fields after it on that line are thrown off.

As an example, here's drush's output for argfilters, a sandbox project at http://drupalcode.org/sandbox/itangalo/1086472.git.  It's .info file has no version, and **drush pmi** doesn't (normally) really care:

```
$ drush pmi argfilters
 Extension     :  argfilters
 Project       :  Unknown
 Type          :  module
 Title         :  Views arguments in filters
 Description   :  Allows you to use argument values in filters, thereby
                  leveraging functionality such as operators.
 Date          :
 Package       :  Experimental
 Core          :  7.x
 PHP           :  5.2.4
 Status        :  enabled
 Path          :  sites/all/modules/contrib/views_arguments_in_filters
 Schema        :  module has no schema
 version
 Files         :  none
 Requires      :  views
 Required by   :  autocompleteinator, autocompleteinator_helper,
                  more_custom_modules_here
 Permissions   :  none
 Configure     :  None
```

_version_ is just shown as empty, above, but with **--format-csv**, it doesn't work; it simply removes the entire field, instead of (correctly) adding a blank field:

```
$ drush pmi --format=csv --fields=extension,version,path,status argfilters
argfilters,sites/all/modules/contrib/views_arguments_in_filters,enabled
```

Four fields were requested, three were returned.  i can't (easily) parse this.

Please fix!  This is in 6.2.0.

greg-1-anderson:
Sorry this was never fixed in the pre-Drush-9 output format engine, which is now in maintenance only. Drush 9 + output formatters should do this correctly.

Issue #333, opened 2013-12-17
Handle --fields better in output formatters that do not explicitly support fields.


bjeavons:
The drush site-alias command with fields flag does not work as expected.

According to the help for sa, --fields should allow specifying the field to print.

Example, for site alias

```
$aliases["example.env"] = array (
  '#name' => 'example.env',
  'site' => 'example',
  'env' => 'env',
  'root' => '/var/www/html/example.env/docroot',
  'remote-host' => 'host.example.com',
  'remote-user' => 'exampleuser',
);
```

calling drush @site.env --fields=remote-host should print host.example.com but it does not.

drush 6.2.0

hansfn:
So, the consensus is that fields shouldn't be supported for formats like json, yaml and similar? In my opinion they should, and it shouldn't be that hard to remove (unset) the non-selected fields. I would like to know, since I plan to submit a documentation patch for the field support for different output formats.

greg-1-anderson:
Maybe fields could be supported for yaml et. al.; someone would need to submit a PR, though.

weitzman:
I also would like to see --fields support for json and yaml.

4garfield:
do you have any plan to support the fields output for json,yaml?
till now, the fields output is still only supported by --format=csv but there is still no document

greg-1-anderson:
The Drush output formatters code has a very messy "middle layer" that needs some work. It's possible that this enhancement could be made without touching that part of the code, but for my part, I have decided to defer all output formatter work in favor of the rewrite at https://github.com/consolidation-org/output-formatters. I was actually considering this feature in that module a short time ago, but as of yet it is unimplemented.

PRs here or there would be welcome.

greg-1-anderson:
This is now supported in the [output-formatters](https://github.com/consolidation-org/output-formatters) project.

PR #2103, opened 2016-03-29
Use Symfony Console application and annotation commands in Drush

greg-1-anderson:
Setting the DRUSH_SYMFONY environment variable makes the Symfony Console Application the main command runner. Otherwise, the original Drush command dispatcher is used first, and the Symfony Console Application is used as a backup.

There's a lot of work needed here before Drush will be fully functional under the Symfony Console Application. One interesting thing about Symfony Console is that option processing is not fully done until `$application->run()` is called. This means that work like site selection must be [done via event listeners](https://github.com/drush-ops/drush/pull/2103/files#diff-96d45678f9dddb3bd857700af7309bffR18).

The list below enumerates what needs to be done, cross-referenced to other issues, PRs or commits that are relevant. Items that have existing code are checked off, but nothing should be considered to be "done" until reviewed, discussed and merged.

- [ ] Command dispatching
  - [x] @hook annotations to allow annotated commands to hook other commands https://github.com/consolidation-org/annotation-command/pull/6https://github.com/consolidation-org/annotation-command/pull/6
  - [x] Command discovery https://github.com/consolidation-org/annotation-command/pull/10
    - [x] Decide how to find built-in command classnames 9410e40
    - [x] Decide how to find and load module cli commands https://github.com/consolidation-org/annotation-command/issues/12 9410e40
    - [x] In Drupal 8, disable filesystem scanning and use the service container to identify commandline tools for modules. 352cec0
  - [ ] Decide how to test requirements, and filter out commands where they are not met
  - [ ] Redispatch to remote servers
- [ ] Bootstrapping
  - [x] If command classes implement [BootstrapAwareInterface](https://github.com/consolidation-org/bootstrap/blob/master/src/BootstrapAwareInterface.php), give them a reference to Boot via setter injection https://github.com/consolidation-org/bootstrap/pull/1
  - [ ] Command classes that do not implement this interface are always fully bootstrapped
  - [ ] Bootstrap object should be able to provide specific implementations of certain services (e.g. SQL access).
- [ ] Option processing
  - [ ] Allow default options to be specfied for global commands
  - [ ] Handle --global-option=0 options
  - [ ] Handle --no-global-option synonyms
- [ ] Configuration file handling
- [ ] Alias file handling
- [ ] Backend invoke
  - [x] Pass annotated command results to backend invoke 4d110d2
- [ ] SQL classes
- [ ] Output formatting
  - [x] Integrate Robo logging with Drush, so that interpolation etc. works the same way https://github.com/drush-ops/drush/pull/2103/commits/e5c357b5c94c209ebcbff8697e0d091d397e7086
  - [x] Always print LogLevel::NOTICE, and move existing NOTICE logs to INFO #2115
  - [x] Formatter plugin interface https://github.com/consolidation-org/annotation-command/pull/7 / [consolidation-org/output-formatters](https://github.com/consolidation-org/output-formatters)
  - [x] Provide existing basic formats
    - [x] string
    - [x] yaml
    - [x] json https://github.com/consolidation-org/output-formatters/pull/2
    - [x] print-r https://github.com/consolidation-org/output-formatters/pull/3
    - [x] var_export https://github.com/consolidation-org/output-formatters/pull/4
    - [x] list https://github.com/consolidation-org/output-formatters/pull/5
    - [x] table https://github.com/consolidation-org/output-formatters/pull/1
    - [x] csv https://github.com/consolidation-org/output-formatters/pull/6
    - [x] php [40f238c](https://github.com/consolidation-org/output-formatters/commit/40f238cf2b068b7be66e79a33d15d6bddf5d51f5)
  - [x] Drop old formats that can now be handled by one of the standard formatters
    - [x] key-value now handled via --table
    - [x] key-value-list now handled via --sections format https://github.com/consolidation-org/output-formatters/pull/17
    - [x] csv-list now handled via the --csv format
    - [x] nested-csv now handled via --csv with a cell renderer
    - [x] csv-or-string only used by nested-csv, and therefore no longer needed
  - [x] Ensure [old Drush outputformat](https://github.com/drush-ops/drush/labels/area-outputformat) bugs are fixed by consolidation/output-formatters (n.b. commands must be rewritten as annotation commands to benefit)
    - [x] #502: csv tables should not skip missing or empty fields
    - [x] #333: handle fields better for output formats that do not explicitly support fields
    - [x] #130: omit field names
  - [ ] Decide what to do with unusual formats
    - [ ] message (*Only used by search-status command. Drop & replace with a key-value --list?*)
    - [ ] variables (*Unused?*)
    - [ ] config (*Default format for site-alias command, pipe format for variable-get command. Drop and replace with yaml?)
    - [ ] labeled-export (*Unused?*)
    - [ ] html (*Used by help*)
  - [ ] Configure Symfony Table styling to a format we like
  - [x] Provide a mechanism to determine which formatters are compatible with a given command
- [ ] Existing command compatibility layer
  - [X] Do legacy dispatching first; dispatch with Symfony Application if no legacy command found https://github.com/drush-ops/drush/pull/2103/commits/0057d55b9ccf6a08e181524786154e950adef686
  - [x] Convert an example command, core-init, to [InitCommand](https://github.com/drush-ops/drush/blob/symfony-annotation-command/lib/Drush/Command/InitCommand.php)
  - [x] Convert core-eval to [EvalCommand](https://github.com/drush-ops/drush/blob/symfony-annotation-command/lib/Drush/Command/EvalCommand.php), so that we can try basic formatters.
  - [ ] Create a wrapper Symfony Command to call old commands and hooks from the Symfony Application dispatcher
    - [ ] Determine mapping from Drush procedural hooks to new hooks #2116
    - [ ] Decide if we should continue to support rollback hooks, or if we can rely on Robo Collections for this.  
  - [ ] Ensure that Drush notify still works with Symfony commands
- [ ] Help changes #2106
  - [ ] The `list` command shows command aliases as duplicate entries; duplicates should be hidden
  - [ ] All aliases and all usages appear in the "Usage" section; should be split up
  - [ ] All global options are shown all the time
  - [ ] @topic is needed, so help text can show related topics
  - [x] Commands should provide their return type, and the formatter manager should provide the valid formats that work with a give return type. [output-formatters 43aeb9a](https://github.com/consolidation-org/output-formatters/commit/f43aeb9a8621b8955879bc24823fe27469fe13e1) + [annotated-command 2d7d52a](https://github.com/consolidation-org/annotated-command/commit/2d7d52a4b585293d6f5a98ad308f7809d958980b)


weitzman:
Nifty. It I like annotation classes as commands.

Great checklist - lots of work to do! Maybe we should split out the major ones into their own issues and link from there. Lets use the drush9 milestone to collect such issues.

This notice is noisy: `Notice: Undefined index: default-value in /Users/moshe.weitzman/c/h/drush/includes/preflight.inc on line 266`

jhedstrom:
Awesome to see this taking shape!

Chi-teck:
> Decide how to find and load module cli commands

You might find this useful:

Drupal Console discovers commands though name conventions.
https://github.com/hechoendrupal/DrupalConsole/blob/master/src/Helper/CommandDiscoveryHelper.php

DCG discovers commands using PHP reflection
https://github.com/Chi-teck/drupal-code-generator/blob/master/src/GeneratorDiscovery.php

There is core conversation about Symfony Console integration with some interesting points about command discovery.
https://www.drupal.org/node/2242947

weitzman:
Thanks a lot for those links. Will read up on command discovery.

I just played with DCG and reviewed the code. Terrific work!

greg-1-anderson:
Thanks for the references; I was unaware of DCG.

I am ambivalent about the DCG discovery technique vs the DrupalConsole discovery technique. Being able to use instanceof is clearly a good thing; however, DCG at its core is still using filesystem iteration to discover classes, and furthemore it instantiates classes it does not need in order to use instanceof.

In the balance, I think I prefer the model that commandfiles must conform to some naming convention in their filenames, e.g. *Command.php, as DrupalConsole does.

Chi-teck:
> *Command.php

There should be a way to distinguish Drush commands from Drupal Console commands and other types of commands. Perhaps _*DrushCommand.php_ or _Drush/*Command.php_

greg-1-anderson:
Actually, it is my goal to make it possible to write a Drupal 8 cli module that is loadable by both DrupalConsole and Drush. So, we'll instantiate any *Command.php, but will ignore anything that is an instanceof Symfony Console.  Drush will only support [annotation commands](https://github.com/consolidation-org/annotation-command) (and legacy procedural Drush commands, of course); DrupalConsole will use instanceof to distinguish between annotation commands and native Drupal Console commands. I'm working on the Drupal Console PR right now; I will cross-reference here when it is ready.

greg-1-anderson:
Converted one command as a proof-of-concept, and got the tests to pass. Still a lot of work remaining, but a good milestone.

lib/Drush/Command/InitCommand.php
```diff
@@ -0,0 +1,174 @@
...
+namespace Drush\Command;
+
+/**
+ * @file
+ *   Set up local Drush configuration.
+ */
+
+use Drush\Log\LogLevel;
+
+class InitCommand extends \Robo\Tasks
```
jhedstrom: I see the `@option` annotations below on the individual methods, but should there be an annotation on the class too for command discovery? Or are those intended to be discovered by other means?

greg-1-anderson: Commandfile discovery hasn't been designed yet, exactly. The basic mechanism is similar to what Drush does today: find all commandfiles, and then register all of the commands they define. Note that annotations on the class, or using instanceof are both possibilities for secondary techniques for command discovery / selection; however, there needs to be an additional initial filter function as well, as we would not want to load every class file available just to determine which are commandfiles.

DrupalConsole searches for files called *Command.php (in certain paths), and registers all of those. We should probably do something similar. Suggestions / PRs welcome.

greg-1-anderson:
Hooks and formatters are now working; I think the next step is to actually load a commandfile provided by a module, and see if we can make that work. 

n.b. Module commandfiles will invariantly given a full bootstrap. We'll support allowing global and site commands to set their bootstrap level later.

greg-1-anderson:
I decided that [Chi-teck was right, above](https://github.com/drush-ops/drush/pull/2103#issuecomment-204073730); annotation commandfiles should use different filename patterns than Symfony commands. Discussion at https://github.com/consolidation-org/annotation-command/issues/12

greg-1-anderson:
I renamed consolidation/annotation-command to consolidation/annotated-command. GitHub will automatically redirect the old links above to their new location in the new project name.

lib/Drush/CommandFiles/EvalCommandFile.php
```diff
@@ -0,0 +1,34 @@
...
+   *   Copies a file whose path is determined by an environment's variable.
+   *   Note the use of double quotes so the variable $HOME gets replaced by
+   *   its value.
+   * @usage php-eval "node_access_rebuild();"
+   *   Rebuild node access permissions.
+   * @default-format var_export
+   */
+  public function phpEval($php, $options =
+    [
+      'format' => 'var_export',
```
weitzman: Why does an annotation command specify @default-format and $options['format']? Also, I have not seen this use of use of an associative array as default value. Is that new in PHP?

greg-1-anderson: I put in both @default-format and `'format' => 'var_export'` because I was experimenting; only one is needed on any given command. If you are using a Symfony Console command (whose parameters are always InputInterface, OutputInterface), then you must use @default-format.

The associative array as a default value, as far as I know, has always worked in php, but I have only verified this as far back as php 5.4.  Its use in this context -- to define the default values for your options -- originated in Robo.

greg-1-anderson:
Going to investigate service tags, similar to what was done in an earlier Drush PR.  See: [tags example](https://github.com/drush-ops/drush/pull/2018/files#diff-dac902dcbdcdc3fc34cb619bd81d3a54R36) and the [compiler pass](https://github.com/drush-ops/drush/pull/2018/files#diff-504a9db85f17a193e7a9b37071fc9904R21).

Actually, it looks like it won't be necessary to use a compiler pass here.

tests/resources/modules/d7/woot/Command/.gitignore
```diff
@@ -0,0 +1 @@
+WootCommands.php
```
weitzman: Is this .gitignore really needed?

tests/resources/modules/d8/woot/src/Command/GreetCommand.php
```diff
@@ -0,0 +1,46 @@
+<?php
+namespace Drupal\woot\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class GreetCommand extends Command
```
weitzman: Lets document that this is an example of a "straight" Console command.

tests/resources/modules/d8/woot/woot.services.yml
```diff
@@ -0,0 +1,14 @@
...
+  woot.commands:
+    - woot.manager
+    - woot.command
+    - greet.command
+services:
+  woot.manager:
+    class: Drupal\woot\WootManager
+    arguments: ['@current_user']
+  woot.command:
+    class: Drupal\woot\Command\WootCommands
```
weitzman: IMO service name and classname for woot.command should both be singular or both be plural

tests/resources/modules/d7/woot/Command/.gitignore
greg-1-anderson: Rather than make two copies of the same 'WootCommands', my scripts copy or symlink the Drupal 8 version into the Drupal 7 test module directory.  I don't want this test artifact to be accidentally committed. It might also be a good idea to just erase the duplicate after the test runs.

tests/resources/modules/d8/woot/woot.services.yml
greg-1-anderson: Yes, good point. The 'woot.commands' parameter adds to this confusion; fortunately, that particular entry will go away with tagged services. I'll clean up the service name after that.

greg-1-anderson:
I merged in #2169, so this PR now uses tagged services to find commands in Drupal 8 modules.  In order for this to work, we need to create a subclass of DrupalKernel.  A core patch that adds the functionality we need is here: https://www.drupal.org/node/2718933

greg-1-anderson:
I squashed and rebased this with the master branch.

greg-1-anderson:
Continued in #2843.

Issue #1523, opened 2015-08-02
Postgresql schema not used by drush


jhmnieuwenhuis:
Hi
When I install drupal 8 rc13 and the latest drush 8, it does not work if I installed drupal with a prefix (schema name other then public).
If I run drush pm-list I get the well known " needs a higher bootstrap level to run"

When I install drupal without a prefix ( so in the public schema) there is no poblem running drush pm-list.

Please advise....

Regards,

Hans Nieuwenhuis

43n79w:
I too am seeing this issue with beta14 (beta15 doesn't appear to install properly using a prefix/schema on postgresql).

lifenstein:
+1

Drush version 8.0.2, PostgreSQL 9.4.5, RHEL 7.2

`drush -d role-list` shows that drush is trying to use the `public` schema:
`SELECT tablename FROM pg_tables WHERE schemaname='public'`

fiasco:
That is hardset in the pgsql driver for drush: https://github.com/drush-ops/drush/blob/9335be17e76b990ee22d6ef203da09d9cdfc5e15/lib/Drush/Sql/Sqlpgsql.php#L5

RoSk0:
Added pull request #2567 to fix this issue.

RoSk0:
Fix from the pull request is going to work only when schema support will be fixed in core with help this issue https://www.drupal.org/node/1060476.

tstoeckler:
Can you elaborate how you're using this? I tried this out locally with Drupal installed in a separate schema and am getting
```
psql:/tmp/drush_QBlGVG:1: ERROR:  unrecognized configuration parameter "tables"
```
when running
```
drush sqlq "SHOW TABLES;"
```

Also, as you yourself note Drupal itself doesn't work with a separate schema, so I'm having a hard time wrapping my head around this.

Would be great to have some insights, thanks!

fiasco:
`SHOW TABLES` is not supported in PostgreSQL - http://stackoverflow.com/questions/769683/show-tables-in-postgresql

tstoeckler:
Hmm... strangely it seemed to be working for me without the patch. In any case would be great to know how to actually use this so I (or seomeone else) can verify the PR works as expected.

Renrhaf:
Tried the patch given in https://github.com/drush-ops/drush/pull/2567 and it fixes the issue for me 👍 
[...]

PR #2109, opened 2016-03-31
#1523 - Allow a dynamic schema on PostgreSQL

fiasco:
This hopefully fixes a breakage in PostgreSQL when using a non "public" schema.

### DIFF
--- a/lib/Drush/Boot/DrupalBoot.php
+++ b/lib/Drush/Boot/DrupalBoot.php
@@ -497,6 +497,13 @@ abstract class DrupalBoot extends BaseBoot {
       if (!is_array($prefix)) {
         $prefix = array('default' => $prefix);
       }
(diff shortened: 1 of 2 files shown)


Renrhaf:
@fiasco did you try this one https://github.com/drush-ops/drush/pull/2567.patch ? I tested and it worked, I was wondering if you had done any other modifications that might be useful.

fiasco:
@Renrhaf this is a pretty old PR that came long before #2567. So no I've not tested it and probably won't as this issue isn't currently affecting me.

Renrhaf:
Ok, thanks ! Maybe we can close this one then ;)

cafuego:
💙  @fiasco for saving my butt with this patch :-)

ipumpkin:
need this for 8.x

Issue #1189, opened 2015-02-17
drush runserver won't generate images for image styles with Clean URLs enabled


dsnopek:
When running a Drupal site with `drush runserver` and Clean URLs enabled, attempting to access an image for an image style (which hasn't yet been generated) will give 404, rather than cause Drupal to generate the image.

Disabling Clean URLs will allow the image to be correctly generated, but obviously affects lots of other things on the Drupal site.

We (the Panopoly project) are using `drush runserver` in our automated Behat tests, and so anything that uses an image style is getting a broken image and messing up the tests. But disabling Clean URLs also breaks a number of our tests which deal with paths, aliases and redirects. And, of course, no one runs with Clean URLs disabled in real life, so it would make our tests more artificial...

This issue was reported on D.o at one time, but got no response before the migration to GitHub: https://www.drupal.org/node/1961656

kevin-coyle:
Just a little more detail. I'm getting this too running it on Ubuntu 14.04. PHP Version 5.5. GD is installed and working.

nvaken:
Same problem here, any suggestions?

kevin-coyle:
Just noticed something. If you run the clean urls test it states for me that it fails... Could that be part of the issue?

Also I'm getting watchdog errors like this 

````
[Fri Jun 12 00:30:01 2015] Watchdog: Source image at public://gemmapfister_head.jpg not found while trying to generate derivative image at public://styles/team_image/public/gemmapfister_head.jpg. | severity: 5 | type: image | uid: 1 | 127.0.0.1 | http://127.0.0.1:8888/?q=sites/default/files/styles/team_image/public/gemmapfister_head.jpg&itok=e_m3PvZ3 | http://127.0.0.1:8888/ | 
````

richardbporter:
I've got this issue as well. Did anyone find a solution/workaround?

badrange:
On my computer it worked after downgrading Drush from 7.0.0 to 6.6.0.

dsnopek:
Hm. When I created this issue originally I was testing with the latest drush 6 release available at the time which appears to be 6.5.0. Is it really possible that this got fixed in 6.6.0? I'll re-test again when I have chance...

badrange:
Turns out I was very unclear in my previous comment. Yes it does work in Drush 6.6.0 but it does not work in Drush 7.0. 

[...]

PR #2118, opened 2016-04-06
#1189 Add a D8 style router include for runserver.

darthsteven:
This should be a fix for #1189 but I'm not sure what other side effects it might have.


weitzman:
I'm not sure either. I think this has been proposed before but nobody could really say if it was a good idea or not.

weitzman:
Dupe of #1658

PR #2128, opened 2016-04-12
Prevent locale-update from executing same operation multiple times

gapple:
Currently each project returned from `locale_translation_get_status()` results in a new entry in `$langcodes`.  When `$langcodes` is used to determine the batch operations, an operation is added for each module and each duplicate value in `$langcodes`.  For 11 translatable modules and one language on a site this means over 300 mostly duplicate operations instead of the 33 required.


tstoeckler:
Thanks! Verified that it fixes the issue.

Notes:
- I generally like to use the $strict = TRUE third param with in_array() because with PHP's type juggling weird bugs can arise otherwise, so I consider that a best practice. But since this is mostly not used in Drupal or Drush merged anyway. Just wanted to note that here.
- I found that we don't validate that locale module is installed. Will fix that later.

Issue #2132, opened 2016-04-15
Drush 7 composer.lock compatibility with PHP 5.3


JorisVanEijden:
The install.md document states Drush 7 requires PHP 5.3.0+
The composer.json also lists PHP >=5.3.0
But the composer.lock file requires symfony/var-dumper 2.8.0 and symfony/yaml 2.8.0 which require PHP >=5.3.9

The symphony/* 2.* requirements in composer.json also require PHP >=5.3.3 so anything less than 5.3.3 isn't going to work anyway.

**Please keep composer.lock compatible with the listed minimum requirements.**
Maybe add to the composer.json:
`    "config": {
        "platform": {
            "php": "5.3.3"
        }
    }`
Or remove the composer.lock from the repository.

See also Symphony's own discussion about this issue: https://github.com/symfony/symfony-standard/issues/858

weitzman:
just read that issue. There isnt a great solution. I'm inclined to remove composer.lock. Any comments?

JorisVanEijden:
Removing it looks like the best solution.

greg-1-anderson:
Drush 7.x does not have a .phar distribution, but this issue, when it recurs for other components in the future, has implications for .phar files.

In cases where `composer require` does **not** bring in the most recent version of all of an apps dependencies when used with the lowest supported version of php, then a different .phar file for each version of php supported. We could build multiple phars (one per php version) via the Travis build, but it would be inconvenient to determine if they were necessary or not.

The composer.lock problem is similar, but Symfony has no facility for supporting multiple lock files. Pinning the php version is not good, for the reasons stated in the Symfony issue.

I think that the only robust solution for the master branch of the project is to ensure that our minimum php >= the minimum php of the latest release of all of our dependencies. This may sometimes be in conflict with our project goals vis-a-vis php support. It would be nice if all of our dependencies would recognize this, and ensure that bumping the php version can only be done with a change in major version number. Semver currently does not require a change in dependencies to be considered breaking unless it is exposed through the library's public API.  For Composer, it is easy for a dependency change to break backwards compatibility, even when the public API does not change -- but it is difficult to determine when this has happened.

weitzman:
Ugh - good point .... We will probably abandon global Drush by the next time this becomes an issue.

PR #2136, opened 2016-04-18
Fix for PHP 5.3.3.

greg-1-anderson:
Backport from master branch: fix for "Exception 'ReflectionException' with message 'Class Drush\Sql\Sql7 does not have a constructor, so you cannot pass any constructor arguments"

See: #544


greg-1-anderson:
See #2132 for discussion of composer.lock.

weitzman:
This branch did not get any fixes for Drupal 6 EOL so we will not get the tests to pass. I think its OK to merge this PR and make a new Drush7 release anyway.

greg-1-anderson:
I think that the folks who are interested in php 5.3 on Drush 7 want it so that they can do Drupal 6 maintenance on old servers still running that version of php.

I think we'll have to rely on community support to get the Drupal 6 EOL fixes backported to this branch. Let's go ahead and merge here, since it gets us one step closer.

PR #2139, opened 2016-04-19
Provide annoted command support for Drush 8.

greg-1-anderson:
Since command hooks are somewhat different in consolidation/annotated-command than in Drush 8, this PR simplifies things and only handles the primary hook for annotated command files. While it would be possible to adapt the new hooks in the old system, that would be more effort; it seems better to let people continue to write their personal policy files using the old hook system until they are ready to upgrade to Drush 9. Supporing only the primary hook in Drush 8 should meet the use case of Drupal 8 module writers who just want to write a simple command line interface that they also want to work in Drush 9.

This code is still in progress; more work is needed to make help commands and validation work right. However, in its current form, this will run the example in the "Modern CLI tools" patch at https://www.drupal.org/node/2703131.


greg-1-anderson:
consolidation/annotated-commands currently requires php 5.5.  It inherits this requirement from Symfony 3.  Note, however, that this PR does not use Symfony/Console, and the components it does use work fine all the way back to php 5.3.

greg-1-anderson:
So, there are three potential routes forward vis-a-vis avoiding the Symfony/Console requirement here:

- Make a new Consolidation project that contains only the command info parser and the command discovery class. Require this project from consolidation/annotated-command.
- Fork a new project in drush-ops that contains only the command info parser and the command discovery class. Use this project from older versions of Drush, and nowhere else.
- Copy the two classes we need into the Drush 8 and Drush 7 source trees.

The first option is what might be considered the "technically correct" solution; however, there's a maintenance code to splitting out unit tests, and lots of very small repositories are harder to maintain. We need to keep the right balance between separability and maintainability / overhead work. That's why I think that the last option might just be best; just orphan the classes, maintain them to a level of functionality sufficient for the reduced functionality in this PR, and let the main project move forward. There is a cost to ensuring that any needed bug fixes are backported, but I don't anticipate many changes to command info and command discovery once we finish with #2103, so perhaps this is okay.

includes/annotationcommand_adapter.inc
```diff
@@ -0,0 +1,101 @@
...
+ * Search for annotation commands at the provided search path.
+ * @see _drush_find_commandfiles()
+ */
+function annotationcommand_adapter_discover($searchpath, $phase = false, $phase_max = false) {
+  if (empty($searchpath)) {
+    return;
+  }
+  // Check to see if this is the Drush searchpath for instances where we are
+  // NOT going to do a full bootstrap (e.g. when running a help command)
+  if (($phase == DRUSH_BOOTSTRAP_DRUPAL_SITE) && ($phase_max < DRUSH_BOOTSTRAP_DRUPAL_FULL)) {
```
weitzman: This bootstrap test has been buggy many times in the past few years. I rather hate to copy the logic here. Would rather have it centralized if feasible.

includes/annotationcommand_adapter.inc
```diff
@@ -0,0 +1,101 @@
...
+      }
+      $command['required-arguments'] = $required_arguments;
+      foreach ($commandinfo->getOptions() as $option => $default) {
+        $command['options'][$option] = $commandinfo->getOptionDescription($option);
+      }
+      $command += drush_command_defaults($command_name, $commandfile, $commandfile_path);
+      $commands[$command_name] = $command;
+    }
+  }
+  drush_set_context('DRUSH_ANNOTATED_COMMANDS', $commands);
```
weitzman: DRUSH_ANNOTATED_COMMANDS represents the set of commands for the last searchpath we traversed? That seems error prone.

weitzman:
Tentatively agreeing that third option makes sense.

includes/annotationcommand_adapter.inc
greg-1-anderson: DRUSH_ANNOTATED_COMMANDS is the aggregation of all of the annotated commands found on all of the searchpaths in the current bootstrap.

As an added optimization, it might be beneficial to cache this list, as we do in the case for ordinary commands; however, I do not expect any Drush 7 site to ever have a large number of annotated commands, so this might not be necessary.

includes/annotationcommand_adapter.inc
greg-1-anderson: I am reluctant to rework how bootstrapping works in the Drush 8 branch; that is what would be necessary to centralize it. Move the code here that finds all of the specific module folders into the bootstrap code. This is not necessary for the existing use case, which searches deeply for .drush.inc files.

includes/annotationcommand_adapter.inc
weitzman: Looks to me like this line overwrites the value of DRUSH_ANNOTATED_COMMANDS with the value for just the current searchpath. I haven't stepped through this in debugger so maybe I am missing something.

weitzman: I think that commandfile cache is important. Its not really relevant if we discover a lot of annotated commandfiles or not. The cache saves us from all the directory traversal.

includes/annotationcommand_adapter.inc
```diff
@@ -0,0 +1,101 @@
...
+function annotationcommand_adapter_discover($searchpath, $phase = false, $phase_max = false) {
+  if (empty($searchpath)) {
+    return;
+  }
+  // Check to see if this is the Drush searchpath for instances where we are
+  // NOT going to do a full bootstrap (e.g. when running a help command)
+  if (($phase == DRUSH_BOOTSTRAP_DRUPAL_SITE) && ($phase_max < DRUSH_BOOTSTRAP_DRUPAL_FULL)) {
+    $searchpath = annotationcommand_adapter_refine_searchpaths($searchpath);
+  }
+  $commands = drush_get_context('DRUSH_ANNOTATED_COMMANDS');
```
greg-1-anderson: $commands starts out with the set of commands discovered the last time this function was called.

greg-1-anderson:
Actually, php 5.4 is not an issue, as it is supported by Symfony 2. Now, both consolidation/annotated-command and consolidation/output-formatters support php 5.4. The latest commit here uses both of these projects directly, so that annotated commands are now dispatched, processed and formatted using the same functions as the master branch. The main difference here is that we still are not using Symfony/Console in this PR.

tests/annotatedCommandTest.php
```diff
@@ -0,0 +1,141 @@
...
+    $this->assertContains('try:formatters --fields=first,third', $output);
+    $this->assertContains('try:formatters --fields=III,II', $output);
+    $this->assertContains('--fields=<first, second, third>', $output);
+    $this->assertContains('Fields to output. All available', $output);
+    $this->assertContains('--format=<table>', $output);
+    $this->assertContains('Select output format. Available:', $output);
+    $this->assertContains('Aliases: try-formatters', $output);
+
+    // Disable the woot module to avoid cross-contamination of the Drupal
+    // test site's database. (not necessary?)
```
weitzman: The Drupal site is wiped after each test class. No need to clean up.

Issue #2143, opened 2016-04-21
PHP Fatal error:  Call to undefined function cache_get() when trying to use cache in file_scan_directory


kenorb:
Version: Drupal 7

I've come up with this [core patch](https://www.drupal.org/node/2710289) which improves the performance of Drupal core which is basically uses `cache_get`/`cache_set` inside `drupal_system_listing()` function.

The problem is that [QA on Drupal.org](https://dispatcher.drupalci.org/job/default/118565/console) fails with:

    08:32:53 Fatal error: Call to undefined function cache_get() in /var/www/html/includes/common.inc on line 5522

I've compared drush's `site_install_7.inc` with `install.core.inc` and it seems in Drupal core we're loading caches before installing the site which isn't the case for drush.

Drupal core in `install_begin_request()` has these lines:

    require_once DRUPAL_ROOT . '/includes/cache.inc';
    require_once DRUPAL_ROOT . '/includes/cache-install.inc';
    $conf['cache_default_class'] = 'DrupalFakeCache';

which loads at least some fake caching in order to prevent some issues. On the other hand, drush loads only few core files such as file, install, common and module inc files. So the idea is to load `cache.inc` along with them in `drush_core_site_install_version()`.

Steps to reproduce the issue:

```
$ drush -y dl drupal-7 --dev && cd drupal-7.x-dev
$ patch -p1 < <(curl -s https://www.drupal.org/files/issues/file_scan_directory-caching-2710289.patch)
$ drush -y si --db-url="sqlite://db.s"
PHP Fatal error:  Call to undefined function cache_get() in ~/temp/drupal7/drupal-7.x-dev/includes/common.inc on line 5522
```

Although it works without applying the patch.

PR: #2144

PR #2144, opened 2016-04-21
GH-2143: Include cache.inc during site-install.

kenorb:
See: GH-2143

This is attempt to fix the issues reproducible by the following steps:

```
$ drush -y dl drupal-7 --dev && cd drupal-7.x-dev
$ patch -p1 < <(curl -s https://www.drupal.org/files/issues/file_scan_directory-caching-2710289.patch)
$ drush -y si --db-url="sqlite://db.s"
PHP Fatal error:  Call to undefined function cache_get() in ~/temp/drupal7/drupal-7.x-dev/includes/common.inc on line 5522
```

### DIFF
--- a/commands/core/drupal/site_install_7.inc
+++ b/commands/core/drupal/site_install_7.inc
@@ -9,10 +9,14 @@ function drush_core_site_install_version($profile, array $additional_form_option
   require_once DRUSH_DRUPAL_CORE . '/includes/install.core.inc';
 
   if (!isset($profile)) {
+    global $conf;
     require_once DRUSH_DRUPAL_CORE . '/includes/file.inc';
     require_once DRUSH_DRUPAL_CORE . '/includes/install.inc';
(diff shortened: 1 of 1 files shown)


weitzman:
Sorry, but that patch has no chance of getting into D7 five years after its release. If you want to get it to pass testbot, Just make sure your caching is never called from during the MAINTENANCE MODE.

kenorb:
The goal of patch is actually calling cache during installing. And it works, but not when installing site with drush. So bug is specifically with drush how it's trying to fake the Drupal bootstrap. Probably patch won't be committed into core, but I'm sure some people will decide to patch their core to achieve some better performance for high-traffic sites when every second counts. Currently I can't pass CI to run the tests, since it's failing with drush. The CI passed for drush, so I don't see any side effects of merging it in into drush. But it's up to you.

weitzman:
Yes, I understand.

Are we missing a load of cache system for Drupal 8? That would need fixing at same time.

PR #2145, opened 2016-04-22
Remove wget dependency from Drush install docs

christopher-hopper:
### Problem/Motivation
The install commands only work on systems where `wget` is already installed and available. 

### Solution
By using PHP to download the `drush.phar` code file instead, we ensure compatibility with any system shell that meets the minimum requirements to use Drush.

We can further simplify the install by saving Drush using the file name `drush`, minus the `.phar` extension. This is the command name that people expect and will use.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -15,7 +15,7 @@ sudo mv drush /usr/local/bin
 # Optional. Enrich the bash startup file with completion and aliases.
 drush init
 ```
-    
+
 * MAMP users, and anyone wishing to launch a non-default PHP, needs to [edit ~/.bashrc so that the right PHP is in your $PATH](http://stackoverflow.com/questions/4145667/how-to-override-the-path-of-php-to-use-the-mamp-path/10653443#10653443).
 * We have documented [alternative ways to install](http://docs.drush.org/en/master/install-alternative/), including [Windows](http://docs.drush.org/en/master/install-alternative/#windows).
 * If you need to pass custom php.ini values, run `php -d foo=bar drush.phar --php-options=foo=bar`
--- a/docs/install.md
+++ b/docs/install.md
@@ -2,15 +2,15 @@ Install/Upgrade a global Drush
 ---------------
 ```sh
 # Download latest stable release using the code below or browse to github.com/drush-ops/drush/releases.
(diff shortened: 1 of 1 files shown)


weitzman:
Sure, this removes a dependecy. However, not a single person has opened an issue about missing wget. I think thats very very. Also, the redirection is a bit obtuse. Am undecided about this PR. Others are welcome to chime in ... Thanks for the contribution.

greg-1-anderson:
I tend to think that "or browse to github.com/drush-ops/drush/releases" sufficiently covers folks who do not have wget.  However, the instructions in this PR are a bit shorter, and should be understandable to most folks.  I'm neutral about it; don't mind leaving the docs as-is, don't mind this way either.

weitzman:
I'm gonna pass on this. Thanks still.

christopher-hopper:
Oh well, worth a try. I saw how they did it over on the Composer download page and thought hey, that's cool. I have run into this when installing on a bare CentOS box before, but as you say, its' really an edge case.

weitzman:
Decided to merge this. I am setting up a new Mac and dont have wget so I figure others will hit this. Thanks for improving our docs.

skyred:
I like removing wget dependency. But with increasing Docker usage, wget is usually not available in the most production oriented images.  

FYI: I am living in China; today I used `php -r "readfile('http://files.drush.org/drush.phar');" > drush` to install on a new server, the command just hangs there for 5 minutes, then finished. Not having a progress feedback is not nice.

I would recommend to use `cURL` as it is more widely available on barebone systems

christopher-hopper:
The intent was a minimal effort install command guaranteed to work on any system meeting the Drush minimum requirements. 

Perhaps due to the size of the Drush download, it does need a progress output. It would mean going back to wget or curl, or writing a simple installer in PHP that then showed progress as it downloaded the full Drush package.

greg-1-anderson:
Maybe it would be better to switch to curl; it seems that `> drush.phar` might make it harder for some folks to diagnose error conditions.

See: http://drupal.stackexchange.com/questions/200670/why-doesnt-drush-respond-to-commands-after-install

PR #2151, opened 2016-04-23
Add missing backslashes for Drupal:: global namespace

theodorosploumis:


weitzman:
In a non-namespaced file, isn't that slash optional? Does omitting it give notices or anything?

theodorosploumis:
Since we use PSR-4 for Drupal 8.x (see http://www.php-fig.org/psr/psr-4/) the Global Namespaces should start with a backslash. It is also documented in D.O. docs but not so clear (search for "global" at https://www.drupal.org/coding-standards/docs).

It is optional if the namespace is already included when a method is called. For example if you call ```\Drupal::foo``` and later you call ```Drupal::bar``` (without \) there is no problem because the global class is already cached. 

Also, it may cause errors and/or warnings if we don't have the backslash (actually, I found this when I have a warning on a Drush command but I can't remember it now).

damiankloip:
No, it would not cause warnings. If the namespace was really needed and it was not present (including global '\' namespace) you would see fatals. The backslash would be completely required to indicate the global namespace for code that is already namespaced. So nothing is functionally broken here.

So yes, the core docs state it should always use the leading backslash, but this is only for consistency. This is just CS clean up.

> For example if you call \Drupal::foo and later you call Drupal::bar (without ) there is no problem because the global class is already cached.

That is not true, you would need to `use` the class first, then you could use a global namespace without the backslashes.

PR #2161, opened 2016-04-26
replace getStorageController() with getStorage()

aaronbauman:


commands/core/views.d8.drush.inc
```diff
@@ -376,7 +376,7 @@ function _views_drush_op($op = '', array $view_names = array()) {
 
   $view_names = MapArray::copyValuesToKeys($view_names);
```
tstoeckler: `MapArray::copyValuesToKeys()` no longer exists. This should be `array_combine($view_names, $view_names)` now. Could you fix that as well?

PR #2168, opened 2016-05-04
Tagged services

greg-1-anderson:
The main question here is whether creating our own subclass of DrupalKernel is a better tradeoff than using service parameters.  I think it probably is.


greg-1-anderson:
Wrong base branch -- closing.  See #2169.

Issue #2173, opened 2016-05-07
sub-option behaves different on command-line than in drushrc


webbj74:
I would like to contribute some enhancements to the `xh.drush.inc`, however it's unclear to me if it is currently working as expected with regard to the sub-options it provides. When I try to specify an xh-path or xh-link via the command-line, I get an error and suggestion to use `--strict=0`. This behavior started in f61531a.

For example:

```
# An expected error:
$ drush --xh
You must provide the xh-path option in your drushrc or on the CLI in order to profile. [error]

# Produces an error since change to help in @f61531a
$ drush --xh --xh-path=$HOME/xhprof
Unknown option: --xh-path.  See `drush help help` for available options. To suppress this error, add the option --strict=0.  [error]

# Produces an error since --xh was introduced 
$ drush --xh --xh-path=$HOME/xhprof status
Unknown option: --xh-path.  See `drush help core-status` for available options. To suppress this error, add the option --strict=0.  [error]
```

Supplying `--strict=0` will produce an xhprof file; but my understanding of [strict options](https://github.com/drush-ops/drush/blob/master/docs/strict-options.md) is that a command using strict options _require that all Drush global option appear on the command line before the Drush command name_. But I'm getting an error regardless.

Also things work as expected if I supply xh-path via drushrc.php, for example:

```
$ cat ~/.drush/drushrc.php
<?php
$options['ssh-options'] = '-o PasswordAuthentication=no -p 22';
$options['xh-path'] = getenv('HOME') . '/xhprof';

$ drush --xh
<shows help and generated an xhprof file in my tmp>

$ drush --xh status
<shows status and generated an xhprof file in my tmp>
```

weitzman:
It looks like suboptions cannot be changed in hook_help_alter() so this code in  `xh_drush_help_alter` does nothing:

```
$command['sub-options']['xh']['xh-link'] = 'URL to your XHProf report site.';
$command['sub-options']['xh']['xh-path'] = 'Absolute path to the xhprof project.';
```

There is similar code in notify_drush_help_alter() so I'm perplexed about how this every worked. I went all the way back to 6.x branch and same errors. ping @greg-1-anderson

greg-1-anderson:
I've never used --xh.  Perhaps this was provided by @msonnabaum?

Option handling is about to get a big overhaul in Drush 9 anyway, although ideally I would expect this should work in the older branches as well.

msonnabaum:
Definitely not me.

[...]

PR #2174, opened 2016-05-07
Adds sub-options for xhprof_enable() flags.

webbj74:
The function `xhprof_enable()` takes flags to control the profiling of CPU, memory, and PHP built-ins. This change would add a drush sub-option for each of the flags.

Related to how sub-options appear to work currently, see #2173.

### DIFF
--- a/tests/xhUnitTest.php
+++ b/tests/xhUnitTest.php
@@ -1,0 +1,108 @@
+<?php
+
+namespace Unish;
+
+/**
+ * Unit tests for xh.drush.inc.
+ *
+ * @group base
+ */
+class xhUnitCase extends UnitUnishTestCase {
+
+  /**
+   * Test various combinations of XHProf flag options.
+   *
+   * @dataProvider xhOptionProvider
+   */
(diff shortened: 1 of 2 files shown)


tests/xhUnitTest.php
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Provides drush XHProf options and the results we expect from xh_flags().
+   */
+  public function xhOptionProvider() {
+
+    if (!defined('XHPROF_FLAGS_NO_BUILTINS')) {
+      define('XHPROF_FLAGS_NO_BUILTINS', 1);
+      define('XHPROF_FLAGS_CPU', 2);
+      define('XHPROF_FLAGS_MEMORY', 3);
```
webbj74: Added these defines because xhprof is not installed by default on Travis CI. Additionally xhprof extension is not available on PHP 7 yet.

weitzman:
This looks great. Sorry it took ages to review. Thanks for the test case as well.

PR #2190, opened 2016-05-18
Fixes #2189: Support 'no-core' tarballs for profile downloads.

### LINKED ISSUE #2189: Support 'no-core' tarballs for profile downloads
We currently have 2 ways to use profiles in Drush Make: either download just the profile, and use recursion to build out the rest, or using the profile as a core, which then bundles in the version of core that was current as of the latest release.

drupal.org publishes `no-core` tarballs too, but we don't currently appear to have a way to use them effectively. Since the `core` tarballs only get built when the distro/profile makes a release, there's no convenient way to use the tarballs when Drupal core gets a security release (for example), and thus requires a new release of the profile/distro.

It'd be nice to have a project option in Drush Make that supported the 'no-core' tarballs, as this would provide a clean work-around for this issue.

ergonlogic:
PR for Travis

### DIFF
--- a/commands/make/make.download.inc
+++ b/commands/make/make.download.inc
@@ -33,9 +33,9 @@ function make_download_pm($name, $type, $download, $download_location) {
     'yes' => TRUE,
     'package-handler' => 'wget',
     'source' => $download['status url'],
-    // This is only relevant for profiles, but we always want the variant to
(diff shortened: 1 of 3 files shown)


weitzman:
Hah, that was simple. Please add docs to example makefile or anywhere else that makes sense. Might be worth a blog post too.

PR #2192, opened 2016-05-19
Add fish shell alias to install-alternative documentation

HongPong:
It was a little hard to find the right fish shell persistent alias command but I think this is correct - and more fun to use than bash. also link to the Oh my fish drush savvy theme for Fishshell Framework.


weitzman:
Can anyone confirm that these instructions work?

damiankloip:
Yes, that command is correct. However, we should make it clear that the call to `set --universal` should not be placed in your fish.config file. It is meant to be be called once, it will then persist for all your subsequent fish sessions.

greg-1-anderson:
Shells other than bash do not get much attention here, regrettably. Closing.

PR #2202, opened 2016-06-01
Fixed method signature in DrupalKernel.

fantastic91:
`\Drush\Drupal\DrupalKernel::createFromRequest()` seems broken  #after https://www.drupal.org/node/2631362. Fixed it here.

### DIFF
--- a/lib/Drush/Drupal/DrupalKernel.php
+++ b/lib/Drush/Drupal/DrupalKernel.php
@@ -13,7 +13,7 @@ class DrupalKernel extends DrupalDrupalKernel {
   /**
    * @inheritdoc
    */
-  public static function createFromRequest(Request $request, $class_loader, $environment, $allow_dumping = TRUE) {
+  public static function createFromRequest(Request $request, $class_loader, $environment, $allow_dumping = TRUE, $app_root = NULL) {
     drush_log(dt("Create from request"), LogLevel::DEBUG);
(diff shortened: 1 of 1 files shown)


weitzman:
Do we have to change the body of the function at all?

lib/Drush/Drupal/DrupalKernel.php
```diff
@@ -13,7 +13,7 @@ class DrupalKernel extends DrupalDrupalKernel {
   /**
    * @inheritdoc
    */
-  public static function createFromRequest(Request $request, $class_loader, $environment, $allow_dumping = TRUE) {
+  public static function createFromRequest(Request $request, $class_loader, $environment, $allow_dumping = TRUE, $app_root = NULL) {
     drush_log(dt("Create from request"), LogLevel::DEBUG);
     $kernel = new static($environment, $class_loader, $allow_dumping);
     static::bootEnvironment();
```
edurenye: We should pass the $app_root to bootEnvironment otherwise when using drush, it will always guess the root, will never use the app_root, so will be like that change in Drupal was never made. IMHO.

edurenye:
We should pass the $app_root to bootEnvironment otherwise when using drush, it will always guess the root, will never use the app_root, so will be like that change in Drupal was never made. IMHO.

weitzman:
Decided to merge this as is. Anyone is welcome to adjust Drush to honor app_root in the future. For now, we avoid a warning.

PR #2209, opened 2016-06-08
Fix issue 2207: fatal error when canceling actions

gggeek:


lib/Drush/Log/Logger.php
```diff
@@ -27,13 +27,30 @@
...
 use  Symfony\Component\Console\Output\OutputInterface;
 
-class Logger extends RoboLogger {
+class Logger extends BaseLogger {
 
     public function __construct(OutputInterface $output) {
-      parent::__construct($output);
+
+      $verbosityOverrides = [
+        // The first 3 lines are kept for backwards compatibility, as thsi class was extending RoboLogger beforehand
```
andrerom: thsi => this

;)

gggeek: geez, those nitpickers! :-D

andrerom: hence the joke :)

weitzman:
Can you please provide steps to reproduce the fatal?

PR #2210, opened 2016-06-08
Ignore phpstorm file in git

gggeek:


damiankloip:
Wouldn't this just be in your global or user level gitignore?

gggeek:
Of course it could, but I think that the cost/benefit ratio justifies it: one line in the gitignore is not a great maintenance burden, and having it there makes the project a tad more friendly for new devs joining.

As a side note: I do not have my per-user .gitignore set up properly because I use a vast amount of vms/containers, and sometimes run the git command, for the same project, from within 3 different envs (windows, linux vm, container). I understand this might setup be uncommon, so I leave it up to you to decide...

weitzman:
OK, but I'm curious why are you making commits inside of the Drush directory? Users are supposed to add config/command/alias files elsewhere?

gggeek:
@weitzman sorry, what I did in order to send PRs to Drush was to: check out Drush in a dedicated folder without even having Drupal installed there, edit files, commit, send PR. Maybe it is not the standard workflow for Drush developers, but it worked for me...

weitzman:
added. had a typo so i did it manually. 

thanks.

PR #2215, opened 2016-06-10
Fixes #2214: Allow forcing of color output.

### LINKED ISSUE #2214: Allow color output to be forced
Some CI environments (e.g., GitLab CI) appear not to have a terminal, and so Drush output cannot be colored. Drush supports a `--nocolor` option, but there's no way (that I've found anyway) to force color output.

The relevant code is [here](https://github.com/drush-ops/drush/blob/master/includes/preflight.inc#L581). This logic doesn't allow us to force color output, as it would interpret `--nocolor=0` the same as if the option wasn't set to begin with.

So, I propose adding a `--color` option.

ergonlogic:
Simple fix for forcing color output.


weitzman:
Have not seen any other requests for this.

PR #2216, opened 2016-06-12
Fixed missing version info from git_deploy.

darrenoh:
The updatedb command is unable to load version info from git_deploy. This causes every module that has a version dependency to generate errors. I have added support for the workaround that is waiting to be committed to Drupal 7 core: https://www.drupal.org/node/2062763

### DIFF
--- a/commands/core/drupal/update_7.inc
+++ b/commands/core/drupal/update_7.inc
@@ -167,9 +167,10 @@ function update_main_prepare() {
 
   // Load module basics.
   include_once $core . '/includes/module.inc';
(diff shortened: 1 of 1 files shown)


weitzman:
I'm inclined to let Drupal fix and not commit a workaround. Thanks for the patch still. It will help folks to see it here.

darrenoh:
The patch is not a workaround. The workaround is adding modules which change system information to settings.php. Both Drush and Drupal have to be patched to support this.

darrenoh:
I have started an 8.x branch in my fork to support those who maintain Drupal 7 with Drush.

PR #2220, opened 2016-06-14
Add .config in $PATH of install-alternative.md

diqidoq:
Update install-alternative.md by adding .config in the export $PATH instruction.


weitzman:
Dupe of #2244

PR #2225, opened 2016-06-17
Add a caster for MarkupInterface

damiankloip:
I'm bored of casting `MarkupInterface` objects (like `TranslatableMarkup`) to a string every time... Let's add a caster for it!

1__fish_____sites_tag1_updates_docroot__fish_


dawehner:
+1

Issue #1581, opened 2015-08-29
'drush up drupal' deleted my .git file in Drupal root


joachim-n:
```
joachim$ git push
Counting objects: 154, done.
Delta compression using up to 2 threads.
Compressing objects: 100% (79/79), done.
Writing objects: 100% (79/79), 11.87 KiB | 0 bytes/s, done.
Total 79 (delta 66), reused 0 (delta 0)
To git@bitbucket.org:joachim-n/foo.git
   33e5438..47d8d1c  master -> master
~/Sites/joachim-accounts joachim$ drush up drupal
Update information last refreshed: Sat, 29/08/2015 - 09:39
 Name    Installed Version  Proposed version  Message
 Drupal  7.38               7.39              SECURITY UPDATE available


Code updates will be made to drupal core.
WARNING:  Updating core will discard any modifications made to Drupal core files, most noteworthy among these are .htaccess and robots.txt.  If you have made any modifications to these files, please back them up before updating so that you can re-create your modifications in the updated version of the file.
Note: Updating core can potentially break your site. It is NOT recommended to update production sites without prior testing.

Do you really want to continue? (y/n): y
Project drupal was updated successfully. Installed version is now 7.39.
No database updates required                                         [success]
'all' cache was cleared.                                             [success]
Finished performing updates.                                         [ok]
joachim$ git add modules/ themes/ includes/ misc/ profiles/
fatal: Not a git repository (or any of the parent directories): .git
```

The only explanation I can see is that in moving things around to update core, Drush wiped the .git folder!

weitzman:
By design. Thats somewhat obtusely referred to in the warning "Updating core will discard any modifications made to Drupal core files".

Related discussion at #83 and #1261

jonhattan:
Also #1464. It provides an example to recover a folder from the backup.

joachim-n:
By design???
The .git file isn't a modification made to a Drupal core file; it's an extra file.

weitzman:
Ok, not By Design. But still this a dupe. See the links posted above.

joachim-n:
I don't really follow how #83 and #1261 are related. Those are about having Drush help with managing the git repo for my site.
But this bug is merely about Drush not deleting the git repo. I wasn't expecting Drush to do a git commit or anything fancy -- just not delete my repository, and with that, a number of local branches with work in them.
I've been using 'drush up drupal' for years now, and it's never done this before. I would expect Drush to leave alone files that are non-Drupal.

[...]

PR #2226, opened 2016-06-17
Issue #1581: Don't delete files pre-update.

damienmckenna:
Don't delete unknown files during pre-update if backups are disabled.
This destroys repository meta files, e.g. ".git" directories, which is
really poor form, especially when it only happens when the "no-backup"
option is set.


damienmckenna:
Going back to the drawing board.

PR #2227, opened 2016-06-17
Issue #1581: Warn about 'no-backup' side effects.

damienmckenna:
Provide a warning that using the "no-backup" option for pm-updatecode will result in possible deletion of files.


weitzman:
Thanks ... You edited the wrong option by accident.

weitzman:
drush8 is on minimal support now.

PR #2236, opened 2016-06-22
Accept 'local-host' in site alias configuration to specify the host to consider "local"

bendoh:
This addresses an issue with Vagrant setups that use port-forwarding on `localhost`, which is detected as a "local" due to `drush_is_local_host()`.

This was discussed in https://github.com/drush-ops/drush/commit/5fd7351, which seems to revert the `#check-local` flag that was added in https://github.com/drush-ops/drush/commit/00d1f4e055ce9218416a2bc41f060661a5ff385f and was made necessary to force drush to check for local.

This PR creates a new acceptable key on the aliases configuration array, `'local-host'`, which when set, will cause drush to treat the target host as local only when the value specified is equal to the value of `gethostname()`.

For my Vagrant environment, this change allows me to set `'local-host' => 'mymachine.dev'` in the alias configuration, and when I invoke the alias from my host machine (NOT mymachine.dev,) drush will correctly connect via ssh. However when I leverage the same alias when using the virtual machine `mymachine.dev`, it will recognize it as the local host.

Before this change, the host machine would be considered local because the IP for the `remote-host` is `127.0.0.1`. 

This change is backwards-compatible in that it only affects behavior if `local-host` is part of the site alias configuration array.

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -1091,6 +1091,8 @@ function drush_backend_generate_sitealias($backend_options) {
  *      Optional. A remote host to execute the drush command on.
  *   'remote-user'
  *      Optional. Defaults to the current user. If you specify this, you can choose which module to send.
+ *   'local-host'
+ *      Optional. If == gethostname(), then force local operation of command.
  *   'ssh-options'
  *      Optional.  Defaults to "-o PasswordAuthentication=no"
  *   '#env-vars'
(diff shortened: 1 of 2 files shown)


marcaddeo:
Has anyone had a chance to look over this? This is affecting our development workflow.

weitzman:
Sorry for the looooong delay. Reopen and submit PR for master if still desired.

PR #2239, opened 2016-06-24
Fix include makefile from git example

ivesdebruycker:


weitzman:
This looks different from the fix i just committed 19ca1ab

weitzman:
similar to other example, and has unresolved conflicts.

PR #2241, opened 2016-06-24
Drush drupal version context for make update

hanoii:
This is an addition to a pull request that will follow and will reference this one, but basically because 'make-update' needs no bootstrap but indirectly relies on the update-status engine, and when parsing version numbers it needs a drupal version to properly parse module versions, so this provides a way to set up a drupal version internally as a drush context. It's harmless and eventually useful.


weitzman:
If make-update needs to know the Drupal version, then change this line in its command definition to:

    'bootstrap' => DRUSH_BOOTSTRAP__DRUPAL_ROOT,

If that resolves the problem, please update tis PR or submit a new one. Thanks!

jonhattan:
@hanoii does this fix an existing bug in make-update or just enables #2242 to work?

PR #2242, opened 2016-06-24
Support for updating code (both upc and make-udpate) for unsupported modules

hanoii:
I still maintain D6 sites (and D5 and D46 but that's just anecdotical) and it's still very useful to be able to update module versions to the last supported version regardless of the unsupported status.

With #2241, this allows for a nice way to update old D6 sites to the last supported-versions.

This also includes a very small fix with a leftover 't()' to 'dt()'

### DIFF
--- a/lib/Drush/UpdateService/StatusInfoDrush.php
+++ b/lib/Drush/UpdateService/StatusInfoDrush.php
@@ -178,7 +178,9 @@ class StatusInfoDrush implements StatusInfoInterface {
           $project_status = DRUSH_UPDATESTATUS_REVOKED;
           break;
         case 'unsupported':
-          $project_status = DRUSH_UPDATESTATUS_NOT_SUPPORTED;
+          if (!drush_get_option('check-unsupported', FALSE)) {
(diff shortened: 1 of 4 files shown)


lib/Drush/UpdateService/StatusInfoDrush.php
```diff
@@ -370,7 +379,7 @@ private function calculateProjectUpdateStatus($project_release_info, &$project_d
     // Bail out early.
     if (!isset($project_data['recommended'])) {
       $project_data['status'] = DRUSH_UPDATESTATUS_UNKNOWN;
-      $project_data['reason'] = t('No available releases found');
+      $project_data['reason'] = dt('No available releases found');
```
hanoii: This should be fixed regardless. I was about to submit a separate pull request and can do so if this is not really something considered for addition, although I hope it is.

hanoii:
Also worth noting this should work the same if the option is not set

jonhattan:
@hanoii does this depend somehow on #2241 ?

greg-1-anderson:
This can now be done with  the [mydropwizard module](https://www.drupal.org/project/mydropwizard)

PR #2244, opened 2016-06-27
Update install-alternative.md for path to composer 

river-fall:
Path to composer binary directory was wrong


weitzman:
.config is the default location? On all OS?

river-fall:
I can't say for the all OS, but in RHEL family yes

PR #2246, opened 2016-06-28
Fix PHP pcntl_exec() Error during install

christopher-hopper:
The current install instructions generate a PHP pcntl_exec() permissions error when run with PHP 5.6. Some people are incorrectly assuming this means drush needs to be installed as root.

Executing the php call to `drush core-status` from outside the path where the drush PHAR file is located fixes the issue.

See this stackexchange post:

- http://drupal.stackexchange.com/questions/201060/does-drush-need-to-be-installed-as-root

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -2,15 +2,15 @@ Install/Upgrade a global Drush
 ---------------
 ```sh
 # Download latest stable release using the code below or browse to github.com/drush-ops/drush/releases.
-php -r "readfile('http://files.drush.org/drush.phar');" > drush
-# Or use our upcoming release: php -r "readfile('http://files.drush.org/drush-unstable.phar');" > drush
+php -r "readfile('http://files.drush.org/drush.phar');" > /tmp/drush
+# Or use our upcoming release: php -r "readfile('http://files.drush.org/drush-unstable.phar');" > /tmp/drush
 
 # Test your install.
-php drush core-status
+php /tmp/drush core-status
 
 # Make `drush` executable as a command from anywhere. Destination can be anywhere on $PATH.
-chmod +x drush
-sudo mv drush /usr/local/bin
+chmod +x /tmp/drush
+sudo mv /tmp/drush /usr/local/bin
 
 # Optional. Enrich the bash startup file with completion and aliases.
 drush init


christopher-hopper:
Here's the PHP error generated by the current install instructions:

```
# php drush --debug core-status
Using the Drush script found at /home/vagrant/drush using pcntl_exec
PHP Warning:  pcntl_exec(): Error has occurred: (errno 13) Permission denied in phar:///home/vagrant/drush/includes/startup.inc on line 359
PHP Stack trace:
PHP   1. {main}() /home/vagrant/drush:0
PHP   2. require() /home/vagrant/drush:10
PHP   3. drush_startup() phar:///home/vagrant/drush/drush:114
PHP   4. pcntl_exec() phar:///home/vagrant/drush/includes/startup.inc:359

Warning: pcntl_exec(): Error has occurred: (errno 13) Permission denied in phar:///home/vagrant/drush/includes/startup.inc on line 359

Call Stack:
    0.0001     224712   1. {main}() /home/vagrant/drush:0
    0.0201     759536   2. require('phar:///home/vagrant/drush/drush') /home/vagrant/drush:10
    0.0206     808760   3. drush_startup() phar:///home/vagrant/drush/drush:114
    0.0293     819856   4. pcntl_exec() phar:///home/vagrant/drush/includes/startup.inc:359

Error has occurred executing the Drush script found at /home/vagrant/drush
(errno 13) Permission denied
```

christopher-hopper:
@greg-1-anderson 
Just tested the old install instructions on Windows 10 bash (via WSL) and the same problem exists there. These new instructions that are part of the Pull Request fix the issue as it stands today.

If the change to the install instructions is no good, perhaps investigating the root of the problem in `pcntl_exec()` is needed. That's above my pay grade right now I'm afraid.

klidifia:
+1. Following the current instructions exactly just doesn't work on Ubuntu 1404/1604.

hansfn:
This problem was introduced by pull request #2145 which was supposed to just remove wget, but also changed the name of the downloaded file from drush.phar to drush. That caused a regression because when calling `php drush core-status` Drush see as Drush script - that isn't executable yet. When the file was called "drush.phar" Drush didn't find a Drush script, but a phar file. Proof:

If the downloaded file is called drush:
```
$ php drush -d core-status
Using the Drush script found at /.../test/drush using pcntl_exec
PHP Warning:  pcntl_exec(): Error has occurred: (errno 13) Permission denied in phar:[...]
```
If the downloaded file is called drush.phar:
```
$ php drush.phar -d status
Phar detected. Proceeding to drush_main(). [...]
```

If the downloaded file is called drush and is made executable:
```
$ php drush -d core-status
Using the Drush script found at /.../test/drush using pcntl_exec
Using the Drush script found at /.../test/drush using pcntl_exec
Using the Drush script found at /.../test/drush using pcntl_exec
Using the Drush script found at /.../test/drush using pcntl_exec
Using the Drush script found at /.../test/drush using pcntl_exec
Using the Drush script found at /.../test/drush using pcntl_exec
Using the Drush script found at /.../test/drush using pcntl_exec
[...]
```
An endless loop.

Please just revert part of 2145 so we use drush.phar as the name for the download again.

weitzman:
This PR works around the problem nicely and I'll merge it if we dont make progress here soon. Thanks. I'm a bit dismayed that when running the drush.phar renamed to `drush`, you can't use drush when you are in its directory. Thats a bug. Here are steps to reproduce the infinite loop:

```
php -r "readfile('http://files.drush.org/drush.phar');" > /tmp/drush
cd /tmp
chmod +x ./drush
./drush core-status -d
```

I suspect [this line](https://github.com/drush-ops/drush/blob/master/includes/startup.inc#L280). Its a bit annoying debugging a Phar because it is read only and even more annoying in this case because PHPStorm can't see inside it and debug when the file lacks a .phar suffix. ping to @greg-1-anderson  who might spot the problem.

hansfn:
I disagree that the using `/tmp/drush` as downloaded filename is a better solution than switching back to `drush.phar`. What if you happen to stand in the `/tmp` directory? Anyway, this isn't really important to me as it doesn't affect the root issue.

> I'm a bit dismayed that when running the drush.phar renamed to drush, you can't use drush when you are in its directory.

Me too. What about detecting "actively" that we are running Drush as a Phar file? The current code - see [line 319 includes/startup.inc](https://github.com/drush-ops/drush/blob/master/includes/startup.inc#L319) - is as we have seen not robust. Couldn't we just use [Phar::running()](http://php.net/phar.running) to detect it?

weitzman:
I added that Phar::running() call but it did not resolve the problem. when the phar is named just `drush`, I think my new code is not even run. @greg-1-anderson any thoughts?

hansfn:
You put the Phar::running() -test inside a code block that is only reached if a Drush script is **not** found. Hence your code isn't reached. If you still struggle, I can write a patch later tonight. I think you can just add en elseif to the if that starts at line 319:

```
ifelse (class_exists('Phar') && Phar::running()) {
      if ($DEBUG) {
        fwrite(STDERR, "Phar detected. Proceeding to drush_main().");
      }
      require __DIR__ . "{$sep}preflight.inc";
      exit(drush_main());
}
```

Totally untested. And of course, it break the DRY principle.

weitzman:
Thanks @hansfn. Refactored a bit and merged to master and 8.x

Sorry this issue languished for so long.

weitzman:
I just relaxed the fix here (see #2312 ) since it was hijacking some legitimate handoffs to site-local Drush from a Phar. Please reopen if you notice any problems.

sw2band:
I get this error while installing Drush 8.1.5, I had to return to 8.1.3.

weitzman:
Please grab a new phar. The 8.1.5 phar just became available. Please let us know if it works for you.

PR #2254, opened 2016-07-04
remove settings which no longer exist

dawehner:


tstoeckler:
I think `'ui.show.display_embed'`still exists, actually. The others do seem to be bogus.

dawehner:
Oh wow, good catch. I was so sure we removed it at some point during the
cycle

dawehner:
So well this setting still exists in the UI, but we just don't do anything
with it:

```

 * @todo: Wait until annotations/plugins support access methods.
 * no_ui => !\Drupal::config('views.settings')->get('ui.show.display_embed'),
 *
 * @ViewsDisplay(
 *   id = "embed",
 *   title = @Translation("Embed"),
 *   help = @Translation("Provide a display which can be embedded
using the views api."),
 *   theme = "views_view",
 *   uses_menu_links = FALSE
 * )
 */
class Embed extends DisplayPluginBase {

```

Cheers Daniel

tstoeckler:
Ahh, that's nice ;'-P

Very weird, but since the list of settings for this command are completely opt-in (i.e. we don't have to cover all of them), seems fine to me.

PR #2264, opened 2016-07-10
Handle RequeueException in queue workers

juampynr:
Drupal 8 added the [RequeueException](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Queue!RequeueException.php/class/RequeueException/8.2.x) to allow workers to mark an item to be re-queued.

[Core's Cron.php handles it](http://cgit.drupalcode.org/drupal/tree/core/lib/Drupal/Core/Cron.php#n168), but Drush's queue processor doesn't. This pull request fixes it.

### DIFF
--- a/tests/queueTest.php
+++ b/tests/queueTest.php
@@ -120,6 +120,16 @@ class QueueCase extends CommandUnishTestCase {
     $this->drush('queue-run', array('woot_requeue_exception'), $options);
 
     // Check that the item was processed after being requeued once.
+    // Here is the detailed workflow of what the above command did.
+    // 1. Drush calls drush queue-run woot_requeue_exception.
+    // 2. Drush claims the item. The worker sets a state variable (see below)
+    // and throws the RequeueException.
+    // 3. Drush catches the exception and puts it back in the queue.
+    // 4. Drush claims the next item, which is the one that we just requeued.
+    // 5. The worker finds the state variable, so it does not throw the
+    // RequeueException this time (see below).
+    // 6. Drush removes the item from the queue.
+    // 7. Command finishes. The queue is empty.
     $this->drush('queue-list', array(), $options + array('pipe' => TRUE));
     $output = trim($this->getOutput());
     $this->assertEquals(str_replace('%items', 0, $expected), $output, 'Queue item processed after being requeued.');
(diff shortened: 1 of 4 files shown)


juampynr:
Related issue at Drupal.org www.drupal.org/node/2705809

weitzman:
Looks good to me. Assigning to @damiankloip for final review and merge (please merge to 8.x as well).

damiankloip:
This looks good to me. We could use a test for this possibly? @juampynr thoughts? :) We could leverage the `cron_queue_test` module from core and specifically the `CronQueueTestException` worker plugin. In drush we already have a `QueueTest`, so we could add another test case to that...

juampynr:
Sounds good @damiankloip. I just added a test.

tests/queueTest.php
```diff
@@ -85,4 +85,57 @@ public function testQueueDelete() {
...
+    // Check that the queue exists and it has one item in it.
+    $expected = 'woot_requeue_exception,%items,Drupal\Core\Queue\DatabaseQueue';
+    $this->drush('queue-list', array(), $options + array('pipe' => TRUE));
+    $output = trim($this->getOutput());
+    $this->assertEquals(str_replace('%items', 1, $expected), $output, 'Item was successfully added to the queue.');
+
+    // Process the queue.
+    $this->drush('queue-run', array('woot_requeue_exception'), $options);
+
+    // Check that the item was processed after being requeued once.
```
damiankloip: Small point: this is just checking that there is still an item in the queue.

damiankloip:
@juampynr added a small point, otherwise it looks great!

tests/queueTest.php
juampynr: Not really, here is how the queue processes this item:

1. Drush calls `drush queue-run woot_requeue_exception`.
2. Drush claims the item. The worker sets a state variable (see below) and throws the RequeueException.
3. Drush catches the exception and puts it back in the queue.
4. Drush claims the next item, which is the one that we just requeued.
5. The worker finds the state variable, so it does not throw the RequeueException this time (see below).
6. Drush removes the item from the queue.
7. Command finishes. The queue is empty.

juampynr: Going to add this to the test for reference.

juampynr: Done.

damiankloip: Oh, yes. ok. That makes sense. This looks good to me then! Thanks @juampynr

PR #2269, opened 2016-07-12
Fix make tests to not rely on sourceforge

jhedstrom:


weitzman:
@jhedstrom - Seems to have fixed on fail. One more to go.

PR #2272, opened 2016-07-14
Symfony console version

asentner:
Any chance you can be more flexible with symfony libs? Cant global install drush via composer because of the strict 2.7.* in composer.json.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,14 +38,14 @@
     "codegyre/robo": "~1.0.0-beta1",
     "consolidation/annotated-command": "~1.0.0-beta1",
     "consolidation/output-formatters": "~1.0.0-beta1",
-    "symfony/console": "2.7.*",
-    "symfony/event-dispatcher": "2.7.*",
+    "symfony/console": "~2.7",
+    "symfony/event-dispatcher": "~2.7",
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
The Drush version of the Symfony libraries should closely match the version required by the most recent version of Drupal 8.  The version constraints should **not** be loosened up to make it easier to globally install Drush.  This strategy may have some short-term benefits for some configurations, but would cause harder-to-diagnose dependency hell problems in other configurations.

Globally installing Drush with wp-cli is a non-goal.  If you want to do this, `composer global remove` Drush and install it again with [cgr](https://github.com/consolidation-org/cgr).  That will work with wp-cli and other potential configurations.

This issue should be closed, not as a dupe, but as a will-not-fix.

Leaving for @weitzman to agree or disagree.  Alternately, we could leave this open to track the documentation / cgr uninstall issue.

weitzman:
Agreed. Thanks for the explanation.

asentner:
The most recent requires allows for 2.8, the purpose of this issue.

https://github.com/drupal/drupal/blob/8.1.6/core/composer.json

I'll just use my fork for now on until Drush works with the rest.

PR #2274, opened 2016-07-15
Check release information (e.g. versions) strictly

pjcdawkins:
1. Somehow end up with a Drush cache that includes out-of-date release info, _not_ containing Drupal core version `7.50`
2. Run `drush make` on a make file (YAML or ini) which asks for Drupal core `"7.50"`
3. Instead of failing and reporting that the version `7.50` could not be found, Drush will select version `7.5`
4. The result could be a site unexpectedly vulnerable to Drupalgeddon.

The actual fix could probably be improved by contributors more knowledgeable about Make internals: I created the PR to illustrate the problem.

### DIFF
--- a/lib/Drush/UpdateService/Project.php
+++ b/lib/Drush/UpdateService/Project.php
@@ -330,7 +330,7 @@ class Project {
   private function searchReleases($key, $value) {
     $releases = array();
     foreach ($this->parsed['releases'] as $version => $release) {
-      if ($release['status'] == 'published' && isset($release[$key]) && $release[$key] == $value) {
+      if ($release['status'] == 'published' && isset($release[$key]) && $release[$key] === $value) {
         $releases[$version] = $release;
       }
     }


jonhattan:
Easy to reproduce:

```
jonhattan@nuro:/tmp$ drush dl drupal-7.4000
 [notice] Project drupal (7.40) downloaded to /tmp/drupal-7.40.
```

pjcdawkins:
So, this is a secondary problem, but: I also found why the cache was out of date. Drush relies on the `ctime` of the cache file, which changes if you change any file metadata. I was building via a script which did `chown` on the cache before building (for obscure reasons), and that resulted in an update to the `ctime` for everything, so of course the cache was always assumed to be fresh. To me, the `mtime` would be more reliable.

jonhattan:
@pjcdawkins please new issue for ctime/mtime ;)

jonhattan:
Tests are failing because the fix enforces type check. What we really want here is to enforce a string comparison. Added a fix in #2293

pjcdawkins:
Closing in favour of #2293

PR #2280, opened 2016-07-19
add $app_root to bootEnvironment

edurenye:
As stated in the pull request #2202 
We should pass the $app_root to bootEnvironment().


lib/Drush/Drupal/DrupalKernel.php
```diff
@@ -16,7 +16,7 @@ class DrupalKernel extends DrupalDrupalKernel {
   public static function createFromRequest(Request $request, $class_loader, $environment, $allow_dumping = TRUE, $app_root = NULL) {
     drush_log(dt("Create from request"), LogLevel::DEBUG);
     $kernel = new static($environment, $class_loader, $allow_dumping);
```
tstoeckler: Please also add `$app_root` as a final argument here

PR #2282, opened 2016-07-20
Refs #2079. Do not pass `$this->query_extra` to `pg_dump`.

xurizaemon:


RoSk0:
Can we please have this PR updated.

RoSk0:
I wouldn't agree that changes in this PR are correct.
It works, yes, but removing default value from that check makes PostgreSQL integration behave differently to other supported DBs and that is bad.
I think we should either remove $query_extra field from Sqlpgsql class completely(I think **preferred** ) and behave as the rest or set to empty string by default.

xurizaemon:
:wave: hi kirill!

Agree that this isn't perfect solution, your proposal makes sense. Happy for you to submit alternate fix, I'm on leave for now.

xurizaemon:
#3267 instead

cafuego:
Please re-open this one, the alternative fix looks to break more than it fixes for me.

xurizaemon:
Done @cafuego

weitzman:
I made the same change in mysql today so it makes sense to do it here. My deepest apologies for the long delay. We are in need of a Postgres maintainer.

PR #2287, opened 2016-07-21
Pass the $app_root to the kernel during boot

tstoeckler:
This is a follow-up to #2202 (and #2280)

Seems innocent enough, but I'm not very knowledgeable about the Drush boot process, so could use some eyes.

Note that passing the `$app_root` explicitly is considered a best practice. It's only optional for backwards-compatibility.


weitzman:
That change looks harmless to me. Feel free to merge once Travis is happy.

Issue #2152, opened 2016-04-24
Drush in Windows gives error: filename, directory name, or volume label syntax is incorrect


Zythyr:
I am unable to properly use Drush 7 and 8 on Windows. Whenever I run command such as "drush pm-update" I get error as below:

> The command could not be executed successfully (returned: The        [error]
> filename, directory name, or volume label syntax is incorrect.
> , code: 1)
> pm-updatestatus failed.

Below is my environment: 
Windows 10x64 
XAMPP with PHP 5.6.19
Cygwinx64
Drush 8.0.5

Below is my Drush status output:

> dev@DEV-PC /cygdrive/c/xampp/htdocs/www/my_drupal_6_site
> $ drush status
>  Drupal version         :  6.16
>  Site URI               :  http://www
>  Database driver        :  mysql
>  Database hostname      :  localhost
>  Database port          :
>  Database username      :  drupal
>  Database name          :  my_drupal_6_site
>  PHP configuration      :  C:\xampp\php\php.ini
>  PHP OS                 :  WINNT
>  Drush script           :  C:\Users\dev\AppData\Roaming\Composer\vendor\drush
>                            \drush\drush
>  Drush version          :  8.0.5
>  Drush temp directory   :  C:\cygwin64\tmp
>  Drush configuration    :
>  Drush alias files      :
>  Drupal root            :  C:\xampp\htdocs\www\my_drupal_6_site
>  Drupal Settings File   :  sites/default/settings.php
>  Site path              :  sites/default

Below is part of the debug output when I run the command: `drush -d pm-update`. 

> 
> Backend invoke:                                                        [command]
> "C:\\Users\\dev\\AppData\\Roaming\\Composer\\vendor\\drush\\drush\\drush"
>  --backend=2 --verbose --debug
> --root="C:\\xampp\\htdocs\\www\\my_drupal_6_site" --uri=http://www
> pm-updatestatus 2>&1 [7.21 sec, 16.15 MB]
> "C:\\Users\\dev\\AppData\\Roaming\\Composer\\vendor\\drush\\drush\\drush"   [n     otice]
>  --backend=2 --verbose --debug
> --root="C:\\xampp\\htdocs\\www\\my_drupal_6_site" --uri=http://www
> pm-updatestatus 2>&1 [7.21 sec, 16.16 MB]
> The command could not be executed successfully (returned: The        [error]
> filename, directory name, or volume label syntax is incorrect.
[...]

PR #2292, opened 2016-07-25
Issue #2152: Fix escaped path bug for Windows

peterromero:
I don't know all the implications of this. I do know that my particular setup chokes if paths are passed surrounded by quotes. Someone who knows more should look into whether there is a better way to accomplish this.

See https://github.com/drush-ops/drush/issues/2152 for details on the error I am trying to fix.

### DIFF
--- a/includes/exec.inc
+++ b/includes/exec.inc
@@ -328,7 +328,7 @@ function drush_escapeshellarg($arg, $os = NULL, $raw = FALSE) {
     return $arg;
   }
   elseif (drush_is_windows($os)) {
-    return _drush_escapeshellarg_windows($arg, $raw);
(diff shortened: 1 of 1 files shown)


weitzman:
Wondering if the double quote fix at https://github.com/drush-ops/drush/pull/1213/files#diff-41635085774a0b06f19002b99b110187R1177 is a better alternative than this. If anyone could test it, that would be appreciated.

peterromero:
A quick, non-thorough check would suggest that this does fix it (just the part in backend.inc). If someone else can verify, that would be great.

weitzman:
I think I just fixed the problem mentioned here. Not sure I got all cases though. Please post here if problems persist.

PR #2293, opened 2016-07-26
Test project versions are checked strictly (no numeric equivalence).

jonhattan:
Expected to fail. Fix is #2274


jonhattan:
Backported to Drush 8.x and 7.x

PR #2294, opened 2016-07-26
Use mtime instead of ctime for cache validity

pjcdawkins:
The mtime tracks the time the file was last modified.

The ctime tracks the time the file was created _or_ when its inode metadata changed. The ctime is thus vulnerable to being reset via things like `chmod` / `chown`.

Motivation here: https://github.com/drush-ops/drush/pull/2274#issuecomment-235212433

This is intended as the simplest possible fix. A more ideal solution might be something akin to the Doctrine [FilesystemCache](https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/FilesystemCache.php), which adds a timestamp to the first line of each cache file.

### DIFF
--- a/includes/drush.inc
+++ b/includes/drush.inc
@@ -768,7 +768,7 @@ function drush_download_file($url, $destination = FALSE, $cache_duration = 0) {
 
   if ($cache_duration !== 0 && $cache_file = drush_download_file_name($url)) {
     // Check for cached, unexpired file.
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
No activity here for a long time, and the referenced issue was closed in favor of a different issue that closed by a different PR being merged, so perhaps this is not needed.

PR #2301, opened 2016-08-03
download theme engines into the sensible default destination

davyrolink:
At the moment when you download a theme engine, e.g. "drush dl tfd7-7.x-3.0-beta2", it ends up in /public/themes/engines. But the sensible default place for theme engines from "contrib" is 
/public/sites/all/themes/engines.


jonhattan:
`sites/all/themes/engines/contrib` also works.

PR #2313, opened 2016-08-18
Extend \Drush\Config\StorageWrapper to apply the filters to all methods

bircher:
The StorageWrapper which applies filters is a very neat concept.
However, to do more serious filtering the other methods on the StorageInterface should also use the filters.

I added a base class for filters that is essentially a transparent filter and lets everything pass. This way the CoreExtensionFilter continues to work almost untouched.


bircher:
An example of a more elaborate filter can be found here: https://github.com/nuvoleweb/config_split/blob/8.x-1.x/src/Config/SplitFilter.php

weitzman:
Still needed/wanted?

bircher:
This is still needed if you want to support storage filters. I was looking at #1964 and found it to be better approach, but still incomplete. I worked on the problem of partial configuration import/export for a while and config_split is what I came up with.

The reason you need to pass all the config storage operations to filters is because you need to set/unset other configuration instead of just editing core.extensions, and if that is supposed to work also with languages then you need also to work with collections.

weitzman:
We are going to remove filter support from Drush commands now that Contrib can take over via the D8 sync service

Issue #2133, opened 2016-04-15
Allow custom storage filters to be created/applied during config import/export


ericduran:
This issue is coming from https://github.com/drush-ops/drush/pull/1522#issuecomment-206711685

@ericduran: 

> @greg-1-anderson @weitzman I came here looking for drush_command_invoke_all('drush_storage_filters');
> 
> I'm working on a solution to allow a multi-site set up to share a majority of configs (Entity/Bundles/Fields/etc..) but keep certain specific settings to be unique per site such as system.site.yml in its own dir.
> 
> Right now I'll need to write new drush commands to selectively import/export certain configurations into certain directories. That being said with the addition of drush_command_invoke_all('drush_storage_filters'); I can easily implement my own storage filters to figure out what goes where.
> 
> Thoughts? Concern? This is an oversimplification of the solution but yeah.
> 
> Should I open a PR? Move the discussion to a new issue?

I'm just moving this into an issue for now, I'll probably probably give this a dry run sometime soon, just starting the conversation.

larowlan:
Thoughts on adding a ```--ignore-list=blacklist.yml``` for a simple generic approach?

Basically you'd add your blacklist (patterns) into .yml file and the import/export would ignore those.

Would that solve the 90% use case?

This could be done with a similar approach to partial.

Eg. for the import:

- export live config into a tmp dir
- copy config to be imported into second tmp dir
- copy blacklist matches from tmp dir into second tmp dir (effectively removing the variations)

Use those to dirs for the StorageComparer

bircher:
An example for custom storage filters in action: https://www.drupal.org/project/config_split

A similar approach could also be useful for #1820 it needs PR #2313 though.

weitzman:
We are discussing bringing back the storage filter hook in #2314

PR #2314, opened 2016-08-18
Add storage filter hook again

bircher:
This reverts commit 7322efc4f21ebbf2aabb6359f01499d165ad691b.

PR  #1522

See issue #2133

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -591,14 +591,16 @@ function drush_config_import_validate() {
 }
 
 /**
- * Presently, the only configuration storage filter that is supported
- * is the 'CoreExtensionFilter'.  If other use cases arise that are
- * not supported by Drupal's configuration override system, then we
- * could add a hook here via drush_command_invoke_all('drush_storage_filters');
- *
- * See: https://github.com/drush-ops/drush/pull/1522
+ * Return storage filters to alter config import and export.
  */
 function drush_config_get_storage_filters() {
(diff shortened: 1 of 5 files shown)


weitzman:
Now that we've seen innovation in Contrib (e.g. https://www.drupal.org/project/config_split and https://www.previousnext.com.au/blog/introducing-drush-cmi-tools), I think it makes sense to bring back this hook. Would this be helpful for config_split and drush_cmi_tools? Feedback is most appreciated. Ping @bircher @webflo @greg-1-anderson @ericduran @larowlan

greg-1-anderson:
New-style annotated hooks should work in the 8.x branch. If that isn't the case, then we should fix it, and implement this as a hook annotation so that we can maintain the same code in 8.x and master.

Config_split is just the thing, but I haven't played with it enough to know what needs to be done here. It already has a Drush command, and seems to satisfy the use-cases I wanted to cover, but never had time to address. Is our goal to bring config_split into core, or leave it as a contrib module? Either way, I'm interested in helping convert to annotations and make sure it still works in 8.x and master. It seems @bircher already has the filter classes he wants implemented; we just need to give him the hooks he needs to integrate. I am very much in favor of this. Kind of busy with personal obligations over the next few days, though.

weitzman:
I merged this into 8.x (d38f6fe) and master (11ed42f). I'll work on moving this into an Annotation hook in both branches when config.drush.inc ports to Annotated (this week).

PR #2315, opened 2016-08-19
Convert browse to an annotated Command (8.x)

weitzman:
- Add discovery of annotated commands
- Converts the following commandfiles. See top of each for @todo.
  - browse


greg-1-anderson:
Looks pretty good to me. I'll look into that bug with optional args.

weitzman:
I thought it might be fixed by https://github.com/consolidation-org/annotated-command/commit/ed279df30b9386fd8e523003dc679421a87c52e0 but updating to latest annotated-command did not  fix the bug and introduced a new one. Looks like its HookManager.php extends Symfony\Component\EventDispatcher\EventSubscriberInterface but no dependency on that component is declared.

weitzman:
I updated composer.lock in both 8.x and master branches. So now we have the latest libraries from Consolidated. I'll fix this PR soon so it applies cleanly again.

weitzman:
Applies cleanly again. As mentioned this PR adds a dependency on "symfony/event-dispatcher" which rightfully belongs to annotated-command due to HookManager.php

greg-1-anderson:
So, regarding the problem with the options ending up in the $path argument, this was not a bug of consolidation/annotated-command. The problem stemmed from the fact that, if no args were specified on the commandline, `drush_dispatch()` added the default options onto the argument list (which was empty) before passing it to `annotationcommand_adapter_process_command()`, which was the function that fixed up default argument values. 

I added a commit to move the addition of the default options to `annotationcommand_adapter_process_command()`, which seems to have cleared up the problem.

weitzman:
I pushed handling of @bootstrap and handle-remote-commands into annotationcommand_adapter_get_commands_for_commandhandler(). I think thats fine, though there are more formal ways to implement like a processBootstrap() function similar to \Consolidation\AnnotatedCommand\Parser\Internal\AbstractCommandDocBlockParser::processUsageTag. 

I welcome thoughts on how to document these annotations which are unique to Drush and not in \Consolidation\AnnotatedCommand. Thats the only @todo, and its not urgent. IMO this is mergeable now.

greg-1-anderson:
The various `internal` functions in AnnotatedCommand, such as the various `process` methods of the DocBlockParser were not intended to be extended. The way you did it is the intended way to add annotations in the application. LMK if you think that a formal way to register processing callbacks should be added to AnnotatedCommand; my current thought is that it's not worth the complexity, but I'm open to ideas if you want to make improvements here.

Regarding documentation, we just need a "writing Drush commands" section on readthedocs or in a Drush topic.

I agree this is good to merge. Didn't look on the failing tests, tho.

weitzman:
Sounds good to me. 

Looking ahead, we need to deal with complete callbacks like image_image_flush_complete(). My assumption is that the equivalent in an annotation command is a hook method with an annotation like `@hook complete image-flush`

I can see how to implement this if we add 'complete' as a new hook type in HookManager. Would you support that? If not, how could Drush add its own Hook?

weitzman:
The test fail at https://travis-ci.org/drush-ops/drush/jobs/157660720 is telling us that we lose this 8.x branch would lose PHP 5.4 compatibility when we start depending on the PHPDocumentor packages. Given that, and the fact that we would need to add backward compat for old-style hooks, it seems like we should leave 8.x alone.

So, that means that we launch a new 9.x branch based off 8.x and this PR will be its first change. My only hesitation with this plan is that we are maintaining 3 branches (8.x, 9.x, master). Thats fine in the short term but we should consolidate soon.

greg-1-anderson:
My preference vis-a-vis completion handler is to have commands create a Robo CollectionBuilder. Even if Robo tasks are not used, function callbacks can be added to the collection; these are all executed when the CollectionBuilder is returned from the command function.  Completion hooks can also be added to the collection builder; these run after all of the other tasks / code on the collection have finished executing, or as soon as any task / code fails. Rollbacks can also be added; these run only if some task fails.

I know I added Robo tasks to the master branch, but I'm not sure if I added them to 8.x. Think so, but I need to run now & don't have time to look at the code just yet.

In general, I agree to the idea of maintaining three branches, unless Robo hasn't been added to the 8.x branch, in which case we might want to revisit our earlier decision and just push for finishing master.  I think three branches is probably better than that.  I'm clearing out my backlog and should be able to help out here shortly.

weitzman:
Sorry, the Composer issue was my own fault. Seems like we can keep PHP 5.4 compat thanks to your earlier work at https://github.com/consolidation-org/annotated-command/commit/d0144c2fce5b4e2b8e662a3764c79c5c0c907914. I pushed a new composer.lock

Your comment about Robo CollectionBuilder makes me wonder if you thought I referred to tasks that run after command execution. My aplogies for being unclear. I was referring to Bash completion of command arguments. I just pushed a commit which implements this in a reasonable way IMO. See bottom of the BrowseCommands file.

greg-1-anderson:
You're right, I misunderstood which kind of `completion` you meant. I would be in favor of adding features to AnnotatedCommand to help implement bash completion. I'll take a look at BrowseCommands.

greg-1-anderson:
Robo 1.x does not support php 5.4, so maybe this means no Robo tasks in the 8.x branch for now.

weitzman:
Perhaps we merge this PR in the name of progress and then any bash completion enhancements made in AnnotatedCommand will be easily incorporated later.

I will see make sure this merges into master easily before merging into 8.x.

weitzman:
Merged locally and pushed to 8.x

PR #2330, opened 2016-08-30
Fix drush make test

jhedstrom:


jhedstrom:
@weitzman I think the remaining fails are unrelated?

jhedstrom:
Merged to `8.x` as well.

PR #2333, opened 2016-09-01
moved 'chmod +x' to come before executing drush

nedyrb:
The step 'chmod +x drush' needs to come before the step where the user executes drush. This is just a simple fix to the order of instructions.


weitzman:
You sure? Its `php` that needs to be executable in the next step.

nedyrb:
Yeah, I'm sure. The line is unedited all I did was move it. Php would already be executable on any system. Chmod +x drush makes the downloaded drush script executable since most Unix systems will default new files to rw only.

greg-1-anderson:
Advice in PR is perhaps clearer (could make `php drush` just `drush`), but unnecessary.

Issue #2337, opened 2016-09-08
user-add-role fails on Drupal 8 when role label specified


colans:
The code in [RoleBase::__construct()](https://github.com/drush-ops/drush/blob/master/lib/Drush/Role/RoleBase.php#L31) is assuming that `user_roles()` will return an array of strings.  This works fine in D7, but in D8 an array of objects is returned.  So `array_search()` fails, and the title can't be converted to the machine name.

``` php
    $this->roles = user_roles();
    if (!is_numeric($rid)) {
      $role_name = $rid;
      if (in_array($role_name, $this->roles)) {
        $rid = array_search($role_name, $this->roles);
      }
    }
```

The workaround is to specify the machine name of the role, not the role label, when running the command.

colans:
I think we need to do something like this:

```php
    $this->roles = user_roles();

    // If we can't find the ID, assume we were given the name.
    if (!isset($this->roles[$rid])) {
      $role_name = $rid;
      if (is_object($this->roles[0])) {
        // D8+: The roles are objects.
        foreach($this->roles as $key => $role) {
          if ($role->label() == $role_name) {
            $rid = $key;
            break;
          }
        }
      }
      else {
        // D7-: The roles are strings.
        if (in_array($role_name, $this->roles)) {
          $rid = array_search($role_name, $this->roles);
        }
      }
    }

```

weitzman:
Lets close this in favor of the pr

PR #2340, opened 2016-09-09
Allow urol setting by label in Drupal 8

colans:
See #2337 for details.


dman-coders:
Related - this causes behat automated testing
https://github.com/jhedstrom/drupalextension

to break in its 'userAddRole()' implementations (in D8) that anticipate that either machine_name or label are allowed.

Related code in the behat side:
```
@Given I am logged in as a user with the :role role(s)
```
calls
`
DrupalContext::assertAuthenticatedByRole()
`

greg-1-anderson:
No activity here for many years. Sorry no one reviewed the work here. If this is still needed / desired, it will need a Drush 9 version. Please open a new PR if there is interest in continuing.

PR #2341, opened 2016-09-11
Convert browse to an annotated command (master).

weitzman:
Same as #2315 but for master branch. Here are a few limitations that can be pushed to a followup I think:

1. drush_complete_get() (and various parts of Drush) assume that drush_get)commands() is a full command list. It looks like annotated commands are not there. I found those at $application->all(). 
1. No handling for @bootstrap and @handle-remote-commands annotations. Not sure where to implement those in master.


greg-1-anderson:
1. Our original plan was to remove all of the old-style commands on the master branch. We could revisit this, in which case we should call the old command hooks and convert them into Symfony objects. (A lot of adapters would be needed to do that.)  In either event, completion will need to be reimplemented.

You are correct that @bootstrap & c. still need to be implemented here. I think we can go ahead and merge this here now anyway, and deal with the other tbds as follow-on tasks.

PR #2343, opened 2016-09-15
Prevent drush_scan_directory from following symlinks.

pdscti:
Drush gets stuck in an infinite loop when scanning the filesystem of a Drupal installation that has subsites configured using symlinks.

I'm fairly sure that this has been suggested before and rejected for some reason, but the simplest fix for this is to stop the drush_scan_directory() function stop recursing through symlinks.


greg-1-anderson:

PR #2353, opened 2016-09-21
Drush complete explanation

adrian-cid:
Added one example to users that don't know where and how configure the shell completion.


weitzman:
In general this text is obsolete. The recommend way to add completion is to run `drush init`. That appends to your ~/.bashrc.

adrian-cid:
do you prefer that I edit the file with the drush init command?

adrian-cid:
I don't have in drush 5.10.0 the drush init command.

weitzman:
Drush 5 is no longer supported. We just released 8.1.5

PR #2354, opened 2016-09-21
Fix PHP 5.3 test on PHP 7 in Drush 7 startup script

greg-1-anderson:
The `drush` script (which is now `drush.laundher` in Drush 8 or later) uses PHP_MIN_VERSION to determine if the php release is PHP 5.3.  This causes the test to spuriously pass on PHP 7.  Adding the extra -d options is innocuous, but noisy.

Running `php -r` has a bad side-effect on Pantheon, though, which doesn't like it if you run php without the php options we've told Drush to use. This recently became an actual operational problem with some APC changes we made; breaks Kalabox when running Drupal 7 + Drush 7 + PHP 7.

This PR fixes all of these problems.


greg-1-anderson:
Applied to wrong branch.  Boo, GitHub.  You can find the branch point, right?

Issue #2336, opened 2016-09-05
Fancy options features for annotated commands (suboptions)


weitzman:
We don't have to keep supporting these weird old behaviors, but if we drop them we should knowingly do so:
- [x] Sharing options across commands inside a commandfile and even across commanfiles. For example, see [drush_sql_get_table_selection_options](http://api.drush.org/api/drush/commands!sql!sql.drush.inc/function/drush_sql_get_table_selection_options/master) is used in archive commands and sql commands. Maybe we allow specifying a callable as value to a new annotation called @optionset?
- [ ]  [sub-options](https://github.com/drush-ops/drush/blob/master/commands/pm/pm.drush.inc#L331)
- [x]  hidden options - supported by annotation_adapter
- [x]  example-value. Symfony always shows the default value for the example value;
- [x] [allow-additional-options](https://github.com/drush-ops/drush/blob/master/commands/pm/pm.drush.inc#L383). Seems easy enough.
- [x] ~~[engines key](https://github.com/drush-ops/drush/blob/master/commands/pm/pm.drush.inc#L370). engines is a way for commands to declare what engines apply to them. Alas, each declaration can have config such as _add-options-to-command_~~ Engines will be removed.

weitzman:
Think @greg-1-anderson made progress on this in AnnotatedCommand

greg-1-anderson:
Progress is being made in AnnotatedCommand, but the above items are still all todo's -- although getting closer.

greg-1-anderson:
The first of these, sharing options, can now be done as follows:

/**
 * @hook pre-command @sql-get-table-selection
 * @option skip-tables-key A key in the $skip_tables array. @see example.drushrc.php. Optional.
 */
public function sqlGetTableSelection(...)
{
    // Code to implement common functionality of skip-tables-key et. al. goes here
}

Then, the above pre-command hook will be applied to any command annotated with `@sql-get-table-selection`.

This same mechanism could potentially be used to implement something similar to the existing Drush engines feature; however, it seems unlikely that this implementation would be perfectly backwards-compatible with the existing implementation. Perhaps this would be acceptable, as existing external Drush commands that referenced Drush engines (if any) could continue to run the original implementation.

Sub options will require implementation of a custom `help` command. There is some question as to what these should look like.

Using `@suboption` as a mode-switch would probably work for most use-cases:
```
/**
 * @option lock Add a persistent lock
 * @suboptions lock
 *     @option lock-message A brief message 
[...]

PR #2357, opened 2016-09-22
Move —drupalicon option to an annotated commandfile

weitzman:
A couple todos before we merge this:

- [ ] The help() method doesn't work yet. [We currently get the global options list via an old-style hook](https://github.com/drush-ops/drush/blob/master/includes/drush.inc#L336) and that ignores AnnotatedCommands.
- [ ] The append method declares itself as post-command hook, and wants to run after any command is issued, not a specific command. Is this possible? Also, this option should ideally work for old style commands as well as annotated commands.


greg-1-anderson:
The way to hook all commands when using the command processor is:
```
/**
 * @hook pre-process *
 */
```
The last argument in @hook is where to hook. This can be a specific command name, `*` for all commands, or empty to hook all commands declared in the same class as the hook.

However, I don't think we should write our hooks using annotations.  We should instead adapt to old-style hooks, figure out what the old function name would be called, and call it.  This is my off-the-cuff answer, aiming to reduce the number of permutations we need to support. We can revisit as needed.  I think in the short term we should concentrate on making progress where it is easy to make progress.  We can't cut a 8.x release until hooks work, or we'd break folks' policy files, but I think we should try to leave our options open for a while. Maybe we should be doing this work on a separate branch.

lib/Drush/CommandFiles/core/DrupliconCommands.php
```diff
@@ -0,0 +1,36 @@
+<?php
+namespace Drush\CommandFiles\core;
+
+class DrupliconCommands {
+
+  /*
+   * @todo not working until we modernize drush_get_global_options().
+   * @hook alter help
+   */
+  public function help($command) {
```
greg-1-anderson: @hook alter isn't called with a `$command` array; it is called with the same parameters as the main command method, which is to say a commandname string.

lib/Drush/CommandFiles/core/DrupliconCommands.php
```diff
@@ -0,0 +1,36 @@
...
+      $command['options']['druplicon'] = [
+        'description' => 'Shows the druplicon as glorious ASCII art.',
+      ];
+    }
+  }
+
+  /**
+   * Print druplicon as post-command output.
+   *
+   * @hook post-command
```
greg-1-anderson: post-command currently doesn't do what one might expect -- but never mind about that, I can and will fix it.  (If you run this today, though, you'll probably see post-command run before the main command hook.)

greg-1-anderson: I just went to make a fix to the post-command hook.  The thing about the command hook is that, in AnnotatedCommand, this hook is called via the Symfony Console `Command` event.  This means that it takes different parameters than the other hooks -- namely, InputInterface and OutputInterface.  The advantage of doing things this way is that it's possible to hook 'native' Symfony commands in a mixed application.  Also, the Command Event happens before argument and option validation, so it's a good place to provide things like default options from configuration.

The odd thing about this is that post-command currently runs before the main command hook, as I alluded to above.  I was going to make this happen afterwards, but the different parameters for this event makes this weird.  I think that maybe the best thing to do is to rename the current 'command' hook to the 'commandevent' or 'symfonycommand' or 'symfony' hook, and make the 'command' hook behave as it does in Drush.

greg-1-anderson:
I think that we should implement druplicon as a separate command rather than as an option to all commands. That would be a lot simpler.

Imagine that we did want to add a hook with an option to just one command, as I am sure that we will (e.g. this will be common is policy files).  That, I think, should look like this:
```
/**
 * @hook pre-command hooked:command
 * @option newoption Some new option 
 */
public function myHookFunction($arg1, $arg2, $options)
{
    if ($options['newoption']) {
        // do my stuff
    }
}
```
Of course, we should still support the equivalent of Drush's `hook help`, but for most simple cases, we could just aggregate the options from all of the hooks.  The hard part is that we wouldn't want this to happen for all commands all of the time; we'd want to lazy-evaluate the dynamic options for a command only if that command is being executed, or if the `help` for that command is being executed.

AnnotatedCommand is close to being able to do this.  I'll try to put it together.

greg-1-anderson:
AnnotatedCommands now support a "Command Event" hook, which is the Symfony Command hook, and also a "Command" hook, which behaves as the Drush command hook always has -- that is, the post-command hook now runs after the primary command hook.

As of version 2.0.0-beta8, [adding @option annotations to hooks](https://github.com/drush-ops/drush/pull/2357#issuecomment-249044561) now attaches those options to the command or commands hooked.  It is also possible to just write an `@hook option` hook that will receive a $command reference that it can add options to directly using code.

c.f. #2336

weitzman:
I pushed a commit which starts using ~2 from Consolidation and updates drupliconCommands as you suggested (I think). The error below prevents from me from seeing if my changes are working. Not sure how to fix it. Feel free to say here or push another commit which fixes it. Thanks @greg-1-anderson.

```
PHP Catchable fatal error:  Argument 2 passed to Robo\Robo::configureContainer() must be an instance of Robo\Config, instance of Symfony\Component\Console\Input\StringInput given, called in /Users/moshe.weitzman/reps/drush/includes/preflight.inc on line 252 and defined in /Users/moshe.weitzman/reps/drush/vendor/consolidation/robo/src/Robo.php on line 103
PHP Stack trace:
PHP   1. {main}() /Users/moshe.weitzman/reps/drush/drush.php:0
PHP   2. drush_main() /Users/moshe.weitzman/reps/drush/drush.php:11
PHP   3. drush_preflight_prepare() /Users/moshe.weitzman/reps/drush/includes/preflight.inc:36
PHP   4. drush_init_dependency_injection_container() /Users/moshe.weitzman/reps/drush/includes/preflight.inc:195
PHP   5. Robo\Robo::configureContainer() /Users/moshe.weitzman/reps/drush/includes/preflight.inc:252
```

greg-1-anderson:
I changed the order of some of the parameters in 2.x for better consistency. I'll pull your commit and fix the constructors.

Note that it will cause problem for Drush clients if we release a stable tag of Drush that uses unstable dependencies, as that would then necessitate that everyone put 'minimum-stability: beta in their Drupal project composer.json files if they also require Drush.  I'll get a stable 2.0.0 release out soon so that we don't need to worry about this.

greg-1-anderson:
I fixed the constructor, so the version of Drush in this PR will now at least run; however, we are still using the native Drush command dispatcher (and I think we must continue to do so on the 8.x branch, as we'd have little hope of maintaining b/c otherwise), and the @hook annotations are not examined and called yet.  I'll work on that next, since hook compatibility is critical to the success of an interoperable branch.

greg-1-anderson:
Seems like annotated commands are completely broken still.

weitzman:
I'm seeing brokenness in 8.x even without this PR:

```
$ drush browse admin/structure

PHP Fatal error:  Uncaught TypeError: Argument 4 passed to Consolidation\AnnotatedCommand\CommandProcessor::process() must be an instance of Consolidation\AnnotatedCommand\AnnotationData, array given, called in /var/www/drupal/public_html/vendor/drush/drush/includes/annotationcommand_adapter.inc on line 191 and defined in /var/www/drupal/public_html/vendor/consolidation/annotated-command/src/CommandProcessor.php:61
Stack trace:
#0 /var/www/drupal/public_html/vendor/drush/drush/includes/annotationcommand_adapter.inc(191): Consolidation\AnnotatedCommand\CommandProcessor->process(Object(Symfony\Component\Console\Output\ConsoleOutput), Array, Array, Array, Array)
#1 /var/www/drupal/public_html/vendor/drush/drush/includes/command.inc(190): annotationcommand_adapter_process_command('admin/structure')
#2 /var/www/drupal/public_html/vendor/drush/drush/lib/Drush/Boot/BaseBoot.php(67): drush_dispatch(Array)
#3 /var/www/drupal/public_html/vendor/drush/drush/includes/preflight.inc(66): Drush\Boot\BaseBoot->bootstrap_and_dispatch()
#4 /var in /var/www/drupal/public_html/vendor/consolidation/annotated-command/src/CommandProcessor.php on line 61
Drush command terminated abnormally due to an unrecoverable error.                                                                                   [error]
Error: Uncaught TypeError: Argument 4 passed to Consolidation\AnnotatedCommand\CommandProcessor::process() must be an instance of
Consolidation\AnnotatedCommand\AnnotationData, array given, called in
/var/www/drupal/public_html/vendor/drush/drush/includes/annotationcommand_adapter.inc on line 191 and defined in
/var/www/drupal/public_html/vendor/consolidation/annotated-command/src/CommandProcessor.php:61
```

greg-1-anderson:
That looks like maybe you didn't run `composer install` when moving from master to 8.x. No matter; I don't think it's worth keeping 8.x on ~1 of annotated command. I'm going to switch 8.x to ~2 right now, and fix whatever API adjustments (e.g. the error above) are needed.

weitzman:
As far as I can tell, that error happenned in a VM after a composer update that brought in latest drush 8.x. Dont think this codebase was ever running drush master.

Thanks a lot for looking into this.

greg-1-anderson:
If that's the case, maybe I accidentally broke b/c in the ~1 branch. Anyway, it doesn't matter to me, as I no longer plan on maintaining ~1.

I fixed the 8.x branch in https://github.com/drush-ops/drush/issues/2366. You can merge that if you want, as long as we get the composer.lock on 2.0.0 stable before making another 8.x stable tag.  That should be available as soon as this afternoon or maybe in the next couple of days.

greg-1-anderson:
With the latest commit, the --druplicon option works with commands that have been ported to annotated commands, but not legacy Drush commands.

Note that in order to get this hook to work, I had to change the annotation to:
```
@option $druplicon Shows the druplicon as glorious ASCII art.
```
The `$` is inherited from the syntax required for `@param`.  Param has to be strict here, because there is an optional type that can be included -- that is, the annotation might be `@param string $myarg`, or just `@param $myarg`.  The code to parse the option annotation uses the same function as the code to parse the param, so it inherits the requirement for the `$`.

I don't like this at all, so I'm going to it that in the upstream; then I will come back here and delete one character, and this should then be ready to merge.

greg-1-anderson:
I still think that --druplicon should become a command, not a hook, but at least this implementation is revealing bugs in the annotated command hook manager. ;)

greg-1-anderson:
There's still `help` work to do here, too, although I think that perhaps hook conversion should be done as a follow-on PR.

weitzman:
What would be the end user effects of changing druplicon to a command? Would it no longer be triggerable on all commands? I think that would limit its already meager raison d'etre.

greg-1-anderson:
The effect would be as you said; the user would have to run it via `drush druplicon` rather than `drush <command> --druplicon`.  It's fine if you prefer to keep it as a hook.

greg-1-anderson:
The help command works now. The problem was that some of the new Annotated Command code worked with Symfony 3, but not with Symfony 2.  That has been fixed in rc6, which is now used in this branch.

We'll see how the tests fare.

greg-1-anderson:
One failure, and yet I can run:
```
$ drush example:formatters --format=json
{
    "en": {
        "first": "One",
        "second": "Two",
        "third": "Three"
    },
    "de": {
        "first": "Eins",
        "second": "Zwei",
        "third": "Drei"
    },
    "jp": {
        "first": "Ichi",
        "second": "Ni",
        "third": "San"
    },
    "es": {
        "first": "Uno",
        "second": "Dos",
        "third": "Tres"
    }
}
```
Will need to look at the test later. Getting close.

weitzman:
I stepped into the test fail. The issue is that the output of an annotated command with a --backend option is too verbose. Specifically, we are wrongly emitting the command output before the usual DRUSH_BACKEND_OUTPUT_START>>> packet. If that output were inside the packet, the output would be parseable. Not sure how you want to fix it. A good place to start is \Consolidation\OutputFormatters\FormatterManager::write

greg-1-anderson:
I guess Drush needs to override the formatter manager and inject its own version into the command factory. Then we can test for --backend prior to calling parent::write.

greg-1-anderson:
Actually, it would probably be better for Drush to just make its own output object in backend mode, and have drush_print() call through to that.  Then, we could stop using ob_start() / ob_end() (save where needed to catch output when bootstrapping Drupal, etc.)

greg-1-anderson:
Fixing backend invoke is going to be a big job. Fortunately, it won't be necessary; I cleaned up some issues with container construction, and the existing code started working again.

I recommend merging here, and handling hook compatibility and global help options in follow-on issues.

greg-1-anderson:
On 8.x, we will need to call new-style hooks from the legacy dispatcher. On master, we'll need to call legacy hooks from the Symfony dispatcher.  (sigh).  I'm going to work on config, esp. command-specific options first, which necessary for the later. I'll merge here first, to avoid being in the way of other efforts.

PR #2366, opened 2016-09-27
Implement --druplicon using new annotated hooks on 8.x branch

greg-1-anderson:
This PR updates the 8.x branch to use Annotated Commands 2.0.0-rc3, to keep it in alignment with what is being used on master (sans Robo).

Note that we should be sure to bump up to 2.0.0 prior to releasing a stable 8.x release -- perhaps even before merging this branch.  This should be available soon.

See also https://github.com/drush-ops/drush/pull/2357


greg-1-anderson:
Adding the druplicon and example:table commands here.  Druplicon command is so that we can get it to work here in this branche; example:table is for convenience, for now; it can ultimately be removed again.

greg-1-anderson:
With the last commit:
```
$ drush help example:table
Demonstrate Robo formatters.  Default format is 'table'.

Examples:
 example:formatters --format=yaml           
 example:formatters --format=csv            
 example:formatters --fields=first,third    
 example:formatters --fields=III,II

Options:
 --druplicon                               Shows the druplicon as glorious ASCII art.                                        
 --fields=<first, second, third>           Fields to output. All available fields are: first, second, third.                 
 --format=<table>                          Select output format. Available: csv, json, list, php, print-r, sections, string, 
                                           table, tsv, var_export, xml, yaml. Default is table.                              
 --french                                  Add a row with French numbers.

Aliases: tf, example-table
```
Hooks are not called yet.

lib/Drush/CommandFiles/core/DrupliconCommands.php
```diff
@@ -0,0 +1,24 @@
...
+class DrupliconCommands {
+
+  /**
+   * Print druplicon as post-command output.
+   *
+   * @hook post-command *
+   * @option $druplicon Shows the druplicon as glorious ASCII art.
+   */
+  public function druplicon($result, $argsAndOptions, $annotationData) {
+    if (isset($argsAndOptions['options']['druplicon'])) {
```
greg-1-anderson: @weitzman Do you hate `$argsAndOptions`?  I think I hate it.  I could split this into separate `$args` and `$options`.  It is tempting to try to make the new post-command hook look like the old post-command hook -- that is, have Drush call it with the same calling conventions as the primary command hook.  However, doing that is hard on hooks that want to attach to every command, as the parameter order can be different for different commands.  Additionally, we want to pass in options as a parameter rather than call a globally-scoped function, and we also have annotation data to provide.

Finally, note that the master branch does not use `isset` in its comparison; at the moment, 8.x is not providing default values for missing options.  I'll try to fix that next, so that we can have the same expectations in 8.x and master.

weitzman: Its a bit ugly, yes. Not that big of a dealbreaker for me though. I'm fine with whatever nastiness we have to do in the 8.x branch. In 9.x, I think we can break backwards compat and head toward a clean API.

greg-1-anderson: Right now the argument list is the same in master. Shall we split it into `$args` and `$options` there? I think that'd be my preference.

weitzman: Sure, split them out. Or make some sort of $request object with $getArguments() and getOptions() methods.

greg-1-anderson:
So, the last thing here (besides waiting to see if there are any test failures to fix) is to consider when the --druplicon hook should fire.

The existing 8.x implementation attached this hook to hook_exit, which prints after all other hooks have been executed.

In this branch, the druplicon hook is attached to post_command, which prints out after most ordinary commands are finished, but before the output from any output formatter has been rendered.

We don't have hook_init and hook_exit in annotated command at the moment.  Are these necessary?

Commandfile classes can now do something akin to hook_init by running code in their constructor.  We could potentially add an explicit `init` method (advertised via an appropriate interface that the commandfile would implement) in case folks want to do a lot in `init`, and don't want to do it in the constructor.  Similarly, if it was sufficiently important, we could also have an interface for 'exit' events, and call these when Drush is finishing up.

Not sure how much we care where the Druplicon prints, or how likely it is that this issue will be needed by other commands.  We can always add more hooks without breaking backwards compatibility, so perhaps the post-command hook will do for now.

greg-1-anderson:
Also, this version kind of goes out of its way to try to not print the Druplicon twice.  LMK if you think it should just let the cards (and Druplicons) fall as they may. This would be cleaner if we used a hook_exit.

greg-1-anderson:
Recommend merging this and #2370, and doing other work in follow-on PRs.  If you think the Annotated Commands library API is good enough now, I'll tag the 2.0.0 stable release and update composer.json here to use it.

greg-1-anderson:
Oh! I just noticed that I forgot to implement a bunch of hooks. Working on that.

includes/annotationcommand_adapter.inc
```diff
@@ -329,19 +354,100 @@ function annotationcommand_adapter_get_commands_for_commandhandler($commandhandl
...
+  if (!empty($command['annotated-command-callback'])) {
+    $commandHandler = $command['annotated-command-callback'][0];
+    $reflectionClass = new \ReflectionClass($commandHandler);
+    $commandFileClass = $reflectionClass->getName();
+    $names[] = $commandFileClass;
+  }
+  return $names;
+}
+
+function annotationcommand_adapter_call_hook_pre_validate($names, CommandData $commandData)
```
weitzman: Not clear what these empty functions are for. Please add Doxygen. A bit more Doxygen would be welcome, in the adapter in general.

lib/Drush/CommandFiles/core/DrupliconCommands.php
```diff
@@ -0,0 +1,42 @@
...
+use Consolidation\AnnotatedCommand\CommandData;
+
+class DrupliconCommands {
+  protected $printed = false;
+
+  /**
+   * Print druplicon as post-command output.
+   *
+   * @hook post-command *
+   * @option $druplicon Shows the druplicon as glorious ASCII art.
```
weitzman: Lets document why this option name begins with a $

greg-1-anderson: Oh, I fixed that in the upstream -- the $ is now optional.  I'll push a new commit to remove it.

includes/annotationcommand_adapter.inc
greg-1-anderson: I implemented all of these.  Let me check and see if I forgot to push that commit.

greg-1-anderson: Whoops, these got mixed in with the next commit. I'll separate them and push it back up in a minute.

greg-1-anderson:
Okay, fixed that up, and also tagged stable releases for annotated-command and output-formatters, and included those in the composer.json here.

greg-1-anderson:
I'll document the annotated command adapter file when I get a chance; might make sense to merge here first, though, since this PR fixes the broken tests on the 8.x branch.

weitzman:
Feel free to merge when you are satisfied. 

One last thought - can we do hidden options in annotated-command? Or "global" options? I'm not too keen on showing --drupalicon in the help for every command. Lets discuss further at #2336.
 
Thanks!

PR #2368, opened 2016-09-29
Backport Composer support into make-convert command

phenaproxima:
This functionality is present in Drush 9.x, but not in Drush 8.x. It'd be really useful to have it here, so this PR backports (actually just copy-and-pastes) it back into the 8.x branch.

### DIFF
--- a/commands/make/make.drush.inc
+++ b/commands/make/make.drush.inc
@@ -679,6 +679,32 @@ function drush_make_convert_project_to_composer($original_project, $core_major_v
 }
 
 /**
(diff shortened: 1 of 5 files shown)


greg-1-anderson:
Since we are planning on ultimately removing make and pm-* from the master branch, it would be good to have this in the 8.x branch.

+1

fuzzy76:
This backport broke 8.x since the command started throwing the output to stdout instead of writing to a file per the documentation. :-/ Se #2399

Issue #2365, opened 2016-09-27
phpdocumentor fatal error since drush 8.1.5


accuraz:
I'll get the same fatal error on every drush command since drush 8.1.5
</br>
That's the log :
</br>
```
$ drush cron
PHP Fatal error:  phpDocumentor\Reflection\DocBlock\Tags\BaseTag cannot implement phpDocumentor\Reflection\DocBlock\Tag - it is not an interface in /var/www/exemple/vendor/phpdocumentor/reflection-docblock/src/DocBlock/Tags/BaseTag.php on line 21
Drush command terminated abnormally due to an unrecoverable error.   [error]
Error: phpDocumentor\Reflection\DocBlock\Tags\BaseTag cannot
implement phpDocumentor\Reflection\DocBlock\Tag - it is not an
interface in
/var/www/exemple/vendor/phpdocumentor/reflection-docblock/src/DocBlock/Tags/BaseTag.php,
line 21
```

Is this some king of a bug or maybe can someone help me out ?
</br>
Thanks,

weitzman:
Are you using a global Drush and/or a Drush local to your project via
composer.json? How did you upgrade? Make sure you do a composer update to
get new dependencies.

accuraz:
I'm using a global drush and I ran composer update after I get the error message and before I post this issue.
</br>
If it might help, I run php7

weitzman:
What version of Drupal? Can you run `drush status --debug` when you are not in a drupal directory?

accuraz:
I run Drupal 8.1.10
</br>
I can run `drush status --debug` when I'm not into a Drupal directory with no errors at output
</br>
If I run `drush status --debug` into a Drupal directory I'm getting the same errors about phpdocumentor

sms20:
Yeah got the same issue as well.

Steps to replicate:
1) Install Drupal 8.1.0 
2) Install AMP Module (https://www.drupal.org/project/amp)
3) Run composer drupal-update
4) Run drush status - now reports error as above

Running PHP 5.5.9

greg-1-anderson:
Drush uses phpDocumentor, but it can use either version 2 or version 3. The trouble is, if you are using a global Drush, it will be using version 3.  This will cause problems for you if your Drupal site is using a module that uses version 2.

Two solutions:

1. Install global Drush via git clone + composer install, then use `composer require` to specifically request version 2 of phpDocumentor.

[...]

PR #2369, opened 2016-09-29
Update drush dependency phpdocumentor/reflection-docblock to 2.x

JTubex:
Drush is currently broken when phpdocumentor/reflection-docblock version 3 is used as seen in #2365

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,8 @@
     "symfony/event-dispatcher": "2.7.*",
     "symfony/config": "~2.2",
     "pear/console_table": "~1.3.0",
-    "webmozart/path-util": "~2"
+    "webmozart/path-util": "~2",
+    "phpdocumentor/reflection-docblock": "^2.0"
   },
   "require-dev": {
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
It would be nice to have a test that fails (would fail) with 3.x, but passes with this PR.

JTubex:
I'm unfamiliar with drush best-practices, but this test did the trick for me.

greg-1-anderson:
While I appreciate your efforts, I don't think that a test of `composer update` is what we need here. The general idea is to exercise some Drupal code that calls phpdocumentor, so that if in the future the version Drush pulls in becomes incompatible with Drupal's version (e.g. if they upgrade to version 3.x), then the test should fail.

We could potentially just merge here if no one has time to do that, but it would be good to have.

weitzman:
Thanks. Decided to commit this to 8.x and master, without the test. Appreciate your efforts.

dsteplight:
I manually removed the folder from my vendors directory. That solved the problem for me.

PR #2370, opened 2016-09-30
Update Druplicon hook and ExampleCommandFile to new hook APIs

greg-1-anderson:
Making the `master` branch like #2366


lib/Drush/CommandFiles/core/DrupliconCommands.php
```diff
@@ -1,16 +1,34 @@
...
+    }
+    $this->printed = true;
+    $annotationData = $commandData->annotationData();
+    $commandName = $annotationData['command'];
+    // For some reason, Drush help uses drush_invoke_process to call helpsingle
+    if ($commandName == 'helpsingle') {
+      return;
+    }
+    drush_log(dt('Displaying Druplicon for "!command" command.', array('!command' => $commandName)));
+    if ($commandData->input()->getOption('druplicon')) {
```
greg-1-anderson: Robo already has a Config object that can be passed in to a command handler via dependency injection.  I intend to make a system similar to Drush's current procedural system that will just set the option values directly on the $input object, so a command object would only need to have the Config object injected if it needed access to more than just its command args and options.

lib/Drush/CommandFiles/ExampleCommandFile.php
```diff
@@ -46,9 +48,9 @@ public function exampleTable($options = ['format' => 'table', 'fields' => ''])
      * @option $french Add a row with French numbers.
      * @usage example:formatters --french
      */
-    public function alterFormatters($result, array $args, AnnotationData $annotationData)
+    public function alterFormatters($result, CommandData $commandData)
     {
-        if ($args['options']['french']) {
+        if ($commandData->input()->getOption('french')) {
```
greg-1-anderson: CommandData is only used in the hook system; there are no other clients of that, so I could bend the semver and change its name in 2.1.0 and it wouldn't hurt anything.

CommandData includes information about the command for a hook.  The information it includes is the Input object (which is where getOption() comes from -- Symfony's InputInterface::getOption()), the Output object, and the annotation data, which includes the command name, but does not include the entire command record.

weitzman: OK thanks. There may be no need to change the name then ... Do you think that --druplicon should write to Output object instead of drush_print(). --druplicon is a silly edge case so maybe we just dont worry about it.

greg-1-anderson:
Oh, changes approved -- guess I'll merge. Still getting used to the new GitHub system too.

PR #2373, opened 2016-10-04
8 to master

greg-1-anderson:
- [x] Need to initialize Robo commandfiles correctly
- [x] Need to bootstrap annotated commands correctly
- [x] Need to support @allow-additional-options

So, I went ahead and added support for `@allow-additional-options` here.  Truthfully, this might be difficult to support in Symfony.  We can postpone removing it until later; however, we should consider whether there is a good use-case for allowing unvalidated options moving forward. Hooks can always declare their additional options explicitly with @option, and that is a validated pattern.  `php-eval` / `php-script` seems to be the one place this might be helpful; however, it might be better to simply de-support `drush_get_option()` in these sorts of scripts, and force folks to define an actual command if arguments and options are desired.

Another area where I broke backwards compatibility is in `php-eval` output.  The output-formatters library only outputs objects, arrays and strings.  Booleans and integers are ignored. Again, I think that output of integers and booleans is not really important in most cases, with `php-eval` being a notable exception.  I figure that we can force folks to use `var_export($foo);` instead of `return $foo;` if they wish to display booleans, though; perhaps this is good enough.

Enough tests have passed that I think this test run will finish up all green.


commands/core/init.drush.inc
```diff
@@ -0,0 +1,173 @@
+<?php
```
weitzman: We already have https://github.com/drush-ops/drush/blob/master/lib/Drush/CommandFiles/InitCommandFile.php. We should not add this old style file back in, I think.

commands/make/make.drush.inc
```diff
@@ -155,7 +155,12 @@ function make_drush_command() {
       'makefile' => 'Filename of the makefile to convert.',
     ),
     'options' => array(
-      'format' => 'The format to which the make file should be converted. Accepted values include make, composer, and yml.'
+      'format' => 'The format to which the make file should be converted. Accepted values include make, composer, and yml.',
+/*
```
weitzman: Debug code? Why commented out?

includes/annotationcommand_adapter.inc
```diff
@@ -0,0 +1,798 @@
...
+function annotationcommand_adapter_get_discovery() {
+  static $discovery;
+  if (!isset($discovery)) {
+    $discovery = new CommandFileDiscovery();
+  }
+  return $discovery;
+}
+
+/**
+ * Initialize and cache the command factory. Drush 9 uses dependency injection.
```
weitzman: Confusing reference to Drush 9, since this code is itself Drush 9.

includes/annotationcommand_adapter.inc
```diff
@@ -0,0 +1,798 @@
...
+    array_merge(
+      drush_get_context('DRUSH_ANNOTATED_COMMANDFILES'),
+      $annotation_commandfiles
+    )
+  );
+}
+
+/**
+ * This function is set as the $command['callback'] for Symfony Console commands
+ * e.g. those provided by Drupal 8 modules.  Note that Drush 9 calls these with
```
weitzman: Perhaps this comment needs updating as this code is Drush 9.

includes/array_column.inc
```diff
@@ -0,0 +1,117 @@
+<?php
+/**
+ * This file was recommended at http://php.net/manual/en/function.array-column.php#refsect1-function.array-column-seealso
```
weitzman: This file is no longer needed since master branch wont support PHP 5.4

includes/command.inc
```diff
@@ -317,14 +331,55 @@ function _drush_invoke_hooks($command, $args) {
     }
   }
 
+  // We will adapt and call as many of the annotated command hooks as we can.
+  // The following command hooks are not supported in Drush 8.x:
```
weitzman: Wrong Drush version mentioned.

lib/Drush/UpdateService/ReleaseInfo.php
```diff
@@ -183,6 +183,19 @@ public function selectReleaseBasedOnStrategy($request, $restrict_to = '', $selec
     }
     $releases = $project_release_info->filterReleases($filter, $version);
 
+    // Special checking: Drupal 6 is EOL, so there are no stable
```
weitzman: I think this whole block is unneeded, as we dont support Drupal 6 in master.

tests/coreTest.php
```diff
@@ -130,6 +130,11 @@ function testCoreRequirements() {
       'ignore' => 'cron,http requests,update,update_core,trusted_host_patterns', // no network access when running in tests, so ignore these
       'strict' => 0, // invoke from script: do not verify options
     );
+    // Drupal 6 has reached EOL, so we will always get errors for 'update_contrib';
```
weitzman: Not needed on master

tests/pmUpdateCodeTest.php
```diff
@@ -68,6 +68,9 @@ public function setUp() {
   }
 
   function testUpdateCode() {
+    if (UNISH_DRUPAL_MAJOR_VERSION < 7) {
```
weitzman: Not needed.

includes/annotationcommand_adapter.inc
greg-1-anderson: I'll fix up the comment. This code will use more DI in the near future, but does not right now.

greg-1-anderson:
kk, did all of that; tests still green.

weitzman:
- Whats BuilderAwareInterface? In general, it would be ideal to detail the command authors should use Robo features. I see that Init command does so, but not others.
- druplicon appears as a command in `drush help`. Do we support a @hidden annotation yet?
- -—druplicon option is showing under all commands. Would be good for this to be a hidden option only visible under global options.
- The end of `drush help` shows a bunch of non end-user commands. See `Other commands: (ExampleCommandFile,EvalCommandFile)`, `All commands in InitCommandFile: (InitCommandFile)`
- Upon running `drush browse -v admin`, I see that \Drush\CommandFiles\core\BrowseCommands::browse is passed the following array as $options:
```
$options = {array} [9]
 no-browser = true
 verbose = true
 browser = false
 druplicon = false
 config-file = "/Users/moshe.weitzman/reps/d7/drush/drushrc.php"
 context-path = "/Users/moshe.weitzman/reps/d7/drush/drushrc.php"
 uri = "http://d7.dd:8083"
 php = "/Applications/DevDesktop/php5_6/bin/php"
 php-options = false
```
Kind of an odd set of keys in there. I wonder if this is the most clear API to be handing our command callback authors.
- Inside that browse method, I see `drush_invoke_process($site_record, 'browse', func_get_args()`. The func_get_args() is no longer correct. Should be `[$path]`, I think.
- Also from browse(), I see 
```
      if (!drush_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_FULL)) {
        // Fail gracefully if unable to bootstrap Drupal. drush_bootstrap() has
        // already logged an error.
        return FALSE;
      }
```
The code comment and behavior look wrong. If I run `drush browse` outside a Drupal site, I see somewhat client error message: `[error]  We could not find an applicable site for that command. `.
- FYI there is a pre-2.0 code comment at \Consolidation\AnnotatedCommand\Hooks\HookManager::callProcessor
- This log message should be inside the conditional below it: `drush_log(dt('Displaying Druplicon for "!command" command.', array('!command' => $commandName)));`
- I’m seeing notices when I run `dr st -v` from outside a site:
```
~/reps/drush (8-to-master<) $ dr st -v
 [info] Undefined index: name annotationcommand_adapter.inc:506
 [info] Loading outputformat engine.
 [info] Undefined index: name annotationcommand_adapter.inc:506
 [info] Undefined index: name annotationcommand_adapter.inc:506
 [info] Undefined variable: annotationData annotationcommand_adapter.inc:735
 [info] Undefined variable: annotationData annotationcommand_adapter.inc:735
 [info] Undefined variable: annotationData annotationcommand_adapter.inc:735
 [info] Undefined variable: annotationData annotationcommand_adapter.inc:735
```
- \Drush\CommandFiles\InitCommandFile should now have @bootstrap NONE annotation and remove related code @todo.

greg-1-anderson:
`@bootstrap none` is currently the default. I guess that's backwards from current Drush; should probably make the default `full` and make folks label it `none` like today -- although if we were starting over, I'd say no annotation == no behavior == bootstrap none.

greg-1-anderson:
Could / should we just start `throw`-ing from commandfiles if we want to stop execution with an error?

greg-1-anderson:
I got a lot of the above comments cleared up, but there are a few open threads on this issue. Could we merge here, and discuss each of the open questions in separate issues / PRs?

weitzman:
- I can get behind BOOTSTRAP_NONE being the default. Its strictly more of a DrupalBoot thing than a Drush thing.
- Yes, throwing exceptions sounds good to me. Also, I think we can ditch rollback hooks when we ditch the legacy runner. 
- I notice that this PR merges 226 commits. I'm ambivalent about whether you squash merge it or not. Your call. Interestingly, this PR lists 30 participants. They must have been added because they submitted a commit.
- Fine with me to merge and open followups.

Thanks @greg-1-anderson.

greg-1-anderson:
All of the commits here are an artifact of the fact that GitHub is confused about the merge that is happening.  I think that all of those commits are everything on the 8.x branch that has been committed since it forked off of master, and all of the participants here are all the folks who made at least one of those commits.

I'll merge with caution through the cli to ensure that what we end up with is rational.

weitzman:
This epic PR has landed. Thanks @greg-1-anderson.

PR #2383, opened 2016-10-08
Allow setting a different terminal environment for an interactive SSH tty

dmsmidt:
Sometimes it is necessary to set a different terminal environment type for an interactive SSH tty. This is the case if the terminal emulator type you use locally is not recognised on the remote machine.

If so, the interactive ssh connection will be closed immediately with a hint: "Try requesting a different terminal environment."

This pull request adds an option to request a different type so we can happily use Drush in any terminal emulator we like (in my case urxvt).

### DIFF
--- a/commands/core/sitealias.drush.inc
+++ b/commands/core/sitealias.drush.inc
@@ -37,7 +37,7 @@ function sitealias_drush_command() {
       'variable-name' => 'aliases',
       'hide-empty-fields' => TRUE,
       'private-fields' => 'password',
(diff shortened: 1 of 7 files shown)


greg-1-anderson:
Sorry I missed this two years ago. Seems reasonable, but would need to be re-done with our latest API changes ([site-process](https://github.com/consolidation/site-process)). See the [RedispatchHook](https://github.com/drush-ops/drush/blob/9.6.0-beta3/src/Runtime/RedispatchHook.php#L92), which would need to be modified to take environment variables from the alias.

I don't think this feature would need to be specific to the TERM environment variable.

weitzman:
[In prior versions of Drush we have accepted #env-vars in a site alias.](https://github.com/drush-ops/drush/blob/8.x/examples/example.aliases.drushrc.php#L253-L254) I think that would solve this.

Not yet possible in Drush9. Would be nice. Process already lets you set env vars for a call so that part is simple.

PR #2384, opened 2016-10-10
Port views-* to Annotated

weitzman:
There are a few @todo in new ViewsCommands class. Help wanted. ping @greg-1-anderson


weitzman:
@damiankloip - currently, views-execute just barfs out a print_r() of each Drupal\views\ResultRow object, one for each row. Is there something nicer we can output here?

damiankloip:
@weitzman Yeah, I'm sure we could do something nicer. The whole results array is just returned currently,so I guess the default drush output is using that? So I guess setting an output format would still be supported? How does that look currently? ResultRow should just have a load of public properties on.

weitzman:
I think there is no same way right now to output the results of a views-execute on the CLI. Maybe if we had our own Views row formatter (or something). For now, I suggest we default to outputting a result count ( @greg-1-anderson how can we return an integer without it being interpreted as a status code) and keep --rendered as an option.

weitzman:
@greg-1-anderson - just pushed a small fix to the adapter in c5797d6. It was setting too many required-arguments.

greg-1-anderson:
@weitzman The output formatters can format arrays, or objects that can be converted into Array (e.g. ArrayObject and subclasses thereof), and also objects that are specifically handled by formatters (e.g. RowsOfFields, ArrayObjects).  While the data-transformation capabilities of the output formatters engine is flexible enough to allow for the conversion of an integer to a string, in general there is no benefit to sending integers through the output formatters, because there is no use-case for translating them into yaml, json etc.

Since returning integer results that are not status result codes is an uncommon use-case, I think it would be better to just use backend set result and drush_print (or $output->writeln) directly.

If you really want to return the integer just to get it stuffed into the backend result consistently, then we could make something like [CommandError](https://github.com/consolidation/annotated-command/blob/master/src/CommandError.php) that held the integer in the `message`.  CommandError wasn't really designed for this use (returning non-error results), since integers aren't really a use-case for output formatters, but something like this could work.

greg-1-anderson:
The Robo [ResultData](https://github.com/consolidation/Robo/blob/master/src/ResultData.php) should work for this, if we want to go this way.  Of course, Robo requires php 5.5 and is not currently used in Drush 8.x, so using Robo classes would be a Drush-9-only solution.

weitzman:
For views-execute, maybe what we can should do is [convert the rendered HTML to markdown](https://github.com/thephpleague/html-to-markdown) for easier CLI readability. Not sure if its a problem to add dependencies on the XML and DOM PHP extensions. There is also good ole [drupal_html_to_mail()](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Mail!MailFormatHelper.php/function/MailFormatHelper%3A%3AhtmlToText/8.2.x) but that doesn't convert tables.

PR #2386, opened 2016-10-11
fix(bootstrap): Do not clear the module_implements() cache on a drush invocation

klausi:
We run drush cron every 10 minutes on a site and we saw a lot of cache clearing on every cron run. The problem was that the cache entry for system_cache_tables was gone after each drush invocation.

Traced that back to drush bootstrap where the module_implements() cache is reset. Drush should never do that when a random drush command is executed. We can inject the logging into the static module implements cache instead.

### DIFF
--- a/lib/Drush/Boot/DrupalBoot7.php
+++ b/lib/Drush/Boot/DrupalBoot7.php
@@ -33,8 +33,14 @@ class DrupalBoot7 extends DrupalBoot {
     // If needed, prod module_implements() to recognize our system_watchdog() implementation.
     $dogs = drush_module_implements('watchdog');
     if (!in_array('system', $dogs)) {
-      // Note that this resets module_implements cache.
-      drush_module_implements('watchdog', FALSE, TRUE);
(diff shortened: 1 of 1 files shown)


damienmckenna:
Did you test to confirm there were no problems with "drush pm-enable", "drush pm-install", "drush pm-disable" and "drush pm-uninstall"? They all call ->add_logger() too.

weitzman:
Cron every 10 minutes? Anyway, this PR looks fine to me.

damienmckenna:
FYI I help run a site that runs cron every 1 minute, though it uses Ultimate Cron to limit which cron hooks are actually triggered.

damienmckenna:
I understand that the code might not be possible to port to DrupalBoot6 due to core's limitations, but has anyone looked to see whether the issue would apply to DrupalBoot8?

weitzman:
Merged to 8.x and master.

Does not apply to DrupalBoot8 as we add a proper Logger alongside Drupal's logger.

Issue #2390, opened 2016-10-13
Make --include-overridden the default


damienmckenna:
In Drupal 7 when you do "drush vget [varname]" it always shows the variable as it would be presented to Drupal's UI, including any variables that are overridden via the appropriate settings.php file. In D8 the equivalent command "drush cget [configname]" does not include config values overridden unless the "--include-overridden" argument is provided. This is a major change to the expected output and what one would expect when querying a specific instance of the site; there isn't a flag to enable in Drupal to "accept values overridden in the settings.php file" so there also shouldn't be a flag for Drush to present the same data structures, it should be automatic.

weitzman:
Lets do conversation in the PR.

PR #2391, opened 2016-10-13
PR for issue #2390 - replace --include-overridden

damienmckenna:
Removes the --include-overridden option the default for "config-get". Adds a new "--exclude-overridden" option to make the previous logic available.


damienmckenna:
This is for issue #2390.

weitzman:
I can see an argument for either way. Lets leave is as.

PR #2393, opened 2016-10-14
Use Request::create() to create the request for drush cr so that the request method is set

Berdir:
To reproduce:

1. Create a menu link pointing to /user/logout in the UI
2. Run drush cr, then access the page as anonymous user. The Logout link is visible
3. Clear cache in the UI, the logout link is not visible as anonymous user.

See also https://www.drupal.org/node/2817911 and possibly other duplicate bug reports caused by this.

Interesting site note: This only behaves like this in 8.2, in 8.1 it worked fine but I think this change makes sense anyway.

There is at least another createFromGlobals() but that's only a fallback that might not actually be used anmyore.


weitzman:
I'm happy to merge this if you think its ready.

dawehner:
Works for me

PR #2394, opened 2016-10-16
Fix "Could not open input file" error under Windows

Fonata:
I installed Drush on Windows using ``composer global require drush/drush:8.*``; however, calling ``drush afterwards yielded this error: 

``Could not open input file: 'C:\Users\Blaeul\AppData\Roaming\Composer\vendor\drush\drush\drush.php'``

The error comes on cmd.exe as well as on MINGW64.

After some digging in the code, I could make Drush work perfectly normal by changing ``drush_is_windows`` in the following way:

````
function drush_is_windows($os = PHP_OS) {
  if (!$os) {
    $os = PHP_OS;
  }
  return strtoupper(substr($os, 0, 3)) === 'WIN';
}
````

Thus it seems like my PHP doesn't like the default value for the $os parameter. I am using 
````
PHP 5.6.8 (cli) (built: Apr 15 2015 15:07:09)
    Zend Engine v2.6.0, 
    with Zend OPcache v7.0.4-dev
````

### DIFF
--- a/includes/startup.inc
+++ b/includes/startup.inc
@@ -128,7 +128,12 @@ function find_wrapper_or_launcher_at_location($location) {
 /**
  * Determine whether current OS is a Windows variant.
  */
-function drush_is_windows($os = PHP_OS) {
+function drush_is_windows($os = NULL) {
+  if (!$os) {
+    // cannot use PHP_OS as default value
+    $os = PHP_OS;
+  }
+  
   return strtoupper(substr($os, 0, 3)) === 'WIN';
 }
 

### CODE REVIEW

**Comment by weitzman** [comment]
Its been valid to use  a constant as default value in function param for a long time in PHP. See http://stackoverflow.com/questions/1229047/using-constants-as-default-function-values-in-php. There must be something else going on.

**Comment by Fonata** [comment]
My incorrect suspicion was Zend OPcache. I checked that is is enabled in CLI using ``echo function_exists('opcache_invalidate');``, disabled it by commenting out ``zend_extension = "C:\xampp\php\ext\php_opcache.dll"``. Now, the echo above is quiet, but the unusual ``drush_is_windows`` behaviour persists without this PR. I have no explanation. :disappointed:

**Comment by Narretz** [comment]
The problem is further up in the code:
This function https://github.com/drush-ops/drush/blob/2b44148ad876b6108dee5a7cbf5927a75b1ce02d/includes/startup.inc#L135-L146 calls drush_is_windows($os), but drush_escapeshellarg has set $os = null.
```
function drush_escapeshellarg($arg, $os = NULL, $raw = FALSE) {
  // Short-circuit escaping for simple params (keep stuff readable)
  if (preg_match('|^[a-zA-Z0-9.:/_-]*$|', $arg)) {
    return $arg;
  }
  elseif (drush_is_windows($os)) {
    return _drush_escapeshellarg_windows($arg, $raw);
  }
  else {
    return _drush_escapeshellarg_linux($arg, $raw);
  }
}
```

The problem is that drush_escapeshellarg is called like this 
`$escaped_args = array_map(function($item) { return drush_escapeshellarg($item); }, $arguments);`
https://github.com/drush-ops/drush/blob/2b44148ad876b6108dee5a7cbf5927a75b1ce02d/includes/startup.inc#L425

which means NULL is passed to https://github.com/drush-ops/drush/blob/2b44148ad876b6108dee5a7cbf5927a75b1ce02d/includes/startup.inc#L131-L133
```
function drush_is_windows($os = PHP_OS) {
  return strtoupper(substr($os, 0, 3)) === 'WIN';
}
```

PR #2397, opened 2016-10-17
Add a base class called DrushCommands

weitzman:
- [ ] Is the unset() in preflight.inc sufficient? Am seeing errors without it.
- [ ] the io() method in DrushCommands currently references a TerminusStyle. Do we want one of these for Drush? I'm especially thinking of of custom table formatting needs.


greg-1-anderson:
Both Drupal Console and Terminus make their own subclass of SymfonyStyle, per http://symfony.com/doc/current/console/style.html#defining-your-own-styles. We could do this in Drush as well.

PR #2416, opened 2016-10-27
[WIP] Compile Twig templates directly, without going through the render engine

Kazanir:
I've been working on a path towards supporting D8 sites with a pre-built code artifact for HHVM's repo authoritative mode. I talked with dawehner about this a bit at BADCamp and we ran into some trouble with the existing Twig compilation command around render contexts. Since we're eventually working towards supporting compilation without database access, we figured it would be simpler to at least change the command to compile the templates directly rather than calling the complete rendering engine.

This is a work in progress, I hope to return to this soon with additional changes to make this command usable without a database connection.

### DIFF
--- a/commands/core/twig.drush.inc
+++ b/commands/core/twig.drush.inc
@@ -1,0 +1,191 @@
+<?php
+
+use Drupal\Core\Site\Settings;
+use Drupal\Core\Config\NullStorage;
+use Drupal\Core\Database\Connection;
+use Drupal\Core\DependencyInjection\ServiceModifierInterface;
+use Drupal\Core\DependencyInjection\ServiceProviderInterface;
+use Drupal\Core\DependencyInjection\ContainerBuilder;
+use Drush\Log\LogLevel;
+
+/**
+ * Implements hook_drush_command().
+ */
+function twig_drush_command() {
+  $items['twig-compile'] = array(
+    'description' => 'Compile all Twig template(s).',
+    'aliases' => array('twigc'),
+    'bootstrap' => DRUSH_BOOTSTRAP_DRUPAL_SITE,
+    'options' => array(
+      'standalone' => 'Run this command on a Drupal codebase without attempt to connect to a working database or installed site. This will process all themes and modules regardless of their enabled state. Requires PHP 7+.',
+    ),
+    'core' => array('8+'),
+  );
+  return $items;
+}
+
+function drush_twig_compile() {
+  $bootstrap = \Drush::bootstrap();
+  if (!drush_get_option('standalone')) {
+    drush_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_FULL);
+    $searchpaths = $bootstrap->commandfile_searchpaths(DRUSH_BOOTSTRAP_DRUPAL_FULL);
+
+  }
+  else {
+    // Scan all modules and themes. By (ab)using DRUSH_BOOTSTRAP_DRUPAL_SITE we
+    // lose the ability to exclude modules, but no longer require database access
+    // to get the info of active themes. Since compiling templates is really
+    // cheap this is a valuable tradeoff.
+    // With this method we need to make sure to search core directories too.
+    $searchpaths = array(
+      'core/themes',
+      'core/modules',
+      'core/profiles',
+      'profiles',
+    );
+    $searchpaths += $bootstrap->commandfile_searchpaths(DRUSH_BOOTSTRAP_DRUPAL_SITE);
+    // We have this command set to use DRUSH_BOOTSTRAP_DRUPAL SITE. This let's us
+    // run the command and intervene before further bootstrapping into the
+    // _CONFIGURATION and _FULL phases which we do here manually.
+    drush_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION);
+
+    // Intervene in settings.php here to skip all database stuff.
+    $settings = Settings::getAll();
+    $settings_object = Settings::getInstance();
+    $ref = new ReflectionProperty($settings_object, 'storage');
+    $ref->setAccessible(TRUE);
+    $settings = drush_twig_custom_settings($settings);
+    $ref->setValue($settings_object, $settings);
+
+    // Add a service provider which will nullify the KeyValue store for us,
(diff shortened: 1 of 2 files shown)


weitzman:
Sounds useful. Looking forward to this Thanks.

Kazanir:
Well.

- This is definitely the gnarliest code I've ever written and I understand the Drupal container, kernel, and bootstrap process a lot better now.
- It was immensely cool to be able to pull it off without having to just put stuff in settings.php anywhere, even though I had to cheat in multiple ways (Reflection, bootstrapping upwards mid-command, fake database classes...) to do it.
- Speaking of which, Good-aligned Drupalists reading this code may need to make a saving throw or be driven mad. You have been warned.
- I might have been able to get away without using anonymous classes, but since this is intended for HHVM users I figured some PHP7 features are fair game.
- Feedback is obviously extremely welcome since I'm deep into bat country at this point.

commands/core/twig.drush.inc
```diff
@@ -0,0 +1,191 @@
...
+      public function register(ContainerBuilder $container) {
+        $container->register('keyvalue.expirable.null', 'Drupal\Core\KeyValueStore\KeyValueNullExpirableFactory');
+      }
+    };
+
+    // Skip the database bootstrapping phase.
+    drush_set_context('DRUSH_BOOTSTRAP_VALIDATION_PHASE', DRUSH_BOOTSTRAP_DRUPAL_DATABASE);
+    drush_set_context('DRUSH_BOOTSTRAP_PHASE', DRUSH_BOOTSTRAP_DRUPAL_DATABASE);
+    // Pretend to have a working site.
+    drush_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_FULL);
```
greg-1-anderson: Failed my saving throw. :)

commands/core/twig.drush.inc
```diff
@@ -0,0 +1,191 @@
...
+      public function databaseType() {}
+      public function driver() {}
+    });
+  }
+
+  require_once DRUSH_DRUPAL_CORE . "/themes/engines/twig/twig.engine";
+  $twig = \Drupal::service('twig');
+
+  foreach ($searchpaths as $searchpath) {
+    foreach ($file = drush_scan_directory($searchpath, '/\.html.twig/', array('tests')) as $file) {
```
greg-1-anderson: Could you do this without `$searchpath`?  Could you just search for all .html.twig classes beneath the Drupal root, for example, without incurring too much of a performance penalty?

commands/core/twig.drush.inc
greg-1-anderson: How about:
```
$bootstrap->bootstrap_drupal_full();
```
You can keep `drush_bootstrap(DRUSH_BOOTSTRAP_DRUPAL_CONFIGURATION);`, but if you just call the `$bootstrap` method directly, then you shouldn't need to call `drush_set_context` to perturb Drush's bootstrapping behavior. However, I wonder what you are really skipping here, as the Drush `database` bootstrap phase [does not do much in Drupal 8](https://github.com/drush-ops/drush/blob/3b1eaaa3972e2e282a61f0f68772906dc23bce37/lib/Drush/Boot/DrupalBoot8.php#L116).  Are you just skipping [database validation](https://github.com/drush-ops/drush/blob/3b1eaaa3972e2e282a61f0f68772906dc23bce37/lib/Drush/Boot/DrupalBoot.php#L475), so that your settings.php file does not need to be configured?

greg-1-anderson:
Looks pretty cool. Didn't try to run it; my suggestions are untested, and may or may not be viable.

Kazanir:
Last thought is -- is this even the right approach? The only reason to bootstrap Drupal this far is to get a functional container, but maybe it would be possible to instantiate the Twig environment directly while still in DRUSH_BOOTSTRAP_DRUPAL_SITE. I guess the question is if this hackery is saving us enough work we'd have to do in manually instantiating the entire DI chain without the container to do it for us.

commands/core/twig.drush.inc
Kazanir: Yeah that's probably for the best. There's also the possibility of stuff in a non-DRUPAL_ROOT vendor directory containing templates -- we haven't crossed that Rubicon yet but I imagine it'll happen sometime in the D8 lifecycle so I should figure out how to handle that too.

commands/core/twig.drush.inc
```diff
@@ -0,0 +1,191 @@
...
+    $settings = Settings::getAll();
+    $settings_object = Settings::getInstance();
+    $ref = new ReflectionProperty($settings_object, 'storage');
+    $ref->setAccessible(TRUE);
+    $settings = drush_twig_custom_settings($settings);
+    $ref->setValue($settings_object, $settings);
+
+    // Add a service provider which will nullify the KeyValue store for us,
+    // because that no longer has a nice convenient $settings override.
+    $GLOBALS['conf']['container_service_providers'][] = new class() implements ServiceProviderInterface, ServiceModifierInterface {
```
greg-1-anderson: If this is a separate extension, then I think it would be fine to require php7

commands/core/twig.drush.inc
Kazanir: Yes, it is all about skipping the validation so that those queries don't need to fire. The ultimate goal here (motivated very obviously by Platform.sh's build step) is to be able to run without a database even present. I think that's a good idea anyway since in theory you should be able to do template compilation without an installed site -- tricking Drupal into running in spite of the missing database was just the easiest path to get there.

greg-1-anderson:
I think that the two main questions are:

- How likely is it for the `FULL`  bootstrap to stop working if the `DATABASE` phase is skipped?
- How likely is it that the DI chain might change, breaking manual construction of the twig objects?

I think that, in terms of the full bootstrap issue, it is somewhat likely that Drush might break the full bootstrap (sans-database), but unlikely that the Kernel initialization will ever require the database. Also, it seems to me that it is likely that, as Drupal becomes more decoupled from its legacy Drupal 7 roots and more class/container based, that we are likely to have more Drush commands that might want to play with the DI container without bootstrapping the database. To that end, I am wondering if we should provide this capability as a standard service in the Drush bootstrapping classes.

Issue #2328, opened 2016-08-29
drush sql-cli returns improper exit code


iBobik:
If database is running exit code is correct:
```
$ drush sqlc <<< ""
$ echo $?
0
```
But when database is stopped exit code should not be 0:
```
$ drush sqlc <<< ""
ERROR 2003 (HY000): Can't connect to MySQL server on 'db' (113 "No route to host")
$ echo $?
0
```

Drush Version   :  8.1.2

iBobik:
Also Drush Version   :  8.1.3

weitzman:
Fixed for drush9. Needs backport.

PR #2417, opened 2016-10-27
Fixed exit code for mysql cli invocations

MartijnBraam:
This makes drush exit 1 if the mysql cli has a non-zero exit code. (See issue #2328 )

Also it seems that the module uses return ... everywhere but the return value from the callback is always dropped.


weitzman:
Too much unrelated change in here. Please skip the indention noise.

weitzman:
Fixed yesterday.

PR #2423, opened 2016-10-31
Fix so drush make acts more like drush pm-updatecode for contrib projects

uberhacker:
When Drupal contrib projects are updated, the current project directory should be cleared out and replaced by the new code; not leaving residue behind.


weitzman:
Make is in maint-only nowawadys

Issue #2426, opened 2016-11-02
dt() pollutes translation system with wrong strings


webflo:
``dt`` replaces the old ! placeholder with the related argument, this creates a new source string for every argument in the Drupal site.

Examples ``drush_log(dt('Downloading release history from !url', array('!url' => $url)));``

Create a new string for every $url. 

```
Downloading release history from https://updates.drupal.org/release-history/menu_admin_per_menu/8.x
Downloading release history from https://updates.drupal.org/release-history/address/8.x
```

**Steps to reproduce**

Run drush up on a Drupal site. Look at locales_source for new source strings, order lid by desc

webflo:
I have useless stuff like ```Backend invoke: env COLUMNS=227 /Users/fweber/.composer/vendor/webflo/drush-shim/bin/drush  --backend=2 --verbose --debug --local --root=/Users/fweber/Sites/project.dev/web --uri=http://default  pm-updatestatus 2>&1```

In my database :/

webflo:
https://github.com/drush-ops/drush/pull/1647/files

webflo:
My cleanup script

```
<?php

use Drupal\locale\StringStorageInterface;

/** @var StringStorageInterface $localeStorage */
$localeStorage = \Drupal::service('locale.storage');

foreach ($localeStorage->getTranslations(['translated' => FALSE], []) as $string) {
  $source = $localeStorage->findString(['lid' => $string->getId()]);
  $source->delete();
}
```

webflo:
Thx

PR #2427, opened 2016-11-02
Convert legacy translation placeholders to valid D8 placeholders

webflo:
PR for #2426


includes/output.inc
```diff
@@ -1,5 +1,6 @@
 <?php
 
+use Drupal\Core\Render\Markup;
```
webflo: Does it break 7? I think i should move it to replace_legacy_dt_args itself.

PR #2433, opened 2016-11-06
Add global and site (Drupal root) locations for Drush extensions.

greg-1-anderson:
This allows Drush extensions with commands and hooks to be declared in global (e.g. ~/.drush) and site (e.g. __DRUPAL_ROOT__/drush) locations.

For each potential search location, the actual commandfiles may be located at `.`, `./Commands`, or `./CommandFiles`.  The namespace calculation is not necessarily correct, though.  For example, a commandfile named `ExampleCommand.php` located in ~/.drush should have a namespace of `Drush`, whereas a commandfile with the same name located in ~/.drush/Commands should have a namespace of `Drush/Commands`. This is an ambiguity, because if the file is distributed as a standalone .php file, its namespace should not have to be modified based on where the end user chooses to install it.

This is not necessarily rational.  Potential improvements:

- Only allow commandfiles to be installed via Composer.
- Restrict global commandfiles to a single directory in each search location (e.g. ~/.drush/Commandfiles) so there is no ambiguity about namespaces. (No deep searching ever.)
- Assume the namespace of every global / site commandfile is `Drush`, regardless of location.
- Prohibit global commandfiles from using namespaces at all.
- Read in the first few lines of each global commandfile and search for the `namespace` line. Allow commandfiles to be namespaced to anything.

### DIFF
--- a/includes/preflight.inc
+++ b/includes/preflight.inc
@@ -773,6 +773,15 @@ function _drush_find_commandfiles_drush() {
   $annotation_commandfiles = $discovery->discover(DRUSH_BASE_PATH . '/lib/Drush', '\Drush');
 
   // And, finally, search for commandfiles in the $searchpath
+  $searchpath = array_map(
+    function ($item) {
+      if (strtolower(basename($item)) == 'commands') {
+        return dirname($item);
+      }
+      return $item;
+    },
+    $searchpath
+  );
   $global_drush_extensions = $discovery->discover($searchpath, '\Drush');
   $annotation_commandfiles += $global_drush_extensions;
 
--- a/includes/preflight.inc
+++ b/includes/preflight.inc
@@ -739,6 +739,9 @@ function _drush_preflight_uri() {
 }
 
 function _drush_find_commandfiles_drush() {
+  // Core commands shipping with Drush
+  $searchpath[] = dirname(__FILE__) . '/../commands/';
+
   // User commands, specified by 'include' option
   $include = drush_get_context('DRUSH_INCLUDE', array());
   foreach ($include as $path) {
@@ -762,14 +765,8 @@ function _drush_find_commandfiles_drush() {
     }
   }
 
-  // Core commands shipping with Drush
-  $drushCoreSearchpath[] = dirname(__DIR__) . '/commands/';
-
-  // Make sure search path is unique
-  $searchpath = array_unique($searchpath);
-
   // @todo the zero parameter is a bit weird here. It's $phase.
-  _drush_add_commandfiles($drushCoreSearchpath + $searchpath, 0);
+  _drush_add_commandfiles($searchpath, 0);
 
   // Also discover Drush's own annotation commands.
   $discovery = annotationcommand_adapter_get_discovery();
--- a/includes/preflight.inc
+++ b/includes/preflight.inc
@@ -561,6 +561,18 @@ function drush_preflight_root() {
   // Load the config options from Drupal's /drush, ../drush, and sites/all/drush directories,
   // even prior to bootstrapping the root.
   drush_load_config('drupal');
+
+  // Search for commandfiles in the root locations
+  $discovery = annotationcommand_adapter_get_discovery();
+  $searchpath = [dirname($root) . '/drush', "$root/drush", "$root/sites/all/drush"];
+  $drush_root_extensions = $discovery->discover($searchpath, '\Drush');
+  drush_set_context(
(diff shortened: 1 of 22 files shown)


greg-1-anderson:
There are two use-cases we need to be concerned with:

1. Policy files and the like -- it should be possible to take an example file and drop it somewhere for customization, with a minimum amount of work (an minimum chance of having the file just not recognized)

2. Extensions that do something -- config extras, for example.

In the later case, it seems like relying on the Composer autoloader would be a good idea. If we are moving to a model that favors site-local Drush, then perhaps we could stipulate that extensions that want to rely on the Composer autoloader should just be Drupal modules. I'm not sure that is entirely satisfying, though.

includes/preflight.inc
```diff
@@ -561,6 +561,18 @@ function drush_preflight_root() {
   // Load the config options from Drupal's /drush, ../drush, and sites/all/drush directories,
   // even prior to bootstrapping the root.
   drush_load_config('drupal');
+
+  // Search for commandfiles in the root locations
+  $discovery = annotationcommand_adapter_get_discovery();
+  $searchpath = [dirname($root) . '/drush', "$root/drush", "$root/sites/all/drush"];
```
weitzman: webmozart's Path::getDirectory() is apparently more cross platform than dirname().

weitzman:
For each potential search location, the actual commandfiles may be located at ., ./Commands, or ./CommandFiles

For the sake of clarity to implementers, I think we should just decide on one or the other. I think Commands makes more sense. We can change our own lib/Drush/CommandFiles dir accordingly.

My quick thought is that folks who want to use Commandfiles located outside the composer project should bring them in via --include or similar. In other words, we change Drush to always act like --local and expect folks to use --include, --alias-path, --config when needed. Hopefully this is helpful.

includes/preflight.inc
greg-1-anderson: There are still > 100 calls to dirname in Drush. Perhaps we should look into switching those over in a separate PR.

greg-1-anderson:
I think that we should backport everything we do in this issue to 8.x, then move on to new things such as changing the default for --local (or removing non-local locations completely).

Next question up is whether we should rename `console.command` to `drush.command` as [previously mused](https://github.com/hechoendrupal/DrupalConsole/issues/2594#issuecomment-248957675). Downside is that this has been in a stable Drush release for six months or so. On the other hand, we never promoted this, and I don't think it is widely used; we could call it a bugfix and make folks change. On the flip side, we could also just call it a DrupalConsole bug that has been fixed, and leave it.

weitzman:
I think its unrealistic to think that folks will be using annotated commands with Drush 8.x. But if you want to backport this I'll go along.

I think we can switch to drush.command as hardly anyone has started using console.command.

OK on doing all the dirname conversions in one future issue.

greg-1-anderson:
I want to allow annotated command policy files to work with 8.x so that folks can make policy files that work across two versions of Drush.  I don't think we should backport any of the core annotated commands to 8.x (and that wouldn't really be possible anyway).

I'll rename console.command to drush.command.

greg-1-anderson:
The `--include` option says: `A list of additional directory paths to search for drush commands.`

As currently implemented in this PR, by default, Drush is looking for "global" annotated commands in `~/.drush/Commands`. No deep searching is done. Annotated commands stored directly in `~/.drush` are ignored.  If we removed this location by default, then users could restore the current behavior via `--include=~/.drush`, or `$options['include'] = "$home/.drush";`. In either case, the namespace folks must use for these files would be `Drush\Commands`.  See the most recent commit (8fc9aa3) for tests.

This gets odd if folks use `--include=~/.drush/Commands`, which is sort of logical, since that's where the commands are located. Doing it as currently implemented is better for consistency, but if we remove the default behavior, then it gets a little questionable what we are remaining consistent with.  On the flip side, if we allow annotated commands directly in the --include directory, what would the namespace be?  Perhaps the most consistent / least confusing thing would be to require the namespace to be Drush\Commands, require that the commands be in a `Commands` folder, and silently strip a trailing `Commands` directory from `--include` if it is provided.

weitzman:
That "most consistent" proposal sounds good to me.

greg-1-anderson:
I think this is good to go. See #2436 for 8.x version.

weitzman:
Both this and @2436 looks good to me. Feel free to merge. Thanks @greg-1-anderson.

PR #2437, opened 2016-11-08
Annotated command implementation of 'core-status' command.

greg-1-anderson:
The handling of default fields is not working right.  If --fields is not specified, an error is returned. If a valid list of --fields is given, they will be printed.

### DIFF
--- a/commands/core/core.drush.inc
+++ b/commands/core/core.drush.inc
@@ -140,42 +140,6 @@ function core_drush_command() {
     ),
     'aliases' => array('conf', 'config'),
   );
-  $items['core-status'] = array(
-    'description' => 'Provides a birds-eye view of the current Drupal installation, if any.',
-    'bootstrap' => DRUSH_BOOTSTRAP_MAX,
-    'aliases' => array('status', 'st'),
-    'examples' => array(
-      'drush core-status version' => 'Show all status lines that contain version information.',
-      'drush core-status --pipe' => 'A list key=value items separated by line breaks.',
-      'drush core-status drush-version --pipe' => 'Emit just the drush version with no label.',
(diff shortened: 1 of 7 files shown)


greg-1-anderson:
Fixed one bug in the upstream, but one more remains. All of the possible fields are displayed, even if some fields contain no data. Fields with no data should be omitted from the output.

weitzman:
@greg-1-anderson - Does your recent output-formatters work fix #130 for annotated commands?

greg-1-anderson:
@weitzman Yes, with the addition of the --include-field-labels option to the status command. This could be a global option, perhaps.

Having a hard time figuring out why the tests are failing. Sometimes the site-under-test ends up at `drush-sandbox/web/web` instead of `drush-sandbox/web`. It seems that maybe archive-dump or archive-restore or something else in the tests is relying on the behavior of `drush core-status` that I am not completely emulating. This is not a great design. Might want to fix it by changing the way the tests work.

greg-1-anderson:
Okay, got the tests using only "standard" forms, no shortcuts, when using `core-status`.

Not implemented yet:

- [ ] --format=tsv blows up (does not handle non-string cell data)
- [ ] --format=csv not working perfectly (again, array handling a bit off)
- [ ] Using arguments in place of --fields=... not implemented (drop support for this?)
- [ ] --pipe selects tsv; should select json to maintain backwards compatibility

weitzman:
I think we can drop support for the argument and drop support for --pipe
option. Folks can be explicit.

weitzman:
@greg-1-anderson  - I'm OK if we merge this without all that checklist complete.

greg-1-anderson:
Latest commit should take care of the primary backwards-compatibility issues.  Let's see how the tests fare.

PR #2439, opened 2016-11-08
Fix error message when unable to copy example.sites.php to sites.php

ssirois:
The error message now helps to understand which file couldn't be copied.


ssirois:
I do see the error here : https://travis-ci.org/drush-ops/drush/jobs/174290298#L354

Could it be that there was a "web hiccough" that prevented the proper download of the file available at https://github.com/downloads/gilbitron/Nivo-Slider/nivo-slider2.7.1.zip for the test?

I have difficulty reproducing Travis' error locally by running `unish.sh --group=make`. Any advice?

PR #2441, opened 2016-11-10
Use Robo 1.0.0 stable.

greg-1-anderson:


greg-1-anderson:
Tests failed because I introduced a bug in output formatters when there is no data to output in a table. Will fix that and update composer.lock here.

greg-1-anderson:
https://github.com/consolidation/output-formatters/pull/40

greg-1-anderson:
Ignore the above comments -- they're for a different project. 😝   Everything is fine here.

PR #2448, opened 2016-11-16
Convert cli command to Annotated command class

damiankloip:


greg-1-anderson:
LGTM +1 way to go. Looks like there was one spurious test failure. Restarted the job and all is green now.

lib/Drush/Commands/core/CliCommands.php
```diff
@@ -0,0 +1,254 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drush\Commands\core\CliCommands.
```
weitzman: IMO we don't need @file declarations. Even Drupal got rid of this standard, no?

commands/core/cli.drush.inc
```diff
@@ -1,25 +1,9 @@
...
-use Drush\Psysh\DrushCommand;
-use Drush\Psysh\Shell;
-
 /**
  * Implements hook_drush_command().
  */
 function cli_drush_command() {
-  $items['core-cli'] = array(
-    'description' => 'Open an interactive shell on a Drupal site.',
-    'remote-tty' => TRUE,
```
weitzman: We are losing this remote-tty declaration. I'm not clear what it does. I do know that we should continue to support `drush @remote php`. Please test that.

commands/core/cli.drush.inc
```diff
@@ -1,25 +1,9 @@
...
-    'aliases' => array('php'),
-    'bootstrap' => DRUSH_BOOTSTRAP_MAX,
-    'topics' => array('docs-repl'),
```
weitzman: We are losing this too. Any thoughts on how topics should be handled going forward? I like the current approach but am happy to change. Anyway, not a blocker for this PR.

greg-1-anderson: We should just add a @topic annotation, and implement that later (soon?).

commands/core/cli.drush.inc
greg-1-anderson: 'remote-tty' causes `drush_do_command_redispatch()` to add `#tty` to the backend options, which causes backend invoke to add -t to the ssh session.  I think that is critical to the remote operation of the shell.

damiankloip:
Yes. The intention was to add these, this is kind of a half finished PR
still :)

weitzman:
@damiankloip See how we handle @hidden annotation in annotationcommand_adapter_get_commands_for_commandhandler(). Add similar support there for @remote-tty and we should be good.

lib/Drush/Commands/core/CliCommands.php
damiankloip: Yeah, this is just automated from my IDE (from the days we needed that). I'll remove!

weitzman:
Lets add `@topics docs-repl` even though it does not work yet. Then I think this is ready.

damiankloip:
I did already add that but I called it '@topic' I'll change to the plural version.

damiankloip:
@weitzman @greg-1-anderson let's do it in a follow up either way, but how do we handle the `docs-repl` command and the callbacks that uses like `drush_print_file` with the repl.md file?

greg-1-anderson:
Long term we want to convert everything to classes, but in the short term, we are just calling existing procedural functions as necessary.

My general desire is to use the [Robo](https://robo.li) tasks to replace much of what is in the `includes` directory.

damiankloip:
@greg-1-anderson ok, thanks. Maybe I'll just port that command in this PR too then..

greg-1-anderson:
@damiankloip Use `@topic`, as in:

```
/**
 * @topic docs-mytopic
 *    This is the description of my topic.
 */
```
There can be more than one, each with a description.

If you port the topic command itself, do it in a separate PR. The existing topic command should work if you just update the annotation adapter.  Also, you'll have to update the annotation adapter if you want the topic to show up in the help output.

damiankloip:
@greg-1-anderson Sorry, I was not thinking of the actual topic command, just the `docs-repl` command I think that was added by @weitzman originally. From this drush command info:

```
$items['docs-repl'] = array(
  'description' => 'repl.md',
  'hidden' => TRUE,
  'topic' => TRUE,
  'bootstrap' => DRUSH_BOOTSTRAP_NONE,
  'callback' => 'drush_print_file',
  'callback arguments' => array(drush_get_context('DOC_PREFIX', DRUSH_BASE_PATH) . '/docs/repl.md'),
);
```

greg-1-anderson:
Oh, okay; that could go in the same commandfile as the cli command itself.

damiankloip:
It's already here. Please review :)

weitzman:
Lgtm. 

What do folks think if getting rid of doc_prefix? That was for Debian or something like that. It's antiquated in a composer world. I'd rather not copy that all over the place.

greg-1-anderson:
Yeah, the DOC_PREFIX thing was so that the Drush docs could be places in /usr/share/docs, or something like that, when installed from a package manager (apt, yum, et. al.).  For `man` files (if we ever produced one), I think that would be fine, but I do not really like the idea of splitting apart the Drush distribution.  That's also contrary to our general strategy of moving away from a global Drush in favor of site-local installs.  I'd say get rid of it.

damiankloip:
Thanks Moshe. I say we ship this now!

PR #2451, opened 2016-11-18
fix [warning] apcu_clear_cache() expects exactly 0 parameters, 1 given

ms-slasher13:
After #2450 with PHP 7
The following error is caused by execution of `drush cr`
[warning] apcu_clear_cache() expects exactly 0 parameters, 1 given cache.drush.inc:280

### DIFF
--- a/commands/core/cache.drush.inc
+++ b/commands/core/cache.drush.inc
@@ -277,7 +277,7 @@ function drush_cache_rebuild() {
     'wincache_ucache_clear',
     'xcache_clear_cache',
   ];
-  array_walk(array_filter($user_caches, 'is_callable'), 'call_user_func');
+  array_map('call_user_func', array_filter($user_caches, 'is_callable'));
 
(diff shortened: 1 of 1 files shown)


weitzman:
Thought this was copy and paste from rebuild.php. Is the bug present there
too? Anyway array_map is fine with me. Will merge this weekend though
@damian is welcome to do so before me. Needs back port to 8.x as well.

damiankloip:
@weitzman Yeah, it looks like core has the same problem too...

damiankloip:
@weitzman cherry-picked into `8.x` too.

ms-slasher13:
ok, I will fix in core, too!

damiankloip:
@ms-slasher13 I already created one https://www.drupal.org/node/2829578 feel free to jump in on there too!

ms-slasher13:
Reviewed & tested by the community

PR #2454, opened 2016-11-21
Maintain compatible output for 'drush status'

greg-1-anderson:


greg-1-anderson:
This PR makes the format of `drush status` remain consistent with what is was in Drush 8 and earlier. (n.b. includes change in the upstream to support this.) I think this is important because I suspect there are a lot of scripts that parse this output to get info about their Drupal site. I'd like to avoid breaking these until folks have a chance to switch to using `--fields` instead.

If you prefer our existing format to the new one, we can also keep it forever.

There's a choice in implementation. We can annotate every command that returns a PropertyList as shown in this PR. We could also switch to using a callback to switch the format of every command that returns a list to use this format.

weitzman:
I have no preference on how this is implemented. PR looks good.

PR #2457, opened 2016-11-22
Adding field sanitation for Drupal 8.

grasmash:


greg-1-anderson:
Yes, classes are better -- this is the direction we are going.  If you'd like to make the existing code more class-based as well, you could [follow the example of commands already ported over](https://github.com/drush-ops/drush/tree/master/lib/Drush/Commands/core).

grasmash:
@greg-1-anderson Ok, I did some minor refactoring of the command. However, the logic for email and password refactoring are very procedural and a little hairy, so I steered clear.

grasmash:
@greg-1-anderson The test results are looking a little cryptic to me. Any guidance on what's causing the failure?

greg-1-anderson:
The `make` failure looks transient; I re-started the test to see if it might work itself out.

Your other failure is caused by ` [error]  Sanitize query failed. ` for Drupal 8. This test is passing for Drupal 7. Didn't investigate further to see why the query was failing.

grasmash:
Ok, tests seem good now. However, the tests don't exactly cover the new functionality. To do that, I'd need to create new fields on the user entity and also enable the comment module and create comments. Are there any existing tests doing things like that in Drush?

weitzman:
A little torn on this. Seems like this overlaps lot with devel_generate. Its identical except for update versus create. devel_generate works with any storage backend, not just sql. seems more future proof.

greg-1-anderson:
I think that testing off of a real (sanitized) copy of the production database is an important use-case. I like to use generated / fixture content for functional tests, but to for pre-production testing (e.g. sign off prior to go-live), it's nice to see something that looks like a copy of the production site.

grasmash:
I'd argue that this serves a different use case than devel generate. `sql-sanitize` serves to take real production data and make it safe for non-production machines. I view it more as a security measure. Devel generate will create new dummy data and will not sanitize anything that's already in the database. 

Anyway, I'm happy with either merging into drush or leaving as a standalone module. @weitzman Please let me know your decision one way or another.

commands/sql/sql.drush.inc
```diff
@@ -691,11 +696,7 @@ function sql_drush_sql_sync_sanitize($site) {
...
   }
 
-  // Seems quite portable (SQLite?) - http://en.wikipedia.org/wiki/Truncate_(SQL)
-  $table = 'sessions';
-  if ($wrap_table_name) {
-    $table = "{{$table}}";
-  }
-  $sql_sessions = "TRUNCATE TABLE {$table};";
-  drush_sql_register_post_sync_op('sessions', dt('Truncate Drupal\'s sessions table'), $sql_sessions);
+  require __DIR__ . '/Sanitize.php';
```
weitzman: Shouldn't we autoload this? We could add it to composer.json

commands/sql/Sanitize.php
```diff
@@ -0,0 +1,233 @@
...
+   *
+   * @param string $query
+   *   The SQL query to execute. Must end in a semicolon!
+   *
+   * @return string
+   *   The output of the query.
+   */
+  protected function query($query) {
+    $current = drush_get_context('DRUSH_SIMULATE');
+    drush_set_context('DRUSH_SIMULATE', FALSE);
```
weitzman: Why do we disable simulate?

commands/sql/Sanitize.php
```diff
@@ -0,0 +1,233 @@
...
+   *   The SQL query to execute. Must end in a semicolon!
+   *
+   * @return bool
+   *   TRUE if command executed successfully.
+   *
+   * @throws \Drush\Sql\SqlException
+   */
+  protected function executeQuery($query) {
+    // Enable prefix processing when db-prefix option is used.
+    if (drush_get_option('db-prefix')) {
```
weitzman: I'm a bit surprised that this is needed.

uberhacker:
Should the title be `sanitization` instead of `sanitation`?

commands/sql/sql.drush.inc
greg-1-anderson: Put it in lib/Drush, and it will not need an extra entry in composer.json

commands/sql/sql.drush.inc
```diff
@@ -691,11 +696,7 @@ function sql_drush_sql_sync_sanitize($site) {
...
+  $sanitizer = new \Drush\Commands\core\SanitizeCommands($site, $major_version);
```
greg-1-anderson: This omits dependency injection, so it's probably not a good pattern to encourage. Perhaps we need a new flavor of drush_invoke to do dispatches like this.

commands/sql/Sanitize.php
grasmash: I was just copying the approach taken in methods like db_exists() and listTables(). Is this not necessary?

commands/sql/Sanitize.php
grasmash: This similarly came from an approach I saw elsewhere. Likely unneeded.

commands/sql/sql.drush.inc
```diff
@@ -159,13 +159,18 @@ function sql_drush_command() {
     'remote-tty' => TRUE,
   );
   $items['sql-sanitize'] = array(
-    'description' => "Run sanitization operations on the current database.",
+    'description' => "Run sanitation operations on the current database.",
```
weitzman: I need a 3rd opinion on this. Sanitization sounds more proper to me.

grasmash:
I'm seeing `[warning] Missing argument 1 for Drush\Commands\core\SanitizeCommands::__construct() SanitizeCommands.php:39` which leads me to believe that Drush is doing some type of automated command registration when it sees SanitizeCommands.php in the Command namespace.

weitzman:
Yes, exactly. If you look at the master branch, you will see many Drush commands ported to AnnotatedCommand. See links at https://medium.com/@weitzman/drush-9-roadmap-9eb5cd227c50. It would be super if you rewrote this as an annotated command. That way, it could go into master and 8.x. We aren't really accepting any more old style commands. I'm happy to help if you get stuck..

grasmash:
Alright, taking a stab at conversion to an AnnotatedCommand.

At least one wrinkle... this command invokes `drush_sql_sync_sanitize()`, which uses `drush_get_option()`. 

Will `drush_get_option()` still work when the options have been passed to an annotated command? If not, I will need to pass `$options` to `drush_command_invoke_all()` as a parameter, which will require that implementations of `hook_drush_sql_sync_sanitize()` change their function signature to include the new parameter.

grasmash:
Wow. I guess that worked :)

grasmash:
@weitzman @greg-1-anderson let me know if I need to perform any changes.

lib/Drush/Commands/core/SanitizeCommands.php
```diff
@@ -0,0 +1,274 @@
...
+   * @option whitelist-fields Exempt these fields from sanitizing.
+   * @aliases sqlsan
+   * @usage drush sql-sanitize --sanitize-password=no
+   *   Sanitize database without modifying any passwords.
+   * @usage drush sql-sanitize --whitelist-fields=field_biography,field_phone_number
+   *   Sanitizes database but exempts two user fields from modification.
+   * @see hook_drush_sql_sync_sanitize() for adding custom sanitize routines.
+   */
+  public function sqlSanitize($options = [
+    'db-prefix' => '',
```
weitzman: Is --db-prefix a string or a boolean. Lets clarify via help text and/or default value in params.

lib/Drush/Commands/core/SanitizeCommands.php
```diff
@@ -0,0 +1,274 @@
...
+    if (!drush_confirm(dt('Do you really want to sanitize the current database?'))) {
+      return drush_user_abort();
+    }
+
+    if ($sanitize_query) {
+      $sql = drush_sql_get_class();
+      $sanitize_query = $sql->query_prefix($sanitize_query);
+      $result = $sql->query($sanitize_query);
+      if (!$result) {
+        return drush_set_error('DRUSH_SQL_NO_QUERY', dt('Sanitize query failed.'));
```
weitzman: Should now be `throw new \Exception(dt());`

lib/Drush/Commands/core/SanitizeCommands.php
```diff
@@ -0,0 +1,274 @@
...
+   * @param string $query
+   *   The SQL query to execute. Must end in a semicolon!
+   *
+   * @return string
+   *   The output of the query.
+   */
+  protected function query($query) {
+    $current = drush_get_context('DRUSH_SIMULATE');
+    drush_set_context('DRUSH_SIMULATE', FALSE);
+    $success = $this->executeQuery($query);
```
weitzman: Do we not need to handle query failure here? Also, can we merge query() and doQuery() especially since SQL API already has a query(). We now have 2 similar methods. Just asking.

tests/sqlSyncTest.php
```diff
@@ -44,12 +44,13 @@ public function localSqlSync() {
       'yes' => NULL,
     );
     $this->drush('user-create', array($name), $options + array('password' => 'password', 'mail' => $mail));
+    $this->drush('pm-enable', array('comment'), $options + array('yes' => NULL));
```
weitzman: Can we test some field sanitization? If it helps, you can install the standard testing profile so you get 2 content types automatically (article and page).

grasmash: @weitzman The field sanitation is performed only on user fields. So, adding content types will not test the field sanitation. Is there a precedent for adding fields to the user entity in drush testing?

lib/Drush/Commands/core/SanitizeCommands.php
grasmash: It would be great if we could specify the argument type in the annotation.

tests/sqlSyncTest.php
weitzman: Ah, thanks for clarifying that we are only talking about fields on user entity. There is no code to copy in this repo for adding Fields to users. However, I found [this code in Drupal that you can copy](https://api.drupal.org/api/drupal/core%21modules%21field%21tests%21src%21Kernel%21Boolean%21BooleanItemTest.php/function/BooleanItemTest%3A%3AsetUp/8.2.x)

lib/Drush/Commands/core/SanitizeCommands.php
grasmash: I don't think we need to do anything with a failure. It will fail and the failure message should be printed to screen.

grasmash:
Hm. Travis is reporting:

```
Fatal error: Class 'Drupal\field\Entity\FieldStorageConfig' not found in /home/travis/build/drush-ops/drush/tests/resources/user_fields-D8.php on line 31
```

I was under the impression that Drupal was bootstrapped during a php-eval call and that FieldStorageConfig would be available.

weitzman:
Think you mean php-script? That  bootstraps as far as it can before running the script. So yes, that class should be available. Does the php-script call work locally for you? Also try locally to run `./unish --filter= testLocalSqlSync`. You can customize unish by copying tests/phpunit.xml.dist to phpunit.xml and edit accordingly. Or set some corresponding env variables.

grasmash:
I encounter the same issue locally. I'll likely need to dig into the `php-script` command to see if it's properly autoloading Drupal's core classes.

weitzman:
I fixed the failing test by enabling field module and a couple others. Unish installs the 'testing' install profile by default since thats much faster than standard.

Next problem was that `drush user-create` doesn't allow you to specify values for custom fields. It would be cool to add that one day. It might look like the name/value pair that site-install takes. Anyway, I moved user creation to inside user_fields-D8.php and now the test is passing.

I pushed these commits to your branch. Hopefully Travis will be happy now.

grasmash:
Ah, well done! Question for you: will this test only execute against Drupal 8?

I see that [this line](https://github.com/drush-ops/drush/pull/2457/commits/116ebf0f4620635073457a64325e1bff094de642#diff-f8a5a30fb8daedb84b97fb3f7dba0756R51) uses `UNISH_DRUPAL_MAJOR_VERSION`. If this is not 8, then the call to `php-script` will fail.

weitzman:
I've [declared that Drush9 will only work with Drupal 8](https://medium.com/@weitzman/drush-9-roadmap-9eb5cd227c50) so those conditionals are no longer needed.

grasmash:
Ah, so that means that this PR will only be part of Drush 9 and not merged into Drush 8? Admittedly I'm a bit murky on the relationship between the master, 8.x, 8-to-master, and the tags for 8 and 9. 

It seems like Drush 8 is actually ahead of Drush 9, at least in terms of releases. That's lead me to recently downgrade from Drush 9 to Drush 8.

weitzman:
Sorry, its a bit confusing. Drush 8 is more stable than Drush 9. Functionality-wise, they should be identical. The main difference is that Drush9 is where we are porting core Drush commands to Annotated. If you want to backport this PR to Drush 8, you can put the conditional back into the test.  Thats so minor that I dont mind it going into 9. Hopefully this PR will backport without much trouble. 

Have you addressed most of my prior feedback? When thats done, this PR is ready IMO.

grasmash:
Not ready yet. While the db tables for the fields are being created, the data isn't being populated: `Field field_user_email is unknown`. Not sure why yet. We are using that same code that is used in other tests.

tests/resources/user_fields-D8.php
```diff
@@ -30,7 +27,17 @@
...
-$user = User::create($values);
+
+$user = User::create([
+  'name' => $args[2],
+  'mail' => $args[3],
+  'pass' => 'password',
+]);
+
+foreach ($values as $field_name => $value) {
+  $user->set($field_name, $value);
```
grasmash: Using this method creates a more obvious failure when the values are not set, which is nice.

grasmash: Yeah I arrived at the same impasse.

weitzman:
Thanks to @Berdir for looking at our php-script and offering this feedback. Hopefully soneone can push this farther now.

```
10:33 AM B<berdir> moshe: I would recommend to do $field_storage = FieldStorageConfig::create(); $field_storage->save(), then you can pass field_storage to FieldConfig::create() and you only need field_storage, bundle, label and settings, see node_add_body_field()
10:33 AM M<moshe> thanks @berdir. will try that
10:33 AM B<berdir> moshe: but yes, that widget code there is bogus, that's now how it works. you might be missing some required storage/config settings and you need to do what node_add_body_field() does for the widget settings
10:34 AM M<moshe> i'm not sure where we got that widget code from
10:34 AM B<berdir> 7.x
10:34 AM M<moshe> don't think we care about widget for this PR
10:34 AM B<berdir> 7.x had widget and formatter settings in the field instance, now they are separate config entities
```

grasmash:
I think that I already tried that approach locally, but let's see what happens.

grasmash:
I think I'm onto something. In the db, the bundle is assigned as " user" with a leading space. No idea why, but it would explain the fields not being associated correctly.

weitzman:
Nice work @grasmash. have you addressed all the code feedback from a while ago? Once you have, this is ready IMO.

grasmash:
@weitzman Yes I believe that I addressed all other feedback.

weitzman:
Merged to 8.x branch.

PR #2458, opened 2016-11-23
Port user commands to Annotated

### LINKED ISSUE #2306: Required password should be argument in user-password command
```
colan@snake[Fri 5 13:31]% drush upwd colan                                                                                                                  
Missing required option: --password.  [...]
```

By definition, options aren't required.  Is there a good reason why the password here is an option, and not an argument?  If not, we should switch this around, making it so.

weitzman:
Pretty straightforward port. A couple minor todos in the code which can be followups. Some notes:

1. Many user commands had an unusual argument where users to operate on could be specified by name, uid, or email. If someone wanted to avoid a little ambiguity (username=2 versus uid=2), they could use a named option like --uid=2. I did away with this convenience in for the sake of clarity. Now these commands take a list of usernames and nothing else. I hope this doesn't bother tooo many people. The help and the code are clearer this way.
1. When trying to preserve the complex option handling above, I tried to use the `@hook option @user_selection` but apparently [that way of adding options is not recognized by annotation_adapter](https://github.com/drush-ops/drush/blob/f6e36a727b77221d7e7f260834faa87aac23d9fa/includes/annotationcommand_adapter.inc#L472). Would be nice to add it will be needed by pm and sql commands.
1. Fixed #2306 - user-password now takes the password as an argument. 
1. The master branch will drop D7 testing with this PR. Someone had to do it :). This opens the way for PRs which remove D7 handling.
1. A follow-up PR could dismantle the UserList and UserVersion classes as those are not needed once we are D8 only.


weitzman:
Fixed tests. Note that the last commit had to add --strict=0 option when using the formatter option , --include-field-labels. @greg-1-anderson - perhaps we not declaring formatter options in the adapter?

greg-1-anderson:
Maybe we need a handy command to easily convert from email / uid to usernames, so that commands that need to specify users by some other means can convert first.

weitzman:
1. Good point. Added --uid and --mail options back to user-information in case folks need to convert to username.
1. Went all the way and deleted the UserList and UserSingle classes as we are D8 only now.

IMO this is ready.

greg-1-anderson:
Did not test, but code LGTM

derimagia:
Want to make sure it's clear, so this is paving the way to drop support for Drupal 7 in Drush 9? Nothing against it just want to check.

Either way, does that mean http://docs.drush.org/en/master/install/ (Drupal Compatibility) needs to be updated?

weitzman:
Yes, thats the plan. See https://medium.com/@weitzman/drush-9-roadmap-9eb5cd227c50.

I just updated that table - c2471d7. Thanks.

PR #2464, opened 2016-11-25
Port site-ssh command to Annotated.

weitzman:
The only real change here is that the optional bash at the end must be quoted to form one argument. I didn't bother trying to figure out how to preserve that. I guess I could remove the strict option handling from this command if this feature goes away. Thoughts?


weitzman:
I just removed a test which which checks for the multiple argument bash variant which i propose not to support. yet the tests suggests that this multiple argument variant is preferred. ping @greg-1-anderson for an opinion on whether to continue supporting it.

weitzman:
Just pushed a change that brings back the multiple argument variant. Tests are green so merged.

PR #2466, opened 2016-11-27
Port site-install to Annotated

weitzman:
Ping @greg-1-anderson. The install() method takes a variable number of arguments due to optional key+value pairs being passed to installer. You can see my struggles in the PR for how to handle this. Tips?

I neutered some options on quick-drupal to get this to run. I guess quick-drupal should become a robo collection. Seems low priority.


lib/Drush/Commands/core/SiteInstallCommands.php
```diff
@@ -0,0 +1,283 @@
...
+   *   Re-install with specified uid1 credentials.
+   * @usage drush site-install standard install_configure_form.site_default_country=FR my_profile_form.my_settings.key=value
+   *   Pass additional arguments to the profile (D7 example shown here.
+   * @usage drush site-install standard install_configure_form.update_status_module='array(FALSE,FALSE)'
+   *   Disable email notification during install and later. If your server has no smtp, this gets rid of an error during install.
+   * @bootstrap DRUSH_BOOTSTRAP_DRUPAL_ROOT
+   * @aliases si
+   *
+   */
+  public function install($profile, array $additional = [], $options = ['db-url' => NULL, 'db-prefix' => NULL, 'db-su' => NULL, 'db-su-pw' => NULL, 'account-name' => 'admin', 'account-mail' => 'admin@example.com', 'account-pass' => NULL, 'locale' => 'en', 'site-name' => 'Drush Site-Install', 'site-pass' => NULL, 'sites-subdir' => NULL, 'config-dir' => NULL]) {
```
greg-1-anderson: This declaration for variable arguments (the last declared `array`) works in consolidation/annotated-command and Robo. If you are having trouble with it here, the problem is in the annotation adapter. I'll look at that when I get a chance.

greg-1-anderson:
So, in Robo / annotated-commands, handling of variable arguments via `array` is currently done by Symfony. Every Command object has an InputDefinition, which defines what the command parameters are. When a command is executed, then the `bind()` method is called on the `$input`, which assigns the input definition to the input object and parses the options again. Iff the input isa ArgvInput, and iff the last parameter is declared to be an array parameter, then the ArgvInput class will put the last N arguments in the last command parameter. In contrast, if the an ArrayInput object is being used, then it is the responsibility of the caller to ensure that it matches the expectation of the command that is being called (e.g. the last parameter should be an array with all of the variable arguments in it).

Currently, in Drush, we are using a DrushInputAdapter, which implements InputInterface, but is not an ArgvInput. Our Command objects **do** have input definitions, but, since we are not dispatching via the Symfony Application, `bind()` is never called.

We need to:

- Rewrite DrushInputAdapter to extend ArgvInput.
- Ensure that we call `bind` on the DrushInputAdapter with the input definition from our command object.

weitzman:
ping @greg-1-anderson. any chance you can help with DrushInputAdapter?

greg-1-anderson:
Yes. Been busy, but this is now at the top of my Drush to-do list. Will try to find some time for it soon.

greg-1-anderson:
Maybe merge #2485 first?

weitzman:
I would rather merge this and then port #2485 to new annotated hook system. He kindly offerred to do that.

weitzman:
Merged ... Opened #2552 for variable param handling. For now that feature is gone.

PR #2471, opened 2016-11-30
Use config.storage.sync to instantiate the FileStorage

webflo:
This should make [config_split](https://www.drupal.org/project/config_split) compatible cex/cim.

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -461,7 +461,12 @@ function _drush_config_export($destination, $destination_dir, $branch) {
   $storage_filters = drush_config_get_storage_filters();
   if (count(glob($destination_dir . '/*')) > 0) {
     // Retrieve a list of differences between the active and target configuration (if any).
-    $target_storage = new FileStorage($destination_dir);
+    if ($destination == CONFIG_SYNC_DIRECTORY) {
+      $target_storage = \Drupal::service('config.storage.sync');
+    }
+    else {
+      $target_storage = new FileStorage($destination_dir);
+    }
     /** @var \Drupal\Core\Config\StorageInterface $active_storage */
     $active_storage = \Drupal::service('config.storage');
     $comparison_source = $active_storage;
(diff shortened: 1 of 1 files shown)


weitzman:
LGTM. We had a spurious test fail that cleared up on re-test.

I can merge this once you think it is complete. I notice that the title says WIP.

webflo:
Lets get some feedback from config_split maintainers first.

weitzman:
Any feedback yet?

swentel:
Not a maintainer, but supporter and user. Patch makes very much sense, so +1!

weitzman:
FYI I added back storage filters hook in #2314 for better integration with config_split. Not sure if this PR is still desirable/needed.

bircher:
I am a maintainer of config_split and I love this PR. In fact this would be a far superior approach to storage filters.
I am moving to an approach that I swap out the `config.storage.sync` service and return the wrapped storage that unites the main directory and the split directories of the active splits. If drush would use that storage I could even think about retiring the custom commands for importing and exporting.
The storage filters - while inspired by the storage filter drush uses - would mean a much more complex interaction, I just recently had to change the interface a bit to fix a bug and that would require a change to drush etc. In addition, if drush and drupal console and drupal core all use the same service all work as expected.

The only thing that might need extra work would be the extra functionality drush cex and cim have such as automatically committing to git etc. But in my opinion that is not really blocking.

webflo:
Alright, i think we have agreement. LGTM :)

weitzman:
Merged into master and 8.x. I think we can get rid of Drush's filter and our filter hook and just expect folks to take over the sync service. What do folks think?

bkosborne:
Now that Drush 8.1.10 has been released and includes this code, what needs to be done on my end to get drush cex/cim aware of config split?

weitzman:
Config split folks are working on that in https://www.drupal.org/node/2855319 and similar issues. They would better know the answer to this question.

PR #2476, opened 2016-12-01
Update symfony to ~3.1

alexpott:
Core has updated Symfony - Drush needs to too.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,20 +29,20 @@
   ],
   "config": {
     "platform": {
-      "php": "5.4.5"
+      "php": "5.5.9"
     }
   },
   "require": {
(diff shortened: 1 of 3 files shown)


weitzman:
I dont see how this can go into Drush 8. Drush 8 supports D6 and D7 which happily run on PHP 5.4

Berdir:
Can we just add the optional argument without forcing the new version? That might be enough to have it work with 3.1?

alexpott:
I don't think that @Berdir's suggestion is going to work since if the drush autoloader gets in before the Drupal autoloader the wrong classes are going to be loaded :(

greg-1-anderson:
Continue discussion at #2477

PR #2477, opened 2016-12-01
Update symfony to 3.1

alexpott:
This PR is against the master branch as opposed to https://github.com/drush-ops/drush/pull/2476


weitzman:
Seems OK to me to commit this to master. Ping @greg-1-anderson.

alexpott:
The problem with applying this to master is that then master does not support Drupal 8.2.x anymore - only 8.3.x and up. Painful.

The only way I can see this working is if Drush has both the old symfony code and the new symfony code

And switches depending on Drupal version

alexpott:
Or it never autoloads these dependencies for Drupal 8 and always uses the ones provided by Drupal 8.

weitzman:
I guess we could mark our master branch as 8.3+ compatible. Not sure other solutions are viable.

greg-1-anderson:
@alexpott @weitzman we need a cli in core. That would be the best solution -- make every command that requires a bootstrap part of the Drupal that has matching dependencies.  The global tool could just be a finder and alias manager. This is the **convergent strategy** -- the cli just becomes part of Drupal.

If we want Drush to continue to be multi-target, then we could stop using Symfony completely, so we had no overlapping dependencies with Drupal at all. Not sure that would be entirely productive, but it would be possible and not terribly difficult. We could just fork to a new interface the minimal set that we need to work -- yaml parser, input interface and so on. It would then become a non-goal to ever use the Symfony Application to dispatch commands. Instead, we would over time wrap our adapter and legacy command dispatcher in classes. This is the **divergent strategy** -- Drush keeps acting like it has always acted. If we went this route, we'd want to reconsider dropping Drupal 7 support.

If we don't decide to do either of those options, I think we should wait until 8.3.0 is released before merging this to master.  We should probably maintain a Drush branch for 8.3.  We could tag an 8.3.0-alpha1 so folks using a site-local-Drush could still ask for an appropriate version.  Everyone using ~8 for Drupal 8.2.x would have to change.  We could put `conflict` directives in our composer.json to catch incompatibilities at `composer update` time.

When I say "we should probably", what I actually mean is that we must do this, and must continue to do this every time Drupal updates its dependencies in a way that is incompatible with the dependencies currently in Drush. This **middle road** strategy is what we're stuck with in the short term, but I don't think it's the best option long-term.

alexpott:
I've managed to get around the Drupal 8.2.x site-install problems when Drush is using Symfony 3.x dependencies

bimsonz:
@alexpott care to elaborate ?

alexpott:
@bimsonz sorry I meant to include this link :)

https://www.drupal.org/node/2843828

bimsonz:
@alexpott thanks man!!

jibran:
This needs re-roll and what's the verdict on https://github.com/drush-ops/drush/pull/2477#issuecomment-264508717?

ms-slasher13:
Can't merge pull request. Created an update of this pull request:
https://github.com/drush-ops/drush/pull/2672

Appveyor failures
 - kb2999226 (exited 1058) - Error while running 'C:\ProgramData\chocolatey\lib\KB2999226\tools\chocolateyinstall.ps1'.
 See log for details.
 - kb2919442 (exited 1058) - Error while running 'C:\ProgramData\chocolatey\lib\KB2919442\tools\ChocolateyInstall.ps1'.
 See log for details.
 - kb3035131 (exited 1058) - Error while running 'C:\ProgramData\chocolatey\lib\KB3035131\Tools\ChocolateyInstall.ps1'.
 See log for details.
Command exited with code 1058

weitzman:
My inclination is to merge this into Drush master ASAP. Drupal 8.3 is almost out. I'll do that unless anyone objects in next 12 hours. 

A reroll would be appreciated.

alexpott:
@weitzman - this was done by @ms-slasher13 in #2672

That said Drupal 8.3.x will not have symfony 3.x so I think we're good here.

weitzman:
Darn. I could use some features from 3.1+.

I got confused. Its easy to miss the tiny word Draft at 'https://github.com/drush-ops/drush/pull/2477#issuecomment-287702077'.

PR #2481, opened 2016-12-02
Append /login to the user password reset URL also in 8.X

Berdir:
Drupal 8 supports this again since 8.1.9 or so, as part of a security improvement.

Not sure if we want some kind of version check.

### DIFF
--- a/lib/Drush/User/UserSingle7.php
+++ b/lib/Drush/User/UserSingle7.php
@@ -32,15 +32,6 @@ class UserSingle7 extends UserSingleBase {
     return $userinfo;
   }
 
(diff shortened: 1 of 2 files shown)


weitzman:
How do control the URL that we go to after clicking the one time login button. The user should be able to control that as first arg to uli command but on D8 it does not work.

Berdir:
I had a look at that. What you describe is now how drush_user_login() works. The single-argument is only when combined with --uid, --name or --mail and then it works fine, at least now with this PR.

If you don't provide any of those, then the first argument is consider the user selector and the second is the path. That also seems to work fine with this PR

Issue #2484, opened 2016-12-03
Give site-install an interactive mode.


serundeputy:
If no `--db-url` option is specified drop into interactive shell prompts to get the required information form the user.

serundeputy:
@greg-1-anderson I wonder if you'd be willing to take a look at this `PR`: https://github.com/drush-ops/drush/pull/2485

In the file: `site_install.drush.inc` I've introduced the `drush_prompt`'s if `--db-url` is empty.

```php
$db_spec['database'] = drush_prompt('Database name');
    $db_spec['driver'] = drush_prompt('Database driver', 'mysql');
    $db_spec['username'] = drush_prompt('Database username');
    $db_spec['password'] = drush_prompt('Database password');
    $db_spec['host'] = drush_prompt('Database host', 'localhost');
    $db_spec['port'] = drush_prompt('Database port', '3306');
    //$db_spec->db_prefix = drush_prompt('Database prefix', NULL);
    //return drush_sql_get_class($db_spec);
    $sql = drush_sql_get_class($db_spec);
```

That seems to get the `$db_spec` and either create the DB or drop the tables as necessary, but then I seem to have to repopulate the `$db_spec` in either `commands/core/drupal/site_install.inc` for D8 or `commands/core/drupal/site_install_7.inc` for D7.

There must be a way I can pass the `$db_spec` info through to those functions, but I can't figure it out ... you think you could give me some direction?

weitzman:
In 8.x, you could try with drush_set_option('db-url, drush_sitealias_convert_db_spec_to_db_url($db_spec));

Note that we'll need PR for master branch first, and then can merge this into 8.x. Thanks for the great improvement.

serundeputy:
@weitzman thanks!

re-rolled against `master`

https://github.com/drush-ops/drush/pull/2485

weitzman:
This went in

PR #2485, opened 2016-12-03
Site install interactive

serundeputy:
#2484

### DIFF
--- a/lib/Drush/Commands/core/SiteInstallCommands.php
+++ b/lib/Drush/Commands/core/SiteInstallCommands.php
@@ -190,10 +190,29 @@ class SiteInstallCommands extends DrushCommands {
         throw new \Exception('The config source is not a directory.');
       }
     }
+    if ($commandData->input()->getOption('db-url') == '') {
+      // Prompt for the db-url data if it was not provided via --db-url.
+      $database = drush_prompt('Database name');
+      $driver = drush_prompt('Database driver', 'mysql');
+      $username = drush_prompt('Database username');
+      $password = drush_prompt('Database password');
+      $host = drush_prompt('Database host', 'localhost');
+      $port = drush_prompt('Database port', '3306');
+      $db_url = "$driver://$username:$password@$host/$database";
 
-    $sql = SqlBase::create($commandData->input()->getOptions());
-    if (!$sql->getDbSpec()) {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Yeah, this looks good in concept -- but why isn't the results from Travis showing up here? I looked briefly and did not see the tests. If you can demonstrate that the tests are passing, then I'd be okay with merging this, porting to 8.x, and then doing the annotated conversion as a follow-on step.

If this were to be done to the annotated version of this command, `@hook interact` would be the thing to use to get the database credentials.

greg-1-anderson:
Tests all green: +1

weitzman:
I just merged #2466, so this needs a re-roll.

1. I'm not sure we can use `hook @interact` after all. If user is re-installing, he/she doesn't need to provide a db-url, and we shouldn't ask for one. We bootstrap to SITE level during 'pre' phase so I we know yet whether a db-url is needed during the interact phase. So, I think the code in pre phase like the PR is now.
1. We need to replace call to drush_set_error() with an exception and use $options instead of drush_get_option()

weitzman:
ping @serundeputy

serundeputy:
@weitzman thanks for the ping.

I've refactored this agains the new master and I've got the prompts working but I'm confused about what `$sql = SqlBase::create($commandData->input()->getOptions());` is expecting.

I've set the options via prompts but it is bailing out with unrecoverable errors.

Can you provide any insight?

thanks,
~Geoff

weitzman:
The only setOption() that you should need is on db-url. You can use regular local variables to briefly store stuff like password, host, etc. 

When you pass `$commandData->input()->getOptions()` to `SqlBase::create()`, its important that 'db-url' is set in the returned array.

What errors are you seeing?

serundeputy:
Thanks @weitzman 
refactored the PR tests are passing :)

Manually tested on current D8 `8.2.6` works great.

Manually tested on current D7 `7.54` but `si` is not working on `master` for D7 at all:

```
[warning] require(/Users/geoff/sites/flat/drupal-7.54/autoload.php): failed to open stream: No such file or directory drupal.inc:17
```

I think the D7 is a separate issue though.

~Geoff

weitzman:
I moved the code around and then added this in 820fb5d. Thanks @serundeputy. Great to have some more interactivity in Drush.

greg-1-anderson:
This could also have been done with `@hook interact`

PR #2492, opened 2016-12-10
Port version and core-execute commands.

weitzman:
Like site-ssh, core-execute also should support variable number of arguments. @greg-1-anderson will look into that as time permits.


weitzman:
Still outstanding-  one last test failure which indicates that --backend requests are not getting their 'output' element populated ('object' is working fine).

PR #2494, opened 2016-12-10
Validate module uninstall eligibility

### LINKED ISSUE #2493: Catch ModuleUninstallValidatorException
`drush pmu` doesn't display the reasons from this exception very well and it's sometimes very difficult to parse.

It should catch `\Drupal\Core\Extension\ModuleUninstallValidatorException` and then parse the reasons (splitting by `;`) into a readable list.

markcarver:
Fixes #2493 

**Before:**
screen shot 2016-12-10 at 5 16 49 pm

**After:**
screen shot 2016-12-10 at 5 08 13 pm

X-ref: https://www.drupal.org/node/2834394

### DIFF
--- a/commands/core/drupal/environment.inc
+++ b/commands/core/drupal/environment.inc
@@ -221,6 +221,7 @@ function drush_module_disable($modules) {
  * @param $modules
  *   Array of module names
  *
+ * @see drush_module_uninstall_validate()
  * @see \Drupal\Core\Extension\ModuleInstallerInterface::uninstall()
  */
 function drush_module_uninstall($modules) {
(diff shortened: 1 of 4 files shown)


weitzman:
Drush has validate hooks for this purpose.

markcarver:
I know and I thought about adding a validation for this in the drush command, but the reality is that any code that invokes `drush_module_uninstall()` will fatal if it's not validated first (because it doesn't catch the exception from drupal). Thus, this validation is specific to Drupal, not Drush, and should live with that code... IMO.

markcarver:
Ok then... I'm interpreting your silence as a firm "no" and went ahead and created the validation for `pm_uninstall` as you so elaborately dictated.

Note: because of Drush's BC nightmare, I had to add even more cruft to keep the < 8.x implementations of this command working.

I really think this is backwards, but meh, whatever. Just needs to be fixed.

weitzman:
ping @jonhattan for review (pm maintainer)

weitzman:
Moved to #2494 for Drush 9.

PR #2499, opened 2016-12-15
Add OptionsCommands for sharing options providers

weitzman:
Adds a trait to DrushCommands class. The new trait holds option providers that are shared across commandfiles.

- [x] Decide if this options trait is the right way to share options providers. Same technique would apply to validation providers.
- [x] Fix \annotationcommand_adapter_add_hook_options. That is currently adding the optionsets to all commands. Need guidance on how best to find the hooked options. 
- [x] InitCommands wants to extend DrushCommands so it gets the trait but then it misses out on \Robo\Tasks. Should \Robo\Tasks be brought in some other way?


greg-1-anderson:
A trait is not necessary here; multiple instances of `@hook option @optionset_proc_build` are not needed. Just define that in any one commandfile; then, any command tagged @optionset_proc_build will get the options. We should probably have a namespace for our hooks. Terminus has this issue; at the moment, that project is just adding hardcoded hook commandfile classnames to the discovered commandfiles list after discovery runs (there are only two -- and one is in an unmerged PR). We could do similarly, or we could run discovery twice.

greg-1-anderson:
\Robo\Tasks is composed with a trait; the trait may be used instead of inheritance as desired.

greg-1-anderson:
I need to experiment to figure out the answer to the `annotationcommand_adapter_add_hook_options` questions. I am not sure off the top of my head how to best do that.

weitzman:
Thanks @greg-1-anderson . I implemented those suggestions, and used your tip to fix up annotationcommand_adapter_add_hook_options().

Can you please take a look at the use statement and class declarations of DrushCommands and InitCommands to see if they represent the best usage of Robo that we can model?

I did not need to change Command Discovery at all.

includes/annotationcommand_adapter.inc
```diff
@@ -470,7 +470,17 @@ function annotationcommand_adapter_add_hook_options(&$command)
   // the extension hook a core command, or have an early-loaded global
   // extension hook a late-loaded extension (e.g. attached to a module).
   $names = annotationcommand_adapter_command_names($command);
-  $names[] = '*'; // we are missing annotations here; maybe we just don't support that? (TODO later, maybe)
+  $names[] = '*';
+  // Add possible hooks.
+  $blacklist = ['command', 'handle-remote-commands', 'strict-option-handling', 'allow-additional-options', 'bootstrap', 'topic', 'hidden', 'hidden-option'];
```
greg-1-anderson: Props for removing annotations we know will not be used as hooks. Maybe should add that to the upstream api too.

lib/Drush/Commands/OptionsCommands.php
```diff
@@ -0,0 +1,26 @@
+<?php
+namespace Drush\Commands;
+
+/*
+ * Common options providers. Use them by adding an annotation to your method.
+ */
+class OptionsCommands extends DrushCommands {
```
greg-1-anderson: If you don't mind mixing hooks with commands, this is a perfectly fine thing to do.

There's nothing wrong with making this extend DrushCommands, but is that necessary? For a commandfile with only hooks, this could perhaps just be `class OptionsCommands`

greg-1-anderson:
LGTM 👍 

Thanks for fixing this up!

PR #2506, opened 2016-12-22
Port docs- commands to Annotated

weitzman:


commands/core/docs.drush.inc
```diff
@@ -1,310 +0,0 @@
...
-  );
-  $items['docs-aliases'] = array(
-    'description' => 'Site aliases overview on creating your own aliases for commonly used Drupal sites with examples from example.aliases.drushrc.php.',
-    'hidden' => TRUE,
-    'topic' => TRUE,
-    'bootstrap' => DRUSH_BOOTSTRAP_NONE,
-    'callback' => 'drush_print_file',
-    'callback arguments' => array($docs_dir . '/examples/example.aliases.drushrc.php'),
-  );
-  $items['docs-ini-files'] = array(
```
jibran: I saw that we removed the example file but `\Drush\Commands\core\EditCommands::phpIniFiles` still supports `DRUSH_INI`, `/etc/drush.ini` and `~/.drush/drush.ini`.

Should we add `example.drush.ini` back?

weitzman: no, those files are no longer used. we need to fix EditCommands.

weitzman: See 2nd tip at http://docs.drush.org/en/master/install/

jibran: So how are we going to update drush specific php config? Can we add php ini override in drush.yml?

greg-1-anderson: What is your use-case for drush-specific php config?

jibran: To setup the memory limit `memory_limit = 256M`

PR #2510, opened 2016-12-23
Use output-formatters:~3

greg-1-anderson:
I'll explain this if the tests pass (as they should).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
     "psr/log": "~1.0",
     "psy/psysh": "~0.6",
     "consolidation/annotated-command": "~2",
-    "consolidation/output-formatters": "~2",
+    "consolidation/output-formatters": "~3",
     "symfony/yaml": "~2.3",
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Over in https://github.com/pantheon-systems/terminus/issues/1488, @bval requested that the dependencies for Terminus and Drush 8 be brought in line. While the specific reason for doing this is not applicable, it did bring up the fact that Drush master is using output-formatters:~3, whereas Drush 8 is still on output-formatters:~2.

These two versions are compatible with each other, save for the fact that the RenderCellInterface changed in ~3, because the initial definition was inadequate for its purpose. Few people are using ouptut-formatters right now, and I'd bet that there are zero clients that depend on the old RenderCellInterface.

I think there's value in having these components consistent in Drush 8 and Drush master, so that when folks do start using output formatters, it will be possible to write Drush commands that behave consistently between Drush 8 and Drush 9.

Issue #2511, opened 2016-12-24
Change $_SERVER['REQUEST_METHOD'] default to 'GET'


kentr:
### Problem / Motivation

After upgrading core from 8.1.x to 8.2.4, I'm experiencing issues with menu items when enabling / disabling modules via drush:

The `menu_tree` table becomes corrupted for menu items that were created with a path alias as the value for **Link**, and the active menu item isn't set correctly in the page output.

#### Setup

1. Assign a path alias to a node.
2. Create a menu item, using that path alias in the `Link` field.
3. Go to the node page in the browser.  The menu item should be highlighted as "active".
4. Observe the `menu_tree` table with the following query.  There should be a record with the new menu item, and the record should have `route_name = 'entity.node.canonical'`, the node id in `route_param_key` & `route_parameters`:

```sql
-- JOINS on `menu_link_content` and `menu_link_content_data` are for convenience,
-- to view human-readable identifiers for the menu entry.
SELECT
  mt.menu_name
  , mt.mlid
  , mt.id
  , mt.route_name
  , mt.route_param_key
  , mt.route_parameters
  , mt.url
  , mlcd.title
  , mlcd.link__uri
FROM menu_tree mt
JOIN menu_link_content mlc
  ON CONCAT('menu_link_content:', mlc.uuid) = mt.id
JOIN menu_link_content_data mlcd
  ON mlcd.id = mlc.id
WHERE
  mt.id LIKE 'menu_link_content:%'
  AND
    (
      route_name IS NULL
      OR route_name = 'entity.node.canonical'
    )
ORDER BY
  route_name
  , route_param_key
```

#### Triggering the issue

1. Enable devel module via drush (tested with drush versions `8.2-dev` & `master`).
2. Reload the node page in the browser.  The menu item now isn't highlighted as "active".
3. Rerun the SQL query.  The record in the `menu_tree` table now has `route_name = NULL`, empty `route_param_key`, & empty `route_parameters`.

#### Repairing the `menu_tree` table

Either of these should correctly repopulate `menu_tree` and correct the menu item's "active" state in the browser:
* Execute `drush cr`.
* Enable / disabling devel via the UI.

I believe the first option works because [`CacheCommands::rebuild()` explicitly sets the request method to 'GET'](https://github.com/drush-ops/drush/blob/master/lib/Drush/Commands/core/CacheCommands.php#L183).

#### Explanation

[...]

PR #2512, opened 2016-12-24
change $_SERVER['REQUEST_METHOD'] default to 'GET'

kentr:
For #2511

### DIFF
--- a/lib/Drush/Boot/DrupalBoot.php
+++ b/lib/Drush/Boot/DrupalBoot.php
@@ -389,7 +389,7 @@ abstract class DrupalBoot extends BaseBoot {
     $_SERVER['PHP_SELF'] = $_SERVER['REQUEST_URI'] . 'index.php';
     $_SERVER['SCRIPT_NAME'] = $_SERVER['PHP_SELF'];
     $_SERVER['REMOTE_ADDR'] = '127.0.0.1';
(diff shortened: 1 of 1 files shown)


kentr:
Happy to write a test for this, but I need suggestions on the details (what specifically to assert, etc).

kentr:
@weitzman What are the steps for getting this into a stable 8.x version?

pjcdawkins:
Thanks for the PR @kentr, this fixed a problem for me with the simple_sitemap module, which took a very long time to debug (https://www.drupal.org/node/2828238).

I'll second that having it in a stable release would be much appreciated.

weitzman:
We create stable releases every couple months unless an emergency arises.

kentr:
@weitzman Thx.

I'm don't know what the criteria are for "emergency" status, but for consideration in that triage:

This issue will break sites...  It prevented my client from upgrading Drupal core b/c the upgrade broke their theme.

It's also [affecting the Simple XML Sitemap module](https://www.drupal.org/node/2828238).

I would guess that the average drush user doesn't understand the intricacies of installing a dev version or patching their project-specific version.

greg-1-anderson:
The situation in #2511 sounds severe enough to me to warrant a release of 8.x. Our last release, 8.1.8 was 29 Nov, so we're about due anyway.

Issue #2524, opened 2016-12-29
How to use @hook interact?


weitzman:
Please advise on how to use @hook interact. We now have two non-functional instances of it. See WatchdogCommands and ImageCommands.

greg-1-anderson:
`@hook interact` is called by the Symfony Console `interact` method during the Symfony Console command dispatch process, which Drush is not using. Beyond the fact that this is not implemented, the usage in WatchdogCommands is correct.

Hook interact is [called in AnnotatedCommand::interact()](https://github.com/consolidation/annotated-command/blob/2.2.2/src/AnnotatedCommand.php#L362). We simply need to call this method, or equivalent functionality, iff interaction has not been disabled. We also need to decide if -n should behave like Drush 8 or like Symfony Console. Terminus has both --yes and -n, the later of which behaves like Symfony Console. In SC, -n defaults most yes/no prompts to `yes`. Terminus defaults destructive prompts to `no` and the rest to `yes`.

greg-1-anderson:
+1 LGTM looks like a good start

weitzman:
We are now using this in a couple commands. Grep for `@hook interact`

PR #2525, opened 2016-12-29
Implement @hook interact and @hook init.

greg-1-anderson:
This is a start. c.f. #2524; we don't have an implementation for --no-interaction yet.  Also, we will need to throw an exception to abort from an interact hook.

Ideally, hook interact should only be used to ask the user to provide values for arguments or options not provided on the commandline. If something is queried in the interact hook, then it should be skipped if the user already provided a usable value. The image commands follow this pattern, but in the case of `watchdog-list`, the interact hook is not necessary -- the interaction could be done as part of the command, before it calls through to `watchdog-show`.  I left this as an interact hook for illustrative purposes, though.

n.b. The interact hook conforms to the interface provided by Symfony Console. If the annotated-command project had supplied CommandData here, then we could make `watchdog-list` an alias of `watchdog-show`, and have @hook interact exit if the command was anything other than `watchdog-list` or `wd-list`.

### DIFF
--- a/includes/annotationcommand_adapter.inc
+++ b/includes/annotationcommand_adapter.inc
@@ -212,6 +212,7 @@ function annotationcommand_adapter_process_command() {
     $args[$key] = $value;
   }
 
+  // TODO: Need to determine if $input is interactive, and ensure that $input->isInteractive() returns the correct result.
   $input = new DrushInputAdapter($args, annotationcommand_adapter_get_options($command), $command['command']);
   $output = new ConsoleOutput();
   $annotationData = $command['annotations'];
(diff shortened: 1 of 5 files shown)


greg-1-anderson:
Failing like https://travis-ci.org/drush-ops/drush/builds/187510546https://travis-ci.org/drush-ops/drush/builds/187510546

Re-building the last passing build results in a pass, but nothing in the commit above appears to be breaking. Any ideas, @weitzman ?

greg-1-anderson:
Seems like a55729ce broke sql-query

greg-1-anderson:
Merging to make master tests pass again. We can address remaining features in follow-on work.

Issue #2526, opened 2016-12-30
Undefined variable: name generate.make.inc:304


gh640:
My environment is:

- PHP: 5.6.20
- Drupal: 7.53
- Drush: 8.1.4

The command `drush generate-makefile` generates warning messages like below.

```
$ drush generate-makefile result.make
Undefined variable: name generate.make.inc:304
Undefined variable: name generate.make.inc:304
Undefined variable: name generate.make.inc:304
```

Seeing the code, the variable $name is not defined in the scope. Probably it should be passed to the function.

```
/**
 * Helper function to check for a non-default installation location.
 */
function _drush_generate_makefile_check_path($name, $project) {
  $info = array();
  $type = $project['type'];
  $path = dirname($project['path']);
  // Check to see if the path is in a subdir sites/all/modules or
  // profiles/profilename/modules
  if (preg_match('@^sites/[a-zA-Z0-9_]*/' . $type . 's/..*@', $path) || preg_match('@^sites/[a-zA-Z0-9_]*/' . $type . 's/..*@', $path)) {
    $subdir = preg_replace(array('@^[a-zA-Z0-9_]*/[a-zA-Z0-9_]*/' . $type . 's/*@', "@/$name" . '$@'), '', $path);
    if (!empty($subdir)) {
      $info['subdir'] = $subdir;
    }
  }
  return $info;
}
```

gh640:
I created a pull request for this after testing the patch in my environment:

https://github.com/drush-ops/drush/pull/2527

Thank you in advance.

greg-1-anderson:
Make is deprecated. Closing old issues.

PR #2527, opened 2016-12-30
Fix the warning Undefined variable: name in generate.make.inc.

gh640:
I believe that this can be fixed only by passing a new parameter to the function. I'd like this to be reviewed. 

The related issue is https://github.com/drush-ops/drush/issues/2526 .


greg-1-anderson:
Closing; Make is still missing a formal maintainer.

PR #2531, opened 2017-01-02
Add Drush::call or re-use taskExec for calling other commands.

weitzman:
For now it wraps drush_invoke_process(). Should we change its signature? 

Inspired by https://laravel.com/docs/5.3/artisan#programmatically-executing-commands. That example doesn't require forking a new process - maybe we shouldn't either? How can we achieve same as Artisan::call(). Maybe the method in this PR should be `Drush::process()` and `Drush::call` comes later.

### DIFF
--- a/lib/Drush/Commands/core/BrowseCommands.php
+++ b/lib/Drush/Commands/core/BrowseCommands.php
@@ -30,7 +30,7 @@ class BrowseCommands extends DrushCommands {
     $alias = drush_get_context('DRUSH_TARGET_SITE_ALIAS');
     if (drush_sitealias_is_remote_site($alias)) {
       $site_record = drush_sitealias_get_record($alias);
-      $return = drush_invoke_process($site_record, 'browse', [$path], drush_redispatch_get_options(), array('integrate' => TRUE));
+      $return = \Drush::call($site_record, 'browse', [$path], drush_redispatch_get_options(), array('integrate' => TRUE));
       if ($return['error_status']) {
         throw new \Exception('Unable to execute browse command on remote alias.');
       }
--- a/lib/Drush/Commands/core/BrowseCommands.php
+++ b/lib/Drush/Commands/core/BrowseCommands.php
@@ -2,8 +2,13 @@
 namespace Drush\Commands\core;
 
 use Drush\Commands\DrushCommands;
+use Robo\Contract\BuilderAwareInterface;
+use Robo\LoadAllTasks;
+use Symfony\Component\Console\Application;
 
-class BrowseCommands extends DrushCommands {
+class BrowseCommands extends DrushCommands implements BuilderAwareInterface {
+
+  use LoadAllTasks;
 
   /**
    * Display a link to a given path or open link in a browser.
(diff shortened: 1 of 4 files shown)


greg-1-anderson:
I think I prefer `Drush::run`, like Symfony Console's `Application::run`.

Drush needs to invoke a new process frequently due to the way that global options work. This shouldn't be an issue with the current API, as long as the code being called uses the options passed in via parameter rather than via `drush_get_option()`.

greg-1-anderson:
Hm, signature is different for `Application::run`. :man_shrugging: Not strongly opinionated on what to call this.  Would `Drush::drush` be too repetitive? `call` is fine.

greg-1-anderson:
Sorry, I should read the PR before commenting. 😝  If you have a function that takes an alias as the first parameter, then it might need to bootstrap a different Drupal site -- ergo, if there is an alias parameter, then it should call through to `drush_invoke_process`. If the `call` function executes in the same process, then it should not take a site alias parameter, and should implicitly operate on `@self`.

Not sure which function should be called what.

weitzman:
Instead of a Drush::call(), how about we create a class called 'Request' or 'Call'. It has methods where you specify the target alias (defaults to @self), arguments, options, etc. After building all that up, you run $request->execute(). This would be a more structured alternative to drush_invoke_process().

greg-1-anderson:
If you want structured, we could consider building off of `taskExec` from Robo.

weitzman:
I looked at TaskExec ([docs](http://robo.li/tasks/Base/), [code](https://github.com/consolidation/Robo/blob/master/src/Task/Base/Exec.php)) and it looks pretty nice to me. Could you elaborate on how we might build upon it? Would we actually re-use its code (how?) or just draw inspiration?

weitzman:
Added a new commit with a little progress. See my inline code comments

lib/Drush/Commands/core/BrowseCommands.php
```diff
@@ -30,7 +35,19 @@ public function browse($path = '', $options = ['browser' => NULL]) {
     $alias = drush_get_context('DRUSH_TARGET_SITE_ALIAS');
     if (drush_sitealias_is_remote_site($alias)) {
       $site_record = drush_sitealias_get_record($alias);
-      $return = drush_invoke_process($site_record, 'browse', [$path], drush_redispatch_get_options(), array('integrate' => TRUE));
+
+      // Experiment 1a.
+      $ret = $this->taskExec('drush browse')->run();
```
weitzman: This works but it looks ugly. Better to pass an object to taskExec()? See next experiment.

lib/Drush/Commands/core/BrowseCommands.php
```diff
@@ -30,7 +35,19 @@ public function browse($path = '', $options = ['browser' => NULL]) {
...
+      // Experiment 1b.
+      $container = \Drush::getContainer();
+      /** @var Application $app */
+      $app = $container->get('application');
+      $command = $app->get('browse');
+      $ret = $this->taskExec($command)->site($site_record)->run();
```
weitzman: This results in `Consolidation\AnnotatedCommand\AnnotatedCommand does not implement CommandInterface, so can't be passed into this task `. How to fix?

Also, note that I have invented a site() method. Thats how I envision a dev indicating that the task should be run on a separate site. Any thoughts on how to implement that?

weitzman:
Yeah, that fixed this nicely - `Drush::drush()`

PR #2540, opened 2017-01-05
Fix #2538: Create command drush pm-uninstall-disabled

### LINKED ISSUE #2538: drush pm-uninstall-disabled command
It would be very helpful if there was a Drush command which would, in one step, uninstall all modules on a site that are disabled. The command could be <code>drush pm-uninstall-disabled</code>. It would take no arguments.

lkmorlan:


weitzman:
This has a limited audience, and doesn't apply to Drupal 8, as you've noted. All in all, I thing this command lives better in Contrib instead of Drush core. Still, thanks for sharing it.

lkmorlan:
OK, thanks. If I want to make this publicly available, should I post it as a drupal.org project or is there another place for projects which are only Drush commands?

PR #2542, opened 2017-01-09
Added dt_plural function

adrian-cid:
As drush don't have a plural string function we can't translate plural strings, with this function you can use a code like this:

$output = dt_plural($node->comment_count, '1 comment', '@count comments');

To translate your plurals strings.

Reference:
http://drupal.stackexchange.com/q/206820/28275

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -179,6 +179,71 @@ function dt($string, $args = array()) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


adrian-cid:
I can't see why the merge failed. Some idea?

greg-1-anderson:
n.b. Drush 8 is in maintenance-only.

PR #2543, opened 2017-01-09
Fix the site list when a site is symlinked.

awm086:
- `$real_root= drush_locate_root($filename)` always return true even if the site directory is symlinked to a directory outside drupal root. This fixes cases where some sites are symlinked.
- For example: a site that is symlinked in the following manner `ls -l DRUPAL_ROOT/sites`  `vanilla -> ../../vanilla/src` where the target is a directory outside the drupal root, the function drush_locate_root return true we end up with a wrong alias.
- I think we should be using the real site directory (`$realsite_dir`) instead of the filename to properly build the `$site_list` for sites that are symlinked.


greg-1-anderson:
This PR looks too old to merge into the latest version of Drush. Site alias code has been rewritten for Drush 10. Please continue on the 8.x branch if this fix is still relevant for the older version of Drush.

PR #2548, opened 2017-01-16
Port config commands to Annotated

weitzman:
Big port. Some notes

1. Replaces the old style storage filter hook with a new style on-event hook. Needs backport to 8.x. 
1. The new hook receives $options as argument. I was hoping to pass something substantive like CommandData but I didn't see how to do that.
1. We have a bootstrap problem during bash completion of a command argument. Lets take that one at #2514.
1. config-list command is gone in favor of interactive choosing of config_name
1. config-get command was simplified to always pull from active config.


weitzman:
Merged. Lets keep discussing the open issues as needed.

PR #2549, opened 2017-01-17
Fix spacing between badges in README

AMDmi3:


weitzman:
Merged into master and 8.x. Thanks for your sharp eyes and willingness to make a PR.

PR #2551, opened 2017-01-18
Fixes #2528: Remove make.

### LINKED ISSUE #2528: Remove make from Drush9
Per https://medium.com/@weitzman/drush-9-roadmap-9eb5cd227c50, we are removing make from Drush 9 and recommending that folks use Composer instead. This issue forms a plan for executing this.

- **Keep** ability for folks to convert to Composer from a makefile in ini or yaml format. Keep tests/makeConvertTest.php
- **Port** make-convert to Annotated.
- **Drop** everything else - Building, writing a lock file, make-generate, ...

Make will keep getting bug fixes in Drush 8, so folks have a long time to transition.

grasmash:
Leaving this here as a place holder. WIP and not ready to merge.

Todo:

- [ ] Audit remaining make code with a code metrics tool to identify unused methods
- [ ] Refactor tests
- [ ] Modify make-convert command so that it uses D.O packagist and accompanying version constraints.
- [ ] Audit make-convert command for deprecated functions


grasmash:
@weitzman 

If we're going to continue supporting make-convert, then we need to support parsing ini files, which may mean support includes, recursive make files, remote make files, etc.

There's a whole slew of functionality associated with parsing the various things that might be found in a make file.

Where do we draw the line? We could greatly simplify the codebase if we, for instance, said that we weren't going to support remote make file references.

weitzman:
Could folks use Drush8 to generate a lock file and then we convert that? That way we remove remote support. Or maybe enhance make-convert in 8.x and then we don't have to ship any make at all in Drush9?

DocDJ:
Why would you want to remove remote support as opposed to leaving it up to the sys-admin?
 
Live long & prosper,
Dennis

weitzman:
I went with my last proposal. Folks can use make in Drush8 and convert using its make-convert when they are ready. See #2644 2644

Issue #1854, opened 2015-12-11
Add pm-uninstall option to automatically uninstall dependent modules


colans:
For Drupal 8, there's no more "drush pm-disable" command to turn off modules & such (say for development sites).  They now need to be uninstalled.  But in Drupal 7-, "dis" automatically uninstalled dependent modules as well.  There's no such functionality in pm-uninstall to do the same sort of thing in Drupal 8.

We should add a pm-uninstall option that takes care of this, something like "--remove-dependents".

danepowell:
The current situation is also somewhat unintuitive, in that the dependency chain of the module arguments isn't even taken into consideration--the modules are just uninstalled in the order they are passed. So if you are uninstalling modules with dependencies on each other, you have to be very careful in what order you pass them as argument to `pmu`. This is especially painful in situations where you may not know _a priori_ what the dependencies are.

danepowell:
See #2566, I think this is actually a regression introduced by #925

Issue #925, opened 2014-10-24
Uninstall should fail if extension is not enabled (in D8)


weitzman:
In D8, I think uninstall should fail if the extension is not enabled. This makes sense because there is no disable operation.

PR #2566, opened 2017-01-24
Issue #1854: Allow uninstall of dependent modules.

danepowell:
Currently, there is no way to uninstall modules that depend on one another because Drush checks the dependency chain for each module individually, without considering any of the other modules you are uninstalling.

It's also counterintuitive that although Drupal's core ModuleInstaller::uninstall() function automatically uninstalls dependent modules, Drush (which ultimately just uses the same function) won't let you do the same thing.

Drush should automatically uninstall depending modules, same as core. I can't imagine a scenario where the current behavior is desired. The only argument I could imagine against this is that there's a small performance hit to computing the dependency chain... but we are already incurring that hit by not passing a second parameter to ModuleInstaller::uninstall().

This is a regression introduced by #925, which ostensibly had nothing to do with checking module dependencies.

### DIFF
--- a/commands/core/drupal/pm_8.inc
+++ b/commands/core/drupal/pm_8.inc
@@ -42,18 +42,6 @@ function _drush_pm_uninstall($extensions) {
       unset($extensions[$extension]);
       drush_log(dt('!extension is already uninstalled.', array('!extension' => $extension)), LogLevel::OK);
     }
-    elseif (drush_extension_get_type($extension_info[$extension]) == 'module') {
-      $dependents = array();
-      foreach (drush_module_dependents(array($extension), $extension_info) as $dependent) {
-        if (!in_array($dependent, $required) && ($extension_info[$dependent]->status)) {
-          $dependents[] = $dependent;
-        }
-      }
-      if (count($dependents)) {
-        drush_log(dt('To uninstall !extension, the following extensions must be uninstalled first: !required', array('!extension' => $extension, '!required' => implode(', ', $dependents))), LogLevel::ERROR);
-        unset($extensions[$extension]);
-      }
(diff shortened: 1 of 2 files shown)


weitzman:
Thanks for the PR. Hopefully @jonhattan will get a chance to look into this bug.

weitzman:
Others are welcome to review/test this. @jonhattan is quite busy.

weitzman:
Related #2494

weitzman:
Will the prompt list the defendant modules that will also be removed? If so I think this makes sense to merge.

danepowell:
No this won't list dependencies. If we want to do that, we would have to manually compute the dependency chain and pass false as the second parameter to drush_module_uninstall.

I think it would be good to merge this as-is and open a follow-up issue to make that change, since the current situation is pretty bad (completely impossible to uninstall any modules with dependencies).

danepowell:
Actually it might not be so hard to make that change, since the existing code already builds the dependency chain, I'll try to take a look later today.

danepowell:
Okay this should list all modules (including dependencies) that will be uninstalled.

weitzman:
Worked for me. Thanks a lot @danepowell 

```
~/reps/drushbare (2566) $ ./drush  -r ~/reps/d8 pmu file
The following extensions will be uninstalled: file, update
Do you really want to continue? (y/n): y
 [ok] file was successfully uninstalled.
 [ok] update was successfully uninstalled.
```

weitzman:
@danepowell Any chance you can backport this to 8.x? I dont think we fixed it there.

danepowell:
This is the 8.x backport: #2717

PR #2567, opened 2017-01-25
#1523: Added schema support for PostgreSQL.

### LINKED ISSUE #1523: Postgresql schema not used by drush

RoSk0:
Fixes #1523 .


Renrhaf:
See also https://github.com/drush-ops/drush/pull/2109

RoSk0:
Fixed by #2109, closing...

RoSk0:
Silly me, that PR wasn't merged. Re-opening..

RoSk0:
Updated code, also added schema support to sql-drop command.

I would recommend testing this PR with patch from https://www.drupal.org/node/1060476.

src/Sql/SqlPgsql.php
```diff
@@ -170,4 +171,46 @@ public function getPasswordFile()
...
+
+    /**
+     * Get schema name.
+     *
+     * @return string
+     */
+    public function getSchemaName()
+    {
+        $schema = 'public';
+        $dbSpec = $this->getDbSpec();
```
Renrhaf: @RoSk0 Hello, this method changed to "db_spec()" in master. The patch does not apply anymore, could you take a look please ? :)

Renrhaf:
@tstoeckler up ? :) this patch can not be merged anymore because of some minor modifications in master branch such as the "db_spec" method name change. But it really helps when using a different schema than the default one "public".

RoSk0:
@Renrhaf Hi, updated branch. `db_spec()` marked as deprecated in master... Not sure I understand your comment.

RoSk0:
Failing tests using MySQL which is not good to test PostgreSQL related functionality. I have tested this command manually, works without issues. Can we test with PostgreSQL?

Renrhaf:
@RoSk0 sorry you're right, I was trying to apply the patch to the 8.1.15 tag, not the master branch. Thanks for your patch.

jibran:
Changes look good to me. RTBC.

weitzman:
FWIW postgres is now fully tested every night by our functional test suite.

RoSk0:
Passing... 

Will test manually locally and then will try to add test for the case when schema is other then `public`.

RoSk0:
Finally those tests fail as expected... in #3848 which is a test-only-wip branch.

But now, the way I wrote them, this depends on #3849

RoSk0:
Is there any reason why we not testing on PostgreSQL on pull requests? Can we enabled it?

RoSk0:
Finally all green!

greg-1-anderson:
Is this still needed? Drush is testing postgres nightly.

RoSk0:
Hi @greg-1-anderson ,

Yes, I believe it is. This is where schema support will be added to Drupal https://www.drupal.org/project/drupal/issues/1060476

src/Sql/SqlPgsql.php
```diff
@@ -178,4 +180,44 @@ public function getPasswordFile()
...
+        if (!empty($dbSpec['prefix'])) {
```
mradcliffe, line 193: I think `!empty` will be a warning on a certain versions of PHP rather than `isset`.

RoSk0: Not according to docs https://www.php.net/empty

RoSk0:
Going to rebase this branch on top of 10.x and push updated soon

RoSk0:
Well, this turned out to be a bit more involved than I expected. I tried to insure that we have proper test coverage by creating test only PR here https://github.com/drush-ops/drush/pull/4507 but it fails and ran out of time for it. 

Unfortunately leaving it in a limbo for now.

PR #2514, opened 2017-01-25
Completion and Annotated

weitzman:
In 28f1adc, we work around the fact that Annotated command that provide global options (e.g. --notify, --druplicon) are not recognized by drush_get_global_options() and thus not offerred by our completion script.

I'm not sure it makes sense to distinguish global options from local anymore. Further, we should look at completion frameworks like https://github.com/bamarni/symfony-console-autocomplete (requires XML help) and https://github.com/stecman/symfony-console-completion

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,8 @@
     "pear/console_table": "~1.3.0",
     "phpdocumentor/reflection-docblock": "^2.0",
     "webmozart/path-util": "~2",
-    "sebastian/version": "~1"
+    "sebastian/version": "~1",
+    "stecman/symfony-console-completion": "^0.7.0"
   },
   "require-dev": {
     "lox/xhprof": "dev-master",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "546d303006c9073ab6f5238326b266e5",
+    "content-hash": "e341e5561b685a97a11d540235ea02fb",
     "packages": [
         {
             "name": "consolidation/annotated-command",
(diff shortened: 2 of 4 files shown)


weitzman:
This PR uses the stecman project from above to provide completion. Enable completion by running `eval $(drush _completion --generate-hook -p drush)` in your bash shell.

The current problem is that only the built-in commands are available for completion. Once we figure out how to bootstrap, this thing will be dog slow. So we'll need to cache command definitions like the current complete.inc does ... Not sure how much time I will put into this.

greg-1-anderson:
Note that the current implementation of the annotated-command library supports caching of CommandInfo classes. The application is responsible for save / restore.  c.f. [Terminus](https://github.com/pantheon-systems/terminus).

greg-1-anderson:
Having good luck in light testing of https://github.com/bamarni/symfony-console-autocomplete.

As long as we support `list --raw` and `help --format=xml`, we should be able to use this project, or something similar, to dump a static autocomplete file.

In the fullness of time, we should customize this project to also dump  our aliases.

Chi-teck:
I think the _bamarni/symfony-console-autocomplete_ can only autocomplete command and option names. It won't help us to autocomplete the values of arguments and options.

weitzman:
I made a small commit (1177d94) and now both of the completion approaches mention here are working. Even though can complete argument and option values, I'm still a bit wary of the stecman. We had full featured completion on Drush for a while and not many people used it because it was slow and poorly maintained. I dont love a solution that calls into a Console command to source its potential completions.

One problem with both solutions is that after completing an option name, a space is added. So, instead of `--uri=dev` you get ``--uri dev". If you submit that request you get 
```
[Symfony\Component\Console\Exception\RuntimeException]  
  The "--uri" option requires a value.       
```

greg-1-anderson:
I just tested `--uri dev` and it worked for me.

I agree that a faster, less-functional autocomplete mechanism is better than a full-featured slow autocomplete mechanism. I used the old Drush autocomplete mechanism for a long time, then eventually just turned it off. Maybe we can look into ways to make autocomplete better by exporting more info; whatever we do, though, we should keep it fast.

greg-1-anderson:
Do we want to just close this and recommend https://github.com/bamarni/symfony-console-autocomplete? Haven't tried autocomplete in a while.

weitzman:
Lets keep this open as we need to add this recommendation to the docs. Also, Symfony Console has built-in autocomplete in 5.4 https://symfony.com/blog/new-in-symfony-5-4-console-autocompletion

weitzman:
Actually lets continue at https://github.com/consolidation/annotated-command/issues/243, as time permits.

PR #2571, opened 2017-01-26
Add drush cexy and drush cimy from Lee Rowlands (larowlan) at PreviousNext into drush 8

scotteuser:
After a quick message to @larowlan requesting permission to help move forward on this, I have created this PR. 

More info: 
- https://github.com/previousnext/drush_cmi_tools
- https://www.previousnext.com.au/blog/introducing-drush-cmi-tools

The benefits of merging this in from my perspective are:
- available to all developers
- eventually visible in https://drushcommands.com/ for easier to browse documentation
- helps keep all drush documentation in one place

This is the code straight from [here](https://github.com/previousnext/drush_cmi_tools), only modified to work within the drush repo and cleaned up from minor drupal 8 phpcs complaints on spacing.

### DIFF
--- a/commands/cmi/cmi_tools.drush.inc
+++ b/commands/cmi/cmi_tools.drush.inc
@@ -1,0 +1,237 @@
+<?php
+
+/**
+ * @file
+ * Config-export-plus and config-import-plus implementation.
+ *
+ * Contains turbo-charged config commands for a better partial workflow. This
+ * has been ported from the work of Lee Rowlands (larowlan) into drush from
+ * drush_cmi_tools on github by scott_euser.
+ *
+ * @author Lee Rowlands (larowlan)
+ *   https://www.drupal.org/u/larowlan.
+ * @link https://github.com/previousnext/drush_cmi_tools/
+ *   Drush CMI Tools on github.
+ */
+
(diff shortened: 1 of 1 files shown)


scotteuser:
@larowlan Just pinging you on this PR as requested, thanks!

weitzman:
Those are terrific features, but they need to be integrated properly into config-import and config-export, not bolted on as separate commands. Note the we have been working on improving flexibility for Contrib to hook into those commands. See end of https://www.drupal.org/node/2830767. config_split is working on integrating into core drush commands. drush_cmi_tools could do same or just merge with config_split

scotteuser:
Thanks for the quick reply @weitzman . Just for clarity for myself, would this be the desired approach:

Add option attributes to use the ignore list, etc to config-export and config-import within commands/core/config.drush.inc and modify larowlan's code to work when those options are used?

Or do you prefer to keep this feature out of drush entirely and as a contrib module hooking in only?

weitzman:
For starters i think it can live outside of Drush core.

scotteuser:
Okay thanks! I think I'll just continue to use it as is for now then.

PR #2580, opened 2017-01-30
Fix issue #1965: Set DRUSH_SELECTED_DRUPAL_SITE_CONF_PATH in preflight.

torotil:
This is another attempt at fixing #1965 . If possible it sets the `DRUSH_SELECTED_DRUPAL_SITE_CONF_PATH` already in `drush_preflight_site()`.

### DIFF
--- a/includes/preflight.inc
+++ b/includes/preflight.inc
@@ -351,6 +351,15 @@ function drush_preflight_root() {
 }
 
 function drush_preflight_site() {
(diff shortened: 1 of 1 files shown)


weitzman:
We deliberately don't select a drupal site until a later bootstrap phase.

torotil:
Is there actually a way to satisfy both https://github.com/drush-ops/drush/pull/2055#issuecomment-193513501 and this?

PR #2582, opened 2017-02-01
Fix issue #1314: drush core-requirements does not output descriptions in render arrays

AdamPS:
Please review new PR following the resolution suggested by @greg-1-anderson in https://github.com/drush-ops/drush/pull/1315#issuecomment-126396642

FYI I couldn't find anyway to override @default-fields so had to temporarily edit the code.  The file http://api.drush.org/api/drush/commands%21core%21outputformat%21topics%21table.html/master suggests a --fields option but that gave me  Unknown option: --fields.

I would be interested in backport to 8.x and have done and tested code changes for that too.

### DIFF
--- a/commands/core/drupal/environment.inc
+++ b/commands/core/drupal/environment.inc
@@ -389,3 +389,21 @@ function drush_get_token($value = NULL) {
 function drush_url($path = NULL, array $options = array()) {
   return \Drupal\Core\Url::fromUserInput('/' . $path, $options)->toString();
 }
+
+/**
+ * Output a Drupal render array, object or string as plain text.
+ *
+ * @param string $data
+ *   Data to render.
+ *
+ * @return string
+ *   The plain-text representation of the input.
+ */
+function drush_render($data) {
+  if (is_array($data)) {
+    $data = \Drupal::service('renderer')->renderRoot($data);
+  }
+
+  $data = \Drupal\Core\Mail\MailFormatHelper::htmlToText($data);
+  return $data;
+}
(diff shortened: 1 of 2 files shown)


weitzman:
How do I see the bug with just core D8 and core Drush?

AdamPS:
Thanks for taking time to look at this PR.  The easiest way I can see to reproduce in D8 core is to trigger the check for "file %file is not protected from modifications and poses a security risk" _with multiple files_.

```
sudo chown root:root sites/default/settings.php sites/default/settings.local.php
sudo chmod 666 sites/default/settings.php sites/default/settings.local.php
drush rq --severity=2
```

Output from latest drush phar 8.1.9
```
 Title                Severity  Description
 Configuration files  Error     Not protected
                                Array

```

Output with drush patch applied (the PR is for master but I also have a D8 patch I could share)
```
 Title           Severity  Description
 Configuration   Error     Not protected
 files                      * The file /sites/default/settings.php/ is not protected from modifications
                              and poses a security risk. You must change the file's permissions to be
                              non-writable.
                            * The file /sites/default/settings.local.php/ is not protected from
                              modifications and poses a security risk. You must change the file's
                              permissions to be non-writable.

```

weitzman:
Removed a little unneeded code and merged. Thanks @AdamPS.

I considered showing the description field by default but it wraps terribly with the line breaks so I kept it hidden, just like this PR.

weitzman:
Feel free to open a new issue for 8.x back port.

AdamPS:
Thanks @weitzman.  Agree that description is best hidden by default.  I would be interested to understand what option to pass to override the default and show the description.

8.x backport PR coming up.

weitzman:
Pass --fields option to control what fields are shown. Works for many commands.

AdamPS:
@weitzman Thanks, but it was drush 9/master that I was struggling with how to control the fields.  Perhaps that's because it's a work in progress.

PR #2587, opened 2017-02-03
Undefined property: DrushMakeProject_Library::$type make.project.inc:154

20th:
`DrushMakeProject` class does not define `type` property but uses it extensively throughout its methods. In most cases, there is no problem because `type` is set inside of class constructor from project definition array:

```
class DrushMakeProject {

  protected function __construct($project) {
    $project['base_contrib_destination'] = $project['contrib_destination'];
    foreach ($project as $key => $value) {
      $this->{$key} = $value;
    }
    // ... snip ...
  }
```

For modules and themes, the `type` key is always present in project definition because, even if it is not specified in the makefile, it will be added by `make_prepare_projects()` function:

```
  foreach ($info['projects'] as $key => $project) {
    // Merge the known data onto the project info.
    $project += array(
      'name'                => $key,
      'type'                => 'module',
      'core'                => $info['core'],

      // ... snip ..
```

But it is different for libraries - the `type` key is not added in `make_prepare_libraries()`:

```
    // Merge the known data onto the library info.
    $library += array(
      'name'                => $key,
      'core'                => $info['core'],
      'build_path'          => $build_path,
      'contrib_destination' => $contrib_destination,
      'subdir'              => '',
      'directory_name'      => $key,
      'make_directory'      => $make_dir,
    );
```

So, if a library does not have a specified version inside of a makefile, site administrators will get an `Undefined property` warning during execution of `drush make` command when `make_download_factory()` function is called in `commands/make/make.project.inc:154` with `$this->type` as a second argument.

### DIFF
--- a/commands/make/make.project.inc
+++ b/commands/make/make.project.inc
@@ -61,6 +61,11 @@ class DrushMakeProject {
   protected $variant = 'profile-only';
 
   /**
+   * Type of downloaded project.
+   */
+  protected $type = 'core';
+
+  /**
    * Set attributes and retrieve project information.
    */
   protected function __construct($project) {
@@ -636,6 +641,12 @@ class DrushMakeProject_Core extends DrushMakeProject {
  * For processing libraries.
  */
 class DrushMakeProject_Library extends DrushMakeProject {
+
+  /**
+   * Override default project type.
+   */
+  protected $type = 'library';
+
   /**
    * Override constructor for libraries to properly set contrib destination.
    */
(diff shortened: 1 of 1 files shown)


20th:
To fix this, the `type` property should be added to `DrushMakeProject` class and every child class. This way, even if `DrushMakeProject` instance is contructed with a `type` key, this property will still have a valid value.

20th:
This bug exists in `master`, `8.x`, `7.x`, `6.x` and `5.x`, but this PR applies only to `master` and `8.x`.

greg-1-anderson:
Closing; Make is still missing a formal maintainer.

PR #2588, opened 2017-02-03
Provide the possibility to use sql-sync even if @source and @target seem the same

yobottehg:
On a containerized infrastructure (like platform.sh) it is likely that all database credentials are the same. Currently there is only the database info and not the ssh info taken into account.

This would be difficult. We can just provide an option to sync the database even it it seems the same.


weitzman:
Lets just remove that validation until it can work better.

yobottehg:
Okay did that.

yobottehg:
closed in favor of #2589

PR #2589, opened 2017-02-04
remove validation for the equalness of @source and @target in sql-sync

yobottehg:
Removed the validation based on #2588


weitzman:
Thanks. Merged to 8.x as well.

PR #2591, opened 2017-02-04
Update from phpunit 4 to phpunit 5.

greg-1-anderson:
Let's try phpunit 6.  This is "php": "^7", so we probably won't want to merge this just yet.  phpunit 5 is "php": "^5.6 || ^7.0", so we might try that for Drush 9.

I didn't check the php version requirements before pushing this branch. Let's see how the tests fare. I've found downgrading from 6 to 5 is pretty easy.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
     "sebastian/version": "^2"
   },
   "require-dev": {
-    "phpunit/phpunit": "^6",
+    "phpunit/phpunit": "^5",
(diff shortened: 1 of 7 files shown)


greg-1-anderson:
Down to phpunit 5.

greg-1-anderson:
OK, phpunit 5 works fine. Ready to upgrade?

weitzman:
LGTM

PR #2596, opened 2017-02-06
Fix uli command annotation

TravisCarden:
The command help for `drush uli` is broken (`drush uli -h` thinks there's a "string" option) due to what looks like an incomplete conversation to an annotated command, and one of the usage examples doesn't use the current options correctly.


weitzman:
Thanks @TravisCarden

PR #2599, opened 2017-02-08
Remove query string processing from D8 rs bootstrap

webflo:
Related issue: https://www.drupal.org/node/2850280

ping @Berdir: D8 takes care of path processing right? Drush should not longer preprocess the incoming request.


Berdir:
?q doesn't exist anymore in D8, so yeah, pretty sure this is not needed. @dawehner might want to +1 too.

dawehner:
+1.0

weitzman:
Thanks. Was already removed in master.

PR #2602, opened 2017-02-10
Fix issue #1314: backport to 8.x

AdamPS:


weitzman:
I think drush_render() needs an implementation in environment_6.inc, even if it does nothing.

AdamPS:
Ah sorry hadn't realised D6 was still supported in drush.  Unfortunately I don't have a D6 server to test against.  However as far as I can see, an exact copy of the D7 version ought to work, as D6 supports both functions that are called here.

PR #2606, opened 2017-02-10
Pass variable args to functions that declare a parameter of type 'array'

greg-1-anderson:
First step: get the arguments to the function.

Does not respect `--` yet.


greg-1-anderson:
Strict option handing is now removed in this PR. So, instead of:
```
drush -v rsync here there --moar-rsync-options
```
Do this:
```
drush rcyns here there -v -- --moar-rsync-options
```
Note also that formerly, `drush_invoke_process` expected `--moar-rsync-options` to be placed in the $options parameter. Now, you do this instead:
```
$values = drush_invoke_process('@self', 'rsync', ['here', 'there', '--', '--moar-rsync-options'], ['v']);
```
Probably need to scrub strict options from more docs; definitely need to document `--`.

greg-1-anderson:
Fixed for rsync.

weitzman:
Great work. Let's add Usage examples if missing to the commands which use this. 

I can't recall if I commented out any testing of the extra form fields handling in site-install. If I did that could be re-enabled.

weitzman:
- I added Usage examples to core-execute and rsync. 
- There is no testing of extra arguments on site-install so lets disregard my comment above.

I did notice that both of the following commands work. Which is preferred, and why?
```
drush core-execute -- git pull origin rebase --no-ff
drush core-execute git pull origin rebase -- --no-ff
```

greg-1-anderson:
`--` is defined to mean "turn off option handling, and treat everything else that follows as arguments." Either of the above is therefore fine; this behavior is consistent with Symfony Console.

For Terminus, I was considering using `--` to disambiguate options. Terminus site specifications are like Drush aliases, but they always come immediately after the command. So:

`terminus some:command my-site.dev -- arg1 arg2 --option`

This could potentially disambiguate `my-site.dev` from `arg1`, when the site specification is optional. I'm not sure that disambiguation is going to be necessary. If it is, the above will require changes to annotated-command.

I don't think Drush needs this disambiguation either, but since you asked about `--` ordering, I thought I'd mention it.

greg-1-anderson:
Going AFK; go ahead and merge this after the tests pass if you want.

Issue #2611, opened 2017-02-13
Drush should be using the Drupal UpdateKernel when running updates


hchonov:
Currently drush is using the normal DrupalKernel when running updates, but it should be using the designated UpdateKernel as through the UpdateServiceProvider Drupal is altering services and using the null cache backend in order not to write into the cache at some places. There is currently an issue which is going to make use of this and really needs that drush is using the UpdateKernel otherwise updates might/will fail - https://www.drupal.org/node/2851529.

greg-1-anderson:
This has been done

PR #2612, opened 2017-02-13
Fix issue #2611: Use the Drupal UpdateKernel when running updates

hchonov:
Unfortunately I am not familiar with the Drush architecture, but the following code is sufficient in order to use the Drupal UpdateKernel during executing updates through drush. Please review it and probably adjust it as needed.


hchonov:
I've messed up my drush fork repository so I had to delete it .. I've reopened a new pull request for this instead - https://github.com/drush-ops/drush/pull/2627

PR #2616, opened 2017-02-14
Better feedback for failing SSH commands.

danepowell:
If you are running Drush commands on a remote host and SSH fails for any reason (say, because you forgot to spin up your DrupalVM or Vagrant Box, or the remote host is unavailable), you have to wait two minutes to find out that there's anything wrong, and then Drush silently exits with no indication of any error. Needless to say, this is a pretty bad UX and causes a lot of confusion and issue churn (just search the issue queue for "SSH").

This is a fairly simple solution. It's a little tricky to determine if a failed command is due to a problem with SSH or something else, so it's possible that this could produce false positives. If we want to make it more selective, we could just check for an error_status of 255, which indicates SSH failed to connect. And maybe verify that SSH was actually used to call the command. Let me know where you want to go with it.

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -1163,7 +1163,7 @@ function _drush_backend_generate_command($site_record, $command, $args = array()
   if (isset($hostname)) {
     $username = (isset($username)) ? drush_escapeshellarg($username, "LOCAL") . "@" : '';
     $ssh_options = $site_record['ssh-options'];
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Sorry that I missed seeing this PR last year. This year, Drush has replaced backend invoke with the Symfony Process component. Please test the latest master branch and see if it is working to expectations.

weitzman:
ssh failures are reported immediately in master.

PR #2621, opened 2017-02-16
Fixes #2617: Perform cache rebuild during updb in same process

### LINKED ISSUE #2617: drush should re-boot Drupal after the update_N's and before the post updates
Currently drush is executing the update_N's and the post updates in a single batch and in between is rebuilding the cache in a separate process, which is causing the static cache of the checksum service to be out-of-sync in the post updates. This is leading to all kind of bugs and there is a currently an issue on drupal.org - https://www.drupal.org/node/2851529.

tstoeckler:
This is a backport of bc8bb69e9e44b958e8e22cb175e7523ad113a281

### DIFF
--- a/commands/core/drupal/update.inc
+++ b/commands/core/drupal/update.inc
@@ -102,6 +102,25 @@ function drush_update_do_one($module, $number, $dependency_map,  &$context) {
   $context['message'] = 'Performing ' . $function;
 }
 
+/**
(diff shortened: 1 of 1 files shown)


hchonov:
Thank you for merging the pull request. Is there any chance that there is a new minor release including this bug fix?

PR #2623, opened 2017-02-19
Update /docs, especially install.md

weitzman:
The install docs now recommend installing Drush into your project via Composer and optionally copying drush.wrapper for convenience. I think this is the simplest approach that can possibly work.


greg-1-anderson:
Looks good.

skyred:
> The install docs now recommend installing Drush into your project via Composer and optionally copying drush.wrapper for convenience. I think this is the simplest approach that can possibly work

Drush is simply not required to run Drupal. Adding drush to composer.json as dependency is a little bit confusing, therefore maybe the "recommend" key word out.

Issue #2622, opened 2017-02-17
drush make has always new timestamps for files and directories of git checkouts - makes syncing more expensive


hchonov:
drush make is not retaining the file timestamps of core and contrib modules, which adds an enormous overhead to rsync when syncing project files.
Currently noticing this with Drush 8.1.9 on OSX 10.12.3.

PR #2626, opened 2017-02-20
Fix issue #2622: Update the timestamps of files and directories of git checkouts to the latest commit timestamp during make

hchonov:


hchonov:
I've made something wrongly during the git process ... Will make a new pull request.

PR #2627, opened 2017-02-20
Fix issue #2611: Use the Drupal UpdateKernel when running updates

hchonov:
This is a "clone" of the pull request https://github.com/drush-ops/drush/pull/2612 as I've messed up my drush fork ... Sorry about this...

Unfortunately I am not familiar with the Drush architecture, but the following code is sufficient in order to use the Drupal UpdateKernel during executing updates through drush. Please review it and probably adjust it as needed.

### DIFF
--- a/lib/Drush/Boot/BaseBoot.php
+++ b/lib/Drush/Boot/BaseBoot.php
@@ -62,6 +62,10 @@ abstract class BaseBoot implements Boot, LoggerAwareInterface {
           $command += $this->command_defaults();
           // Insure that we have bootstrapped to a high enough
           // phase for the command prior to enforcing requirements.
+          if ($command['command'] == 'updatedb-batch-process') {
+            global $drupal_update;
+            $drupal_update = TRUE;
+          }
           $bootstrap_result = drush_bootstrap_to_phase($command['bootstrap']);
           $this->enforce_requirement($command);
 
--- a/lib/Drush/Boot/DrupalBoot8.php
+++ b/lib/Drush/Boot/DrupalBoot8.php
@@ -6,7 +6,9 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Psr\Log\LoggerInterface;
 use Drupal\Core\DrupalKernel;
+use Drupal\Core\Update\UpdateKernel;
 use Drush\Drupal\DrupalKernel as DrushDrupalKernel;
+use Drush\Drupal\DrupalUpdateKernel as DrushDrupalUpdateKernel;
 use Drush\Drupal\DrushServiceModfier;
 use Symfony\Component\DependencyInjection\Reference;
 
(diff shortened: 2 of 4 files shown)


lib/Drush/Boot/BaseBoot.php
```diff
@@ -62,6 +62,10 @@ function bootstrap_and_dispatch() {
           $command += $this->command_defaults();
           // Insure that we have bootstrapped to a high enough
           // phase for the command prior to enforcing requirements.
+          if ($command['command'] == 'updatedb-batch-process') {
```
greg-1-anderson: Shouldn't check `$command['command']` here; instead, invent another property that indicates what kind of Kernel the command needs. Maybe pass that in to `drush_bootstrap_to_phase()`.

lib/Drush/Boot/DrupalBoot8.php
```diff
@@ -118,15 +120,26 @@ function bootstrap_drupal_database() {
...
 
   function bootstrap_drupal_configuration() {
+    global $drupal_update;
     $this->request = Request::createFromGlobals();
     $classloader = drush_drupal_load_autoloader(DRUPAL_ROOT);
     // @todo - use Request::create() and then no need to set PHP superglobals
     $kernelClass = new \ReflectionClass('\Drupal\Core\DrupalKernel');
     if ($kernelClass->hasMethod('addServiceModifier')) {
-      $this->kernel = DrupalKernel::createFromRequest($this->request, $classloader, 'prod', DRUPAL_ROOT);
+      if (!$drupal_update) {
```
greg-1-anderson: Rather than having Drush bootstrap and then call commands, I think it would be better to have commands call `$bootstrap->...` as needed. Maybe for Drush 10; keeping this parameterized is fine for now.

lib/Drush/Boot/BaseBoot.php
greg-1-anderson: I haven't really reviewed this patch in context, so I'm not sure at this point what the best method for passing this information down to `bootstrap_drupal_configuration()` is.

lib/Drush/Boot/DrupalBoot8.php
```diff
@@ -118,15 +120,26 @@ function bootstrap_drupal_database() {
...
+        $this->kernel = UpdateKernel::createFromRequest($this->request, $classloader, 'prod');
+      }
     }
     else {
-      $this->kernel = DrushDrupalKernel::createFromRequest($this->request, $classloader, 'prod', DRUPAL_ROOT);
+      if (!$drupal_update) {
+        $this->kernel = DrushDrupalKernel::createFromRequest($this->request, $classloader, 'prod');
+      }
+      else {
+        $this->kernel = DrushDrupalUpdateKernel::createFromRequest($this->request, $classloader, 'prod');
```
greg-1-anderson: Is `Drupal\Core\Update\UpdateKernel` a new addition to Drupal 8, or has it been around for a while? If there isn't a version of Drupal that has an `UpdateKernel` that does not have an `addServiceModifier()` method, then we don't need this adapter. Since Drush has not supported this capability at all until now, I am wondering if we can perhaps save ourselves an extra adapter and just fail here for old versions of Drupal.

hchonov: The UpdateKernel is part of Drupal since the 8.0.0 release, but the DrupalKernel and the UpdateKernel both don't have an addServiceModifier() method. I have no idea why drush is implementing an adapter to add this method and add a service modifier through it. The DrupalKernel way of adding service modifiers is through A) defining in a module ModuleNameServiceProvider class or B) putting the class or the service object under $GLOBALS['conf']['container_service_providers'].

lib/Drush/Boot/BaseBoot.php
hchonov: You've beat me here .. I am not familiar with the drush architecture that well and those were my first baby steps .. So I guess probably someone else could take a deeper look into it?

lib/Drush/Boot/DrupalBoot8.php
hchonov: I think this would be better yes.

weitzman:
See #3206

PR #2628, opened 2017-02-20
Fix issue #2622: Update the timestamps of files and directories of git checkouts to the latest commit timestamp during make

hchonov:
Instead of leaving the timestamps of files and directories to the current time set the timestamps to the timestamp of the latest git commit from the cloned git repository.

### DIFF
--- a/commands/make/make.download.inc
+++ b/commands/make/make.download.inc
@@ -353,6 +353,10 @@ function make_download_git($name, $type, $download, $download_location) {
     chdir($cwd);
   }
 
+  // Set the timestamps of files and directories to the latest commit timestamp
+  // of the git checkout directory.
+  drush_update_git_checkout_timestamps($tmp_location);
+
   // Move the directory into the final resting location.
   drush_copy_dir($tmp_location, $download_location, FILE_EXISTS_OVERWRITE);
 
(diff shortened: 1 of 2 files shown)


weitzman:
Can you elaborate on why this is needed?

hchonov:
Because when we build the project in a temporary location and then sync it with the destination all the files will have to be transported and synced because on each subsequent build even without changes in the make file the files will now have new timestamps from the time when the git clone has been made, it is just because of git that it does not keep the modification times of the files. Here we have a difference between git download and a normal download, as the normal download retains the timestamps. 
I use the same syncing process for local development as well and after drush make my IDE freezes until it has reindexed everything, so sometimes I wait longer for my IDE to reindex everything than I had to wait for drush make to finish ... 
But if we use the timestamp of the latest git commit then each build will have the same timestamp for the files and only the patched files will have the timestamp of the time when the patch has been applied and so only the patched files will have to be transported or/and reindexed.

hchonov:
I hope that this makes sense for you or at least to make it possible through an option of drush make? This will be fine for me as well.

weitzman:
Make is not under active development anymore. Obvious bug fixes only, and I don't think this qualifies. Feel free to post-process the build as needed.

hchonov:
But there are still a lot of projects which are going to use drush make for a long time until they are ready to switch to composer and covering the case with the timestamps is big performance improvement for local development and syncing projects across servers so that I think it is a bug in make that when I sync two builds I have to update all the files without any real changes.

Issue #2197, opened 2016-05-26
Add config-export option to skip hashes


colans:
On exporting configuration for use in installation profiles, we don't want to see diffs like this in the export:

```php
+uuid: 74fefaf8-cc75-4ba5-af97-8beb8a94708f
 langcode: en
 status: true
 dependencies:
   module:
     - taxonomy
 _core:
-  default_config_hash: '-PPKjsNQPvoIDjOuUAvlLocYD976MNjb9Zpgyz5_BWE'
+  default_config_hash: rps99NkTWYwTbGfk-jw_yYH6ytj8ngYdrvCVaeVASR8
```

So instead of having to remove the UUID lines, and the *default_config_hash* changes, let's simply stop these from being exported by adding a `--skip-hashes` option.

The [Features](https://www.drupal.org/project/features) module has already done [something similar](https://www.drupal.org/node/2668526).

weitzman:
a PR would be welcome. Also would be nice to know why core added this key (so we can evaluate when to omit it).

greg-1-anderson:
I think that the hashes are included so that the configuration manager can differentiate between `delete-and-add-another` from `update`. Features an installation profiles don't need this included. I suspect it must be up to the user to decide whether or not to omit the hashes.

PR #2630, opened 2017-02-20
config-export --skip-hashes

nevergone:
Issue #2197


greg-1-anderson:
+1 LGTM; tests sure would be nice.

weitzman:
I'd like to see how the config_split integration finalizes before merging this.

bircher:
config_split will work as long as the read/write interaction happens with the `config.storage.sync`, so it would work too.
But of course this functionality could also be achieved with a ConfigFilter from the config_filter module (which is a dependency for the next config_split release)

weitzman:
Lets rely on config_filter instead of coding it here. Hope thats OK.

PR #2631, opened 2017-02-21
Adjust 'site-install' command for better use in Continuous Integration

greg-1-anderson:
This PR makes two behavior changes to the site-install command to improve its usability with Continuous Integration servers.

## Hide account password

If the CI server creates a Drupal site that persists after the test suite runs, then it is undesirable to have the uid1 password visible in the test logs. With this PR:

- The account password is by default hidden if it is provided on the commandline. Usually, users will not need to know the password if they already explicitly specified what it should be. If a CI script stores the account password in an environment variable, then it may be passed via `drush si --account-pass=$PASS` without exposing the value in the CI logs.
- The option `--show-passwords` / `--show-passwords=0` is available to override the default.

## Make empty --config-dir a warning

Currently, `drush site-install --config-dir=../config` will fail if the configuration directory is empty.

Use case: in a Drupal site template that includes CI scripts to test Drupal, it is desirable to import the configuration files if they are available.  If the template does not handle this automatically, then the user must know where in the script to add the `--config-dir` option once they have exported their configuration. With this change, the template may simply include --config-dir, and everything will start working correctly once the configuration has been exported. Without this change, the template file could still effect this same behavior by conditionally providing `--config-dir`, but the cost is more conditionals in the CI script.

### DIFF
--- a/lib/Drush/Commands/core/SiteInstallCommands.php
+++ b/lib/Drush/Commands/core/SiteInstallCommands.php
@@ -27,7 +27,6 @@ class SiteInstallCommands extends DrushCommands {
    * @option site-mail From: for system mailings. Defaults to admin@example.com
    * @option sites-subdir Name of directory under 'sites' which should be created. Only needed when the subdirectory does not already exist. Defaults to 'default'
    * @option config-dir A path pointing to a full set of configuration which should be imported after installation.
-   * @option show-passwords Show uid1 account and password once installation completes. Defaults to 'false' if --account-pass is provided.
    * @usage drush site-install expert --locale=uk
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Also, if this is okay for `master`, I'd also like to backport it to the 8.x branch. The config behavior affects an error condition, and therefore does not change B/C. The password display change will not affect scripts, and for users not providing --account-pass, there will be no change in behavior. Seems like a minimal impact to users, and an improvement to CI to hide the password when provided as an option.

weitzman:
The --config-dir warning could be moved to validate(), no? We areal do some config-dir validation there. 

--show-passwords is already a global option. Its going to be very rarely needed here so lets just not declare it for site-install.

PR #2643, opened 2017-02-27
Use annotated-command cache.

### LINKED ISSUE #2608: Adopt command caching from Annotated-Command
Per https://github.com/consolidation/annotated-command/pull/72

greg-1-anderson:
In my testing, this seems to save about 0.02s when using Drush core commandfiles. The difference for Terminus was much greater, though, so I think it really depends on what sort of code you have in your commandfiles.

This cache code does a good job at noticing when methods are added or removed from a commandfile, so there's not much downside to using it. There could be some potential upside if some external commandfiles contain more complicated implementations that take longer to parse.

Fixed #2608.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "league/container": "~2",
     "consolidation/robo": "~1",
     "symfony/config": "~2.2",
-    "consolidation/annotated-command": "^2",
+    "consolidation/annotated-command": "^2.4.2",
     "consolidation/output-formatters": "~3",
     "symfony/yaml": "~2.3",
     "symfony/var-dumper": "~2.7",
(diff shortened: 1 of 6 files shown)


greg-1-anderson:
Looks like tests are broken on the `master` branch. I'll rebase here when those are fixed.

greg-1-anderson:
It seems that Terminus isn't really benefiting from the cache any longer either. A while ago, Terminus depended on a component that was holding it back to Symfony 2.x; this has been cleared up, and Symfony and other dependencies are now being brought in at a higher version. It could be that the cache was only compensating with slow code in an older component.

weitzman:
FYI, make has been removed (#2644), and thus tests are green again.

greg-1-anderson:
Still ambivalent about whether we should merge this or not; at a minimum, we should take ca5e073

greg-1-anderson:
Let's just merge this, so that it can be kept "fresh" by being active in our unit tests. It's easy enough to disable or remove later if we decide it is not needed.

PR #2651, opened 2017-03-04
Unish builds Site-Under-Test via Composer

weitzman:
pm-download and its release and version control engines will be removed as a follow-up in #2654.


includes/preflight.inc
```diff
@@ -775,9 +776,9 @@ function _drush_find_commandfiles_drush() {
 
   // Also discover Drush's own annotation commands.
   $discovery = annotationcommand_adapter_get_discovery();
-  $annotation_commandfiles = $discovery->discover(DRUSH_BASE_PATH . '/lib/Drush', '\Drush');
+  $annotation_commandfiles = $discovery->discover(DRUSH_BASE_PATH . '/src', '\Drush');
```
greg-1-anderson: Did I forget to do this when I relocated everything to /src?  If so, why didn't everything break? Odd.

weitzman:
Ready for review.

Some highlights:
1. phpunit.sut.php. This script builds a drupal-project style codebase using Composer. This is called the System Under test (SUT).
1. unish.phpunit.php. This script runs `phpunit` against the SUT above.
1. phpunit.clean.php. For convenience, this script calls each of the scripts above.
1. Travis uses the top two scripts above to complete its run. This is good "documentation" on how the system works.
1. $this->drush still calls exec() in order to run Drush against the SUT. A follow-up can change this to execute Annotated Command classes directly.
1. pm-download is now removed from the test suite, and will be removed in a followup
1. a follow-up can move the code in tests/bootstrap.inc to class properties and getters on UnishTestCase

PR #2663, opened 2017-03-09
Fix #2658: Run invalidateContainer() only when needed

### LINKED ISSUE #2658: $this->kernel->invalidateContainer(); increases Drush/Drupal bootstrap by 300%
I was debugging Drush 8.1.10 bootstrap to figure out why it works slower then before even on basic operations like `drush status` and came up to the reason. Now before Drush starts Drupal kernel in \Drush\Boot\DrupalBoot8::bootstrap_drupal_full it invalidates the container making Drupal to build it from scratch.

I'm sure this has been done for a good reason but I can't find appropriate issue. But one thing worries me: it is a **huge** makeweight to every drush run. Look, before we could use Drush w/o this invalidation and it worked just fine. Now we need to rebuild container for every even tiny operation which costs us 250-300% of execution time grow (calculated on `drush status`).

So my question is: do we **really** needed it?

Can't it be at least OPTIONAL and enabled by some `--force-rebuild` global switch? And even without one, we can always do `drush cr` if it's really needed.

OnkelTem:


OnkelTem:
Sorry I'm not a big expert in running tests, so haven't launched them locally :-/

greg-1-anderson:
Only one 'make' test failing here. Probably a spurious failure; I restarted that one test.

greg-1-anderson:
Failed the same way -- although I can download the file on my local system just fine.

Note that `make` has been remove from the `master` branch. Please always start with all PRs against `master`, and then backport to 8.x once merged on master.

weitzman:
I re-ran the last failed job and now the PR is green. Unless @greg-1-anderson objects in the next day or two, I'm inclined to merge this.

greg-1-anderson:
Figured it would eventually turn green.

I am strongly in favor of this change. We should do the same on the `master` branch as well. Code should be very similar.

Issue #2603, opened 2017-02-10
Improve table wrapping


weitzman:
We have unfortunate table wrapping. Examples:

- [ ] drush status (try a bootstrapped site if available)
- [ ] drush help site-install

ccharlton:
And output for `drush`/`drush help` that lists all available commands.

weitzman:
![terminal](https://cloud.githubusercontent.com/assets/7740/25601953/d619c412-2ebd-11e7-8f0b-b2c9484032fe.gif)

My wish for long cells is that they add a new row and the wrapped text starts in the proper cell, not the leftmost cell. The red boxes in this screenshot should be in column 2, not in column 1.

weitzman:
![terminal_ _-bash_ _122x49](https://cloud.githubusercontent.com/assets/7740/25626570/f184c2a4-2f2d-11e7-8acb-c1f13c866465.gif)

greg-1-anderson:
Yes, and the strange thing about the status example ^^ is that there are enough unused characters on each row that the algorithm could have made both columns full-width. This use case, at least, should be easy to fix.

ccharlton:
(This is seriously my current favorite Drush ticket.)

greg-1-anderson:
tl/dnr: Wrapping is hard.

I have a new PR (https://github.com/consolidation/output-formatters/pull/55) that improves wrapping considerably. It fixes the narrow column width in the first column in the `drush status` example, at the expense of reducing the last column to a single character in width (!).

The problem with the current algorithm stems mostly from oddities with the way that PropertyList data types are handled. RowsOfFields seem to be wrapping well, but the wordwrap algorithm is confused about what the headers are in the PropertyList case. They should be `Key` and `Value`, or equivalent (or empty), but instead the entries from the first column are being used as the headers. This is more rational than it sounds -- these items would in fact be the header if the table were laid out horizontally. (PropertyList is essentially a RowsOfFields with a single row, rotated vertically.) There are also problems when the column keys are numeric, as they are in `drush status`. This is probably because various PHP array operations renumber array indexes, which never happens when the keys are strings. Not sure what to do about that -- I don't think I can prohibit numeric column keys. I'll probably have to remap them to strings in the word wrapper.

Oh, and the thing about the output of `drush help` is that it seems that it is not being sent through the word wrapper at all. Not sure why that is.

greg-1-anderson:
[...]

PR #2670, opened 2017-03-13
Fix #2106. Port help commands to Annotated.

### LINKED ISSUE #2106: Help command for annotation classes
We need to implement our own sublcass of Application so that we can provide our own [Application::getDefaultCommands](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Console/Application.php#L919). From there, we can provide our own implementation of the Help command.

The existing Symfony help command is not very extensible.  It uses a [Descriptor](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Console/Descriptor/Descriptor.php) that formats the help sections in code. Each different output format is implemented as a subclass of Descriptor. To change the way help works using this model, we would need to make our own subclass of Descriptor in order to change the help structure, and then copy all of the existing formatters, so that they extended our subclass instead of Symfony's descriptor.

A better way to do it would be to follow the existing Drush model, and have the help command build a data structure containing the structured help contents. This could then be formatted by a formatter.

weitzman:
A start at porting help and helpsingle commands. I'm happy with the output of Console's list command (only gripe is that it wants to list all global options which is a currently long list in our case.)

The output appears to need a colspan for the category titles. How to do this with Consolidated/OutputFormatters? Or should I bypass that and Output via Console's Table class directly? Ping @greg-1-anderson


weitzman:
Now depends on 
- https://github.com/consolidation/output-formatters/pull/52
- https://github.com/consolidation/annotated-command/pull/92

weitzman:
I'm now reusing HelpDocument from Annotated-Command project. Drush is supplying its own formatter for CLI presentation of help. Ping @greg-1-anderson for input on the list below.

Bugs
=========
1.  [ ] We have a legitimate test fail at https://github.com/drush-ops/drush/blob/helpAnnotated/tests/annotatedCommandTest.php#L70. For some reason, when help gets an $application, the woot command is not in there. We are successfully bootstrapping Drupal.
1. [ ] Global options are appearing in command help. I'd rather show a subset and move those to own section. I can hack a way around this in the formatter but not sure if we should incorporate a new section in HelpAlter
1. [ ] Table wrapping is poor in both list and help commands. Related #2603.

Nice To Have
=============
1. [ ] HelpDocument doesn't supply the default value for arguments. I think this nice improvement over Console help.
1. [ ] `help` isn't yet making use of the description and the "long description" of a command.

PR #2672, opened 2017-03-15
Update symfony to 3.2

ms-slasher13:
update of pull request https://github.com/drush-ops/drush/pull/2477

see https://www.drupal.org/node/2712647#comment-11981323

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,14 +33,14 @@
     "psy/psysh": "~0.6",
     "league/container": "~2",
     "consolidation/robo": "~1",
-    "symfony/config": "~2.2",
+    "symfony/config": "^3.2",
     "consolidation/annotated-command": "^2.4.2",
     "consolidation/output-formatters": "~3",
-    "symfony/yaml": "~2.3",
-    "symfony/var-dumper": "~2.7",
-    "symfony/console": "~2.7",
-    "symfony/event-dispatcher": "~2.7",
-    "symfony/finder": "~2.7",
+    "symfony/yaml": "^3.2",
+    "symfony/var-dumper": "^3.2",
+    "symfony/console": "^3.2",
+    "symfony/event-dispatcher": "^3.2",
+    "symfony/finder": "^3.2",
     "pear/console_table": "~1.3.0",
     "phpdocumentor/reflection-docblock": "^2.0",
     "webmozart/path-util": "~2",
@@ -49,7 +49,7 @@
   "require-dev": {
     "lox/xhprof": "dev-master",
     "phpunit/phpunit": "^4",
-    "symfony/process": "2.7.*"
+    "symfony/process": "^3.2"
   },
   "suggest": {
       "ext-pcntl": "*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "546d303006c9073ab6f5238326b266e5",
+    "content-hash": "fcca24bb200af96c9b6a8edad45e6b0f",
     "packages": [
         {
             "name": "consolidation/annotated-command",
@@ -808,24 +808,24 @@
         },
         {
             "name": "symfony/config",
-            "version": "v2.8.18",
+            "version": "v3.2.6",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/config.git",
-                "reference": "06ce6bb46c24963ec09323da45d0f4f85d3cecd2"
+                "reference": "741d6d4cd1414d67d48eb71aba6072b46ba740c2"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/config/zipball/06ce6bb46c24963ec09323da45d0f4f85d3cecd2",
-                "reference": "06ce6bb46c24963ec09323da45d0f4f85d3cecd2",
+                "url": "https://api.github.com/repos/symfony/config/zipball/741d6d4cd1414d67d48eb71aba6072b46ba740c2",
+                "reference": "741d6d4cd1414d67d48eb71aba6072b46ba740c2",
                 "shasum": ""
             },
             "require": {
-                "php": ">=5.3.9",
-                "symfony/filesystem": "~2.3|~3.0.0"
+                "php": ">=5.5.9",
+                "symfony/filesystem": "~2.8|~3.0"
             },
             "require-dev": {
-                "symfony/yaml": "~2.7|~3.0.0"
+                "symfony/yaml": "~3.0"
             },
             "suggest": {
                 "symfony/yaml": "To use the yaml reference dumper"
@@ -833,7 +833,7 @@
             "type": "library",
             "extra": {
                 "branch-alias": {
-                    "dev-master": "2.8-dev"
+                    "dev-master": "3.2-dev"
                 }
             },
             "autoload": {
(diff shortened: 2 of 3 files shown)


alexpott:
@ms-slasher13 can you confirm that drush site-install works on both 8.3.x and 8.4.x + https://www.drupal.org/node/2712647 with this applied?

jibran:
@weitzman it more than 24 hrs now as you said in https://github.com/drush-ops/drush/pull/2477#issuecomment-287679510.
No pressure ;-)

weitzman:
I don't know when to merge this, but it isn't now. It isn't until Drupal 8.4 that we are going to need that. Therefore we have some time before Drush needs to deal with it. Once we do this, master won't work with anything but an unreleased drupal and thats non-good.

My comment on the other issue happened when I mistakenly thought Drupal 8.3 was on Symphony 3.1.

alexpott:
It's tricky. We are planning to commit the Symfony 3.x patch to 8.4.x asap. 8.3.x still works with this because of the autoloader fixes to Drupal 8 - done in https://www.drupal.org/node/2843828 - that was committed to 8.2.x too. So all supported versions of Drupal 8 should work with this patch. At least for site-install.

greg-1-anderson:
If worse came to worse, maybe we could get rid of the DrushInputAdapter and just compose an ArrayInput object. That's a very attractive option from a code standpoint; however, I'm not sure that it would be very efficient. We have a lot of global options.

It seems that the tests here are all working, though. I would assume there would be more differences between Symfony 2.8 and 3.2 -- at least, I seemed to remember this being more of a problem in the past, with some function prototypes changing. In general, Symfony 2.x and 3.x are fairly compatible. It's fairly common to see '^2 || ^3' & variants in various Symfony-using components.

+1 to merging this.

weitzman:
Merged. Thanks everyone.

weitzman:
I just reverted this. It turns out that Unish, Drush's test suite, tested this PR with Symphony 2.x and Drupal 8.2. You can see that at https://travis-ci.org/drush-ops/drush/jobs/213928304 (the relevant section is the script unish.sut.php. That script builds a Site-Under-Test which is a Composer assembled codebase just like drupal-project. Since no version of Drupal currently is at Symphony 3.2, that forced the Site-Under-Test to use old versions of Symfony. Any PR here is going to also need to modify the composer.json and composer.lock that define the Site-Under-Test. Those are in our [tests/resources/codebase directory](https://github.com/drush-ops/drush/tree/master/tests/resources/codebase). 

I'm going to add a safeguard so this doesn't happen again, and then work on this PR. This Site-Under_test methodology is brand new so a few bumps like this are to be expected.

greg-1-anderson:
In a way that's good; the tests are working as they should, to test the recommended configuration. We just need to add a Drupal 8.4.x install to the test matrix. This could be done a-la the highest/lowest test methodology, [as the Robo tests do](https://github.com/consolidation/Robo/blob/cccb98ca4ad52688200cfb06f5ab9e9c9f441742/.travis.yml#L28) -- just add a `composer require --no-update drupal/drupal:dev-master` (or similar) to modify the composer.json prior to running `composer update`.

weitzman:
I like that idea. 

My one hesitation is that we currently ship with a composer.lock that defines our SUT. Thats at https://github.com/drush-ops/drush/tree/master/tests/resources/codebase.  We do that in order to speed up our test runs. Each of the parallel test runners has to build a SUT so giving up the lock file will slow down our tests a lot. Adding a `composer require` effectively negates the lock file and Composer's dependency resolution will kick in. Maybe this slowness is just the price we pay for correctness and broad coverage.

greg-1-anderson:
When I commented on this in #2688, my thought was that Drush would not be benefiting much from the composer.lock file, since a lot of its dependencies would be driven by what Drupal requires. However, if there is a sufficient speed increase in installing with the lock file, then we could continue to maintain it. The thing to do would be to populate the composer.lock file with the dependencies from the most common version of Drupal under test. When testing this default version, then `composer require` and `composer update` are not called. See the Robo tests I linked to above.

Another alternative would be to maintain multiple composer.lock files. This might be useful if we wanted to test both a stable version of 8.2.x and a stable version of 8.3.x. Maintaining the lock file is not useful for dev versions of Drupal, though, as we should expect that dependencies may change at any time there, so we should definitely ensure that an update happens under the `highest` scenario. It's likely that we only care about one stable version and dev, though, in which case there is no point to maintaining two lock files. But if we did, we'd just call it something like `tests/composer.lock-^8.3`, and move it into place in the root directory whenever `tests/composer.lock-$DRUPAL_VERSION` exists.

weitzman:
Some pre-work for this issue is going on in #2690. Folks are welcome to help there.

weitzman:
I'll note that someone needs to prepare a git branch with Drupal 8.4.x and the patch applied on top. It is doesn't work for cweagans/composer-patches to [apply the patch during composer install](https://github.com/cweagans/composer-patches#patches-containing-modifications-to-composerjson-files)

ms-slasher13:
done https://github.com/ms-slasher13/drupal/tree/symfony32
```
    "repositories": [
	    {
            "type": "vcs",
            "url": "https://github.com/ms-slasher13/drupal"
        }
    ],
    "require": {
        "drupal/drupal": "symfony32"
    },
```

weitzman:
thanks @ms-slasher13. 

unfortunately, [the package used by Drush's test suite](https://github.com/drush-ops/drush/blob/master/tests/resources/codebase/composer-highest.json) is drupal/core, not drupal/drupal. See [drupal/core+8.4.x](https://github.com/drupal-composer/drupal-core/tree/8.4.x). If you could fork and patch that, I think we can proceed quickly.

ms-slasher13:
Sorry, it's only a typo. It's "drupal/core". Please use
```
    "repositories": [
	    {
            "type": "vcs",
            "url": "https://github.com/ms-slasher13/drupal"
        }
    ],
    "require": {
        "drupal/core": "symfony32"
    },
```

weitzman:
Your first code built fine but i ended up with drupal in the vendor dir, not under /web.

Your second code doesn't build for me. I think you have to name your repo 'core' instead of drupal or else drupal/core won't match the custom composer repo. or maybe an alias can do it. anyway, see https://github.com/drush-ops/drush/pull/2692 for the code i am using. you can clone that branch and run `COMPOSER=composer-highest.json ./unish.sut.php'` to get a working build. then `./unish.clean.php` builds and runs all tests.

```
~/reps/drush (symfony-32 *=) $ COMPOSER=composer-highest.json ./unish.sut.php
Executing: composer  install --no-interaction --no-progress --no-suggest --working-dir '/Users/moshe.weitzman/tmp/unish/drush-sut'
Loading composer repositories with package information
Updating dependencies (including require-dev)           
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - The requested package drupal/core dev-symfony32 exists as drupal/core[8.0.0, 8.0.0-beta10, 8.0.0-beta11, 8.0.0-beta12, 8.0.0-beta13, 8.0.0-beta14, 8.0.0-beta15, 8.0.0-beta16, 8.0.0-beta6, 8.0.0-beta7, 8.0.0-beta8, 8.0.0-beta9, 8.0.0-rc1, 8.0.0-rc2, 8.0.0-rc3, 8.0.0-rc4, 8.0.1, 8.0.2, 8.0.3, 8.0.4, 8.0.5, 8.0.6, 8.0.x-dev, 8.1.0, 8.1.0-beta1, 8.1.0-beta2, 8.1.0-rc1, 8.1.1, 8.1.10, 8.1.2, 8.1.3, 8.1.4, 8.1.5, 8.1.6, 8.1.7, 8.1.8, 8.1.9, 8.1.x-dev, 8.2.0, 8.2.0-beta1, 8.2.0-beta2, 8.2.0-beta3, 8.2.0-rc1, 8.2.0-rc2, 8.2.1, 8.2.2, 8.2.3, 8.2.4, 8.2.5, 8.2.6, 8.2.7, 8.2.x-dev, 8.3.0-alpha1, 8.3.0-beta1, 8.3.0-rc1, 8.3.0-rc2, 8.3.x-dev, 8.4.x-dev, dev-master] but these are rejected by your constraint.
```

weitzman:
Lets close this PR and discuss #2692.

PR #2675, opened 2017-03-16
Adopt form changes for update notification during site installation (8.x version)

pfrenssen:
In [Issue #2691971: "Update Notifications" #title repeated for drupal 8 installation in configure site page](https://www.drupal.org/node/2691971) the form got reworked that allows to configure whether notifications should be sent about module updates during site installation. This will soon be released with Drupal 8.3.0.

This PR brings drush back in-line with Drupal.

This is a version that is compatible with the 8.x branch. I will also make a version that works with the master branch.


weitzman:
Is this backward compatible with Drupal 8.2? If not I think we have to hold off on Drush 8.x branch at least until 8.3.0 is released.

pfrenssen:
Technically it is backwards compatible, the old form fields will fall back to their default values so it will work fine, but it will show the wrong help message which can be confusing.

I would also prefer to wait until 8.3.0 is released before merging this. The people that are already on 8.3.0-rcX can use this PR for the time being.

PR #2676, opened 2017-03-16
Adapt form changes for update notification during site installation (master branch)

pfrenssen:
In [Issue #2691971: "Update Notifications" #title repeated for drupal 8 installation in configure site page](https://www.drupal.org/node/2691971) the form got reworked that allows to configure whether notifications should be sent about module updates during site installation. This will soon be released with Drupal 8.3.0.

This PR brings drush back in-line with Drupal.

This is a version that is compatible with the master branch. There is also a sister PR for the 8.x branch in #2675.


pfrenssen:
It would be best to wait until Drupal 8.3.0 is released before merging this in so people will not be confronted with an incorrect help text.

PR #2684, opened 2017-03-20
Remove --skip-modules from 'cim' and 'cex'

bircher:
And remove storage filters and all the other related code in favor of using [Config Split](https://www.drupal.org/project/config_split)


docs/config-exporting.md
```diff
@@ -33,29 +33,5 @@ you can use Drush configuration filters.
...
-## More Complex Adjustments
-
-Drush allows more complex changes to the configuration data to be made
-via the configuration filter mechanism.  In order to do this, you must
-write some code inside a Drush extension.
-
-See [Drupal Configuration Filtering](config-filter.md) for more information
-on how to do this.
+Use the [Config Split](https://www.drupal.org/project/config_split) module to
+split off development configuration in a dedicated config directory.
```
bircher: We might want to write a better documentation here, and refer to Config Filter for more advanced use cases.

tests/configTest.php
```diff
@@ -59,30 +59,11 @@ function testConfigExportImport() {
     $page = $this->getOutputFromJSON('system.site:page');
     $this->assertContains('unish partial', $page->front, '--partial was successfully imported.');
 
+    // @TODO: this may be irrelevant now.
```
bircher: review this if tests all pass.

src/Commands/config/ConfigExportCommands.php
```diff
@@ -3,28 +3,26 @@
...
+use Consolidation\AnnotatedCommand\Events\CustomEventAwareTrait;
 use Drupal\Core\Config\StorageComparer;
 use Drupal\Core\Config\FileStorage;
 use Drush\Config\StorageWrapper;
 use Drush\Commands\DrushCommands;
 
 class ConfigExportCommands extends DrushCommands implements CustomEventAwareInterface {
 
-  use ConfigTrait;
+  use CustomEventAwareTrait;
```
weitzman: We can now remove all lines that mention CustomEvent in this file. Same for config-import

greg-1-anderson:
This never worked perfectly; +1 for removal

greg-1-anderson:
We might also want to consider backporting removal to the 8.x branch as a "bug fix". Would need to weigh the odds that it works well enough for anyone to be upset by its removal vs. the time folks might waste struggling with its deficiencies if it is left in.

bircher:
Yes I agree to remove it also from 8.x. it never worked and I think it is worth removing it.
I will make a survey by show-of-hands in the CM session at the drupaldevdays.
We should also agree on what to say in the documentation, it is quite sparse right now.

LightGuard:
While I appreciate the "this never really worked right anyway" argument, removing an option in a patch release should not have happened. If you wanted to NO-OP it, great, but to completely pull it?  Please don't. I would have expected a change like this in a minor or major version bump, but not a patch.

PR #2686, opened 2017-03-21
Remove --skip-modules from cim and cex in 8.x. (Backport #2684)

bircher:


commands/core/config.drush.inc
```diff
@@ -109,10 +109,8 @@ function config_drush_command() {
         'example-value' => 'branchname',
       ),
       'destination' => 'An arbitrary directory that should receive the exported files. An alternative to label argument.',
-      'skip-modules' => 'A list of modules to ignore during export (e.g. to avoid listing dev-only modules in exported configuration).',
```
greg-1-anderson: Maybe in the 8.x branch, we should leave the flag definition in place. This will give us a place to point at config-split, and also prevent breaking scripts with `--skip-modules=0`. If someone runs with `--skip-modules`, it should log a warning rather than failing.

weitzman: That makes sense if the value was 0. How frequent is that though? If the value actually listed modules, it makes sense to error out and ideally discuss config-split. All in all, I'm OK with this PR as is.

greg-1-anderson: Sounds fine to me.

weitzman:
Folks should use the config_split module instead.

PR #2695, opened 2017-03-27
Remove old outputformat engine and the engine system itself!

weitzman:


greg-1-anderson:
Yay! 👍

claudiu-cristea:
I used a lot `drush_print_r()` for inspecting variables. What is the best way to do this now?

PR #2697, opened 2017-03-27
Don\'t add profile dependencies to required_by array of modules

balsama:
Using the patch in [D.O Issue#1356276](https://www.drupal.org/node/1356276), it's possible for more than one profile to be installed. Dependencies of profiles have never prevented a module from being uninstalled, but the logic used in pm-uninstall only makes that exception for the active profile.

This patch completely removes profiles from the `required_by` key in the modules list provided by `drush_get_modules()`. If the patch in D.O Issue#1356276 is applied, it uses the `profile_handler` service to get an array of installed modules. Otherwise, it uses `drupal_get_profile()` to emulate that behavior.

### DIFF
--- a/commands/core/drupal/environment.inc
+++ b/commands/core/drupal/environment.inc
@@ -23,8 +23,20 @@ use Drupal\Core\Logger\RfcLogLevel;
  */
 function drush_get_modules($include_hidden = TRUE) {
   $modules = system_rebuild_module_data();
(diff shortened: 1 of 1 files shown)


weitzman:
Looks reasonable at first glance. We need a patch for master branch before this gets merged. Master branch is much simpler. See https://github.com/drush-ops/drush/blob/master/src/Commands/PmCommands.php

PR #2692, opened 2017-03-26
Test with a patched Drupal 8.x.4 (Symphony 3.2)

weitzman:

PR #2705, opened 2017-04-04
Test with Symfony 3.2

webflo:
Fix for https://github.com/drush-ops/drush/pull/2692


webflo:
@weitzman Not sure why it fails but i build it on my local system and it worked.

weitzman:
I got this working in #2724. Lets discuss there.

PR #2707, opened 2017-04-06
Fix bug in updatedb

alexpott:
The batch uses drush_update_cache_rebuild() whereas core via the UI uses drupal_flush_all_caches().

This causes reproducible bugs, steps to reproduce:

1. git co 8.1.1
2. composer install
3. drush si minimal
4. drush en hal config -y
5. drush config-edit rest.settings and change the link_domain - it actually not important but it is what I did.
6. git co 8.3.0
7. composer install
8. drush updatedb -y
9. BOOM

### DIFF
--- a/commands/core/drupal/update.inc
+++ b/commands/core/drupal/update.inc
@@ -231,7 +231,7 @@ function drush_update_batch() {
   // Apply post update hooks.
   $post_updates = \Drupal::service('update.post_update_registry')->getPendingUpdateFunctions();
   if ($post_updates) {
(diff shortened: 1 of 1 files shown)


weitzman:
We just fixed a bug in this area. See https://github.com/drush-ops/drush/issues/2617. 

Anyway, all drush_update_cache_rebuild() does is call flush_caches() so I'm not sure what this helps. Maybe you were using an older version of drush?

Ping @hchonov and @tstoeckler for comments/review.

alexpott:
Ah you are right I was using an outdated version

Issue #1427, opened 2015-06-03
drush updb does not respect hook_requirements


czigor:
On drush 7.0.0 and drupal 7 I have a module A with a hook_update_N() enabling a module B. Both enabling module B and running the rest of the updates are crucial for the further operation of module A so I created a hook_requirements checking if module B can be enabled (i.e. the module_enable() returns FALSE or not). However, running 'drush updb' ignores hook_requirements(). (Running updates from a browser does not.)

The problem is that update_check_requirements() in commands/core/drupal/update_7.inc explicitly states:
  // Check the system module and update.php requirements only.

Is this by design? If so, what is the recommended way to run such updates with drush?

DanChadwick:
I'm using drush 6.3.0 and have the above issue and another related issue.

1) drush updatedb does not call hook_requirements the way that executing update.php via the UI does. update.php prevents the update in the presence of errors of severity REQUIREMENT_ERROR.

2) drush updatedb further does not enable dependent modules declared in the modules.info file. The update.php script does.

Steps to reproduce:
1) Install and enable webform 7.x-3.x. Ensure that views is NOT installed. 
2) Install the code base for webform 7.x-4.x.
3) Run drush updatedb. Note that views is NOT installed and enabled.

Repeat with update.php. The .info file is sufficient to cause update.php to install and enable views.

If I create a requirement for views in hook_requirements, then update.php will refuse to run at step 3. drush updatedb will however execute the updates, ignoring hook_requirements.

alexpott:
I just noticed this today. Totally agree this is a bug.

weitzman:
Lets close in favor of #2708

PR #2708, opened 2017-04-07
Check requirements before running updatedb

alexpott:
See https://github.com/drush-ops/drush/issues/1427

### DIFF
--- a/commands/core/drupal/update.inc
+++ b/commands/core/drupal/update.inc
@@ -130,6 +130,11 @@ function update_main() {
   drupal_load_updates();
   update_fix_compatibility();
 
+  // Check requirements before updating.
+  if (!drush_update_check_requirements()) {
+    return;
+  }
+
   // Pending hook_update_N() implementations.
   $pending = update_get_update_list();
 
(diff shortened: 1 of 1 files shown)


commands/core/drupal/update.inc
```diff
@@ -130,6 +130,11 @@ function update_main() {
   drupal_load_updates();
   update_fix_compatibility();
 
+  // Check requirements before updating.
+  if (!drush_update_check_requirements()) {
```
damiankloip: Maybe this should be a validation callback instead, so it's run before we hit the actual command? Then `drush_update_check_requirements ` can just be used directly or the validation command can call out to that as the code is now.

weitzman:
I get that update.php does this, but updatebd doesnt have to do everything that update.php does. updatedb lacks tests and i'm loath to put more weight on it without tests.

alexpott:
The problem is that drush will run updates when core would not.

commands/core/drupal/update.inc
alexpott: Well then we have to go through the rigmarole of loading all the .install files in the validation

weitzman:
Right. Some other other command could validate requirements and then careful folks would chain calls to that command and then updatedb. Just saying that updatedb doesn't have to do it all.

weitzman:
Merged into 8.x and master in order to help media entity contrib project.

marcoscano:
Thanks!

mootari:
When is the next release due?

gideoncresswell:
This still does not work on a Drupal 7 site. The `update_7.inc` file runs the `update_check_requirements()` function but doesn't check if that returns TRUE or FALSE. And it doesn't call any `hook_requirements()` at all - just `system_requirements()`.

Issue https://github.com/drush-ops/drush/issues/1427 was closed in favour of this ticket, but the bug is still present.

Issue #2710, opened 2017-04-13
Add a cache clear type for page cache


gh640:
I'd like to add an option to clear page cache.

(This is my first attempt to create an issue in this repository and if I'm wrong, I'm sorry.)

gh640:
I read the contributing guide https://github.com/drush-ops/drush/blob/master/CONTRIBUTING.md but I couldn't understand how to "simply open an issue to be added as an Outside Collaborator". 

How should I do that? Is it just to create an issue labeled like "Please add me as an outside collaborator"?

weitzman:
Welcome! Thats not needed. I just removed that line from the docs.

weitzman:
As for page cache, I'd like to not add new cache types. Feel free to add new ones from outside cache.drush.inc. See hows Views does it in Drush core.

gh640:
@weitzman thanks for your response and explanation. I'll try to anticipate in other issues.

PR #2711, opened 2017-04-13
Issue #2710: Add a cache clear option for page cache.

gh640:


weitzman:
I'm not inclined to add more granular caches in core drush. feel free to add them from contrib commandfiles.

Issue #2709, opened 2017-04-12
Errors and exceptions not reported consistently during config-import


danepowell:
On our site, we are using Lightning Media, which provides a few fields in its /config/install directory, such as field.field.media.document.field_media_in_library. Whenever this module is installed, the uuid for this field is going to be set to a random value.

We have previously installed this module and exported the configuration to our sites /config/default directory. Note that this exported configuration is identical to what's provided by Lightning Media, _except_ our version of the field now has a uuid, as expected.

The problem comes when we install the site with the config-dir argument, i.e. `drush si --config-dir=config/default`. I would expect this to delete and then recreate the media_in_library field (because the uuid has changed from the default config). And indeed, the logs indicate that this is what _appears_ to happen:

> Synchronized configuration: delete field.field.media.document.field_media_in_library. [ok]
..
> Synchronized configuration: create field.field.media.document.field_media_in_library. [ok]

However, despite what's indicated in the log, the media_in_library field doesn't exist on the site. It doesn't appear that it was ever actually recreated.

If we subsequently run a config-import, the field does finally get created successfully. But this shouldn't be necessary--I would expect Drush to handle it correctly as part of the initial config sync.

weitzman:
Actually processing the incoming configuration is the responsibility of Drupal core. The log messages you see are just Drush parrotting out what core is reporting. If the results seem wrong to you, you'll have to dig into the core sync service. Follow the money ...

danepowell:
I tried to eliminate Drush from the picture using these steps:

1. Install the site without the config-dir argument
2. Observe that the media_in_library field exists, as expected (but with a randomized UUID)
3. Try to import ("synchronize") configuration from config/default via the Drupal UI.
4. Observe that the import crashes with an exception like this:
> A fatal error occurred: Exception thrown while performing a schema update. Cannot rename media__field_mime_type to field_deleted_data_14304c55cb: table field_deleted_data_14304c55cb already exists.

I suspect that if the import were ever to make it as far as importing field_media_in_library, it would crash with the same exception.

I suspect what's happening here is that Drupal has a major core bug that prevents it from deleting and recreating a field in the same process without an interim garbage collection step. I found and posted in this [core issue](https://www.drupal.org/node/2622160#comment-12038215) that seems to be just such a bug.

[...]

PR #2713, opened 2017-04-18
Issue #2709: Fixed non-reporting of config-import errors.

danepowell:
Fixes (at least partially) #2709 

If an error occurs during config import, Drupal will catch it internally and then clear the exception so that the import proceeds gracefully: http://cgit.drupalcode.org/drupal/tree/core/lib/Drupal/Core/Config/ConfigImporter.php#n762

This means that Drush can't assume that an import was successful just because an import finished without throwing an exception. It must check if any errors were reported.


weitzman:
Merged into 8.x and master. Thanks for hunting this down, Dane.

Issue #1811, opened 2015-11-27
Create drush config-status command


Chi-teck:
I am looking for a way to display overridden configuration. It is something similar to `drush fl` in Drupal 7. At the moment it is possible with `drush cex` or `drush cim` commands but you have to cancel the export / import actions each time. So that's not convenient. Perhaps we can just add some options to `drush config-list` command.

Chi-teck:
Well `drush cex -n` fit my needs.

greg-1-anderson:
I think a config status command is still a good idea.

weitzman:
I think its a good idea as well. The code would like similar to the way pm-updatecode embeds a call to pm-updatestatus and updatedb embeds updatedb-status. This could also bring a confirmation to config-pull in the same way that the confirm happens after pm-updatestatus call

danepowell:
Note that `drush cex -n` does not work on read-only file systems (such as Acquia Cloud). So we do need some other way of generating a list of overridden configuration.

I was thinking either a "dry-run" option for cex that does not try to write to the file system, or a separate config-status command. It sounds like folks would prefer config-status so I might try to whip that up.

weitzman:
For read-only, I suggest the --destination option of config-export or use [config-pull](https://drushcommands.com/drush-8x/config/config-pull/) to bring the config local where you can commit it, etc. [This command was built with Acquia Cloud in mind](https://docs.acquia.com/acquia-cloud/develop/config-d8).

danepowell:
I made a mistake during testing, I think `drush cex -n` runs fine on Acquia Cloud if all you want is a list of configuration changes. Regardless though, a config-status command would be nice. PR imminent.

weitzman:
Lets talk in the PR

PR #2716, opened 2017-04-20
Issue #1811: Added config-status command.

danepowell:
I think there's some overlap between the helper functions I added here and the code in config-import, with a little effort I think config-import could be simplified. But I wanted to limit the amount of refactoring in this PR.

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -209,6 +209,14 @@ function config_drush_command() {
     'topics' => array('docs-aliases', 'docs-config-exporting'),
   );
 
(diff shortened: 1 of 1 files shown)


weitzman:
At first pass this looks great. A prerequisite is a PR for master branch too, if you get a chance. Sorry the code has moved around bit.

danepowell:
Yeah, that's going to have to be a ground-up rewrite :unamused: And I'm going to need to find an environment where I can test Drush HEAD. I'll see what I can do.

weitzman:
FYI this is now in Drush9.

PR #2724, opened 2017-04-26
Console 3.2 Compat and use drupal/core for composer-highest.

weitzman:


weitzman:
Its green, and RTBC IMO. Any last thoughts? Ping @alexpott, @webflo @greg-1-anderson @catch56.

webflo:
Awesome, i will open a follow-up to remove the dependency on webflo/drupal-core-8.4.x once the core patches is committed to 8.4.x.

Thanks!

alexpott:
This has caused https://www.drupal.org/node/2874827 which means that master can't install both 8.3.x and 8.4.x

Drush 8.x can't install 8.4.x atm which is probably okay as long as drush 9 has a release before 8.4.0.

webflo:
@weitzman Its is feasible to backport the changes to drush 8?

weitzman:
I'm hesitant to try to get Drush 8 to work with Drupal 8.4. Many folks use Drush8 via the Phar and then the bundled Symfony classes will be 3.x while the project will be Symfony 3. This is a recipe for trouble.

How about we add conflict statements in the composer.json of Drush 8 and Drush 9 so folks are directed to the right Drush version? 

Drush 9 will definitely have a release before 8.4.0

alexpott:
@weitzman so we just suck it up and say there is no Drush version that'll work with both 8.3.x and 8.4.x? That is going to be painful. At the moment 9.0.0-beta1 does work against both 8.3.x and 8.4.x. 

Imo we should only be making this change once 8.4.0 is out.

weitzman:
Drush master should work with both Drupal 8.3 and 8.4 but you have to use a project-specific instance of Drush, not a global one. Blame the PHP community's Composer-fever if you dont like this. I'm just the messenger :)

PR #2726, opened 2017-04-26
Issue #1811: Added config status command and refactored config commands.

danepowell:
This corresponds to #2716 but against master. For the master branch, it was harder to add this command without significantly refactoring the existing configuration commands (to avoid duplicating quite a bit of code.)

I haven't had a chance to test this yet, but let me know if you have any immediate feedback.

### DIFF
--- a/src/Commands/config/ConfigCommands.php
+++ b/src/Commands/config/ConfigCommands.php
@@ -4,6 +4,7 @@ namespace Drush\Commands\config;
 use Consolidation\AnnotatedCommand\CommandError;
 use Consolidation\AnnotatedCommand\CommandData;
 use Drupal\Core\Config\FileStorage;
+use Drupal\Core\Config\StorageComparer;
 use Drush\Commands\DrushCommands;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\ConsoleOutputInterface;
--- a/src/Commands/config/ConfigCommands.php
+++ b/src/Commands/config/ConfigCommands.php
@@ -181,6 +181,7 @@ class ConfigCommands extends DrushCommands {
    * @command config-status
    * @param $label A config directory label (i.e. a key in \$config_directories array in settings.php).
    * @interact-config-label
+   * @bootstrap DRUSH_BOOTSTRAP_DRUPAL_FULL
    */
   public function status($label = NULL) {
     $directory = $this->getDirectory($label, '');
--- a/src/Commands/config/ConfigCommands.php
+++ b/src/Commands/config/ConfigCommands.php
@@ -175,6 +175,85 @@ class ConfigCommands extends DrushCommands {
   }
 
   /**
+   * Display status of configuration (differences between stored and active configuration).
+   *
+   * @command config-status
+   * @param $label A config directory label (i.e. a key in \$config_directories array in settings.php).
+   * @interact-config-label
+   */
+  public function status($label = NULL) {
+    $directory = $this->getDirectory($label, '');
+    $storage = $this->getStorage($directory);
+    $change_list = $this->getChanges($storage);
+    if (empty($change_list)) {
+      $this->logger()->notice(dt('The active configuration is identical to the configuration in the export directory (!target).', array('!target' => $directory)));
+      return;
+    }
+    drush_print("Differences of the active config to the export directory:\n");
+    ConfigCommands::configChangesTablePrint($change_list);
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


src/Commands/config/ConfigCommands.php
```diff
@@ -175,6 +175,85 @@ public function delete($config_name, $options = []) {
...
+   *
+   * @param string $label
+   *   A configuration directory label.
+   * @param string $directory
+   *   A configuration directory.
+   */
+  function getDirectory($label, $directory) {
+    // If the user provided a directory, use it.
+    if (!empty($directory)) {
+      if ($directory === TRUE) {
```
danepowell: I don't understand how `$directory` could ever be a boolean, but I'm preserving this legacy code path just in case I've missed something.

danepowell:
This should be good, I don't think the failing test is related to this PR. However, I haven't been able to test this since I haven't been able to get a stable local install using the master branch.

src/Commands/config/ConfigCommands.php
weitzman: Yeah we need to keep that code path as its used by config-pull internally.

src/Commands/config/ConfigCommands.php
```diff
@@ -175,6 +176,85 @@ public function delete($config_name, $options = []) {
   }
 
   /**
+   * Display status of configuration (differences between stored and active configuration).
```
weitzman: The word "stored" is not too intuitive here. I think we might want to explain the default case and say "differences between the filesystem configuration and database configuration". Dunno.

src/Commands/config/ConfigCommands.php
```diff
@@ -175,6 +176,85 @@ public function delete($config_name, $options = []) {
...
+   *
+   * @command config-status
+   * @param $label A config directory label (i.e. a key in \$config_directories array in settings.php).
```
weitzman: What about viewing the changelist as a diff? Thats already offerred by config-import. Can be a follow-up.

src/Commands/config/ConfigCommands.php
```diff
@@ -175,6 +176,85 @@ public function delete($config_name, $options = []) {
...
+   * @interact-config-label
```
weitzman: You have to add a @bootstrap annotation like many other commands. Otherwise your command is called unbootstrapped you get the fatal I am seeing now - Fatal error: Call to undefined function config_get_config_directory()

src/Commands/config/ConfigCommands.php
```diff
@@ -175,6 +176,85 @@ public function delete($config_name, $options = []) {
...
+   */
+  public function status($label = NULL) {
```
weitzman: As a follow-up, it would be great to support output formatters in some way for config-status. I think callers will want programmatic access to the $change list.

weitzman:
Needs a reroll since Config commands moved under src//Drupal. Sorry.

danepowell:
All of the rebasing and refactoring has kind of killed my interest in this, sorry. Using `drush config-export -n` feels hacky but is a decent enough workaround for me right now so I'm probably going to continue using that instead.

weitzman:
No worries. Thanks for taking it this far. Many hands make light work.

weitzman:
@bircher - Any chance you could make a config-status command for master? I think you know the code flow very well. I've gotten a few requests for it, most in the context of CI systems. You probably should start fresh, pulling ideas from this PR as needed.

weitzman:
The use case for config:status that I have heard is in a deploy script. Specifically imagine a script that puts site in maintenance mode if there are pending config operations or pending db updates. Otherwise the script just deploys the new code (because its safe to do so). In this use case config:status merely needs to communicate a  binary yes/no.

danepowell:
Note that `drush cex --no` is no longer a tenable workaround in Drush 9, due to #3014. So a config:status command is more of a necessity now (unless it's possible to replicate this with something like `config:export --simulate`, which is currently broken).

I think the most important use case for this is to ensure that configuration fully imports correctly. If configuration remains overridden following an import, it's an indication that the stored configuration is corrupt in some way. See:

- http://blt.readthedocs.io/en/latest/readme/configuration-management/#ensuring-integrity-of-stored-configuration
- https://github.com/acquia/blt/blob/8.9.6/src/Robo/Commands/Setup/ConfigCommand.php#L183

danepowell:
Trying to resolve merge conflicts on this is fairly hopeless, I started from scratch in #3020

Issue #2728, opened 2017-04-30
Notice in core-cli - update for new command list


weitzman:
core-cli command needs to use code below to get list of commands (already categorized). This code comes from ListCommands

```
$application = \Drush::getApplication();
annotation_adapter_add_legacy_commands_to_application($application);
$all = $application->all();
```

damiankloip:
Yes! I saw this recently, then it slipping my mind. Looking into this now.

PR #2729, opened 2017-05-01
Update core-cli command to use new command list

damiankloip:
See https://github.com/drush-ops/drush/issues/2728

We can remove quite a bit of code that was needed before to convert the drush command config arrays into input options etc.. now we already have those object on the original commands, which is nice :)

### DIFF
--- a/src/Drupal/Commands/core/CliCommands.php
+++ b/src/Drupal/Commands/core/CliCommands.php
@@ -82,7 +82,6 @@ class CliCommands extends DrushCommands {
    */
   protected function getDrushCommands() {
     $application = \Drush::getApplication();
-    //annotation_adapter_add_legacy_commands_to_application($application);
     $commands = $application->all();
 
     $ignored_commands = ['help', 'drush-psysh', 'php-eval', 'core-cli', 'php'];
--- a/src/Drupal/Commands/core/CliCommands.php
+++ b/src/Drupal/Commands/core/CliCommands.php
@@ -53,19 +53,8 @@ class CliCommands extends DrushCommands {
     }
 
     // Add Drush commands to the shell.
-    $commands = [new DrushHelpCommand()];
-
-    foreach (drush_commands_categorize($this->getDrushCommands()) as $category_data) {
-      $category_title = (string) $category_data['title'];
-      foreach ($category_data['commands'] as $command_config) {
-        $command = new DrushCommand($command_config);
-        // Set the category label on each.
-        $command->setCategory($category_title);
-        $commands[] = $command;
-      }
-    }
-
-    $shell->addCommands($commands);
+    $shell->addCommands([new DrushHelpCommand()]);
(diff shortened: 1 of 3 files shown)


src/Commands/core/CliCommands.php
```diff
@@ -99,25 +88,34 @@ public function cli(array $options = ['version-history' => FALSE]) {
    * @return array
    */
   protected function getDrushCommands() {
-    $commands = drush_get_commands();
+    $application = \Drush::getApplication();
+    //annotation_adapter_add_legacy_commands_to_application($application);
```
damiankloip: @weitzman using the legacy command adapter here produces a ton of warnings (looks like one foe each command) like:

```
 [warning] array_key_exists() expects parameter 2 to be array, string given annotationcommand_adapter.inc:503
 [warning] Illegal string offset 'default_value' annotationcommand_adapter.inc:504
 [warning] Illegal string offset 'default_value' annotationcommand_adapter.inc:506
 [warning] Illegal string offset 'description' annotationcommand_adapter.inc:506
 [warning] Illegal string offset 'default_value' annotationcommand_adapter.inc:506
```

src/Drupal/Commands/core/CliCommands.php
```diff
@@ -92,25 +81,34 @@ public function cli(array $options = ['version-history' => FALSE]) {
    * @return array
    */
   protected function getDrushCommands() {
-    $commands = drush_get_commands();
+    $application = \Drush::getApplication();
+    //annotation_adapter_add_legacy_commands_to_application($application);
```
damiankloip: @weitzman using the legacy command adapter here produces a ton of warnings (looks like one foe each command) like:

```
 [warning] array_key_exists() expects parameter 2 to be array, string given annotationcommand_adapter.inc:503
 [warning] Illegal string offset 'default_value' annotationcommand_adapter.inc:504
 [warning] Illegal string offset 'default_value' annotationcommand_adapter.inc:506
 [warning] Illegal string offset 'description' annotationcommand_adapter.inc:506
 [warning] Illegal string offset 'default_value' annotationcommand_adapter.inc:506
```

weitzman:
I'm fine with merging this and dealing with the legacy commands some day in the future.

damiankloip:
OK - let's follow up on that. I'm also OK with not really supporting legacy commands, all of our stuff will be converted when drush 9 is out anyway?

damiankloip:
I'll remove the commented out line for now.

damiankloip:
@weitzman there are also still a load of command classes that have not been moved to the `Drush\Commands` namespace, they are still in the `Drush\Drupal\Commands` namespace. Is that intentional?

weitzman:
The \Drupal namespace is new and yes, we deliberately moved commands into there just last week. Those are the core BOOTSTRAP_FULL commands that can get Drupal dependencies injected.

PR #2740, opened 2017-05-04
Add option to disable post updates in drush updatedb.

mbm80:


commands/core/core.drush.inc
```diff
@@ -97,6 +97,7 @@ function core_drush_command() {
     ),
     'options' => array(
       'entity-updates' => 'Run automatic entity schema updates at the end of any update hooks. Defaults to --no-entity-updates.',
+      'post-updates' => 'Run hook_post_update_X() implementations. Defaults to --post-updates.',
```
weitzman: Better to say "Use --post-updates=0 to bypass post-updates.".

Also needs PR for master branch. Thanks.

mbm80: Fixed & add PR https://github.com/drush-ops/drush/pull/2742

fago:
Reason for skipping post-updates: Run a config import first, then do the post-update. Post-updates are for content migration, thus they might need config changes (fields) for being able to run!

weitzman:
Not sure I understand the reasoning. Wouldn't those fields have been changed during the config-import that just ended?

fago:
>Not sure I understand the reasoning. Wouldn't those fields have been changed during the config-import that just ended?

That's the point. Consider an update that needs newly added fields available, e.g., a data migration. The new field must be imported before the update can run.

weitzman:
OK, thanks for clarifying. I understand the problem now. However, once https://www.drupal.org/node/2762235 goes in, it will be impossible to do a config-import without having already run post-updates. Is that patch mistaken?

alexpott:
Well where is the content coming from? It should be coming from a site where the updates have already been run. I'm suspicious of where this has got out-of-sync. And how was the config changed? If post-updates depend on it why is it not being changed in a hook_update_N?

weitzman:
Please reply and I can reopen as needed.

mbm80:
The case we need this for is to set default values in the database for fields which are created (by contrib modules) when config import runs. So basically we run:
drush updatedb --entity-updates --no-post-updates -y && drush cim -y && drush updatedb

In drupal 7 we can feature_revert_module and then update the database in another module, but this is not possible for drupal 8.

fago:
@weitzman Indeed, I think the core issue is mistaken. I replied there.

>Well where is the content coming from? It should be coming from a site where the updates have already been run.
Content is life, e.g. on a production site - which runs older config lacking my new field X.

Now, for example I add my new field X and I need to write an update function which initializes the field values in a certain way. For that to work, I need the field to exist - so the content-update needs to run *after* the config import.

> I'm suspicious of where this has got out-of-sync. And how was the config changed? 
It's just regular development, nothing out of sync.

>If post-updates depend on it why is it not being changed in a hook_update_N?
Yes, that would be possible. Add everything that is needed in an update function. But if more config parts are needed, that would mean replicating the confg sync in an update function. But would I replicate there what the config import already does??

To clarify, I'm talking about project development here. If this is about a contrib module, possibly required changes would have to be done in update functions, yes. But for project development, this is not practical.

weitzman:
@fago @alexpott @bircher - Seems like https://github.com/bircher/drupal-config_import_n is attacking this problem in an organized way.

bircher:
Hello
We use the project @weitzman mentioned above to solve this problem.

I think we should not allow to split the execution of the update and the post_update hooks because then we would "abuse" the post_update hooks for configuration deployment rather than cleaning up the database fixes from the update hook. (separation of concerns and such)
Arguably we should rather add the post_config_import hook to core.

fago:
> Arguably we should rather add the post_config_import hook to core.
I'm not convinced this is needed. Commented at https://www.drupal.org/node/2762235#comment-12113213 - let's discuss there.

PR #2741, opened 2017-05-05
Include improved-wrap PR from output-formatter project.

greg-1-anderson:
Addresses #2603 (partially).

Note that the output from `drush list` and `drush help` are not being processed by the word wrapper at all at the moment, even with this PR. However, other table output should be wrapped much better with this version.


greg-1-anderson:
c.f. https://github.com/consolidation/output-formatters/pull/55

PR #2742, opened 2017-05-05
Add option to disable post updates in drush updatedb.

mbm80:


weitzman:
See discussion in elsewhere. This does not look needed.

PR #2744, opened 2017-05-05
Fix site-install command

webflo:
- Handle exception from drupal_get_profile when container ist not initialized and db connection not defined yet.
- Reimplement distribution detection (reverts 7b609a4)

---

I did some debugging for https://github.com/drupal-composer/drupal-project/pull/270 and found the root cause in commit 7b609a4.

### DIFF
--- a/src/Commands/core/SiteInstallCommands.php
+++ b/src/Commands/core/SiteInstallCommands.php
@@ -2,6 +2,7 @@
 namespace Drush\Commands\core;
 
 use Consolidation\AnnotatedCommand\CommandData;
+use Drupal\Core\Database\ConnectionNotDefinedException;
 use Drush\Commands\DrushCommands;
 use Drush\Exceptions\UserAbortException;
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks for debugging. 

I am under the impression that Drush doesn't need the distro code anymore because core doesn't write profile to settings.php if its read only.

webflo:
The correct the Drupal installer takes care of it (``install_write_profile()``). But this patch is about reading the install profile from settings prior installation to detect which install profile should be used. 

- ``drupal_get_profile()`` reads it from the database or settings file.
- ``_install_select_profile()`` reads its from profile metadata (distro, or only one profile etc)

webflo:
The important change is ``drupal_get_profile()``. Drush called it before, but it throw a exception because the container was not initialized and not db connection was available.

weitzman:
Thanks @webflo

PR #2746, opened 2017-05-07
Use tableFormatter in help and cli commands.

weitzman:
Looks good. One last problem is that the list command is not wrapping correctly in the table containing command list. Use a narrow window and see command description wrap like it used to.


weitzman:
The remaining bug was slayed by setting the terminal width before calling formatter in list command.

PR #2748, opened 2017-05-08
info files with "refname 'HEAD' is ambiguous"

lkmorlan:
In certain cases, the following can be added to info files by Drush:

datestamp = "warning: refname 'HEAD' is ambiguous."

datestamp should be an integer. This is a fix.

To reproduce the problem, use drush make to build a makefile which references a repository containing a tag "HEAD". The error is because tag "HEAD" is the same as Git's use of HEAD to indicate the current position.


weitzman:
make is in maint only mode.

PR #2758, opened 2017-05-14
PSR-2 formatting. 

greg-1-anderson:
- [x] Reformat for PSR-2
- [x] Rename methods to camel caps format
- [x] Put \Drush class in namespace Drush
- [ ] Enforce PRS-2 in ~~Travis~~ Circle CI tests


greg-1-anderson:
If you see an error `Fatal error: Class 'Drush' not found`, then:

- `$ composer dumpautoload`
- `$ ./unish.clean.php`

This will rewrite the autoload files and correct the stale references to the `Drush` class, which moved into a new namespace.

To fix up a PR or other new code to conform to PSR-2:

- `$ composer cbf`

includes/bootstrap.inc
```diff
@@ -273,7 +274,7 @@ function drush_bootstrap_validate($phase) {
         break;
       }
       if ($phase_index > $validated_phase) {
-        $current_phase .= '_validate';
+        $current_phase .= 'Validate';
```
weitzman: I'm a bit concerned about the BC implications of changing bootstrap hook names. Maybe c'est la vie, but I just want to discuss.

src/Commands/config/ConfigPullCommands.php
```diff
@@ -4,89 +4,91 @@
...
-        throw new \Exception('There are uncommitted changes in your git working copy.');
-      }
+    /**
+     * @hook validate config-pull
+     */
+    public function validateConfigPull(CommandData $commandData)
+    {
+        if ($commandData->input()->getOption('safe')) {
+            $return = drush_invoke_process($commandData->input()
+            ->getArgument('destination'), 'core-execute', array('git diff --quiet'), array('escape' => 0));
```
weitzman: Should a wrapped line like this be indented more?

includes/bootstrap.inc
greg-1-anderson: If we are going to adopt all of PSR-2, then we must camelCase() our methods. This is definitely b/c breaking, but I think that now is a good time to just do it, as folks will hopefully be doing rewrites to annotated anyway. Drush 8 is still available for those who have trouble with the upgrade.

greg-1-anderson:
I'll merge if you're comfortable with the backwards compatible story (or lack thereof) of camelCase() methods. Again, I think we should do it now, as if we do not, then we will likely never be PSR-2 compliant (with consistent decisions in this area).

includes/bootstrap.inc
greg-1-anderson: Also, the only bootstrap class in use in Drush 9 is DrupalBoot8, so I think the impact here is negligible.

weitzman: Oh, I misread. I thought the drush invoke function names were changing. I've merged this PR.

Chi-teck:
https://github.com/drush-ops/drush/blob/master/CONTRIBUTING.md#coding-style
The documentation still enforces contributes to use Drupal coding standards.

greg-1-anderson:
Thanks. See #2779

PR #2759, opened 2017-05-15
Experiment with DCG

weitzman:
To test:

```
git clone https://github.com/drush-ops/drush.git --branch generate-single
COMPOSER=composer-highest.json  ./unish.sut.php
// A convenient way to install Drupal in the SUT
./unish.phpunit.php tests/pmEnDisUnListInfoTest.php
sut list
sut d8:controller
```

- [ ] --yes does not skip the prompt before overwriting file.
- [ ] Decide what to do when no generator name is specified.
- [ ] Add generator for creating an Annotated Drush command (done). @todo generate a  drush.services.yml file. Blocked by DCG here. Is also needed by its other:drupal-console-command command.
- [x] Remove -d shortcut from option definition of new commands. Or ask DCG to remove this shortcut. Avoids a conflict with Drush's -d
- [x] interact() for module names could have better UX. Its better than DCG since you now pick from an autocomplete list. Perhaps improve the UX of additional questions?
- [x] Our path handling is naive. Note wrong path when calling `sut gen-module-configuration-entity`, for example.
- [x] When creating a module (d8:module-file), it makes no sense to pick from the enabled modules list
- [x] The paths reported after generation are relative to extension. Prefer absolute or drupal root relative.
- [ ] See [test fails](https://travis-ci.org/drush-ops/drush/jobs/233248777) that are unique when using Drupal 8.4.x. Looks like Symfony Filesystem 3.2 is no longer compatible with \DrupalCodeGenerator\GeneratorDiscovery::__construct


src/Boot/DrupalBoot8.php
```diff
@@ -215,4 +221,39 @@ public function terminate()
...
+        $drushGenerator = DrushGenerator::create($twig_directories);
+        $nameNew = str_replace(['d8:', 'other:'], 'gen-', $name);
+        $nameNew = str_replace(':', '-', $nameNew);
+        $drushGenerator = $drushGenerator->setName($nameNew);
+        $drushGenerator = $drushGenerator->setDescription($generator->getDescription());
+        $drushGenerator->generator = $generator;
+        $drushGenerators[] = $drushGenerator;
+        $def = $generator->getDefinition();
+        $optiondef = $def->getOption('destination');
+        // How to remove shortcut from $optiondef?
```
Chi-teck: $options = $def->getOptions();
// Do something with options here.
$def->setOption($options);

// Or maybe just remove both options (destination, answers) at all.
$def->setOptions([]);

src/Generate/DrushGenerator.php
```diff
@@ -0,0 +1,50 @@
...
+  // See interact().
+  protected $files;
+  protected $destination;
+
+  public static function create(array $twig_directories)
+  {
+    return parent::create($twig_directories);
+  }
+
+  protected function interact(InputInterface $input, OutputInterface $output)
```
Chi-teck: Seems like this method can be replaced with _input_preprocessor_ helper.

Chi-teck:
There might be some performance issue with _addGeneratorsToApplication_() since it is invoked for any drush _command_ . It does a quite heavy work including file system scanning, reflection and instantiating dozens objects. At least worth benchmarking.
I would consider implementing a single `drush generate` command that accepts generator name as an argument.

weitzman:
Closed in favor of #2772

PR #2772, opened 2017-05-23
Experiment with DCG (Part 2)

weitzman:
To test:

```
git clone https://github.com/drush-ops/drush.git --branch generate-single
COMPOSER=composer-highest.json  ./unish.sut.php
// A convenient way to install Drupal in the SUT
./unish.phpunit.php tests/pmEnDisUnListInfoTest.php
sut list
sut d8:controller
```

- [x] --yes does not skip the prompt before overwriting file.
- [x] Decide what to do when no generator name is specified.
- [x] The new generator for creating an Annotated Drush (drush-commandfile) command requires and enhancement to DCG for creating a drush.services.yml file in the owning module (needed for its Drupal Console command as well)
- [x] Write a Drush topic (i.e. docs)
- [x] Write a test
- [x] Add API for contrib/custom code to add a directory for discovering its generators
- [x] Update to PSR2 code style


weitzman:
I added a commit showing how modules may supply generators via the drush.services.yml file. Also added an end to end test for DCG+Drush. Feedback welcome.

I updated the checklist in first comment. I think this is ready for code style fixup and merge (Drush recently changed to PSR2). We can use follow-ups as needed.

PR #2778, opened 2017-05-29
Fixes #2777: prevent updates to 8.4.0 from lower versions

### LINKED ISSUE #2777: Prevent drush up from conducting an 8.3.x to 8.4.x upgrade
There is currently no way for Drush 8 to upgrade the codebase then perform schema updates. See https://www.drupal.org/node/2874827.

cilefen:


greg-1-anderson:
This is untenable. The solution is https://www.drupal.org/node/2874827#comment-12107459.

We can proceed here if it becomes impossible to simultaneously support Symfony 2.x and Symfony 3.x in Drush 8. I'll give it a shot.

greg-1-anderson:
Actually, we might be forced to take the route proposed in this PR. The problem we have is that the *global drush* must know whether to load Symfony 2.x or Symfony 3.x classes at `include autoload` time -- since the pm-* code is for global Drush (site-local Drush being updated via Composer).

It might be possible for us, if necessary, to make the site-local Drush adapt to Symfony 2 or Symfony 3, based on which one is installed in its vendor directory. However, giving the same flexibility to the global Drush would be much more difficult. We'd need two vendor directories, for one thing. Not sure that's going to fly.

Chi-teck:
One more route to consider. Drop support for Symfony 2 at all and create a new release tagged say 8.50 that works only with Drupal 7 and Drupal 8.4+. Keep Drush 8.1 to maintain Drupal 8.1 - 8.3 sites.

greg-1-anderson:
That's not completely out of the question. However, if we go that route, it might be better to rename the master branch to Drush 10, and use version 9 to identify the version of Drush that uses Symfony 3.

weitzman:
As it stands today, I think all we need is a 'conflicts' entry in the the drush 8.x composer.json indicating incompat with drupal/core:8.4. Site-local drush installs need to move to drush9. If you use a global drush, you can keep using drush8 even after the drupal 8.4 upgrade (not supported, but does work right now).

weitzman:
I'm no longer sure that a 'conflicts' line would be helpful. I think Composer will explain (in its usual obtuse way) that there is a conflict when user upgrades to drupal 8.4

So, nothing to do here anymore.

PR #2781, opened 2017-05-30
Let generators define destination generated code

Chi-teck:
**Problem:**
At the moment both input handler and input preprocessor rely on hardcoded mapping of DCG generator names.
https://github.com/drush-ops/drush/blob/master/src/Commands/generate/Helper/InputHandler.php#L29
https://github.com/drush-ops/drush/blob/master/src/Commands/generate/Helper/InputPreprocessor.php#L40

There are two major problem with this approach.

1. The mapping has to be synchronized with DCG. For example if DCG gets new theme generator, Drush generate may not be able to handle it automatically. Currently for all unknown generators we assume that if a generator asks for [machine name](https://github.com/drush-ops/drush/blob/master/src/Commands/generate/Helper/InputHandler.php#L56) the generated code can be stored to existing module. This is too optimistic.

2. Third party generators has no such mapping at all. So that they can only provide code for existing modules.

**Solution:**
Generators should be able to define destination for the generated code themselves. I've just [added](https://github.com/Chi-teck/drupal-code-generator/blob/master/src/Command/GeneratorInterface.php#L40) a way that can help helpers get clue about generated code.

### DIFF
--- a/src/Commands/generate/Helper/InputHandler.php
+++ b/src/Commands/generate/Helper/InputHandler.php
@@ -19,51 +19,51 @@ class InputHandler extends BaseInputHandler
     {
         $vars = parent::collectVars($input, $output, $questions);
         if (empty($input->getOption('directory'))) {
-            /** @var \Symfony\Component\Console\Command\Command $command */
+            /** @var \DrupalCodeGenerator\Command\GeneratorInterface $command */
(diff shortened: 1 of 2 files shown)


weitzman:
LGTM - merged. Thanks.

PR #2786, opened 2017-05-31
Add generator for a migration

weitzman:


mikeryan776:
Couple of thoughts:

1. "migrate-source" is a bit misleading, since it's generating an entire migration plugin in the .yml, not just the source.
2. It assumes a SQL source. But, that can be the first pass - other sources can be added as options later.

PR #2797, opened 2017-06-06
Fix #2753: Prevent loops when unable to download extensions

### LINKED ISSUE #2753: Infinite loop in pm-install when sites/all is read-only
In our environment, we have sites/all read-only and do not install anything there. If pm-install cannot find a module, it attempts to download to sites/all, fails, and loops endlessly. Drush should instead exit will a failure status.

lkmorlan:
This fixes #2753 by breaking out of the loop if downloads fail.

### DIFF
--- a/commands/pm/pm.drush.inc
+++ b/commands/pm/pm.drush.inc
@@ -982,6 +982,7 @@ function drush_pm_enable_validate() {
   $extension_info = drush_get_extensions();
 
   $recheck = TRUE;
+  $last_download = NULL;
   while ($recheck) {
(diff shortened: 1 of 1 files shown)


lkmorlan:
This solution has two block of identical code. If this approach is accepted, I would combine the block into a new function so as to not repeat code.

lkmorlan:
Would you like me to make a patch merging the two identical "Prevent from looping" blocks?

weitzman:
I think its OK this way. Its only a few lines, and this code has already been removed from master.

PR #2800, opened 2017-06-07
Allow Symfony 3 components for site-local installs of Drush 8.x

greg-1-anderson:
While keeping global installs at Symfony 2.

If we do not allow Symfony 3 components, then site-local installs of Drush 8.x will not work with Drupal 8.4.x. However, if we simply add Symfony 3 to the require section, then we will break global installs of Drush for Drupal 8.3.x. To prevent this, we pin Symfony components to Symfony 2 in require-dev.

c.f. https://github.com/drush-ops/drush/pull/2799

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,7 +92,7 @@ matrix:
     - php: 5.4
       env: UNISH_DRUPAL_MAJOR_VERSION=8 UNISH_DRUPAL_MINOR_VERSION=3.7 PHPUNIT_ARGS=--group=quick-drupal
     - php: 5.4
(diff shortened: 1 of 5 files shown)


greg-1-anderson:
Drush 8.3.x will not support site-local installation for Drupal 8.4.x or later. Use Drush 9 for that.

greg-1-anderson:
Let's revisit this for 8.1.15

greg-1-anderson:
Here are the installation permutations with this PR:

- `composer require` of Drush8 outside a Drupal site gets Symfony 2. Global Drush 8 w/ Symfony 2 works with Drupal 8.3 and Drupal 8.4.
- `composer require` of Drush8 inside a Drupal 8.4 site gets Symfony 3
- `composer require` of Drush8 inside a Drupal 8.3 site gets Symfony 2
- The phar gets Symfony 2, just like the first example

The composer.lock file has Symfony 2.

grasmash:
> composer require of Drush8 inside a Drupal 8.4 site gets Symfony 3

Much needed!

xjm:
Added a followup for this in core: https://www.drupal.org/node/2916204

PR #2802, opened 2017-06-09
Add module machine names to autocomplete for Module label

weitzman:
Ultimately, we might want to omit the Module name question for virtually all generators. Its redundant with machine_name.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
   "repositories": [
     {
       "type": "vcs",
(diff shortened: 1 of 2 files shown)


Chi-teck:
This needs to be rerolled according to the last changes in DCG. Input preprocessor is receiving array of Question objects now.

weitzman:
Adjusted and merged into master.

PR #2807, opened 2017-06-14
PR: Only close SSH pipe to prevent deadlock when host machine runs Windows (#2805)

### LINKED ISSUE #2805: drush fails to receive output from non interactive ssh commands in some circumstances
Hi,

I'm using docker and the following docker image to provide an SSH container https://hub.docker.com/r/jeroenpeeters/docker-ssh/ 

I want to access a docker based Drupal site using drush and drush aliases. Everything works apart from drush commands that execute SSH commands non interactively.  An example being user-login e.g. `drush @docker uli`

Degbugging shows the SSH command is executed but drush running outside of the container receives no output from the command. If I manually run the SSH command drush creates, it works fine so its just that the output is not ending up at the local version of drush running outside the container.

On further investigation I found that in backend.inc, the process pipe gets deliberately closed. If I comment this fclose out then the command works as expected.

The line is in `_drush_backend_proc_open()`. In drush 8.1.12 its line 389. I've copied it below along with the comment that goes above it:-

        // If we do not close stdin here, then we cause a deadlock;
        // see: http://drupal.org/node/766080#comment-4309936
        // If we reimplement interactive commands to also use
        // _drush_proc_open, then clearly we would need to keep
        // this open longer.
        fclose($process['pipes'][0]); // <-- if I comment this line out drush @docker uli works as expected

Clearly this works in most environments, I was wondering what might be special with my setup that requires the removal of this line?  What is it even there for? I couldn't work it out from the comment!

If others happen across this issue then maybe an optional toggle can be added to an alias to prevent this?

johnennewdeeson:
Initial go at a fix based on @greg-1-anderson suggestion of only closing this if the host system is Windows - see issue #2805

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -381,12 +381,15 @@ function _drush_backend_proc_open($cmds, $process_limit, $context = NULL) {
         if ($post_options) {
           fwrite($process['pipes'][0], json_encode($post_options)); // pass the data array in a JSON encoded string
         }
-        // If we do not close stdin here, then we cause a deadlock;
-        // see: http://drupal.org/node/766080#comment-4309936
-        // If we reimplement interactive commands to also use
-        // _drush_proc_open, then clearly we would need to keep
-        // this open longer.
-        fclose($process['pipes'][0]);
+
+        if ($is_windows) {
+          // If we do not close stdin here, then we cause a deadlock;
+          // see: http://drupal.org/node/766080#comment-4309936
+          // If we reimplement interactive commands to also use
+          // _drush_proc_open, then clearly we would need to keep
+          // this open longer.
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Looks like build is failing on master too. Maybe rebase here after that is fixed.

weitzman:
We'll have that fixed in next  24 hours or so. We need to react to upstream changes in DCG.

markpdeeson:
I'm seeing the same issue as #2805 when offloading local drush execution to drush within a docker container (Mac host, Alpine container).

This resolves the issue and the change looks good.

greg-1-anderson:
The master branch is now using Symfony Process. The change proposed here is therefore obsolete.

PR #2809, opened 2017-06-14
Update drush generate according to DCG changes

Chi-teck:
Eventually DCG Question wrapper has been removed. At the moment the only point of keeping Question wrapper is [lack of setter methods](https://github.com/symfony/symfony/issues/23147) in original class.
I've added a small trait with helper functions to set question text and default value.

Generators have to use native Symfony console questions. All three types of questions are supported (_Question_, _ConfirmationQuestion_, _ChoiceQuestion_).

Also updated MigrationGenerator and fixed a small bug with missing 'source' key in DrushCommandFile generator.

### DIFF
--- a/src/Commands/generate/Generators/Migrate/MigrationGenerator.php
+++ b/src/Commands/generate/Generators/Migrate/MigrationGenerator.php
@@ -6,6 +6,7 @@ use DrupalCodeGenerator\Command\BaseGenerator;
 use DrupalCodeGenerator\Utils;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
 
 /**
  * Implements `generate migration` command.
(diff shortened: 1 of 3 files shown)


src/Commands/generate/Generators/Migrate/MigrationGenerator.php
```diff
@@ -22,9 +23,9 @@ class MigrationGenerator extends BaseGenerator
      */
     protected function interact(InputInterface $input, OutputInterface $output)
     {
-        $questions = Utils::defaultQuestions() + Utils::defaultPluginQuestions() + [
```
weitzman: We are losing the defaultQuestions() call. Its needed to get machine_name and thus destination, I think.

Chi-teck: Default plugin questions [include](https://github.com/Chi-teck/drupal-code-generator/blob/master/src/Utils.php#L132) them.

PR #2811, opened 2017-06-14
Save theme related files to a correct destination

Chi-teck:
Should work the same way as for modules. The location can be either in a directory of existing theme or global _themes_ directory.


Chi-teck:
You can use `drush generate breakpoints` command for testing.

PR #2816, opened 2017-06-16
Set extension name automatically when possible

Chi-teck:
This should work equally for module and theme generators.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
     "league/container": "~2",
     "consolidation/robo": "~1",
     "symfony/config": "~2.2|^3",
-    "chi-teck/drupal-code-generator": "^1.17.2",
+    "chi-teck/drupal-code-generator": "^1.17.3",
     "consolidation/annotated-command": "^2.4",
     "consolidation/output-formatters": "^3.1.10",
     "symfony/yaml": "~2.3|^3",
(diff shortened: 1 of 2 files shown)


Chi-teck:
Added condition for current working directory when defining default value for name questions.
Do we still need _machineToLabel_ normalizer?

weitzman:
I think we can remove machineToLabel normalizer. Good bye, brief friend!

weitzman:
I really like this PR. Thanks.

At least part of the test fails here happen because [we explicitly pass a --directory](https://github.com/drush-ops/drush/blob/master/tests/annotatedCommandTest.php#L60) but that gets overwritten to modules/%. We should keep honoring --directory I think.

Chi-teck:
DCG dumper did some alteration of $directory just before dumping files. I've just [moved](https://github.com/Chi-teck/drupal-code-generator/commit/f118cad6d6ed7ee4877ec2cc63486398b4c4059b?diff=unified) it to BaseGenerator class. Tests have become green but I am not sure if _directory_ option is still working.

Chi-teck:
Also wonder if it would be a good idea to use absolute path for the directory.

weitzman:
Merged this. We can keep improving it as needed. I dont follow what you are saying about absolute dir.

PR #2820, opened 2017-06-22
Update webmozart/patch-utils minim version

balsama:
[environment.inc](https://github.com/drush-ops/drush/blob/master/includes/environment.inc#L566) calls the getHomeDirectory method from webmozart/patch-utils Path class. That method didn't exist until the 2.1.0 release of patch-utils, but drush's composer.json file allows the 2.0.0 release.

This PR just bumps the minimum version of patch-utils to 2.1.0 while still allowing everything up to, but not including, 3.0.0.


weitzman:
Thanks @balsama

PR #2827, opened 2017-06-30
Fix #2460. Use $application instead of $command in option verify

### LINKED ISSUE #2460: Declare formatter options in annotation_adapter
The CSV formatter has an option --include-field-labels but thats unusable right now without also sending --strict=0.

weitzman:
Thus removing some no longer needed code from annotation_adapter.inc

### DIFF
--- a/includes/command.inc
+++ b/includes/command.inc
@@ -523,11 +523,22 @@ function drush_handle_command_output($command, $structured_output) {
  */
 function _drush_verify_cli_options($command) {
 
-  // Start out with just the options in the current command record.
-  $options = _drush_get_command_options($command);
+  if (annotation_adapter_is_annotated($command)) {
+    // The canonical options list is now in $application
+    $application = Drush::getApplication();
+    annotation_adapter_add_legacy_commands_to_application($application);
+    $console = $application->get($command['command']);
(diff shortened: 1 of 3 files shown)


includes/command.inc
```diff
@@ -523,11 +523,22 @@ function drush_handle_command_output($command, $structured_output) {
  */
 function _drush_verify_cli_options($command) {
 
-  // Start out with just the options in the current command record.
-  $options = _drush_get_command_options($command);
+  if (annotation_adapter_is_annotated($command)) {
+    // The canonical options list is now in $application
+    $application = Drush::getApplication();
+    annotation_adapter_add_legacy_commands_to_application($application);
```
greg-1-anderson: Might we end up doing this multiple times, or is _drush_verify_cli_options() guaranteed to be called only once (including redispatches, etc)? If called multiple times, would that cause any problems?

greg-1-anderson:
Do you know what is causing `Failed asserting that 'Calling proc_open('git pull' origin --rebase);' contains "Calling proc_open(git pull origin --rebase);".`? Quoting errors in various calls to `exec` are causing a lot of test failures. The changes in this PR seem unrelated to the code that composes commands, though.

Issue #2832, opened 2017-07-07
dt function call in user-cancel command passes string instead of array


mradcliffe:
### Problem/Motivation
Running drush ucan fails on PHP 7.1 because of a strict TypeError. The command passes in a string instead of an array.

> drush  ucan <username> --delete-content -y"
> WD php: TypeError: Argument 2 passed to t() must be of the type array, string given, called in /opt/drush/7/vendor/drush/drush/includes/output.inc    [error]
on line 153 in t() (line 1735 of /path/to/Drupal-7.56/includes/bootstrap.inc).

#### Versions
**drush**: 7.4.0
**Drupal**: 7.56
**PHP**: 7.1

### Proposed Resolution

The issue is already fixed in 8.x and greater although I could not figure out the correct search terms to find an existing issue or PR for that.

* Fix the issue in 7.x so that drush remains compatible with PHP 7.1 through Drupal 7 life cycle:
* Looked using `grep -R "dt('[^\s]\+', '" commands`, which was a weak regex search that only found that one instance (it finds a false positive hence weak regex).

https://github.com/drush-ops/drush/blob/7.x/commands/user/user.drush.inc#L342

weitzman:
fixed

PR #2833, opened 2017-07-07
Issue #2832: Fix passing incorrect type to dt/t causing error in PHP 7.1

mradcliffe:
See #2832


mradcliffe:
These seem like random fails, and 7.x branch hasn't been tested in a while.

weitzman:
Agree. Merged.

PR #2838, opened 2017-07-15
See if we can get the Drush 7 make tests to stop timing out

greg-1-anderson:
… by just disabling a bunch of them. We're about to de-support this whole branch soon anyway. :P

### DIFF
--- a/tests/makeTest.php
+++ b/tests/makeTest.php
@@ -453,23 +453,23 @@ class makeMakefileCase extends CommandUnishTestCase {
     }
   }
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
http://download.gna.org/wkhtmltopdf/0.12/0.12.2.1/wkhtmltox-0.12.2.1.tar.bz2 can no longer be downloaded. This, however, is not the only blocker. 😞

greg-1-anderson:
A more profitable continuation would be to compare the Drush 8 and Drush 7 tests, as the former are working and the later are not.

PR #2839, opened 2017-07-16
pm-enable doesn't check required php version for a module

kyvour:
Added a checking of php version before module installation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "consolidation/robo": "~1",
     "symfony/config": "~2.2|^3",
     "chi-teck/drupal-code-generator": "^1.17.3",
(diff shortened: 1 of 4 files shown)


weitzman:
I think this can now be done by composer instead. not so inclined to add drush code for it.

kyvour:
But not all people use composer for the module downloading. BTW even if you has downloaded the module with incorrect php version in the requirements, you can't enable it via admin interface, because UI form has validation for all dependencies and for php version. But drush checks only dependencies. I think that drush pm-enable command should be as closest as possible to the drupal UI functionality

weitzman:
A composer built site is presumed  for drush9. Its part of the design.

PR #2842, opened 2017-07-17
Fix drush docs, link to 8.x docs from within the 8.x documentation

webflo:
I think the docs for Drush 8 should reference 8.x examples. A few link are broken because Drush 9 (master) diverged from Drush 8 (8.x)


weitzman:
Makes sense. Thanks.

PR #2852, opened 2017-07-24
Allow psysh version 0.8.x.

greg-1-anderson:
Drupal Console allows `"psy/psysh": "0.6.* || ~0.8",` which can cause Composer angst if Drupal Console is added to a project before Drush.

There appear to be no API changes in 0.8, at least not that affect Drush.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "php": ">=5.6.0",
     "ext-dom": "*",
     "psr/log": "~1.0",
-    "psy/psysh": "~0.6",
+    "psy/psysh": "0.6.* || ~0.8",
     "league/container": "~2",
     "consolidation/robo": "~1",
     "symfony/config": "~2.2|^3",


chx:
This is not it, the drush.phar already ships with psy 0.8.0 or so it seems.

greg-1-anderson:
~0.6 already includes ~0.8. This PR may improve some edge cases. The Drush 8.x version (#2853) has a couple of minor build fixes; these should be merged separately if the psysh version constraint is not desired.

damiankloip:
I'm not overly sure we need the two constraints here? I think Drupal console added the `0.6.*` constraint for drush originally maybe - I'm not 100% sure because I use drupal console...never. I don't think we should do the same oddities they do? Drupal console literally does nothing with their psysh setup, so any version should be ok for them to be honest.

damiankloip:
Seems like a good idea to just sync with the Console folks and lock it down from there? I thought Drupal console was not trying to do the same things as Drush... 🤦‍♂️

damiankloip:
From https://www.drupal.org/node/2391219

> There might be some overlap, but the console maintainers were smart enough to keep the console niche and concentrate on scaffolding, which is not something that drush core does

sure 🤔

thom8:
Could a metapackage help here which can have tighter constraints and even sit behind latest versions to ensure compatibility.

for example -- https://packagist.org/packages/thom8/drupalcli

greg-1-anderson:
Per @damiankloip, keeping ~0.6 as our constraint. This includes the latest release (until 1.0.0).

PR #2853, opened 2017-07-24
Allow psysh version 0.8.x.

greg-1-anderson:
8.x branch version of https://github.com/drush-ops/drush/pull/2852

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,15 @@
+machine:
+  timezone:
+    America/Chicago
+  php:
+    version: 7.0.11
+
+dependencies:
+  cache_directories:
+    - ~/.composer/cache
+  override:
+    - composer install --prefer-dist --no-interaction
+
+test:
+  override:
+    - composer lint
--- a/composer.json
+++ b/composer.json
@@ -66,6 +66,11 @@
       "Unish":        "tests/"
     }
   },
+  "scripts": {
+    "lint": [
+      "find includes -name '*.inc' -print0 | xargs -0 -n1 php -l"
+    ]
+  },
   "extra": {
     "branch-alias": {
         "dev-master": "8.0.x-dev"
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
   "require": {
     "php": ">=5.4.5",
     "psr/log": "~1.0",
-    "psy/psysh": "~0.6",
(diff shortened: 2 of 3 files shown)


chx:
This is not it, the drush.phar already ships with psy 0.8.0 or so it seems. Original issue at https://github.com/hechoendrupal/drupal-console/issues/3448

greg-1-anderson:
tl;dnr: This PR does not address chx's issue, but is probably still worth merging.

Ah. If you are mixing the global Drush (whether this is the Drush .phar, or Drush installed globally by some other means), then you are encountering the dependency hell that results when combining two Composer projects together. See [the trouble with two autoloaders](https://pantheon.io/blog/trouble-two-autoloaders). Your interim solution, to pin your Drupal project psysh version to the exact version that Drush is using has the weakness that it must be adjusted every time you upgrade your global Drush. The solution to that problem is to install Drush inside your Drupal project, just as you do with Drupal Console.

Global Drush is potentially vulnerable to this problem even in absence of Drupal Console, although the addition of the later project makes the problem more acute, as a) there are more packages in common, giving a larger chance for overlap, and b) Drush and Drupal Console are not tested together.

An attempt to solve the global Drush problem was started at https://github.com/drush-ops/drush/pull/2787; this PR would take a lot of effort to finish. There have also been a couple attempts at collaboration, e.g. https://github.com/drush-ops/drush/pull/1337, but unfortunately they did not work out.

The root cause, though, is Composer -- or, more to the point, the fact that Composer's ambitions outstrips php's capabilities, making it impossible to support per-library dependencies. Maybe things will get better in PHP 8 -- but not sure PHP will go there, as Composer builds would get a lot bigger and perhaps slower than they already are.

chx:
The real big question is, why does the include go outside of phar?

greg-1-anderson:
Drush must include Drupal's autoloader if it is to run any of Drupal's code. This is its purpose, so it must do this.

The better situation would be for Drupal to provide a cli in core that provided a minimal viable set of APIs for operating on Drupal core (including an API for adding commands to modules). Then, external tools that wanted to provide additional functionality could `exec` core's cli to do their business.

greg-1-anderson:
Per @damiankloip, keeping ~0.6 as our constraint. This includes the latest release (until 1.0.0).

I committed just the php lint changes to the 8.x branch.

Issue #2847, opened 2017-07-19
Drush fails to report error for bad alias within group


AdamPS:
I recently discovered that my nightly backup had been silently failing for nearly a week.  The backup runs from cron calling a custom drush command like this:
`cronwrap drush -y @allbackup my-backup-command`

I had mistakenly mistyped an alias in the group @allbackup.  Drush printed a warning message and exited **without running the command on any server** yet **with code 0**.  cronwrap deduced the command had succeeded so I didn't get any alert.

To reproduce the bug, set up aliases.drushrc.php as follows:
`$aliases['myalias'] = ['site-list' => ['@badalias']];`

Then run the sequence of shell commands below

```
$ drush @myalias status
You are about to execute 'status' non-interactively (--yes forced) on all of the following targets:
  @badalias
Continue?  (y/n): y
Not found: @badalias                                                                                                                     [warning]
$ echo $?
0

```

In contrast, drush correctly reports an error when running directly with a bad alias.

```
$ drush @badalias status
Could not find the alias @badalias                                                                                                      [error]
$ echo $?
1

```

greg-1-anderson:
It would indeed be preferable if Drush aborted with an error if any alias in a group did not exist.

AdamPS:
@greg-1-anderson Thanks for the reply.  I have raised PR #2854

PR #2854, opened 2017-07-26
#2847 Drush fails to report error for bad alias within group

AdamPS:


weitzman:
Critical fixes only for sitealias.inc

PR #2857, opened 2017-07-28
Workaround for #2065: Additional use_pcntl_exec check

bleeDev:
Workaround for #2065

### DIFF
--- a/includes/startup.inc
+++ b/includes/startup.inc
@@ -383,7 +383,7 @@ function drush_startup($argv) {
   }
 
   // Always use pcntl_exec if it exists.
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This allows folks to use the workaround for  #2065, but it does not fix #2065. See #2815.

I'm still in favor of merging, if the tests pass.

weitzman:
doesn't the PR currently say to use pcntl_exec even when it is in the disable_functions list?

bleeDev:
@weitzman You are correct, updated with correct logic.

jhedstrom:
Does this need to go into master as well?

weitzman:
This code is removed from master. There is no finder anymore. Note the
drush-launcher repo.

PR #2859, opened 2017-07-28
Factor out fallback dt() code into StringUtils::interpolate().

greg-1-anderson:
A simple refactor, for rebasing into the Symfony Dispatch PR.

### DIFF
--- a/src/Utils/StringUtils.php
+++ b/src/Utils/StringUtils.php
@@ -35,6 +35,13 @@ class StringUtils
      * Examples:
      *   interpolate('Hello, {var}', ['var' => 'world']) ==> 'Hello, world'
      *   interpolate('Do !what', ['!what' => 'work'])    ==> 'Do work'
+     *
+     * @param string $message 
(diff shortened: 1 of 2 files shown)


src/Utils/StringUtils.php
```diff
@@ -28,4 +28,41 @@ public static function csvToArray($args)
         //
         return array_map('trim', array_filter(explode(',', is_array($args) ? implode(',', $args) : $args)));
     }
+
+    /**
+     * Replace placeholders in a string.
```
weitzman: Approved. Ideally we add @param annotations to the new method. (I feel odd asking the annotation maintainer about annotations)

greg-1-anderson: lol np. I'll go back and annotate the other PR too. I often procrastinate on this when parameters are in flux, but that can lead to never doing it.

greg-1-anderson:
Oh, it's [@return](https://phpdoc.org/docs/latest/references/phpdoc/tags/return.html), not @returns.

Issue #2285, opened 2016-07-21
Drush fails to detect sites other than default when a site-local drush is installed


fago:
Similar as reported in #2070, detecting multi-site directories does not work when a site-local drush is in use. The reason is that a global drush 8 automatically picks the site-local drush up, but loses the directory context while doing so.

Steps to reproduce:
- Install site-local drush via composer
- Install global drush 8.*
- Go into a working multi-site directory, like sites/my-site and run "drush st".

**Expected:**
A output of  `Drupal Settings File          :  sites/my-site/settings.php`  
**Actual behaviour:**
`Drupal Settings File          :  sites/default/settings.php`

fago:
Looks like there is simply a chdir() call causing that. It looks like thes chdir() is even unnecessary, in my tests removing it did not cause any troubles and fixed the problem.

See PR2286

weitzman:
I'm working on a D8 multisite now, and I have not experienced this problem. I also call a global Drush which hands off to a site-local Drush. On D8 a sites.php is required in order for multisite to work. I mention that in case you are running D8.

Any chance you can zip up your site and send to me? Don't think I need the DB or uploaded files. If not, running with --debug might provide insights.

fago:
The problem appears on a Drupal 7 site. Unfortunately I cannot share the site, but it has sites.php configured as well. I can check whether this is redrocuable on a plain drupal 7, but I very much think so.

Here is the drush st --debug output in case it helps:

```
drush st --debug 
Using the Drush script found at /path/to/drupal/root/vendor/drush/drush/drush.launcher using pcntl_exec
Loading drushrc "/home/fago/.drush/drushrc.php" into "home.drush" scope. [0 sec, 3.79 MB]                                        [bootstrap]
Loading drushrc "/home/fago/.drush/drudrush/drushrc.php" into "custom" scope. [0.01 sec, 3.79 MB]                                [bootstrap]
Cache HIT cid: 8.0.5-commandfiles-0-26b8001111e3480fa6c27b7b14ecc42c [0.01 sec, 3.87 MB]                                             [debug]
Loading drushrc "/path/to/drupal/root/web/sites/all/drush/drushrc.php" into "drupal" scope. [0.03 sec, 7.96 MB]          [bootstrap]
Bootstrap to phase 0. [0.17 sec, 10.85 MB]                                                                                       [bootstrap]
Trying to bootstrap as far as we can. [0.17 sec, 10.86 MB]                                                                           [debug]
[...]

PR #2860, opened 2017-07-31
#2285: Restore actual CWD after finding the Drush launcher

BR0kEN-:
All details can be found in https://github.com/drush-ops/drush/issues/2285.


BR0kEN-:
Unfortunately, it does not solve the problem in 100% cases.

PR #2862, opened 2017-08-02
Fix infinite recursion when empty path is provided to drush_mkdir

njbooher:
Encountered this while trying debug an issue with site install.

### DIFF
--- a/includes/filesystem.inc
+++ b/includes/filesystem.inc
@@ -318,6 +318,12 @@ function drush_move_dir($src, $dest, $overwrite = FALSE) {
  * halted if the required directory cannot be created.
  */
 function drush_mkdir($path, $required = TRUE) {
+   if ($path === '') {
+    if (!$required) {
+      return FALSE;
+    }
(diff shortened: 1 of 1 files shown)


weitzman:
Seems like we should fix the caller.

njbooher:
This causes segfaults depending on php settings and it took me a while to get here. I'll report the upstream error that leads to it when I find it.

njbooher:
mtiymzaymja2njczmtg2mdcz

greg-1-anderson:
+1 on merging this.

weitzman:
OK. The PR currently handles empty string. Should it handle NULL?

greg-1-anderson:
Yeah, `empty($path)` would be an improvement.

njbooher:
empty is return true for the string '0' which is a valid folder name. I added a type check for $path and made a new pull request https://github.com/drush-ops/drush/pull/2864 since I can't seem to re-open this one or make changes to it.

PR #2865, opened 2017-08-05
Load the autoloader as soon as the Drupal root is identified.

greg-1-anderson:


includes/drupal.inc
```diff
@@ -13,14 +13,17 @@ use Drush\Drush;
...
-      // "Point autoload.php to the proper vendor directory" - https://www.drupal.org/node/2404989
-      $autoloader = drush_get_context('DRUSH_CLASSLOADER');
-    }
+  $autoloadFilePath = $drupal_root .'/autoload.php';
+  if (!$autoloader && file_exists($autoloadFilePath)) {
+    $autoloader = require $autoloadFilePath;
+  }
+
+  if ($autoloader === TRUE) {
+    // The autoloader was already requireed. Assume that Drush and Drupal share an autoloader per
```
webflo: Typo in requireed

includes/preflight.inc
```diff
@@ -497,19 +497,9 @@ function drush_preflight() {
...
-  // that might allow us to find our alias record now
+  // that might allow us to find our alias record now.
   if (empty($alias_record)) {
     $alias_record = _drush_sitealias_set_context_by_name($target_alias);
-
-    // If the site alias settings changed late in the preflight,
-    // then run the preflight for the root and site contexts again.
-    if (!empty($alias_record)) {
-      $remote_host = drush_get_option('remote-host');
-      if (!isset($remote_host)) {
```
greg-1-anderson: Edge case: a site alias defined in an alias file in a local Drupal site (identified via `--root` or the cwd), where the root defined in the alias is different than the root for the Drupal site containing the alias.

This code re-runs the preflight steps that sets up to bootstrap the selected Drupal site. I feel this makes the preflight harder to maintain, and prefer to remove it.

greg-1-anderson:
This PR makes the preflight process more linear, by removing a seldom-used feature (see comment above). This allows us to bring in the autoloader a little earlier in the preflight, which will allow us to share the same bootstrap classes between the current `master` branch and #2843.

The tests are all green, but I need to do a bit more testing to see if removing the code here causes any issues should anyone actually be using this feature (which, I think, is fairly unlikely).

greg-1-anderson:
I added a failing test to show the functionality that is not yet working in this PR.

greg-1-anderson:
@weitzman was +1 on this earlier in slack. I'm therefore going to go ahead and merge, so that I can rebase #2843 and continue with the bootstrap refactoring.

PR #2867, opened 2017-08-08
Update webflo/drupal-finder

webflo:


kmajzlik:
Please tag new 9.x beta version. Console made same change but i am not able to update it (and probably clean install of project with both Drush and Console will be problematical).

PR #2869, opened 2017-08-10
Move more bootstrap code from bootstrap.inc to the bootstrap manager.

greg-1-anderson:
Another interim PR for #2843


greg-1-anderson:
Presuming this is a non-controversial refactoring.

PR #2877, opened 2017-08-14
Loosen phpdocumentor constraint

greg-1-anderson:
Per https://www.previousnext.com.au/blog/using-composer-debug-issues-updating-drupal-core

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,6 @@
     "symfony/event-dispatcher": "~2.7",
     "symfony/finder": "~2.7",
     "pear/console_table": "~1.3.0",
-    "phpdocumentor/reflection-docblock": "^2.0",
     "webmozart/path-util": "~2"
   },
   "require-dev": {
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
This PR would break the global Drush, I think; I'm not convinced we should merge it. See: #2369

greg-1-anderson:
I am +1 on removing this constraint from the `master` branch, but -1 on merging this PR into 8.x. Weakening this constraint is better for Composer-managed Drupal + Drush installations, but leaving it as it is is better for global Drush issues.

The global Drush + dep hell problem is being worked on at #2787 for 8.x and https://github.com/drush-ops/drush/pull/2843 for the master branch.

Note that Drupal still pins phpdocumentor to version 2.0.4. Folks generally should not update a Drupal site with an unbounded `composer update`. See https://github.com/webflo/drupal-core-strict for a better approach.

greg-1-anderson:
The phpdocumentor constraint was added in https://github.com/drush-ops/drush/issues/2365

PR #2882, opened 2017-08-17
fix namespace for Drush command sql-query with option "--db-prefix" in use (branch 8.x)

deminy:
I ran following command with Drush 8.x:

> drush.sh sql-query --db-prefix "TRUNCATE TABLE {file_managed}"

and noticed following error message:

> Error: Class 'Database' not found in .../vendor/drush/drush/lib/Drush/Sql/SqlBase.php on line 204 ...

The commit is to fix the issue in branch 8.x. I have another pull request to address the issue for branch master.

### DIFF
--- a/lib/Drush/Sql/SqlBase.php
+++ b/lib/Drush/Sql/SqlBase.php
@@ -2,6 +2,7 @@
 
 namespace Drush\Sql;
 
+use Drupal\Core\Database\Database;
 use Drush\Log\LogLevel;
 use Webmozart\PathUtil\Path;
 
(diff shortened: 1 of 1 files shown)


jorrit:
Didn't this break Drupal 7?

deminy:
It won't. Class _\Drupal\Core\Database\Database_ is for Drupal 8 only, and it's called only when used by Drupal 8, as you can see from the code [here](https://github.com/drush-ops/drush/blob/8.x/lib/Drush/Sql/SqlBase.php#L205).

deminy:
@jorrit  You are right, I didn't read the code properly. Just had a new change addressing that and got it tested with Drupal 7. Not sure if a new pull request is needed or not.

jorrit:
Thanks, that was something that I would have proposed also. How will it get into drush when you don't make a PR of it?

deminy:
New pull request #3529 created to address the issue. Thanks for the comments and doing a code review on that.

PR #2890, opened 2017-08-22
8.x Fix serving existing files with encoded characters like `%20`

travismiller:
Files with spaces (and other percent encoded characters) are not served from the files directory when using `drush runserver`. This change simply decodes the REQUEST_URI path.

Same change as #2889 for the 8.x branch.


weitzman:
Seems reasonable to me. Any thoughts, @grugnog?

grugnog:
Merged - thanks!

PR #2908, opened 2017-08-28
Convert browse command to use SiteAliasManager.

weitzman:


src/Commands/core/BrowseCommands.php
```diff
@@ -28,11 +31,20 @@ class BrowseCommands extends DrushCommands
      */
     public function browse($path = '', $options = ['browser' => null])
     {
+        // TODO: Remove 2nd branch when no longer needed.
+        if ($this->hasSiteAliasManager()) {
+            $aliasRecord = $this->siteAliasManager()->getSelf();
+            $is_remote = $aliasRecord->isRemote();
+            $site_record = $aliasRecord->exportConfig()->get('options', []);
```
greg-1-anderson: Maybe we should have an API for this (`exportConfig()` + `get('options')`). Until we rewrite backend invoke, this pattern will probably appear a lot. Optional.

greg-1-anderson:
+1 LGTM just fix `composer cs` failure.  `composer cbf` should take care of this particular notice.

You can just merge ported commands right into the Symfony dispatch branch if you want, and just use PRs when changing common code like preflight / bootstrap / site alias manager et. al.; I'm also happy to review ported commands you want feedback on tho

src/Commands/core/BrowseCommands.php
```diff
@@ -28,11 +31,20 @@ class BrowseCommands extends DrushCommands
...
+            $is_remote = drush_sitealias_is_remote_site($alias);
+            $site_record = drush_sitealias_get_record($alias);
+        }
         // Redispatch if called against a remote-host so a browser is started on the
         // the *local* machine.
-        $alias = drush_get_context('DRUSH_TARGET_SITE_ALIAS');
-        if (drush_sitealias_is_remote_site($alias)) {
-            $site_record = drush_sitealias_get_record($alias);
+        if ($is_remote) {
             $return = drush_invoke_process($site_record, 'browse', [$path], drush_redispatch_get_options(), array('integrate' => true));
```
greg-1-anderson: Ah, I see, we are still using `drush_redispatch_get_options()` here. Ideally this method goes away once all of the commands are ported. Maybe use `$options` here?  As previously mentioned, `$options` may have items that do not appear in `drush_redispatch_get_options()` (default values for options that do not appear on the commandline).

Maybe we need to add a method to the PreflightArgs object that contains the original user options from the commandline. Then the question is, do we inject this only for commands that need it, or add an accessor to DrushCommands?

weitzman: I changed this to $options but now the receiving Drush (i.e. the remote in the $site_record) is throwing an error (it happens to be  Drush9 beta3 in this case) - `Unknown options: --version, --ansi, --no-ansi, --no-interaction.  See drush help browse for available options. To suppress this error, add the option --strict=0. `. So, backend seems to be passing along too many options. Thoughts?

greg-1-anderson: Yes, that is exactly what I was referring to in my comment above. Symfony adds in any option with a default value to `$options` (well, Symfony puts it in `$input`, and annotated-commands translates that into `$options`). We need a way for commands that need to redispatch to inject the original commandline arguments. Or perhaps we create a redispatch object that has the original commandline arguments.

PR #2913, opened 2017-08-29
Update Psysh to version 0.8.11

damiankloip:
This got missed and never backported into the 8.x branch. It's late, yes. But worth doing anyway :) We wanted 0.8.2 originally.

https://github.com/drush-ops/drush/pull/2649


damiankloip:
Looks like this has made the 8.x branch build fail in the mean time, reverting for now.

PR #2921, opened 2017-08-31
Made _drush_escapeshellarg_windows escape set to TRUE when then detec…

slefevre-ccad:
…ted OS is Windows


greg-1-anderson:
Not sure if this is still relevant. Will need passing tests to merge. Please re-open against latest version of Drush if continued work desired here.

Issue #34, opened 2013-08-26
drush sql-query should exit with failure status if query fails


rfay:
Currently, drush sql-query exits with a 0 even if there was an error in the query:

```
[rfay@dev-web11 current]$ drush sqlq 'select * from nothing;'
ERROR 1146 (42S02) at line 1: Table 'xxx_dev.nothing' doesn't exist
[rfay@dev-web11 current]$ echo $?
0
```

I think it would be useful to all of us if it returned an error status in a case like this.

Thanks!

plach79:
I'm still experiencing this on 8.x, apparently `drush_sql_cli()` does not set an explicit error when the process fails and returning `FALSE` is not enough to set a non-zero exit status.

PR #2925, opened 2017-09-01
Ensure "sql-cli" command sets an error on non-zero child process exit status

plach79:
See https://github.com/drush-ops/drush/issues/34#issuecomment-326548684


plach79:
Thanks!

PR #2927, opened 2017-09-02
Handle --yes and --debug

weitzman:
1. --yes behaves like --no-interaction. The effect is that prompts just proceed with defaults. At this time we lose Drush's feature to print out the question and response even during a --yes request. --no is still not supported.
1. --debug maps to -vv. Folks wanting more should use -vvv which maps pretty well to Drush's old DEBUG_NOTIFY level.

The new code in Preflight.php could easily be moved elsewhere if anyone has thoughts on that.

### DIFF
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -206,7 +206,7 @@ class Preflight
         }
         // Use -vvv for even more verbose logging.
         if ($input->getParameterOption(['--debug', '-d'], false, true)) {
-            $output->setVerbosity(Output::VERBOSITY_VERBOSE);
+            $output->setVerbosity(Output::VERBOSITY_DEBUG);
         }
 
         // Run the Symfony Application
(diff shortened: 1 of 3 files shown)


src/Preflight/Preflight.php
```diff
@@ -199,6 +200,15 @@ protected function doRun($argv)
         $runner = new \Robo\Runner();
         $runner->registerCommandClasses($application, $commandClasses);
 
+        // Process legacy Drush global options.
+        if ($input->getParameterOption(['--yes', '-y'], false, true)) {
+            $input->setInteractive(false);
+        }
+        // Use -vvv for even more verbose logging.
+        if ($input->getParameterOption(['--debug', '-d'], false, true)) {
+            $output->setVerbosity(Output::VERBOSITY_VERBOSE);
```
greg-1-anderson: VERBOSITY_VERBOSE is `-v`. VERBOSITY_VERY_VERBOSE is `-vv`. `VERBOSITY_DEBUG` is `-vvv`. I think we should use `VERBOSITY_DEBUG` here.

PR #2928, opened 2017-09-03
Fix exit code handling in sql-cli

### LINKED ISSUE #2824: sql-create returns no proper exit code
The exit code for sql-create is always 0 even when it fails because the db config is wrong.

webflo:
Fixes #2824

Looking at the documentation of drush_shell_proc_open. I think the conditions should be the other way around.

ping @jibran for review


jibran:
Yup, this works fine for me.

webflo:
Related fix in Drush 8 https://github.com/drush-ops/drush/commit/67254bd23a98db836b06ec89169a6fad92373d40

Issue #2855, opened 2017-07-26
Support busybox diff


rodrigoaguilera:
When I run 
```
drush cim --preview=diff
```
I get
```
diff: unrecognized option: x
BusyBox v1.26.2 (2017-06-11 06:38:32 GMT) multi-call binary.

Usage: diff [-abBdiNqrTstw] [-L LABEL] [-S FILE] [-U LINES] FILE1 FILE2

Compare files line by line and output the differences between them.
This implementation supports unified diffs only.

	-a	Treat all files as text
	-b	Ignore changes in the amount of whitespace
	-B	Ignore changes whose lines are all blank
	-d	Try hard to find a smaller set of changes
	-i	Ignore case differences
	-L	Use LABEL instead of the filename in the unified header
	-N	Treat absent files as empty
	-q	Output only whether files differ
	-r	Recurse
        --no-dereference Don't follow symlinks
	-S	Start with FILE when comparing directories
	-T	Make tabs line up by prefixing a tab when necessary
	-s	Report when two files are the same
	-t	Expand tabs to spaces in output
	-U	Output LINES lines of context
	-w	Ignore all whitespace
```

Is it possible to support busybox version of diff?

weitzman:
Interested folks can use `drush cim --verbose` to see the shell command that cim uses and can figure out if a more universal way is possible.

bircher:
This issue can be closed now too since the diff command is now `drush_shell_exec('diff -u %s %s', $temp_active_dir, $temp_sync_dir);` which is compatible with busybox diff.

rodrigoaguilera:
Woah. Thank you :)

PR #2940, opened 2017-09-14
Clean up config import and export methods to fix diff #2939

### LINKED ISSUE #2939: Config Import --preview=diff does not produce a complete diff.
When importing with `drush cim --preview=diff` configuration that is not in the sync storage is not listed as removed and language files are ignored because the process that creates the diff skips them.

In addition the diff is incorrect when a decorated sync storage is used, such as with [config_filter](https://www.drupal.org/project/config_filter).

Both of these issues as well as #2855 could be fixed by exporting the active storage to a temporary directory and exporting the sync storage to another temporary directory and diff the two.

bircher:
Cleaning up the config import to fix #2939

As a side effect the temporary filders will not be in git and therefore #2855 is solved too.


weitzman:
Code looks good. I'll merge when green. Thanks.

PR #2964, opened 2017-09-26
Use Symfony-style command names

greg-1-anderson:
Follow Symfony command naming conventions, but preserve original names as aliases.

This PR also updates the `list` command to group by first category using `:` instead of `-`. The original command name alias (e.g. `foo-bar` for `foo:bar`) is removed from the list of aliases in the `list` command only, to save space. All aliases are still listed in help, though.


weitzman:
helloworld.script is a real, working php-script. it should stay IMO.

We need a plan for how to handle the case where Drush9 does a drush_invoke_process to pm:enable on Drush8.

greg-1-anderson:
I mistook hello world as a standalone script; I put it back.

I think we should alter Drush 8 to convert any `group:command` into `group-command`. Then, Drush 9 will be able to call drush_invoke_process on new versions of Drush 8, but not old.

To handle old versions of Drush 8, perhaps we could have a config setting that Drush 9 could check, and, if set, it could convert from `group:command` to `group-command` at backend invoke time. I think this facility, if adopted, should be off by default, though.

PR #2966, opened 2017-09-26
Find uri from site directory basename

greg-1-anderson:
If URI is not specified, and cwd is inside a site directory (something with settings.php), then use the basename of the sites directory as the URI. Also, inject alias manager into Application so that we may later run a remote command that does not necessarily exist on the local machine.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -171,7 +171,7 @@ class Application extends SymfonyApplication implements LoggerAwareInterface
 
     public function aliasManager()
     {
-        return $this->aliasManager();
+        return $this->aliasManager;
     }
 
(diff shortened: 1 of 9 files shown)


weitzman:
Its working for me in the SUT. I'm up for merging this now and improving it in a followup, if you think thats reasonable.

greg-1-anderson:
I'll fix it here now.

greg-1-anderson:
This version now has delegated the responsibility for finding the site directory to the Bootstrap manager (or more specifically, the Boot object it creates). I think it even works, but we'll see what the tests say.

PR #2979, opened 2017-09-28
Remove most of sitealias.inc.

greg-1-anderson:


greg-1-anderson:
Looks like drupal-directory is still using the old site alias path resolver, removed in this PR. Shouldn't be too hard to update that to use the new APIs already in use in the rsync command.

PR #2980, opened 2017-09-28
Test Drupal 8.4.x and 8.5.x.

greg-1-anderson:


weitzman:
LGTM

greg-1-anderson:
Still not sure if we should merge this or not. Since these are the primary versions we support right now, it makes sense to prioritize testing on these versions. On the other hand, it might also be nice to catch breakage with 8.3.

Pausing to think what our coverage should be. At the moment I think 8.3 + 8.4 might be better. Or maybe split the difference and do 8.3 + 8.5 for a while, then switch to 8.4 + 8.5.

greg-1-anderson:
It's past time to revisit this PR.  Now that 8.4.0 is stable, our tests are testing 8.4.x twice.  If we merge this PR, then we will be testing ^8.4 stable and 8.5.x-dev.  Included in this PR is an unused composer-lowest for testing 8.3.7.

PR #2981, opened 2017-09-28
Fall back to uri on alias record for site install

joelpittet:

PR #2982, opened 2017-09-28
Take default value for --sites-subdir in site install command

greg-1-anderson:
This will be done only if the URI is a simple identifier (e.g. "dev" or "stage"), or if stripping the `http://` off the head of the URI results in the name of a folder in `sites` that already exists.

Continuation of #2981


weitzman:
This looks like an improvement. Thanks for looking into this tangled code. I removed it hastily in https://github.com/drush-ops/drush/commit/20749fb1de43759eb9d8f3b50dae2166a4e72a0d#diff-ca1bf65048418d6ca1d1d82c13f88eb7. Perhaps you could look at that function and see if we are replacing what we need to replace. In particular we should load sites.php and respect it during install and re-install.

greg-1-anderson:
Yes, I reviewed that function last night, but I missed the support for `sites.php`. I'm going to merge this now and add support for that in a follow-on pull request.

PR #2991, opened 2017-09-30
Move responsibility of site alias path building to the configlocator

joelpittet:
Move responsibility of site alias path building to the configlocator and only add environment if the config is --local.


greg-1-anderson:
Looks great - thanks!

PR #2995, opened 2017-10-01
Fix #2914. Remove global discovery of aliases/commandfiles

### LINKED ISSUE #2914: By default, remove global discovery of aliases/commandfiles
That is, stop looking in /etc/drush and /usr/share except for config. Folks who want this capability can configure their Drush to look there via --include, --alias-path. The config for doing so can go into a top level /drush folder as proposed in [drupal-project](https://github.com/drupal-composer/drupal-project).

The goal here is to nudge users to commit their aliases and commands to the composer project.

weitzman:


weitzman:
Got the aliases discovered with the following drush.yml. I'm happy to change it as needed. Might be discouraged for config files to set runtime.* as I'm doing here
```
runtime:
    context:
        alias-path: [/Users/moshe.weitzman/tmp/unish/drush-sandbox/etc/drush]
```
This PR also changes Unish to write a single group alias file (in yml format). It used to write multiple  alias.drushrc.php files in php

greg-1-anderson:
We definitely want to allow config files to set alias-paths &c. as shown here. (Also, both the list form and the single-path form should work.)  The `runtime` namespace might not be the best place to put this configuration, though.  Let's discuss and then move to a better place.

greg-1-anderson:
Reviewing this it appears that this change:

- Maintains both site-local and global locations for configuration files
- Maintains site-local locations for alias files and command files, but removes global locations for same

If I have made no mistake in the reading of the diff, then this is +1 LGTM. Thanks for the improvements to the alias file handling in the tests; much better to have these modernized.

weitzman:
You understood correctly. Merged.

Issue #2912, opened 2017-08-29
 failed to open stream: No such file or directory JSONCache.php:25


Jackson8Zhang:
[warning] file_put_contents(C:/Users/Administrator/.drush/cache/factory/Drush\Commands\ValidatorsCommands.cache): faile
d to open stream: No such file or directory JSONCache.php:25

drush:9.0.0-beta4
php:7.1.8
drupal:8.3.7

newsec:
Same errors after updating to 9.0.0-beta4. 
Seems that Drush can't create subdirectories in "factory/Drush" directory in Windows environment.

newsec:
Think so

weitzman:
I pushed a quick fix for this to 8.x and master. I didn't try on Windows so not 100% sure. Please post here if this fixes or does not fix the problem. I'll reopen if needed.

weitzman:
My fix caused test fails, so reverted both branches. Someone needs to work on a better fix.

newsec:
drush_mkdir() function ensures only ".drush\cache\factory\Drush" path and not path from $cid.

weitzman:
Pushed a different fix to 8.x and master.

PR #2998, opened 2017-10-01
#2912 - Create full chain of directories for cache files

newsec:


weitzman:
Thanks. I just fixed the bug differently.

PR #3000, opened 2017-10-02
Fix cannot find Drupal's root when run in a subdirectory on Windows

david-garcia-garcia:
Fix drush not able to find Drupal's root when run in a subdirectory on Windows.

### DIFF
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -169,6 +169,10 @@ function _drush_shift_path_up($path) {
   if (empty($path)) {
     return FALSE;
   }
+  // Normalize directory separators to fix windows paths
+  // that are converted to unix style in _drush_convert_path().
+  $path = str_replace('/', DIRECTORY_SEPARATOR, $path);
+  // Break into directories.
   $path = explode(DIRECTORY_SEPARATOR, $path);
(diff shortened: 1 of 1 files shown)


ArakTaiRoth:
This fixed the issue I was having with Drush 8 on a windows machine. Before I was unable to run any drush command within the site's directory due to it not being able to bootstrap, which was caused by it not looking in the correct directory for base drupal files.

includes/environment.inc
```diff
@@ -169,6 +169,10 @@ function _drush_shift_path_up($path) {
   if (empty($path)) {
     return FALSE;
   }
+  // Normalize directory separators to fix windows paths
+  // that are converted to unix style in _drush_convert_path().
+  $path = str_replace('/', DIRECTORY_SEPARATOR, $path);
+  // Break into directories.
   $path = explode(DIRECTORY_SEPARATOR, $path);
```
greg-1-anderson: Could we explode on `'/'` rather than `DIRECTORY_SEPARATOR` here? It is inelegant converting back and forth between native and unix-style paths.

weitzman:
Is this needed on master? if so has to go in there first.

PR #3008, opened 2017-10-03
Rename drush.services.yml to drush9.services.yml

greg-1-anderson:
This will require corresponding changes to devel, dcg, etc.


greg-1-anderson:
Discussed with Moshe; we will do this a different way, post the 9.0.0-beta5 release.

PR #3012, opened 2017-10-04
Add error handling for when config-export fails for a single item

### LINKED ISSUE #3011: Drush does not complete export when there is an error in exporting a config item.
I run into the following error sometimes while running the `drush config-export` command.
```
TypeError: Argument 2 passed to Drupal\Core\Config\FileStorage::write() must be of the type array,[error]
boolean given
```

In such a case, drush completely errors out and does not complete the config export.

jonnyeom:
This adds error handling to prevent #3011 from occuring.
If a config-item is not writable, drush will log it as an error and continue with the config-export.

Fixes #3011

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -515,7 +515,13 @@ function _drush_config_export($destination, $destination_dir, $branch) {
   }
 
   foreach ($source_storage->listAll() as $name) {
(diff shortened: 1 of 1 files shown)


weitzman:
Before merging this, the reason has to be clearer (use case for failure) and code should be for master branch first. Thanks.

PR #3020, opened 2017-10-05
Added config:status command.

danepowell:
This replaces #2726 

This is completely untested. I'm still trying to sort out a number of other Drush 9 bugs that prevent me from testing this.


danepowell:
Okay this has now been tested and works great.

weitzman:
Nice! 

Any chance you can exercise this command during [this test](https://github.com/drush-ops/drush/blob/master/tests/configTest.php#L26-L57)? Once thats in, rename that method to add Status.

If I'm pushing my luck too far just say so. No worries.

src/Drupal/Commands/config/ConfigCommands.php
```diff
@@ -196,6 +197,86 @@ public function delete($config_name, $key = null)
     }
 
     /**
+     * Display status of configuration (differences between the filesystem configuration and database configuration).
+     *
+     * @command config:status
+     * @interact-config-label
+     * @param string $label A config directory label (i.e. a key in \$config_directories array in settings.php).
+     */
+    public function status($label = NULL)
```
weitzman: $label needs an @param declaration in doxygen

src/Drupal/Commands/config/ConfigCommands.php
```diff
@@ -196,6 +197,86 @@ public function delete($config_name, $key = null)
...
+    {
+        $directory = $this->getDirectory($label, '');
+        $storage = $this->getStorage($directory);
+        $change_list = $this->getChanges($storage);
+        if (empty($change_list)) {
+            $this->logger()->notice(dt('The active configuration is identical to the configuration in the export directory (!target).', array('!target' => $directory)));
+            return;
+        }
+        drush_print("Differences of the active config to the export directory:\n");
```
weitzman: Use $this->output()->writeln().

src/Drupal/Commands/config/ConfigCommands.php
```diff
@@ -196,6 +197,86 @@ public function delete($config_name, $key = null)
...
+     *   1. User-provided $directory.
+     *   2. Directory path corresponding to $label (mapped via $config_directories in settings.php).
+     *   3. Default sync directory
+     *
+     * @param string $label
+     *   A configuration directory label.
+     * @param string $directory
+     *   A configuration directory.
+     */
+    function getDirectory($label, $directory) {
```
weitzman: Maybe $directory should default to null or ''?

src/Drupal/Commands/config/ConfigCommands.php
danepowell: There's already a `@param string $label ...` above this. Am I missing something?

weitzman: oops, my mistake. @param usually comes right below @command.

bircher:
it looks quite nice. a small nit pick maybe: I would use dependency injection instead of using `\Drupal::service('config.storage')`, `\Drupal::service('config.manager')` and `\Drupal::service('config.storage.sync')`.

Issue #3005, opened 2017-10-03
Port the `config-list` command to Drush 9


pfrenssen:
The `config-list` command is not available in Drush 9 and it is sorely missed. The `config-get` command now outputs an interactive list but I didn't find a way to print this to the terminal non-interactively like `config-list` was able to do.

I have two use cases that are no longer possible now:

We are developing a distribution and store all of our config in custom modules. I was using a script that does a diff of the list of config on my local environment with a diff on my reference environment to check that I didn't forget to export any new config to modules. This script relies on `config-list` and will no longer work once we move to Drupal 8.4 + Drush 9.

Secondly, I am very much UNIX / command line oriented so I am used to piping the output of commands and string them together to do useful stuff. One thing I do at least once per day which is no longer possible in Drush 9 is to output the list of newly created config to a module info file for using with the Config Devel module. For example if I create a field with the machine name `field_my_field_name` then it generates a bunch of config files, and I used to be able to dump them in the info file with a simple UNIX command chain:

```
$ drush cli | grep field_my_field_name >> ./path/to/module.info.yml
```

On our Drush 9 branch it is not clear how I can get a list of config items that are related to the field name.

weitzman:
Thanks for using Drush9! ... Those are two pretty good just justifications. If anyone is able to make a PR, I'll review and merge.

greg-1-anderson:
Maybe this is a little off-topic, since this has been a thing for a while now; however, it seems a bit odd to me that the `cli` shortcut is neither `core:cli` nor `sql:cli`, but instead is a command that has nothing to do with a command line interface. Would it be a problem to support `config:list`, `config-list` and `clst`, and drop `cli` in Drush 9?

greg-1-anderson:
(Or `cls`)

weitzman:
I had that thought too. go with `clst`.

pfrenssen:
For me personally it's no problem, the script I am using is just for my own use and I can adapt it.

I have also considered this before that `drush cli` is perhaps not the best shortcut for it. At the moment it is ingrained in my muscle memory though :)

I always thought `drush cli` should be the a synonym for `drush php`.

pfrenssen:
This is fixed by #3026, thanks @Chi-teck for implementing it!

DavideBruner:
You can try typing **drush config:get** or **drush cget**, it should work.

MatthieuScarset:
How do we add this command to the documentation?

[...]

PR #3026, opened 2017-10-06
Enhance config:status command

Chi-teck:
This would allow to filter configuration items by its operation and display results using different formatters.

config-status-output

Related issue #1811.

### DIFF
--- a/tests/configTest.php
+++ b/tests/configTest.php
@@ -51,7 +51,7 @@ class ConfigCase extends CommandUnishTestCase {
     $this->assertContains('unish', $page->front, 'Config was successfully imported.');
 
     // Test status of identical configuration.
-    $this->drush('config:status', array(), $options);
+    $this->drush('config:status --format=list', array(), $options);
     $this->assertEquals('', $this->getOutput(), 'config:status correctly reports identical config.');
       
     // Similar, but this time via --partial option.
--- a/tests/configTest.php
+++ b/tests/configTest.php
@@ -51,7 +51,7 @@ class ConfigCase extends CommandUnishTestCase {
     $this->assertContains('unish', $page->front, 'Config was successfully imported.');
 
     // Test status of identical configuration.
-    $this->drush('config:status --format=list', array(), $options);
+    $this->drush('config:status', array(), $options + ['format' => 'list']);
     $this->assertEquals('', $this->getOutput(), 'config:status correctly reports identical config.');
       
     // Similar, but this time via --partial option.
--- a/src/Drupal/Commands/config/ConfigCommands.php
+++ b/src/Drupal/Commands/config/ConfigCommands.php
@@ -208,7 +208,7 @@ class ConfigCommands extends DrushCommands
      *   Display configuration items that need to be synchronized.
      * @usage drush config:status --state=Identical
      *   Display configuration items that are in default state.
-     * @usage drush config:status --state='Only in sync directory' --prefix=node.type.
+     * @usage drush config:status --state='Only in sync dir' --prefix=node.type.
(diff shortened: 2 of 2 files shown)


Chi-teck:
Seems like `drush config-status --operation=all --format=list` produces the same output as `drush config-list` in Drush 8. This might also resolve #3005.

Chi-teck:
I just thought that "Operation" term seems confusing in this context. It came from `drush cex` and `drush cim` commands where it is used to describe ongoing action. However `drush config-status` is not actionable command.

I propose changing the terminology as follows.

- Operation -> State
- none -> default
- create -> new
- update -> modified
- delete -> deleted

tests/configTest.php
```diff
@@ -51,7 +51,7 @@ function testConfigExportImportStatus() {
     $this->assertContains('unish', $page->front, 'Config was successfully imported.');
 
     // Test status of identical configuration.
-    $this->drush('config:status', array(), $options);
+    $this->drush('config:status --format=list', array(), $options);
```
weitzman: Instead, do $options + ['format' => 'list]

weitzman: FYI you can run test with `./unish.clean.php --filter testConfigExportImportStatus`

weitzman:
That terminology is not totally clear yet. When it says "deleted", it could mean that active config has deleted that item or it could mean that the next import will delete the item from Active.

Any thoughts, @danepowell or @bircher or other?

danepowell:
Config import and export imply a certain causality, which works for verbs like "delete", "create", etc.

That context is completely missing for config-status, so I agree that we need to choose new terminology.

I know this doesn't quite match with existing convention, but I think to be perfectly clear, it would be best to rename the column to "State" and have four states:
* Only in active configuration
* Only in stored configuration
* Stored and active configuration differ
* Identical in stored and active configuration (normally you wouldn't see these rows)

Obviously that takes a lot of screen real estate, I'm trying to think of shorter ways to express that, but it's tough.

weitzman:
I like that. Maybe we should assume default config storage. That way we can say

Only in DB
Only in sync directory
Different
Identical

A site that varies config storage can alter the output of this column pretty easily.

Chi-teck:
Too verbose though. And as far as I know active storage does not have to be in DB.

Chi-teck:
Maybe we can omit "only".
- In active storage
- In sync directory
- Different
- Identical

weitzman:
I prefer my version. This is not a wide table so not too worried about verbosity.

Chi-teck:
By the way `drush feature-list` uses term "Overridden" for "different / modified" features.

danepowell:
I agree that for the majority of users, "Only in DB" and "Only in sync directory" is going to make waaay more sense. Even as a power user, it takes some mental gymnastics to process "active store" / "source storage" / etc.

I'm fine with either "overridden" or "different". I know "overridden" is more standard, but again, it implies causality (which one is overridden?), which doesn't make sense in this context.

Chi-teck:
Well, I just realized. This literals are used in an option to filter out results.
Current implementation `drush config-status --operation="create"` would become `drush config-status --state='In sync directory'`.
Is it fine with you?

weitzman:
Yes, I think those literals are ok for filtering. We should enumerate them
in @usage or @option

Chi-teck:
config-status-output-states

weitzman:
Our custom table wrapping does not play well with the color tags. Can we shorten sync directory to sync dir? That will help a little. The rest of the help will come when we fix wrapping.

pfrenssen:
Great, thanks @Chi-teck ! :+1:

PR #3034, opened 2017-10-08
Rough implementation of @hook obsolete.

greg-1-anderson:
Here is a prototype for `@hook obsolete`.

Any commandfile may implement `@hook obsolete command:name`. If a user then runs `drush command:name`, then the method annotated with `@hook obsolete` will be called. It is expected that this method should throw an exception to indicate that the command is obsolete.

This happens prior to the point where the $input object is parsed, so the obsolete command may contain any number of arguments or options, and the disposition of the items on the commandline will have no effect on the calling of the `@hook obsolete` method.

The one shortfall here is that a separate hook method is needed for every alias / variation of the obsolete command. Perhaps the annotated command library should be enhanced to allow a single method to be annotated with multiple hooks to allow for this use-case.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -35,6 +35,9 @@ class Application extends SymfonyApplication implements LoggerAwareInterface
     /** @var AliasManager */
     protected $aliasManager;
 
+    /** @var HookManager */
+    protected $hookManager;
+
     /** @var RedispatchHook */
     protected $redispatchHook;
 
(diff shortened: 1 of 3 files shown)


weitzman:
Looks great to me. Thanks Greg.

greg-1-anderson:
I thought of a better way to do this. We can let the obsolete functions continue to be commands; that way they can have aliases, just like other commands. After that, we merely let the `find` method find the command. Once we have the command object in hand, we can check it for an `obsolete` annotation, and throw an exception with the obsolete message if found.

PR #3035, opened 2017-10-08
Move alias conversion to own command

weitzman:


src/Commands/core/SiteCommands.php
```diff
@@ -134,6 +136,62 @@ public function siteAlias($site = null, $options = ['format' => 'yaml'])
...
+         *  - review public/private class property changes
+         *  - support custom paths instead of standard
+         *  - mention this command in docs and examples
+         *  - add a test
+         *  - add logging which explains disposition of each file
+         *  - suggest to user that she commits aliases to project after conversion
+         *  - allow a custom destination to deal with read only filesystems
+         */
+
+        $config = Drush::config();
```
greg-1-anderson: This can be injected via ConfigAwareInterface

src/Commands/core/SiteCommands.php
```diff
@@ -134,6 +136,62 @@ public function siteAlias($site = null, $options = ['format' => 'yaml'])
...
+            $config->get('drush.system-dir'),
+        ];
+        if ($siteRoot = Drush::bootstrapManager()->getRoot()) {
+            $paths = array_merge($paths, [ dirname($siteRoot) . '/drush', "$siteRoot/drush", "$siteRoot/sites/all/drush" ]);
+        }
+
+        // Configure alias manager and convert all.
+        $manager = $this->siteAliasManager();
+        $manager->addSearchLocations($paths);
+        $legacyFiles = $manager->legacyAliasConverter->discovery->findAllLegacyAliasFiles();
```
greg-1-anderson: If this is a separate command, we should probably pass the search paths in rather than modify the alias manager search paths. Not such a big deal, since it's only for this one command.

src/Commands/core/SiteCommands.php
```diff
@@ -134,6 +136,62 @@ public function siteAlias($site = null, $options = ['format' => 'yaml'])
...
+        $convertedFiles = $manager->legacyAliasConverter->convert();
+        $rows = [];
+        foreach ($legacyFiles as $legacyFile) {
+            $rows[$legacyFile] = [
+                'legacy' => $legacyFile,
+                'disposition' => array_key_exists($legacyFile, $convertedFiles) ? 'Yes' : 'No',
```
greg-1-anderson: This is deceptive. Any `foo.aliases.drushrc.php` may potentially create multiple bar.alias.yml and baz.alias.yml and so on, depending on what is inside the file. There is also a marker file foo.aliases.yml created, so this column should always say 'yes' for everything, even though the marker files are empty.

A more interesting output would show all of the generated files produced from each converted file, which may be one or more files. The API for the legacy converter would need to be updated for this.

greg-1-anderson:
+1 LGTM, although I would prefer to keep autoconvert for a while.

PR #3048, opened 2017-10-10
Only put the reflection-docblock constraint in require-dev. 

greg-1-anderson:
Let the version float for site-local installs.


greg-1-anderson:
We can just upgrade the annotated command library in Drush 8, and this library will no longer be necessary.

PR #3052, opened 2017-10-10
Fixes #2918: Add site-specific search paths 

### LINKED ISSUE #2918: Command files aren't getting searched deeply enough in DRUPAL/drush
Follow-on from https://github.com/drush-ops/drush/issues/2911:

> If admins setup their Composer project via drupal-project, and you classify Codebuilder as a drupal-drush type, then Codebuilder will be installed into the right place

That's putting the command file into the right place, but annotated-command isn't picking it up.

My project has been placed in: `/Users/joachim/Sites/8-drupal/drush/contrib` so I have this file with my commands in:

```
/Users/joachim/Sites/8-drupal/drush/contrib/drupal-code-builder-drush/CodeBuilderCommands.php
```

However, debug output from \Consolidation\AnnotatedCommand::discoverCommandFilesInLocation() shows that the discovery process doesn't actually go into the 'contrib' folder:
```
"== discoverCommandFilesInLocation: /Users/joachim/Sites/8-drupal/drush, == 0, \Drupal"
"== discoverCommandFilesInLocation: /Users/joachim/Sites/8-drupal/drush/Command, <= 2, \Drupal\Command"
"== discoverCommandFilesInLocation: /Users/joachim/Sites/8-drupal/drush/CliTools, <= 2, \Drupal\CliTools"
"== discoverCommandFilesInLocation: /Users/joachim/Sites/8-drupal/drush/CommandFiles, <= 2, \Drupal\CommandFiles"
"== discoverCommandFilesInLocation: /Users/joachim/Sites/8-drupal/drush/src, == 0, \Drupal"
"== discoverCommandFilesInLocation: /Users/joachim/Sites/8-drupal/drush/src/Command, <= 2, \Drupal\Command"
"== discoverCommandFilesInLocation: /Users/joachim/Sites/8-drupal/drush/src/CliTools, <= 2, \Drupal\CliTools"
"== discoverCommandFilesInLocation: /Users/joachim/Sites/8-drupal/drush/src/CommandFiles, <= 2, \Drupal\CommandFiles"
```

greg-1-anderson:
- Adds __DRUPAL_ROOT__/drush and __PROJECT_ROOT__/drush as commandfile search paths
- Finds directories containing one or more of:
  - composer.json
  - Commands/
  - src/Commands/
- Does **not** support drush.services.yml for these locations (must make a module for DI)
- Adds commands and hooks in DRUSH_BOOTSTRAP_NONE

Support for drush.services.yml not added for simplicity, and because commandfiles with dependency injection cannot be instantiated until DRUSH_BOOTSTRAP_FULL.

### DIFF
--- a/src/Config/ConfigLocator.php
+++ b/src/Config/ConfigLocator.php
@@ -8,6 +8,7 @@ use Consolidation\Config\Util\ConfigOverlay;
 use Consolidation\Config\Util\EnvConfig;
 
 use Drush\Preflight\PreflightArgsInterface;
+use Symfony\Component\Finder\Finder;
 
 /**
  * Locate Drush configuration files and load them into the configuration
@@ -400,21 +401,69 @@ class ConfigLocator
      *
      * @return array
      */
-    public function getCommandFilePaths($commandPaths)
+    public function getCommandFilePaths($commandPaths, $root)
     {
-        // Start with the built-in commands.
-        $searchpath = [
+        $builtin = $this->getBuiltinCommandFilePaths();
+        $included = $this->getIncludedCommandFilePaths($commandPaths);
+        $site = $this->getSiteCommandFilePaths(["$root/drush", dirname($root) . '/drush']);
+
+        return array_merge(
+            $builtin,
+            $included,
+            $site
+        );
+    }
+
+    /**
+     * Return all of the built-in commandfile locations
+     */
+    protected function getBuiltinCommandFilePaths()
+    {
+        return [
             dirname(__DIR__),
         ];
+    }
 
+    /**
+     * Return all of the commandfile locations specified via
+     * an 'include' option.
+     */
+    protected function getIncludedCommandFilePaths($commandPaths)
+    {
+        $searchpath = [];
         // Commands specified by 'include' option
         foreach ($commandPaths as $commandPath) {
             if (is_dir($commandPath)) {
                 $searchpath[] = $commandPath;
             }
         }
-
         return $searchpath;
+    }
+
+    /**
+     * Return all of the commandfile paths in any '$root/drush' or
+     * 'dirname($root)/drush' directory that contains a composer.json
+     * file or a 'Commands' or 'src/Commands' directory.
+     */
+    protected function getSiteCommandFilePaths($directories)
+    {
+        $result = [];
+
+        $directories = array_filter($directories, 'is_dir');
+
+        // Find projects
+        $finder = new Finder();
+        $finder->files()
+            ->ignoreUnreadableDirs()
+            ->path('#composer.json$|^src/Commands|^Commands#')
+            ->in($directories)
+            ->depth('<= 3');
+
+        foreach ($finder as $file) {
+            $result[] = dirname($file->getRealPath());
+        }
+
+        return $result;
     }
 
     /**
--- a/src/Config/ConfigLocator.php
+++ b/src/Config/ConfigLocator.php
@@ -451,6 +451,10 @@ class ConfigLocator
 
         $directories = array_filter($directories, 'is_dir');
 
+        if (empty($directories)) {
+            return $result;
+        }
+
         // Find projects
         $finder = new Finder();
         $finder->files()
--- a/src/Preflight/ArgsRemapper.php
+++ b/src/Preflight/ArgsRemapper.php
@@ -6,13 +6,16 @@ namespace Drush\Preflight;
  */
 class ArgsRemapper
 {
+    protected $remapOptions;
+    protected $remapCommandAliases;
+
     /**
      * ArgsRemapper constructor
      */
-    public function __construct($remap, $remove)
+    public function __construct($remapOptions, $remapCommandAliases)
     {
-        $this->remap = $remap;
-        $this->remove = $remove;
+        $this->remapOptions = $remapOptions;
+        $this->remapCommandAliases = $remapCommandAliases;
     }
 
     /**
@@ -24,43 +27,14 @@ class ArgsRemapper
     public function remap($argv)
     {
         $result = [];
+        $sawCommmand = false;
         foreach ($argv as $arg) {
-            $arg = $this->remapArgument($arg);
+            $arg = $this->checkRemap($arg, $sawCommmand);
             if (isset($arg)) {
                 $result[] = $arg;
             }
         }
         return $result;
-    }
-
-    /**
-     * Apply any transformations to a single arg.
-     *
-     * @param stinrg $arg One arguent to remap
-     * @return string The altered argument
-     */
-    protected function remapArgument($arg)
-    {
-        if ($this->checkRemoval($arg)) {
-            return null;
-        }
-        return $this->checkRemap($arg);
-    }
-
-    /**
-     * Check to see if the provided argument should be removed / ignored.
-     *
-     * @param stinrg $arg One arguent to inspect
-     * @return bool
-     */
-    protected function checkRemoval($arg)
-    {
-        foreach ($this->remove as $removalCandidate) {
-            if ($this->matches($arg, $removalCandidate)) {
-                return true;
-            }
-        }
-        return false;
     }
 
     /**
@@ -70,11 +44,30 @@ class ArgsRemapper
      * @param stinrg $arg One arguent to inspect
      * @return string The altered argument
      */
-    protected function checkRemap($arg)
+    protected function checkRemap($arg, &$sawCommmand)
     {
-        foreach ($this->remap as $from => $to) {
+        if (!$sawCommmand && ctype_alpha($arg[0])) {
+            $sawCommand = true;
+            return $this->remapCommandAlias($arg);
+        }
+        return $this->remapOptions($arg);
+    }
+
+    protected function remapOptions($arg)
+    {
+        foreach ($this->remapOptions as $from => $to) {
             if ($this->matches($arg, $from)) {
                 return $to . substr($arg, strlen($from));
+            }
+        }
+        return $arg;
+    }
+
+    protected function remapCommandAlias($arg)
+    {
+        foreach ($this->remapCommandAliases as $from => $to) {
+            if ($arg == $from) {
+                return $to;
             }
         }
         return $arg;
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -81,7 +81,7 @@ class Preflight
      * Eventually, we might want to expose this table to some form of
      * 'help' output, so folks can see the available conversions.
      */
-    protected function remapArguments()
+    protected function remapOptions()
     {
         return [
             '--ssh-options' => '-Dssh.options',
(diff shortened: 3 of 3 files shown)


src/Config/ConfigLocator.php
```diff
@@ -400,24 +401,76 @@ public function getSiteAliasPaths($paths, Environment $environment)
...
+            return $result;
+        }
+
+        // Find projects
+        $finder = new Finder();
+        $finder->files()
+            ->ignoreUnreadableDirs()
+            ->path('#composer.json$|^src/Commands|^Commands#')
+            ->in($directories)
+            ->depth('<= 3');
```
weitzman: is there a chance we scan large dirs like node_modules and such?

src/Config/ConfigLocator.php
```diff
@@ -400,24 +401,76 @@ public function getSiteAliasPaths($paths, Environment $environment)
...
         }
-
         return $searchpath;
     }
 
     /**
+     * Return all of the commandfile paths in any '$root/drush' or
+     * 'dirname($root)/drush' directory that contains a composer.json
+     * file or a 'Commands' or 'src/Commands' directory.
+     */
```
weitzman: Could we add text that says that CommandFile discovery happens in \Drush\Application::configureAndRegisterCommands. This function is only about discovering paths.

joachim-n:
This doesn't pick up my command, but it's getting there!

A dump($site) in getCommandFilePaths() shows:

```
array:1 [
  0 => "/Users/joachim/Sites/8-drupal/drush/contrib/drupal-code-builder-drush"
]
```

So that's my package's folder getting picked up!

However, the command isn't being found.

The problem is now the namespace, I think. In discoverCommandFilesInLocation(), I've added dump statements for $directory and this is what I get:

```
"searching $directory: /Users/joachim/Sites/8-drupal/vendor/drush/drush/src"
"with $baseNamespace: \Drush"
"searching $directory: /Users/joachim/Sites/8-drupal/vendor/drush/drush/src/Commands"
"with $baseNamespace: \Drush\Commands"
"searching $directory: /Users/joachim/Sites/8-drupal/drush/contrib/drupal-code-builder-drush"
"with $baseNamespace: \Drush"
"searching $directory: /Users/joachim/Sites/8-drupal/drush/contrib/drupal-code-builder-drush/src"
"with $baseNamespace: \Drush"
```

greg-1-anderson:
@joachim-n Put your commandfile in `src/Commands/CodeBuilderCommands.php`, and make the namespace `\Drush\Commands`.

greg-1-anderson:
I'd like to make one more change here, and **require** a composer.json in every Drush extension, whether it is registered with Composer or not. The advantage of this is that we could then use the composer.json `extra` section for purposes such as designating the supported Drush versions for the extension, etc.; 

I'm not sure if we can do this, though. Does Drupal packagist overwrite / replace composer.json files in modules? I don't remember.

weitzman:
drupal-packagist creates a composer.json if none is present but it does not
alter ones that are present. so i think your strategy sounds good. I guess
this requirement would be true for policy files as well, right?

joachim-n:
Hurrah!

That works!

My main command crashes, but for new reasons to do with picking up Drupal core version, which I'll go and figure out...

A quick test command in that file works properly:

```
  /**
   * Test it just registers.
   *
   * @command cb-foo
   */
  public function commandFoo() {
    dump('hello!');
  }
```

One thing though:

> make the namespace \Drush\Commands.

Is this ideal? It forces Drush command packages to pseudonamespace their command classes (in the old D7 style, where every class had to be called MyModuleMyClass). E.g. I have to call my class \Drush\Commands\DrupalCodeBuilderCommands to make sure I don't clash with some other package's \Drush\Commands\Commands class. That seems a bit backwards and not actually making use of namespaces the way they should work.

Could we have it work with a namespace of \Drush\Commands\Foo, or \Drush\Foo\Commands, where Foo is arbitrary?

> I'd like to make one more change here, and require a composer.json in every Drush extension

+1 from me, though I don't know about the d.org packagist question.

src/Config/ConfigLocator.php
greg-1-anderson: This only scans:
- The contents of the ./drush folder
- The contents of the ./drush/contrib folder (or whatever it is called)
- The contents of each ./drush/contrib folder (all the individual extensions)

If someone manually commits an extension into ./drush, then it is possible that a drush extension might be found immediately inside ./drush.  It is expected that only policy files should be installed like this; larger sitewide extensions (that are not modules) should be installed via composer.

If someone adds a directory with 20,000 files immediately inside ./drush, then we will scan it. Don't do that. Only Drush things should be inside ./drush, so it shouldn't be a problem.

We could, perhaps, exclude `site-aliases` and other special Drush folder names.

greg-1-anderson:
The original concept was that large sitewide extensions would just be made as modules; then, you get a drush.services.yml file, and you can use whatever namespace you want for your classes.

The problem with namespacing is that Drush doesn't know what your base namespace should be. If large extensions are modules and only small policy files are stored in the ./drush directory, then namespacing all of the policy files `\Drush` doesn't seem so bad. Maybe we need to be more flexible with namespaces, though; I'm not sure, as I sort of feel that if having a unique namespace is really important, the extension could always be packaged as a module.

There are a number of ways we could allow custom namespaces.

- Use a drush.services.yml for global extensions. At the moment this is a rejected solution. Having services means you bootstrap to "full". Allowing sitewide extensions that might or might not have services seems too complex. Having policy files that do not need bootstrap full to be discovered is also really important.
- Open each commandfile and read lines until "namespace ..." or some line that is not a comment and not a "namespace" line is found. Extract the namespace from the source.
- Read composer.jon and somehow get the namespace from there.

I'm undecided on this point, though.

joachim-n:
> The original concept was that large sitewide extensions would just be made as modules; then, you get a drush.services.yml file, and you can use whatever namespace you want for your classes.

In that concept, what sort of commands would you put into DRUPAL/drush/ ?

There are other commands besides mine which up until now have been Drush extensions but not modules -- https://www.drupal.org/project/grn for example.

greg-1-anderson:
This is getting a little philosophical, but the essence of the theory:

- Commands that do not bootstrap Drupal at all (or exec other Drush commands to do so): don't make a Drush command at all, make a Robo application (https://robo.li/framework).
- Commands that bootstrap Drupal, but do not need Dependency Injection, and often are used with multiple sites: make a global command and put it in ~/.drush
- Hooks that control how other commands work (policy files): put them in a site's ./drush directory
- Anything that uses DI: just make it a module

I'm not really sure why a generator would be site-local, and not global (~/.drush). But we're at least a little bit flexible about what folks do, and much of this is still up in the air.

joachim-n:
> I'm not really sure why a generator would be site-local, and not global (~/.drush). But we're at least a little bit flexible about what folks do, and much of this is still up in the air.

I did used to run it in ~/.drush, but I got the impression that on Drush 9, commands could no longer be kept there. Is that incorrect?

greg-1-anderson:
Global commands are still supported; however, the end user must decide where to put their global commands, and add the search location in their global `drush.yml` configuration file.

The fact that the global location is not enabled by default might cause extension writers to document everything as site-local, I guess. The reason the global locations were removed by default is that global commandfiles would sometimes mess up individual sites, so you have to opt-in to global includes now.

greg-1-anderson:
See: https://github.com/drush-ops/drush/blob/9.0.0-beta7/examples/example.drush.yml#L67:L75

joachim-n:
Ah ok. Thanks for the explanation1

> The fact that the global location is not enabled by default might cause extension writers to document everything as site-local, I guess.

Yes, I'll certainly recommend people install as a site-local Drush command, since that way they can install it with Composer, along with the dependency on Drupal Code Builder.

Once this issue's fixed and I rewrite the installation instructions in the command's README, I will also mention global install, as when you have multiple projects on one machine it's rather handy to have this command (and others) installed once for everywhere.

A thought on namespaces...

> Read composer.jon and somehow get the namespace from there.

In my command I'm using composer's autoload to specify PSR-4 loading:

```
  "autoload": {
    "psr-4": {
      "Drush\\Commands\\CodeBuilder\\": ""
    }
  }
```

I realize now that that isn't needed, as from what I can gather, AC is loading class files as it finds them. 

But could that declaration be used?

For example, you could say that any namespace listed in the composer file under 'psr-4' and that begins with 'Drush\\Commands' will get included.

I'd obviously have to rewrite that line so it makes sense for PSR-4 with the file location that AC expects, so it would be, I think:

```
      "Drush\\Commands\\CodeBuilder\\": "src/Commands"
```

greg-1-anderson:
We could use the psr-4 for namespacing site-wide Drush commands if we used https://github.com/consolidation/Robo/pull/604 to load and identify extensions.

joachim-n:
Isn't that the plugin mechanism which you said wouldn't be ready before Drush 9.0?

If that's the case, what I would suggest is:

- go with this PR as is rather than have it help up
- file a follow-on issue to allow command namespaces from psr-4 
- add documentation about the namespace that non-module commands currently must use, mentioning:
-- command providers should pseudonamespace their command class name to prevent clashes, e.g. \Drush\Commands\MyPackageCommands
-- use of namespaces below \Drush\Commands\ for a class name such as \Drush\Commands\MyPackage\Commands will be possible in a future release of Drush.

That means I'd currently have to call my class \Drush\Commands\DrupalCodeBuilderCommands, and when Drush 9.1 (say) comes out, I rename it to the cleaner \Drush\Commands\DrupalCodeBuilder\Commands. But that's ok as it doesn't break anything public, as nobody will be inheriting from it or referring to it.

greg-1-anderson:
That PR is basically ready, but hasn't been merged because no client has decided to use it. From Drush's perspective, that PR is not necessary for modules, which already autoload. For policy files / global commands, we need to decide on some trade-offs. Since we already differentiate the loading mechanism for modules from the loading mechanism for policy files, it does not seem like we should have more than one loading mechanism for policy files. We also want to avoid requiring `composer require` for policy files, which that plugin mechanism requires. So, our current plan of record is to make policy files lightweight drop-in (but capable of `composer require` if they happen to be packaged via packagist) & not use another plugin mechanism.

We could change that plan with sufficient reason, but if we do, it should happen prior to 9.0.0-rc1, which is to say in this PR, not after.

My plan for this PR is to add some composer.json parsing to prevent commands from being loaded by the wrong Drush & then merge it like that, maybe today or tomorrow. Our goal for rc1 is BADCamp timeframe, i.e. next week, plus or minus, if we can resolve all of the issues by then.

weitzman:
> So, our current plan of record is to make policy files lightweight drop-in (but capable of composer require if they happen to be packaged via packagist) & not use another plugin mechanism.

Sounds good to me.

> My plan for this PR is to add some composer.json parsing to prevent commands from being loaded by the wrong Drush & then merge it like that, maybe today or tomorrow.

IMO, that could be a follow-up. Your call.

greg-1-anderson:
Sure, I'll update this branch to master and merge once green again, and add composer.json limits as a follow-up.

joachim-n:
Is there a follow-up for allowing arbitrary namespaces defined in PSR-4?

I'd like to add other classes to my command package -- for instance a custom Symfony Question command, but having \Drush\ as my top-level namespace doens't seem right -- and there's the risk of class name clashes.

greg-1-anderson:
Our current strategy is for complex commands that need autoloading, DI &/or custom namespaces to be implemented by modules.

This has been discussed a few times in recent issues, but I don't remember if any current open issue is tracking this feature request. There is no immediate plan to improve non-module plugins at the moment.

greg-1-anderson:
Actually, if your plugin is of type drupal-drush, and if you instruct folks to install it only via `composer require org/project`, then you'll get autoloading for free.

Then you can just `new MyNamespace\MyClass` all you want.

joachim-n:
> Our current strategy is for complex commands that need autoloading, DI &/or custom namespaces to be implemented by modules.

I really don't understand why you're making non-module commands into a second-class citizens. It's rather frustrating when a while back I specifically decoupled what are now the Drupal Code Builder Drush commands from the Module Builder module.

> Actually, if your plugin is of type drupal-drush, and if you instruct folks to install it only via composer require org/project, then you'll get autoloading for free.

Nope, it's broken -- see https://github.com/drush-ops/drush/issues/3073.

joachim-n:
Any chance we could get a new beta with this in it?

I have this working ok on my existing setup that I used for working on Drupal Code Builder. However, I tried installing a completely fresh Drupal site using the Composer build, so see if I could get DCB installed properly, and my command is not getting picked up. I assume this is because this fix isn't in the latest beta, and I did have a go at forcing Composer to use -dev versions, but it's quite possible that I failed or missed something somewhere -- I find Composer really hard to use.

PR #3053, opened 2017-10-11
Fix wrong state mapping in config:status command

Chi-teck:
"Different" state should stand for "update" operation and be displayed in yellow to match `drush cex` / `drush cim` commands.

### DIFF
--- a/src/Drupal/Commands/config/ConfigCommands.php
+++ b/src/Drupal/Commands/config/ConfigCommands.php
@@ -230,8 +230,8 @@ class ConfigCommands extends DrushCommands
         $storage = $this->getStorage($directory);
         $state_map = [
             'create' => 'Only in DB',
-            'update' => 'Only in sync dir',
-            'delete' => 'Different',
+            'update' => 'Different',
+            'delete' => 'Only in sync dir',
         ];
         foreach ($this->getChanges($storage) as $collection) {
             foreach ($collection as $operation => $configs) {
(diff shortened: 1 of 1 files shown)


src/Drupal/Commands/config/ConfigCommands.php
```diff
@@ -230,8 +230,8 @@ public function status($options = ['state' => 'Only in DB,Only in sync dir,Diffe
         $storage = $this->getStorage($directory);
         $state_map = [
             'create' => 'Only in DB',
-            'update' => 'Only in sync dir',
-            'delete' => 'Different',
+            'update' => 'Different',
+            'delete' => 'Only in sync dir',
```
weitzman: Is there any chance that Create and Delete are flipped? In the context of an import, if a file is only in sync dir that will lad to a 'create', not a 'delete'.

Chi-teck: I suppose it only depends on the order of parameters passed to [StorageComparer](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Config%21StorageComparer.php/function/StorageComparer%3A%3A__construct/8.5.x) constructor.

Drush passes active storage as first parameter, so it is considered as source (storage object used to read configuration).
https://github.com/drush-ops/drush/blob/9.0.0-beta7/src/Drupal/Commands/config/ConfigCommands.php#L320

PR #3059, opened 2017-10-11
Improve option definitions.

weitzman:


src/Commands/OptionsCommands.php
```diff
@@ -9,12 +9,14 @@
...
 
+    const REQ=InputOption::VALUE_REQUIRED;
+
     /**
      * @hook option @optionset_proc_build
      * @option ssh-options A string of extra options that will be passed to the ssh command (e.g. "-p 100")
      * @option tty Create a tty (e.g. to run an interactive program).
      */
-    public function optionsetProcBuild($options = ['ssh-options' => InputOption::VALUE_REQUIRED, 'tty' => false])
+    public function optionsetProcBuild($options = ['ssh-options' => self::REQ, 'tty' => false])
```
greg-1-anderson: Are these default values respected? The documentation for hook option (and the code that calls hook option) passes the Command object and the AnnotationData.  I don't think that the parameters from the option hook are used in any way. Are you finding any change of behavior with this code?

weitzman: Yes, they are. I can see the difference when running help. Compsre the value of --runner in the two grabs below. I think whats happenning is that AC's reflection is doing its job, even before the params you mention are passed in.

```
sql:sync [--no-dump] [--no-sync] [--runner RUNNER] [--create-db] [--db-su DB-SU] [--db-su-pw DB-SU-PW] [--target-dump TARGET-DUMP] [--source-dump [SOURCE-DUMP]] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-d|--debug] [-y|--yes] [--remote-host REMOTE-HOST] [--remote-user REMOTE-USER] [-r|--root ROOT] [-l|--uri URI] [--simulate] [--pipe] [-D|--define DEFINE] [--skip-tables-key SKIP-TABLES-KEY] [--structure-tables-key STRUCTURE-TABLES-KEY] [--tables-key TABLES-KEY] [--skip-tables-list SKIP-TABLES-LIST] [--structure-tables-list STRUCTURE-TABLES-LIST] [--tables-list TABLES-LIST] [--druplicon] [--notify [NOTIFY]] [--xh-link XH-LINK] [--] <command> <source> <destination>
```

```
sql:sync [--no-dump] [--no-sync] [--runner [RUNNER]] [--create-db] [--db-su DB-SU] [--db-su-pw DB-SU-PW] [--target-dump TARGET-DUMP] [--source-dump [SOURCE-DUMP]] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-d|--debug] [-y|--yes] [--remote-host REMOTE-HOST] [--remote-user REMOTE-USER] [-r|--root ROOT] [-l|--uri URI] [--simulate] [--pipe] [-D|--define DEFINE] [--skip-tables-key SKIP-TABLES-KEY] [--structure-tables-key STRUCTURE-TABLES-KEY] [--tables-key TABLES-KEY] [--skip-tables-list SKIP-TABLES-LIST] [--structure-tables-list STRUCTURE-TABLES-LIST] [--tables-list TABLES-LIST] [--druplicon] [--notify [NOTIFY]] [--xh-link XH-LINK] [--] <command> <source> <destination>

```

PR #3066, opened 2017-10-14
Fixes #3054: Only replace command aliases for commands

### LINKED ISSUE #3054: Drush 9 converts command arguments as command aliases
When passing arguments named like existing command aliases, they are converted to the full name of the command. 

Example:

* `drush en fr foo` passes `[‘fr’, ‘foo’]` to the  command, 
* `drush en fr en` passes `[‘fr’, ‘pm-enable’]` to the command, expanding the parameter as if it was a command alias

Also happens when using `drush_invoke_process()`:

* `drush_invoke_process('@self', 'language:export:translation', ['fr', $file], $exportOptions)` invokes `exportTranslations()` with `fr`
* `drush_invoke_process('@self', 'language:export:translation', ['en', $file], $exportOptions)` invokes `exportTranslations()` with `pm-enable`

greg-1-anderson:
Command aliases (e.g. 'en' + 'si') should only be replaced when they appear as the first non-option / non-alias parameter on the argv list.

### DIFF
--- a/src/Preflight/ArgsRemapper.php
+++ b/src/Preflight/ArgsRemapper.php
@@ -6,13 +6,16 @@ namespace Drush\Preflight;
  */
 class ArgsRemapper
 {
+    protected $remapOptions;
+    protected $remapCommandAliases;
+
     /**
      * ArgsRemapper constructor
      */
-    public function __construct($remap, $remove)
+    public function __construct($remapOptions, $remapCommandAliases)
     {
-        $this->remap = $remap;
-        $this->remove = $remove;
+        $this->remapOptions = $remapOptions;
+        $this->remapCommandAliases = $remapCommandAliases;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


TessaBakker:
This is a bug in Symfony Console (\Symfony\Component\Console\Application::find)

A sort is needed in de code of version 3.2.0:
```
$list = array_values($commands);
sort($list);
$suggestions = $this->getAbbreviationSuggestions($list);
```

In de latest version, aliases won't work if there isn't a command that begins like the alias. This is a bug in the same method, but then in the last if-statement

`if (count($commands) > 1 && !$exact) {`
Here is the count not needed.

When fixed, 'drush en abc' or 'drush si' works perfect when removed from \Drush\Preflight\Preflight::remapCommandAliases

greg-1-anderson:
@TessaBakker Is there a PR for this in Symfony?

TessaBakker:
@greg-1-anderson I couldn't find a PR for this issue (https://github.com/symfony/symfony/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+console)

PR #3070, opened 2017-10-16
Check for null field storage definition in sql:sanitize

kimpepper:
We need to check the storage definition exists when looping through user fields and checking `->isBaseField()` otherwise you get:

```
Error: Call to a member function isBaseField() on null in Drush\Drupal\Commands\sql\SanitizeUserFieldsCommands->sanitize()
```


weitzman:
Thanks Kim.

PR #3075, opened 2017-10-17
Inject options from core:rsync and sql:sync alias parameters.

greg-1-anderson:
Includes tests for both, and documentation in example.aliases.yml.

### DIFF
--- a/examples/example.aliases.yml
+++ b/examples/example.aliases.yml
@@ -160,14 +160,16 @@
 #   if you set a 'remote-host' value, and your remote OS is Windows, if you
 #   do not set the 'OS' value, it will default to 'Linux' and could cause
 #   unintended consequences, particularly when running 'drush sql-sync'.
-# - 'ssh-options': If the target requires special options, such as a non-
+# - 'ssh': If the target requires special options, such as a non-
 #   standard port, alternative identity file, or alternative
-#   authentication method, ssh-options can contain a string of extra
-#   options that are used with the ssh command, eg "-p 100"
+#   authentication method, the `option` entry under the `ssh` item may
+#   contain a string of extra options that are used with the ssh command,
+#   e.g. "-p 100"
 # - 'paths': An array of aliases for common rsync targets.
 #   Relative aliases are always taken from the Drupal root.
 #   - 'files': Path to 'files' directory.  This will be looked up if not
 #     specified.
+#   - 'drush-script': Path to the remot Drush command.
 # - 'command': These options will only be set if the alias
 #   is used with the specified command.  In the example below, the option
 #   `--no-dump` will be selected whenever the @stage alias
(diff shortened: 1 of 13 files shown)


weitzman:
consider adding docs to example.drush.yml and/or example.aliases.yml.

greg-1-anderson:
Right now I am entombed in the hellish quagmire called backend invoke, which resists divulging the secret of `drush-script`.

greg-1-anderson:
Ah. `$site_alias_record->legacyRecord()` converts things back into the form backend invoke expects, but it leaves off some critical information, like `drush-script` and ssh-options.

examples/example.aliases.yml
```diff
@@ -177,6 +179,36 @@
 #     - `drush sql-sync @live @stage`
 #   NOTE: Setting boolean options broke with Symfony 3. This will be fixed
 #     in a future release. See: https://github.com/drush-ops/drush/issues/2956
+# - 'alias-parameters': These options will only be set if the alias is
+#   used as the specified parameter. `sql:sync` and `core:rsync` are the two
+#   core commands that use this entry. These commands both have `source`
+#   and `target` parameters.
```
weitzman: can we add config-pull to this list?

weitzman (commented):
I'll test this and possibly add more feedback. Thanks for your hard work here (and Consolidation). 

My one thought is to pause and ask if we really need  this alias-parameters feature. It looks analogous to the source-/target- prefixes of old. That feature was little used. My thinking is that folks can get used to the rule that command specific options will be honored depending where the command runs. That is:

1. `drush sql-sync @foo @bar`. local command-specific is honored
1. `drush @bar sql-sync @foo @self`. \@bar's command-specific is honored

greg-1-anderson:
`1. drush sql-sync @foo @bar. local command-specific is honored`

Perhaps in this scenario, if we also took the command-specific options from @foo and @bar, that might be sufficient. I used to use `source-command-specific` and `target-command-specific` to handle use cases that config split uses now. With rsync, this feature could be useful to exclude a path when an alias is a target but not the source, or visa-versa.

This feature only takes a small amount of code now, so it shouldn't be too hard to maintain. I'm not convinced that it's critical, but I'm reluctant to give it up. Maybe take a survey and see if it is used?

bar:
@greg-1-anderson @weitzman please, don't use my handle to document your code :p

greg-1-anderson:
Seems like it is safe to use `@` without making an `@`-mention as long as the reference is inside of a code block. If we always embed alias references in code blocks, then we won't need to worry about sending spurious notifications to unrelated GitHub users.

examples/example.aliases.yml
```diff
@@ -160,14 +160,16 @@
...
-#   authentication method, ssh-options can contain a string of extra
-#   options that are used with the ssh command, eg "-p 100"
+#   authentication method, the `option` entry under the `ssh` item may
+#   contain a string of extra options that are used with the ssh command,
+#   e.g. "-p 100"
 # - 'paths': An array of aliases for common rsync targets.
 #   Relative aliases are always taken from the Drupal root.
 #   - 'files': Path to 'files' directory.  This will be looked up if not
 #     specified.
+#   - 'drush-script': Path to the remot Drush command.
```
benjifisher: Typo.  `:s/remot/remote/`.

geek-merlin:
FTR from #5675:
> My one thought is to pause and ask if we really need this alias-parameters feature. 

For some use cases, default command options defined by aliases can NOT be replaced by default command options defined by drush config, because in a deployment use case, there is NO drush config before the first deployment.

PR #3080, opened 2017-10-19
Fix syntax errors in example.drush.yml

eiriksm:
Hello!

I was literally copy-pasting from the example.drush.yml file when i encountered the fact that it is actually not valid yml on some of the lines.

Here is a PR that fixes that

Thanks for drush, everyone!


greg-1-anderson:
Cut-copy-pasta error. Thanks for fixing it.

Issue #2956, opened 2017-09-25
Cannot use boolean values in config


weitzman:
For example, the following does have any effect as seen by the command callback not getting right value in $options.

```
command:
  image-flush:
      options:
        all: true
```

greg-1-anderson:
See https://github.com/symfony/symfony/issues/24314

weitzman:
Now that Symfony pr is merged, what do we need to do to support these options in our config?

greg-1-anderson:
This feature is only in Symfony 5+ (so Drupal 10+). We need support in consolidation/config; I think it should be easy, but I don't remember the code very well, will need to look at it. I don't think that Drush will need to do anything.

marcinmaruszewski:
In my opinion, this problem still occurs in the following configuration:
- Drush 12.4.4.0
- Drupal 11.2

```
asm89/stack-cors                         v2.2.0   Cross-origin resource sharing library and stack middleware
chi-teck/drupal-code-generator           3.4.0    Drupal code generator
composer/installers                      v2.2.0   A multi-framework Composer library installer
composer/semver                          3.4.0    Semver library that offers utilities, version constraint parsing and validation.
consolidation/annotated-command          4.9.2    Initialize Symfony Console commands from annotated command class methods.
consolidation/config                     2.1.2    Provide configuration services for a commandline tool.
consolidation/filter-via-dot-access-data 2.0.2    This project uses dflydev/dot-access-data to provide simple output filtering for applications built with annotated-command / Robo.
consolidation/log                        3.0.0    Improved Psr-3 / Psr\Log logger based on Symfony Console components.
consolidation/output-formatters          4.3.2    Format text by applying transformations provided by plug-in formatters.
consolidation/robo                       4.0.6    Modern task runner
consolidation/self-update                2.2.0    Provides a self:update command for Symfony Console applications.
consolidation/site-alias                 4.0.1    Manage alias records for local and remote sites.
consolidation/site-process               5.2.0    A thin wrapper around the Symfony Process Component that allows applications to use the Site Alias library to specify the target for a remote call.
dflydev/dot-access-data                  v3.0.2   Given a deep data structure, access data by dot notation.
doctrine/annotations                     1.14.3   Docblock Annotations Parser
doctrine/deprecations                    1.1.3    A small layer on top of trigger_error(E_USER_DEPRECATED) or PSR-3 logging with options to disable all deprecations or selectively for packages.
[...]

PR #3084, opened 2017-10-20
Introduce —no and modernize AFFIRMATIVE/NEGATIVE

weitzman:
And stop overloading -n|--no-interaction.


greg-1-anderson:
I don't think we need to handle `yes` and `no` during preflight. They can be ordinary global options, I would expect.

weitzman:
Do you think that we should set runtime.affirmative and runtime.negative configuration during Preflight? Or maybe all we need is Drush:affirmative() and Drush::negative() which calculate the right return value when called?

weitzman:
I just added Drush::affirmative() and Drush::negative() but the global option value is not set properly for some reason. I'm running `sut si --no` and still getting stopped at a prompt. I would expect an immediate cancellation to happen

greg-1-anderson:
Copying global options not handled in preflight is the responsibility of the [GlobalOptionsEventListener](https://github.com/consolidation/Robo/blob/master/src/GlobalOptionsEventListener.php#L44). Anything in the config `options` space will fetch its value from the `$input` option. We have to pre-populate `options` with all of our global options. Currently we do this in drush.yml.

It might be better to not have to name our options twice; however, I don't think that we can tell the difference between global options and local options just by looking at $input.

Maybe we should [have our Application object](https://github.com/drush-ops/drush/blob/master/src/Application.php#L53:L98) initialized the config `options` section. It could just loop over the definition object after initializing it.

greg-1-anderson:
We are also somewhat limited by https://github.com/drush-ops/drush/issues/2956.

src/Application.php
```diff
@@ -26,13 +29,17 @@
...
+class Application extends SymfonyApplication implements LoggerAwareInterface, ConfigAwareInterface
 {
     use LoggerAwareTrait;
+    use ConfigAwareTrait;
 
     /** @var BootstrapManager */
     protected $bootstrapManager;
 
+    /** @var ConfigInterface */
+    protected $config;
```
greg-1-anderson: ConfigAwareTrait provides this

greg-1-anderson:
+1 LGTM

Could remove unnecessary `protected $config;`

greg-1-anderson:
Not sure about the cause of the test failures but this looks good once that is cleared up.

src/Application.php
```diff
@@ -105,8 +105,11 @@ public function configureGlobalOptions()
             );
 
         foreach ($this->getDefinition()->getOptions() as $key => $option) {
-            // The config set here will be populated with resolved values by a later listener.
-            $this->getConfig()->set("options.$key", $option->acceptValue() ? $option->getDefault() : null);
+            // Don't override top level keys set via PreflightArgs::optionsWithValues.
+            if (!$this->getConfig()->has("$key")) {
```
greg-1-anderson: `"options.$key"`?

weitzman:
Just to see what would happen, I commented out the line that was setting config to a default value for the global options. Now we have only one failure, and I'm unclear if its really undesired behavior. @greg-1-anderson - see last commit and last test run. Thanks.

weitzman:
OK, that test fail was due to an unneeded change so I reverted it. This should come back green. Awaiting input from @greg-1-anderson about why this line is needed at all - https://github.com/drush-ops/drush/blob/0a06d83892f2a75f7a39a166f4edfd98acdd2b62/src/Application.php#L111

greg-1-anderson:
Without that line we have:
```
$ ./drush ev 'return \Drush\Drush::config()->get("options");'
array (
  'width' => '89',
)
```

greg-1-anderson:
I'm going to work on making this better in Robo, so that we do not need these extra machinations at all.

danepowell:
Just to be clear, once this is merged we'll have three options for non-interactive calls to Drush:
* `-n|--no-interaction`: assume default values for all prompts
* `--yes`: assume affirmative for all prompts
* `--no`: assume negative for all prompts

Is that correct?

greg-1-anderson:
I removed the `"options.$key"` stuff, which is not necessary with the improved versions of consolidation/config and Robo included in this PR.

weitzman:
@danepowell - no, we cant do that. we dont know if a question is affirmative or negative. So behavior of this PR is:

CORFIRM PROMPT
-n|--no-interaction: proced
--yes: assume default proceed
--no: abort

CHOICE PROMPT
-n|--no-interaction: use default value
--yes: assume default use default value
--no: No effect, AFAICT. user is asked as usual.

greg-1-anderson:
In Drush 8, `--no` caused Drush to abort with an error if a multi-select choice came up.

I'm not sure whether 18eac45 finishes everything here; I'll test a bit more and see. Getting close.

greg-1-anderson:
This seems to be working.  

Continue using default values:
```
$ drush -n si
 [notice] Starting Drupal installation. This takes a while. Consider using the --notify global option.
```

Accept all confirmation prompts:
```
$ drush --yes si

 // You are about to DROP all tables in your 'unish_stage' database. Do you want to      
 // continue?: yes.                                                                      

 [notice] Starting Drupal installation. This takes a while. Consider using the --notify global option.
```

Abort all confirmation prompts:
```
$ drush si --no
                                                                                     
 [WARNING] You are about to DROP all tables in your 'unish_dev' database. Do you want to 
           continue?: no.                                                                
                                                                                         
 [error]  Cancelled. 
```

It appears that a test failure from the master branch was merged in here.

greg-1-anderson:
I think this is all fixed up. Go ahead and merge when ready.

greg-1-anderson:
Oh, actually I should tag a release in Robo before we merge here. Please review and let me know if you're ready for that.

PR #3087, opened 2017-10-20
Adding security-updates command.

grasmash:


grasmash:
I'd like to add a test for this. The test would need to be executed in a context where `$composerRoot = Drush::bootstrapManager()->getComposerRoot()` returns a real composer.json, and the composer.lock contains an insecure module like `"drupal/ctools": "3.0.0.0-alpha27"`. Any pointers?

greg-1-anderson:
This looks great -- really good idea.

The unit tests automatically set up a "site-under-test". You only need to call `setUpDrupal()` if you need a database; in this case I don't think that would be necessary.  Just set up your tests by calling `composer require drupal/ctools:3.0.0.0-alpha27` and then run your test.

In your test setup code, `$this->webRoot()` will return the web root, and of course the parent of the web root will be the Composer root.  Once you get into the code you are testing, getComposerRoot will work as you expect.

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,80 @@
...
+     * @aliases sups
+     * @bootstrap none
+     */
+    public function securityUpdates()
+    {
+        $security_updates = [];
+        $client = new Client();
+        $response = $client->get('https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/8.x/composer.json');
+        if ($response->getStatusCode() != 200) {
+            $this->logger()->error("Received {$response->getStatusCode()} when attempting to fetch drupal-security-advisories information.");
```
greg-1-anderson: For an error, `throw new \Exception()` rather than log and return.

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,80 @@
...
+            return 1;
+        }
+        $response_body = $response->getBody();
+        $data = json_decode($response_body, TRUE);
+        $composer_root = Drush::bootstrapManager()->getComposerRoot();
+        $composer_lock_file_path = Path::join($composer_root, 'composer.lock');
+        if (!file_exists($composer_lock_file_path)) {
+            $this->logger()->error("Cannot find composer.lock file.");
```
greg-1-anderson: This should throw too

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,80 @@
...
+                    }
+                }
+            }
+        }
+        if ($security_updates) {
+            $this->logger()->error("One or more of your dependencies has an outstanding security update. Please apply update(s) immediately.");
+            $table = new Table($this->output);
+            $table->setHeaders(['Name', 'Installed version', 'Suggested version'])
+                ->setRows($security_updates)
+                ->render();
```
greg-1-anderson: Instead of manually rendering the table, return a RowOfFields and allow the output formatter to convert it to the user-specified format.  You should also declare a `@fields` annotation to map you machine names to table headers.

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,80 @@
...
+            if (!empty($data['conflict'][$name])) {
+                $conflict_constraints = explode(',', $data['conflict'][$name]);
+                foreach ($conflict_constraints as $conflict_constraint) {
+                    if (substr($conflict_constraint, 0, 1) == '<') {
+                        $min_version = substr($conflict_constraint, 1);
+                        if (Comparator::lessThan($package['version'], $min_version)) {
+                            $security_updates[$name] = [
+                                $name,
+                                $package['version'],
+                                $min_version,
```
greg-1-anderson: Once you are using RowOfFields (see below), then should be an associative array similar to:
```
$security_updates[$name] = [
                                'name' => $name,
                                'version' => $package['version'],
                                'min-version' => $min_version,
```

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,80 @@
+<?php
+namespace Drush\Commands\core;
+
+use Composer\Semver\Comparator;
+use Drush\Commands\DrushCommands;
+use Drush\Drush;
+use GuzzleHttp\Client;
```
weitzman: I think this would be a new dependency. I'm not too excited about it when we have the shell at our disposal. Maybe borrow the bits from http://api.drush.org/api/drush/includes%21drush.inc/function/_drush_download_file/8.0.x. Thats no longer in Drush core so you could just keep it local to this command for now.

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,80 @@
...
+{
+    /**
+     * Check Drupal Composer packages for security updates.
+     *
+     * This uses the Drupal security advisories package to determine if updates
+     * are available.
+     *
+     * @see https://github.com/drupal-composer/drupal-security-advisories
+     *
+     * @command security-updates
```
weitzman: How about pm:security?

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,80 @@
...
+                        }
+                    }
+                    else {
+                        $this->logger()->warning("Could not parse conflicting version constraint $conflict_constraint for package $name.");
+                    }
+                }
+            }
+        }
+        if ($security_updates) {
+            $this->logger()->error("One or more of your dependencies has an outstanding security update. Please apply update(s) immediately.");
```
weitzman: It would be cool to log an example composer command which fixes the issue. I think its a matter of `composer update foo/bar baz/bax ...`

src/Commands/core/SecurityUpdateCommands.php
greg-1-anderson: For such a simple use-case as this, `file_get_contents($url);` would suffice.

grasmash: Yeah my bad, I thought that this was an existing dependency because PHPStorm was registering it, but I see now that it was not a drush dependency but rather a dependency in my test project.

grasmash:
Need to add composer/semver package and cleanup `use` statement.

grasmash:
At present, output looks like:
```
$ drush sec
One or more of your dependencies has an outstanding security update. Please apply update(s) immediately.
Try running: composer require drupal/alinks:^1.1 --update-with-dependencies
If that fails due a conflict then you must update one or more root dependencies.
+---------------+-------------------+-------------------+
| Name          | Installed Version | Suggested version |
+---------------+-------------------+-------------------+
| drupal/alinks | 1.0.0             | 1.1               |
+---------------+-------------------+-------------------+
```

grasmash:
Hm, this failure is unenlightening:
```
PHP Fatal error: Class 'Unish\UnishProcessFailedError' not found in /home/travis/build/drush-ops/drush/tests/CommandUnishTestCase.php on line 219
/Fatal error: Class 'Unish\UnishProcessFailedError' not found in /home/travis/build/drush-ops/drush/tests/CommandUnishTestCase.php on line 219
The command "if [ -z "$TEST_DIR" ] ; then ${PWD}/unish.phpunit.php $PHPUNIT_ARGS ; fi" exited with 255.
```

greg-1-anderson:
I'm not sure what is causing your failure.

I think we will need to enhance Drush to support all three of your requests in the previous comment.

grasmash:
Here's a fourth: can we log a warning without affecting json output for --format=json?

weitzman:
1. I think that `drush_set_context('DRUSH_EXIT_CODE', 1);` will still work. Add a @todo so we remember to change it when Drush modernizes this bit.
1. You can add a post-command hook where you log the composer command for remedying the  SA. I think you can stash the command text in a class property.
1. I think your composer.json changes for repository and drupal/alinks should go right into boht files at https://github.com/drush-ops/drush/tree/7c7cab051a02b4a325948ac1397e67601d479251/tests/resources/codebase. Then you can remove them from your test. I dont like the SUT to vary depending on whether your test has run yet or not. This would give same SUT codebase for all tests.
1. Logger calls go to STDERR and should not affect output. They should not be present in a tests's getOutput() call. Not sure why that would be going wrong.

grasmash:
@weitzman Thanks. PR updated.

> You can add a post-command hook where you log the composer command for remedying the SA. 

I've done this, but the logged output still appears before the table.

greg-1-anderson:
The command result is processed and printed after all of the hooks fire. It would be possible to install a shutdown handler, but this would be inelegant.

greg-1-anderson:
Adding a `@hook terminate` would do the trick. (Symfony terminate command handler, not a shutdown handler.)

tests/SecurityUpdatesTest.php
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * Test that insecure packages are correctly identified.
+   */
+  function testInsecurePackage() {
+    $this->drush('pm:security', array(), array(), NULL, NULL, self::EXIT_ERROR);
+    $this->assertContains('One or more of your dependencies has an outstanding security update. Please apply update(s) immediately.', $this->getErrorOutput());
+    $this->assertContains('Try running: composer require drupal/alinks:^1.1 --update-with-dependencies', $this->getErrorOutput());
+    $this->assertContains('drupal/alinks', $this->getOutput());
+    $this->assertContains('1.0.0', $this->getOutput());
+    $this->assertContains('1.1', $this->getOutput());
```
weitzman: Use --format=json option and then your asserts can be more precise. Example: https://github.com/drush-ops/drush/blob/747ae29a782f8c51ab3eb3d19268811eed6fb7ff/tests/configTest.php#L49-L51

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,108 @@
...
+    public function preExampleHello($result, CommandData $commandData) {
+        if (isset($this->securityUpdates)) {
+            $suggested_command = 'composer require ';
+            foreach ($this->securityUpdates as $package) {
+                $suggested_command .= $package['name'] . ':^' . $package['min-version'] . ' ';
+            }
+            $suggested_command .= '--update-with-dependencies';
+            $this->logger()->warning("One or more of your dependencies has an outstanding security update. Please apply update(s) immediately.");
+            $this->logger()->notice("Try running: <comment>$suggested_command</comment>");
+            $this->logger()->notice("If that fails due a conflict then you must update one or more root dependencies.");
```
weitzman: missing word 'to'

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    /**
+     * Emit suggested composer command for security updates.
+     *
+     * @hook post-command pm:security
+     */
+    public function preExampleHello($result, CommandData $commandData) {
+        if (isset($this->securityUpdates)) {
```
weitzman: Use !empty() or else you get output when a command failed like at https://travis-ci.org/drush-ops/drush/jobs/290712023. I'm re-running that test now so that failure might fix itself.

src/Commands/core/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,108 @@
+<?php
+namespace Drush\Commands\core;
```
weitzman: Lets relocate this file to src/Commands/pm/SecurityUpdateCommands.php

grasmash:
The only thing that I don't like about this PR is that we are relying on https://github.com/drupal-composer/drupal-security-advisories.

That project is updated daily by @webflo's Circle CI server, [using data from the Drupal.org API](https://github.com/drupal-composer/drupal-security-advisories/blob/master/build/build.sh).

IMO, Drupal.org should provide this service directly.  

@CashWilliams @xjm I wonder your opinions (and also @weitzman @greg-1-anderson)?

grasmash:
@weitzman All but one build is passing, not sure what's up with that one.

weitzman:
Yes, it was a drupal-finder issue. I submitted a PR at https://github.com/webflo/drupal-finder/pull/32.

To get this PR to green (and thus mergeable),I suggest you edit our composer.json and add a repository that points to my fork of drupal-finder (my branch name is `composer-env-variable`). The technique is described at https://getcomposer.org/doc/05-repositories.md#vcs

Yes, I'd like drupal.org to offer this service eventually but I'm fine leading the pack and relying on webflo's service.

weitzman:
The 'no updates' message should use notice() instead of info(). that way it gets shown by default.

grasmash:
Well, that seems to have blown up even unrelated tests:

> Drush not symlinked in the System-Under-Test.

src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -7,6 +7,7 @@
 use Drush\Commands\DrushCommands;
 use Drush\Drush;
 use Exception;
+use function str_replace;
```
weitzman: er, i have not seen a regular function in a use statement. i have seen functions prefixed with a slash to indicate that it is a global function. lets do that if needed.

src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,118 @@
...
+                    }
+                    else {
+                        $this->logger()->warning("Could not parse drupal-security-advisories conflicting version constraint $conflict_constraint for package $name.");
+                    }
+                }
+            }
+        }
+        if ($this->securityUpdates) {
+            // @todo Modernize.
+            drush_set_context('DRUSH_EXIT_CODE', 1);
```
weitzman: I see that we have a constant for this `DRUSH_FRAMEWORK_ERROR`

src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -0,0 +1,118 @@
...
+            $this->logger()->notice("<info>There are no outstanding security updates for Drupal projects.</info>");
+        }
+    }
+
+    /**
+     * Emit suggested composer command for security updates.
+     *
+     * @hook post-command pm:security
+     */
+    public function preExampleHello($result, CommandData $commandData) {
```
weitzman: method name copy-pasta

weitzman:
Merged. Awesome work,  @grasmash.

webflo:
Fancy :)

gitressa:
This works for contrib modules, but is there a way to check for Drupal 8 core updates? Also, this command only checks for an update, but doesn't actually do the update. Will this feature be added at some point?

What I am looking for is a way to apply security updates with Composer, like it is possible with `drush up --security-only` for Drupal7, or Drupal 8 with Drush 8. I have used a cron-triggered `drush up --security-only` to automatically roll-out security-updates on a few web sites for a year now. So far it has worked perfectly, but I of course have plenty of back-ups, should anything ever break.

weitzman:
>
> The intent is for this to work for core. There was a bug with it initially
> but that was fixed a couple releases ago. Do you have latest release of
> Drush (and have a pending Drupal security release)?


Composer is the now tool for updating your code. Drush no longer does this
and has no plans to change. You will want to script a composer update call
when pending security updates are present.

gitressa:
Thanks for a fast answer. Do you know of such a script, which issues a composer update call when pending security updates are present for a contrib module or Drupal core? And if yes, do you have a link or a code example?

As far as I recall, Composer can't discern between normal updates and security updates, so is automatically applying security updates only even possible?

PS: I am not sure if the first quoted part is just FYI or a question to me, but I was testing with Drush 9.4 and Drupal 8.5.3:

```
$ composer --version
Composer version 1.7.2 2018-08-16 16:57:12

$ mkdir drupal && cd drupal && curl -sSL https://ftp.drupal.org/files/projects/drupal-8.5.3.tar.gz | tar -xz --strip-components=1

$ composer require drush/drush

$ vendor/bin/drush st
 Drupal version : 8.5.3                                               
...
 Drush script   : /var/www/public/drupal/vendor/bin/drush             
 Drush version  : 9.4.0                                               

$ vendor/bin/drush site-install --db-url=mysql://root:root@localhost/d853 -y;

$ composer require 'drupal/alinks:1.0'
$ vendor/bin/drush en alinks

$ vendor/bin/drush pm:security
 [warning] One or more of your dependencies has an outstanding security update. Please apply update(s) immediately.
 [notice] Try running: composer require drupal/alinks:^1.1 --update-with-dependencies
 [notice] If that fails due to a conflict then you must update one or more root dependencies.
+---------------+-------------------+-------------------+
| Name          | Installed Version | Suggested version |
+---------------+-------------------+-------------------+
| drupal/alinks | 1.0.0             | 1.1               |
+---------------+-------------------+-------------------+
```
... security update for Drupal 8.6.1 isn't listed.

PR #3089, opened 2017-10-21
Load Drush services for themes as well as modules.

NickWilde1990:
In Drush 8.x themes could define Drush commands as well as modules. In 9.x, this is no longer the case.
This patch re-implements that.


src/Drupal/DrupalKernel.php
```diff
@@ -106,7 +106,11 @@ public function discoverServiceProviders()
         // necessary that the class files in these commands are available
         // in the autoloader.
 
-        // Also add Drush services from all modules
+        // Also add Drush services from all modules & themes.
+        $extensions = $this->getConfigStorage()->read('core.extension');
+        $this->moduleList += isset($extensions['theme']) ? $extensions['theme'] : [];
```
greg-1-anderson: I don't know if this is prudent. This field is not only used by Drush, but is used throughout Drupal. Adding the themes to the module list might have unintended consequences. It would probably be much better to simply build a list of themes in local variables, and then have a separate loop over `$theme_filenames` to call `addDrushServiceProvider`.

src/Drupal/DrupalKernel.php
```diff
@@ -124,4 +128,36 @@ protected function addDrushServiceProvider($serviceProviderName, $filename)
...
+            $profiles = array($parent_profile => $all_profiles[$parent_profile]) + $profiles;
+        }
+
+        $profile_directories = array_map(function ($profile) {
+            return $profile->getPath();
+        }, $profiles);
+        $listing->setProfileDirectories($profile_directories);
+
+        // Now find themes.
+        $this->moduleData += $listing->scan('theme');
```
greg-1-anderson: Same comment - not sure that it is safe to modify `$this->moduleData` like this.

greg-1-anderson:
This is great, thanks! I'm not 100% sure about my comments, but some of the field access here looks potentially unsafe. If it would work equivalently to use local variables in those places, that would be preferable.

src/Drupal/DrupalKernel.php
NickWilde1990: hmmm... yes that does make sense; I wasn't considering that - I was thinking this only affects Drush. Ah well... quite easy to modify. I'll swap over to local/adjust the implementation. A bit more code/close to duplication but seems very worth it to not be altering core data without tons of testing
thanks for the review :)

greg-1-anderson: You could have less duplication by merging the module list and theme list, but I don't think you should do that either. You're not supposed to ever have a module and a theme with the same name, but I've seen that violated before, so best keep the lists separated (unless Drupal is actually preventing that now -- have not tried recently).

NickWilde1990: Yeah I've seen that before too, which should have made me think about keeping it separate

src/Drupal/DrupalKernel.php
```diff
@@ -60,6 +63,7 @@ protected function initializeContainer()
         if ($this->shouldDrushInvalidateContainer()) {
             $this->invalidateContainer();
         }
+        $this->classLoaderAddMultiplePsr4($this->getModuleNamespacesPsr4($this->getThemeFileNames()));
```
NickWilde1990: This has to be added *somewhere* to add the theme folders to the autoloader. I did a bit of investigating and I think this is the best spot but it's the item I'm least sure about.

greg-1-anderson: Did this happen as a side effect of adding the themes to the module list before?

Does the active theme have its autoloader added?

NickWilde1990: Didn't need it before because, one of the uses of `$this->moduleList` was propagating the autoloader.

All enabled themes get their autoloader added by this - in Drupal core, themes don't get added to the autoloader at all.

ericmulder1980:
Always great to find out there is a working patch for problems you come across. We really need this patch for the ZURB Foundation theme as creating subthemes currently does not work. More info at : https://www.drupal.org/node/2920472

I can confirm that the patch works. Any possible ETA on when this could be merged? Perhaps a beta-9 release?

src/Drupal/DrupalKernel.php
greg-1-anderson: This part is the hang up for me. I'm worried about the unpredictable consequences of Drush adding theme classes that Drupal does not add to the autoloader. Maybe this is no big deal; each theme should be uniquely namespaced, and if no one references the class, then it should not matter if it's autoloadable. I am concerned by the fact that this happens all the time when using Drush, whether or not a theme command is being executed. If this did cause a change in behavior, it could be really difficult to track down.

It might be safer to only add the autoloader when a theme command is executed. This is what the [Terminus plugin mechanism does](https://github.com/pantheon-systems/terminus/blob/1.6.1/src/Plugins/PluginAutoloadDependencies.php#L35:L46). The consequence of doing something similar for themes is that theme commands would have access to all of the classes defined in the theme, whether or not the theme was enabled, but hooks in the theme only have access to the command / hook classes in disabled themes (but should still have access to all classes in the enabled theme).

NickWilde1990: hmm... I'm 95% sure I can simply implement that with a bit of copy/paste from there and editing. Investigating, thanks :)

greg-1-anderson: Termins has an autoload file that it can include, which is probably not going to be the case here. You may need to instantiate your own class loader. The Composer class loader works well in this capacity. I've done so successfully in the past, but at the moment I can't quite remember where, or I'd provide a link.

weitzman:
The PR wont be merged as is. The next step is in https://github.com/drush-ops/drush/pull/3089#discussion_r149240981

NickWilde1990:
Finally got some time today to work on this more.

I didn't end up implementing a runtime autoload function for the theme namespaces. A quick investigation revealed that was going to be more complicated than I was first suspecting.
So I'm implemented it slight differently:
- *Only* themes with a Drush services yml will be added to the autoloader
- *Only* the theme/src/Commands folder is added to the autoloader.

I believe this should reduce the risk of accidental collisions sufficiently as well as reducing possibility for accidental effects. Anyways, I look forward to hearing your thoughts on whether that resolves your concerns.

src/Drupal/DrupalKernel.php
```diff
@@ -203,4 +214,77 @@ protected function addDrushServiceProvider($serviceProviderName, $serviceYmlPath
...
+     * @return array
+     *   Array where each key is a theme name, and each value is a path to the
+     *   respective *.info.yml file.
+     */
+    protected function getThemeFileNames()
+    {
+        if ($this->themeNames) {
+            return $this->themeNames;
+        }
+        $extensions = $this->getConfigStorage()->read('core.extension');
```
weitzman: Lets use a Drupal API like `\Drupal::service('theme_handler')->listInfo();` or `system_list('theme')`

src/Drupal/DrupalKernel.php
```diff
@@ -203,4 +214,77 @@ protected function addDrushServiceProvider($serviceProviderName, $serviceYmlPath
             $this->serviceYamls['app'][$serviceProviderName] = $serviceYmlPath;
         }
     }
+
+    /**
+     * populates theme data on the filesystem.
```
weitzman: Start with upper case.

src/Drupal/DrupalKernel.php
```diff
@@ -203,4 +214,77 @@ protected function addDrushServiceProvider($serviceProviderName, $serviceYmlPath
...
+     *
+     * @see Drupal\Core\DrupalKernel::moduleData().
+     */
+    protected function themeData($theme_list)
+    {
```
weitzman: I'm not too clear why we need this code. Why is iterating over all enabled themes sufficient? Are we wanting to catch the case where Drush integration is in the parent thee? What does simpletest have to do with anything?

src/Drupal/DrupalKernel.php
NickWilde1990: Fixed

src/Drupal/DrupalKernel.php
NickWilde1990: Can't do that unless we load core/includes/module.inc. At this point in the bootstrap process, it isn't yet which means `system_list()` is unavailable and `->listInfo()` uses it.

src/Drupal/DrupalKernel.php
NickWilde1990: I've simplified this slightly. This is necessary to populate the Drupal path cache so that drupal_get_path() and similar function work for themes. I think we can remove the simpletest stuff as we're not loading themes for testing - it is in the core `Drupal\Core\DrupalKernel::moduleData()` but reviewing it, we don't need it for this use case, so removed.

NickWilde1990:
@weitzman or @greg-1-anderson either of you have any time to give this a re-review with those changes?
thanks

weitzman:
I took a quick look and still dont feel especially comfortable with this. I'm very sensitive to maintenance/complexity costs these days. Lets see what @greg-1-anderson thainks.

weitzman:
Folks wanting this should work on modernizing Drupal's theme system. See https://www.drupal.org/project/drupal/issues/2002606, for example.

weitzman:
i'm going to cll this 'wont fix'. please work in drupal core to resolve this.

NickWilde1990:
Keep forgetting to comment back. I totally understand wanting to minimize stuff like this diverging from Drupal loader/maintainability etc and makes sense. Thanks for all the review and discussion. When I have time I'll check out that issue on the core queue you linked, thanks

PR #3095, opened 2017-10-23
When an update fails, updatedb is still exiting with 0

claudiu-cristea:


tests/updateDBTest.php
```diff
@@ -46,6 +49,42 @@ public function testUpdateDBStatus() {
...
+
+    // Force re-run of the post-update woot_post_update_failing().
+    $this->drush('sql:query', ["SELECT value FROM key_value WHERE collection = 'post_update' AND name = 'existing_updates'"], $options);
+    $functions = unserialize($this->getOutput());
+    unset($functions[array_search('woot_post_update_failing', $functions)]);
+    $functions = serialize($functions);
+    $this->drush('sql:query', ["UPDATE key_value SET value = '$functions' WHERE collection = 'post_update' AND name = 'existing_updates'"], $options);
+
+    // Run updates. woot_post_update_failing() is failing.
+    $return = $this->drush('updatedb', [], $options);
```
weitzman: Thanks!

$this->drush() asserts for a zero exit code. Once we are actually returning wrong code, we will have to pass a param at the end - \Unish\CommandUnishTestCase::EXIT_ERROR

claudiu-cristea:
@weitzman the problem is in `\Drush\Commands\core\UpdateDBCommands::updateBatch()` where the batch 'finished' is set to an non existent method (probably left there from Drush 8). But if I change that to `\Drush\Commands\core\UpdateDBCommands::updateFinished()` that is making the whole batch process to run while is exhausting the memory. `::updateFinished()` is called again and again in `_drush_backend_batch_process()` here:

```php
    while (!$finished) {
      $data = drush_invoke_process('@self', $command, $args);

      $finished = drush_get_error() || !$data || (isset($data['context']['drush_batch_process_finished']) && $data['context']['drush_batch_process_finished'] == TRUE);
    }
```
because, I think, `$data['context']` is empty. Have no idea how to fix this.

weitzman:
Do we know if this worked before my refactor last week? 

The batch code is hard to follow. I may not be able to dig into this soon.

claudiu-cristea:
@weitzman it surely worked with Drush 8 but with 9 I don't know. We switched only for few days. Bunt now this is critical with Drupal 8.4.0 because we rely on the exit code on CI.

claudiu-cristea:
@weitzman

> Do we know if this worked before my refactor last week?

I went back and restored the `UpdateDBCommands` file from 73659a338370d8d258c899ccda17a3ab5a8ab287 which is 17 days ago. I tried all file versions in this interval. With all, this test is timing-out, running `woot_update_8101()` indefinitely . I cannot go back too much as the old versions are using removed functions. Seems an old issue, I wonder why it hasn't been discovered -- probably because is so silent.

weitzman:
Fixed in #3188. We used much of the code from here. Thanks.

PR #3100, opened 2017-10-25
Mark the update as complete when finished is more than 1

jerbob92:
I also created a patch for this in D8 Core:
https://www.drupal.org/node/2918613


weitzman:
Reopen If Drupal merges it's issue

PR #3101, opened 2017-10-25
When drush is in a vendor dir, try to automatically find root.

weitzman:


src/Preflight/Preflight.php
```diff
@@ -267,7 +267,11 @@ protected function findSelectedSite()
         // TODO: If we want to support ONLY site-local Drush (which is
         // DIFFERENT than --local), then skip the call to `$preflightArgs->selectedSite`
         // and just assign `false` to $selectedRoot.
-        $selectedRoot = $this->preflightArgs->selectedSite($this->environment->cwd());
+
+        // Try two approaches.
+        if (!$selectedRoot = $this->preflightArgs->selectedSite(DRUSH_COMMAND)) {
+            $selectedRoot = $this->preflightArgs->selectedSite($this->environment->cwd());
+        }
         return $this->setSelectedSite($selectedRoot, $this->environment->vendorPath());
```
greg-1-anderson: `$this->environment->vendorPath()` is used later to fall back to finding the Drupal site from the vendor directory. If this isn't working correctly in the SUT (e.g. due to symlinking), then perhaps change this expression to something else. For example, maybe:
```
$selectedRoot = $this->preflightArgs->selectedSite($this->environment->cwd());
$fallbackPath = $this->preflightArgs->selectedSite(DRUSH_COMMAND);
return $this->setSelectedSite($selectedRoot, $fallbackPath)
```
I haven't tried this, but that's how `setSelectedSite()` is supposed to work, anyway. "Selected root" is the root the user selected via cwd et. al., and the fallback path is information used to find the site from other information e.g. vendor path.

PR #3108, opened 2017-10-27
Fix typo and strangely formatted line about global commands

eiriksm:
Hello.

I was curious as to why a global command should have the namespace `Drush\Commands[dir-name]` (notice the missing backslash), when I noticed the source actually contained this said character. So here is a PR making it forced code block, so it actually shows :)

The other change is a typo for what you should read about. There is no `docs-configuration`, only `docs:configuration`


greg-1-anderson:
Thanks

PR #3110, opened 2017-10-27
WIP: Add recursion detection to RedispatchToSiteLocal

webflo:
Add recursion detection to RedispatchToSiteLocal to mitigate faulty configurations with site-local Drush.

Related to https://github.com/webflo/drupal-finder/issues/34

### DIFF
--- a/src/Preflight/RedispatchToSiteLocal.php
+++ b/src/Preflight/RedispatchToSiteLocal.php
@@ -2,11 +2,6 @@
 
 namespace Drush\Preflight;
 
-use Consolidation\AnnotatedCommand\Hooks\InitializeHookInterface;
-use Drush\Drush;
-use Symfony\Component\Console\Input\InputInterface;
-use Consolidation\AnnotatedCommand\AnnotationData;
-use Drush\Log\LogLevel;
 use Webmozart\PathUtil\Path;
 
 /**
@@ -18,6 +13,8 @@ use Webmozart\PathUtil\Path;
  */
 class RedispatchToSiteLocal
 {
+    const REDISPATCH_RECURSION_LIMIT = 5;
+
     /**
      * Determine if a local redispatch is needed, and do so if it is.
      *
@@ -54,7 +51,11 @@ class RedispatchToSiteLocal
         }
 
         // Redispatch!
-        $command = $siteLocalDrush;
+        $recursion_limit = (int) getenv('DRUSH_REDISPATCH_RECURSION_LIMIT');
+        if ($recursion_limit > static::REDISPATCH_RECURSION_LIMIT) {
+            throw new \Exception('Redispatch limit reached. Drush could not bootstrap the site-local Drush. Please review your project setup.');
+        }
+        $recursion_limit++;
         array_shift($argv);
         $args = array_map(
             function ($item) {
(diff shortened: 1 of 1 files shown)


webflo:
I think the fix is necessary because the logic in ``::redispatchIfSiteLocalDrush`` and ``::findSiteLocalDrush`` is not bullet proof again wrong configurations. Here is one example (https://gist.github.com/webflo/fe2105bf0ebbd22f234e4ea12e867ba2) how the whole setup blows if people convert their site from default drupal layout to drupal-project and don't cleanup everything properly.

Drush picks up ``../vendor/drush/drush/drush`` as site-local Drush even if this composer autoloader is not used in the Drupal installation.

greg-1-anderson:
The recursion limit here should be one -- Drush should never redispatch to a site-local Drush after it has already redispatched to a site-local Drush. You might get two exec's if there is a finder/launcher starting things off, but ideally Drush should always get the redispatch right on the first try.

Drush allows us to pass the vendor directory in as an environment variable. Maybe we should always do this if there is a site-local redispatch, and never redispatch again if this environment variable has been set.

Regarding confusion over whether the vendor directory is part of the current site or not, do we need to examine Drupal's `__ROOT__/autoload.php` to do this correctly? I'd rather not have to parse this file, but I'm not sure how to detect if `__ROOT__/../vendor` is not part of the Drupal site.

webflo:
Getting from ``autoload.php`` back to the vendor folder is in general very hard and error prone. Because composer does not provide the metadata, because it should not matter on runtime. ``drupal-finder`` does various things but avoids parsing/loading the ``autoload.php`` (see https://github.com/webflo/drupal-finder/blob/master/src/DrupalFinder.php#L88).

I think its better to keep current code the complexity is already very high. It not possible to support every configuration and especially if its broken or misconfigured.

webflo:
Decreasing the limit is a good idea. Maybe we could display a log message if drush is invoked with ``--debug``. Currently its not possible to see that redispatching is happening.

segi:
I was on of person who reported the infinite loop problem. In my situation and other fix would have been enough. 
If we add realpath() here like:
``` php
// If the site-local Drush is us, then we do not need to redispatch.
if (Path::isBasePath(realpath($vendor), realpath($siteLocalDrush))) {
  return false;
}
```
but maybe this solution is more general and it can prevent more problem.

weitzman:
I agree that it would be good to log a message when we redispatch. We can't right now, as the logger is not available yet. We should fix this ASAP. Would it be ok to just write to stderr for now?

greg-1-anderson:
Ideally we would like to log preflight messages in `--debug` mode, and do nothing when debug is not set. At the moment, `--debug` isn't captured by preflight args. We could set debug in preflight args, and then maybe create a PreflightLog class that writes to stderr iff in debug mode. We could manually inject PreflightLog into RedispatchToSiteLocal.

I would say that PreflightLog should not be derived from psr/log.

jurgenhaas:
@weitzman thanks for referring me over from #3143 - I was having a look and it sounds that my issue is similar to what @segi describes. Using realpath() is what I assumed to be the correct way forward as well, when I analysed my other issue. What kind of help is required here to move this forward?

weitzman:
Added a logger per @greg-1-anderson  mini spec - https://github.com/drush-ops/drush/pull/3154

grasmash:
This issue is a blocker for one of our products. In our case, merging #3235 is sufficient to resolve the issue, as @segi indicated.

weitzman:
#3235 fixed this i think

LichP:
I came across this issue while experiencing a drush recursion which was due to a symlinked vendor folder. I'm using a current version of drush with #3235 applied - in my case applying the fix from @segi 's comment above resolved the problem. This use case is pretty niche so probably doesn't warrant reopening this issue - just posting this for reference.

PR #3115, opened 2017-10-29
Fix #3063. Add example of dynamic alias alter

### LINKED ISSUE #3063: Dynamically set options, aliases
In Drush 8, I could conditionally set configuration options:

```
if (getenv('PROBO_ENVIRONMENT') && getenv('BUILD_DOMAIN')) {
  $options['uri'] = $_ENV['BUILD_DOMAIN'];
}
if (getenv('TUGBOAT_URL')) {
  $options['uri'] = $_ENV['TUGBOAT_URL'];
}
// Include current directory. Will add policy.drush.inc.
$options['include'][] = __DIR__;
// If we are on Acquia Cloud, add Acquia specific commands.
if (file_exists('/usr/local/drush8/commands')) {
  $options['include'][] = '/usr/local/drush8/commands';
}
```
Now that Drush 9 uses drush.yml in place of drushrc.php, how can this be accomplished?

Similarly, how might one dynamically set aliases?

```
// Local environment.
$aliases['${project.machine_name}.local'] = array(
  'root' => '/var/www/${project.machine_name}/docroot',
  'uri' => '${project.local.uri}',
  );
// Add remote connection options when alias is used outside VM.
if ('vagrant' != $_SERVER['USER']) {
  $aliases['${project.machine_name}.local'] += array(
    'remote-host' => '${project.local.hostname}',
    'remote-user' => 'vagrant',
    'ssh-options' => '-o PasswordAuthentication=no -i ' . drush_server_home() . '/.vagrant.d/insecure_private_key'
  );
}

```

weitzman:


greg-1-anderson:
Already works. How about that.

examples/Commands/SiteAliasAlterCommands.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
rudolfbyker, line 1: Where would one place this file? I couldn't find any documentation on this.

PR #3122, opened 2017-10-31
Make it possible to set the uri option in drush.yml. Fixes #3121.

### LINKED ISSUE #3121: Setting uri in drush.yml does not set the base uri
Added a `drush.yml` file to all of these location, containing the following:

```
options:
  uri: 'http://devsite.dev'
```

Locations:

- ./drush.yml
- ./drush/drush.yml
- ./drupal/drush/drush.yml
- ./drupal/sites/default/drush.yml

Based on the example.drush.yml file: https://github.com/drush-ops/drush/blob/master/examples/example.drush.yml#L92

Expected result:

The uri should be set, for example for `uli` commands.

eiriksm:


src/Application.php
```diff
@@ -162,7 +162,9 @@ public function refineUriSelection($cwd)
         $uri = $selfAliasRecord->uri();
 
         if (empty($uri)) {
-            $uri = $this->bootstrapManager()->selectUri($cwd);
+            if (!$uri = $this->config->get('options.uri')) {
```
greg-1-anderson: Style nit: factor this out into `$this->selectUri($cwd)`. Avoid putting assignments inside of `if` statements. Use `return $uri` to exit the selectUri() method early if you get a valid from config, and then otherwise `return $this->bootstrapManager()->selectUri($cwd);`

greg-1-anderson:
Looks like your change is interfering with bootstrapping. Maybe config is returning a value for 'uri' that is evaluating as `true` even though it is not useful (wrong) vis-a-vis bootstrapping.

eiriksm:
Got it. Good starting point :) Will try to look at it tomorrow

eiriksm:
Refactored to a method.

This should hopefully pass now. This should definitely have a test, though. Is there another test that deals with config somewhere?

greg-1-anderson:
The best place to put the test would be to make another test similar to `testSiteSelectionViaCwd` in coreTest.php.

src/SiteAlias/SiteAliasManager.php
```diff
@@ -252,13 +252,18 @@ protected function buildSelf(PreflightArgsInterface $preflightArgs, Environment
...
+        // alias or site specification used by the user. Also note that if we
+        // pass in a falsy uri the drush config (i.e drush.yml) can not override
+        // it.
+        $uri = $preflightArgs->uri();
+        $data = [
+            'root' => $root,
+            'uri' => $uri,
+        ];
+        if (!$uri) {
+            unset($data['uri']);
```
greg-1-anderson: Another minor style point: I'd rather see 'uri' set conditionally than unset conditionally. If there were a number of such elements, then I might suggest `$data = array_filter($data);` might be better, but in this case I'd favor `if ($uri) { $data['uri'] = $uri; }`

weitzman:
This is working for a uri specified in my site's /drush/drush.yml. An automated test would still be appreciated.

eiriksm:
And needed. I don't have time to look at it today, but I will try to make time tomorrow

eiriksm:
OK, added a test for it as well.

Interestingly enough, it still is not possible to set URI in sites/default/drush.yml, but we can look at that in a follow up I guess. It seems config is not loaded at all from that directory, even though the docs say otherwise. At least now it's possible to override, and that's a start.

greg-1-anderson:
Test looks good, once it is passing. Looks like you need to `@mkdir` the `drush` directory.

Also, note that we re-use the Site-Under-Test in the functional tests. You should delete your config file after you run the drush command (but before asserting) to ensure that it is cleaned up.

eiriksm:
OK, let's see if this passes, then :)

greg-1-anderson:
LGTM. Updating branch with master.

tests/coreTest.php
```diff
@@ -94,4 +95,23 @@ function testSiteSelectionViaCwd() {
     }
     chdir($cwd);
   }
+
+  function testOptionsUri() {
+    // Put a yml file in the drush folder.
+    $drush_directory = sprintf('%s/drush', $this->webroot());
+    $drush_config_file = sprintf('%s/drush.yml', $drush_directory);
```
weitzman: whats the benefit of using sprintf here instead of concatenation? Or`Path::join($this->webroot(), 'drush/drush.yml');` Thats the webmozart/path library .. .Actually, if you use the ../drush, you have an equivalent directory that already exists. Thats where sut.alias.yml lives. Its one directory above drupal root.

tests/coreTest.php
```diff
@@ -94,4 +95,23 @@ function testSiteSelectionViaCwd() {
...
+    $test_uri = 'http://test.uri';
+    $options_with_uri = [
+      'options' => [
+        'uri' => $test_uri,
+      ],
+    ];
+    @mkdir($drush_directory);
```
weitzman: We usually use $this->mkdir()

tests/coreTest.php
```diff
@@ -94,4 +95,23 @@ function testSiteSelectionViaCwd() {
...
+    file_put_contents($drush_config_file, Yaml::dump($options_with_uri, PHP_INT_MAX, 2));
+    $this->drush('core-status');
```
weitzman: Don't you need to specify a root and uri for this to work? Thats how we do this all over the test suite. Without that Drush can't find the 'dev' site.

Also, its generally best to send --format=json and then you you dont a regex to get the data you need out. See \Unish\userCase::testBlockUnblock, for an example.

greg-1-anderson: That's a good question. I assumed the test was chdir'ing like `testSiteSelectionViaCwd` does. Shouldn't work without some sort of initial site selector. Maybe the chdir from the previous test persists?

eiriksm: Well, the output contains the expected URI. I guess if cwd is "project root" it finds it anyway?

I can pass in the root as the other tests of course, to make it more uniform.

Good tip about the json format! The reason I was regexing was because I was looking at the test above this one :)

eiriksm:
Thanks for feedback, updated per comments.

tests/coreTest.php
eiriksm: Hm, not sure why my comment was not added here.

I think I wrote something along the lines of this:

There is no real benefit, it is just a matter of taste and coding style. Will update it. :)

Also, great suggestion about the root drush folder. Will use that.

tests/coreTest.php
eiriksm: Ah thanks, did not know this. However, I removed this instead after re-using the root drush dir

tests/coreTest.php
```diff
@@ -94,4 +95,24 @@ function testSiteSelectionViaCwd() {
...
+    $options_with_uri = [
+      'options' => [
+        'uri' => $test_uri,
+      ],
+    ];
+    $options = [
+      'format' => 'json',
+    ];
+    file_put_contents($drush_config_file, Yaml::dump($options_with_uri, PHP_INT_MAX, 2));
+    $this->drush('core-status', [], $options);
```
weitzman: You typically have to specify --root option in order for a drush() call to work. This is the second issue where I have seen PRs omit that and seemingly the test passes. Not sure whats going on but I'd like to find out.

weitzman:
Merged. Thanks so much for fixing this. I wanted it as well.

I looked into the lack of --root and --uri in the test.

1. Because of https://github.com/drush-ops/drush/pull/3101, Drush is now autolocating root since we call Drush from within the vendor dir. So all our tests can now stop setting --root explicitly.
1. This test doesn't need to set --uri because of the feature it adds. All other tests still need to set it.

We should consider removing these options from most tests and auto-choosing 'dev' as uri unless otherwise specified.

eiriksm:
I figured it had to be something like that :)

Thanks for merging, keep up the good work!

PR #3124, opened 2017-10-31
#3123 - Global D8 container is not available during container rebuild.

### LINKED ISSUE #3123: Global D8 container is not available when rebuilding the container
Typically when the container is being rebuilt, it's within the context of a normal Drupal page request. At that point, Drupal has already booted and the global container is available, leaving services available for use in certain other steps, like defining events that your `EventSubscriber` subscribes to. Unfortunately, Drush's `DrupalKernel` will immediately rebuild the container without having the global container initialized with the existing one.

The particular use case we've run into that is causing us headaches is the Rules module (though there may be others). It utilizes the State API to determine which events one of its `EventSubscriber`s should subscribe to. If the global container is not yet available, then it can't get the state service to subscribe to the correct events and then the rules you expect to trigger never do anything.

kevin-dutra:
Here's the suggested change to fix #3123


greg-1-anderson:
This looks compelling, but I don't have the background to know if this is the right thing vis-a-vis Drupal's expectations & your comments in #3123. Maybe @weitzman knows better.

Do you think that perhaps you could provide a test that passes with this PR, but fails without it?

weitzman:
Unfortunately, I dont know.

How fast o slow slow is this new code? Trying to evaluate any perf impact.

kevin-dutra:
In terms of performance, the added code does not adds very little overhead, and it only runs when the container is already being rebuilt, which in theory should not be too often.

As for a test, I'll have to think about that a little. It would be tricky because we'd need to somehow get the container built within the normal context of a Drupal request so that the Drush add-ons were **not** there. That way the next Drush command that gets run would trigger the container rebuild to add the Drush pieces in (thereby executing the new code).

kevin-dutra:
Alrighty, I managed to craft a test that passes with the change and fails without it.

jhedstrom:
Looks like this needs to merge/rebase the latest `master`.

This is the related Rules issue: https://www.drupal.org/node/2816033

tests/containerTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Tests that the existing container is available while Drush rebuilds it.
+     */
+    public function testContainer() {
+        $sites = $this->setUpDrupal(1, TRUE);
+        $root = $this->webroot();
+        $options = array(
+            'root' => $root,
+            'uri' => key($sites),
```
weitzman: We now use $this->getUri() instead.

PR #3126, opened 2017-11-01
Replace tilde for PreflightArgs.

weitzman:


src/Utils/StringUtils.php
```diff
@@ -77,4 +79,22 @@ private static function interpolationKey($key)
...
+     * Replace tilde in a path with the HOME directory.
+     *
+     * @param $path
+     *   A path that may contain a ~ at front.
+     *
+     * @return string
+     *   The path with tilde replaced, if applicable.
+     */
+    public static function replaceTilde($path) {
+        $replacement = Drush::config()->get('env.home') . '/';
```
greg-1-anderson: Drush::config() is not available until after the dependency injection container is set up, which happens at the end of preflight. The upshot is that you can't use the static accessor pattern for any object that has dependencies.

You can get around this limitation by removing the dependency of the `replaceTilde` method on config by passing in the replacement value (`env.home`) as a parameter. The tilde hook could get this value from Config, and the calls in preflight could get it from the Environment object.

greg-1-anderson:
Sometimes replacing the tilde during preflight can be destructive, e.g. if the command is remote, and $HOME is different on the target machine.  The limited replacements you have done here seem like they're probably okay, but we should be careful.

PR #3132, opened 2017-11-02
Allow spaces in PHP override values

### LINKED ISSUE #3129: drush.launcher removes spaces from drush.ini settings
`drush.launcher` will remove spaces from values set in `drush.ini`. This caught me by surprise because I didn't know there were multiple launchers. Now I'm using `drush.php` which works fine. I was using `drush` finder which was then calling `drush.launcher`. Drush is installed via composer using the drupal-composer/drupal-project project using the 7.x branch.

I did spend some time trying to debug the `drush.launcher` script and found the problem line to be https://github.com/drush-ops/drush/blob/8.x/drush.launcher#L119 although no matter what shell tricks I tried to use, I couldn't figure out how to produce a `exec` command that would work as expected.

If the code is left as is, then spaces will be removed. For example I had `sendmail_path='/usr/bin/msmtp -a myvhost -t'` set in `drush.ini` and it would be changed to `sendmail_path='/usr/bin/msmtp-amyvhost-t'`. If I left the space in, then the `exec` command will fail a few lines down in the script.

dmp1ce:
Should fix #3129

### DIFF
--- a/drush.launcher
+++ b/drush.launcher
@@ -116,7 +116,7 @@ if [ -n "$drush_php_ini" ] ; then
   php_options="--php-ini $drush_php_ini"
 fi
 if [ -n "$drush_php_override" ] ; then
-  php_options=`grep '^[a-z_A-Z0-9.]\+ *=' $drush_php_override | sed -e 's|\([^ =]*\) *= *\(.*\)|\1="\2"|' -e 's| ||g' -e 's|^|-d |' | tr '\n\r' '  '`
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Given that Drush 9 no longer supports drush.ini, and Drush 8 is in maintenance fixes only, I'm reluctant to accept this feature. Up to @weitzman tho.

weitzman:
Agreed. Thanks still for posting it. Its here for those who need it.

PR #3134, opened 2017-11-02
Fix a typo in selectUri method docs.

eiriksm:
Just noticed this typo in a doc comment while going through the code.


greg-1-anderson:
Not a typo. "iff" is an abbreviation for "if and only if"

eiriksm:
Wow, did not know that :)

greg-1-anderson:
I'm old. Maybe no one uses that any more. 😝

PR #3135, opened 2017-11-02
Fix #3131. Stop using further() in sql commands. Use @bootstrap max instead.

### LINKED ISSUE #3131: @validate-file-exists doesn't seem to take --root into account
I tried `drush @docker sql-query --file=../private/$DUMP` and got `[error]  File(s) not found: ../private/2017-11-02_17:12.sql `
The same command work if I execute it directly inside the container, without any alias.
After some search, I found out that if I remove "@validate-file-exists file" from SQLCommands.php function query(), it works. I tried a getcwd() in validateFileExists(), before the call to file_exists() and got the wrong dir. I think it's because validateFileExists() doesn't take --root into account, which is added when the command is executed through SSH.

weitzman:


greg-1-anderson:
I think that setting `cwd` during preflight would be a preferable fix.

PR #3137, opened 2017-11-03
Move canonical storage of tmp dir to config()->get(env.tmp)

weitzman:
As a next step, I was thinking of moving the following functions to a static methods on new FileUtils class, in same directory as StringUtils. Feedback welcome on that.

drush_save_data_to_temp_file()
drush_tempnam()
drush_tempdir()
drush_register_file_for_deletion()
_drush_delete_registered_files()

### DIFF
--- a/src/Sql/SqlBase.php
+++ b/src/Sql/SqlBase.php
@@ -5,12 +5,15 @@ namespace Drush\Sql;
 use Drupal\Core\Database\Database;
 use Drush\Drush;
 use Drush\Log\LogLevel;
+use Robo\Common\ConfigAwareTrait;
+use Robo\Contract\ConfigAwareInterface;
 use Webmozart\PathUtil\Path;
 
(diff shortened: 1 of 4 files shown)


src/Sql/SqlBase.php
```diff
@@ -5,12 +5,15 @@
 use Drupal\Core\Database\Database;
 use Drush\Drush;
 use Drush\Log\LogLevel;
+use Robo\Common\ConfigAwareTrait;
+use Robo\Contract\ConfigAwareInterface;
 use Webmozart\PathUtil\Path;
 
-class SqlBase
+class SqlBase implements ConfigAwareInterface
```
greg-1-anderson: SqlBase is neither a commandfile, nor is it created from the DI container. It will therefore be necessary to manually inject the config manager in any code that creates an instance of a class derived from SqlBase.

PR #3139, opened 2017-11-05
Plan for the future (see #3130): Allow for phpunit 5.x.

### LINKED ISSUE #3130: Change the sebastian/version dependeny
For some development work I wanted to install dealerdirect/qa-tools on my D8 site which currently also has Drush 9 beta8. Both can not be installed together because of a version conflicht regarding sebastian/version:

- Drush requires version ~1
- dealerdirect/qa-tools requires version ^2.0

I couldn't find any issues with the 2.x version and wonder if Drush could change their requirement to "~1 | ^2" to allow both and not breaking any other dependencies?

greg-1-anderson:
I started to fix #3130, but then stopped when I realized that mixing Drush with Drupal prevents us from using phpunit 5.x, and erg sebastion/version 2.x, making the upgrade pointless.

However, I realized that at some point, Drupal will upgrade to phpunit 5.x.  While we cannot anticipate every upgrade that Drupal might make, phpunit 4.x is very old, so this is probably overdue, or at least likely to happen eventually.

This PR sets up highest / lowest testing for the isolation tests, and allows the isolation tests to run across different versions of phpunit and sebastion/version. This will give us the flexibility to test with multiple versions of other libraries in the future, if we so desire.

The value for Drush being ahead of the game on upgrades is that it allows newer versions of Drupal to be used with more versions of Drush, making `composer update` easier to get through when there are minor Drupal upgrades. It behoves us to do this due to the limitations in Composer that can come into play when the dependency analysis of an upgrade is complicated (e.g. as it was for the upgrade to Drupal 8.4.0).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,9 @@ sudo: false
 
 env:
   matrix:
-    - TEST_DIR=isolation
+    - TEST_DIR=isolation ISOLATION_INSTALL='install --prefer-dist'
+    - TEST_DIR=isolation ISOLATION_INSTALL='update --prefer-dist --prefer-lowest'
+    - TEST_DIR=isolation ISOLATION_INSTALL='update --prefer-dist'
     - PHPUNIT_ARGS=--group=base
     - PHPUNIT_ARGS=--group=commands
     - PHPUNIT_ARGS=--exclude-group=base,commands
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
While there are more tests here, the isolation tests run really quickly (~1m each, in parallel), and the value of highest / lowest is large.

PR #3146, opened 2017-11-06
Stop setting root/uri/n during Unish::drush() calls

weitzman:


tests/CommandUnishTestCase.php
```diff
@@ -270,6 +271,9 @@ function drush($command, array $args = array(), array $options = array(), $site_
           $hide_stderr = TRUE;
           $value = NULL;
         }
+        if ($key == 'uri' && $value == 'OMIT') {
```
greg-1-anderson: Maybe `=== false` would be better?

weitzman: That would work. But I get confused between NULL and FALSE for avlues since they each works like the other in drush_invoke_process() versus Unish::drush() . So I chose an ugly but obvious alternative. Its only used twice on our source code so no big deal.

PR #3149, opened 2017-11-06
Add DrushConfig class with accessors for common keys

weitzman:


greg-1-anderson:
Instead of `DrushConfig::tmp();`, we should use `Drush::config()->tmp();` & c.

This would allow folks who are using dependency injection access to the convenience functions. Static classes should be provided to make it faster to port code that isn't using DI yet; we shouldn't encourage folks to use static classes over DI.

greg-1-anderson:
Our config object is instantiated here:
https://github.com/drush-ops/drush/blob/9.0.0-beta8/src/Config/ConfigLocator.php#L96

Just create a DrushConfig instance (non-static) here instead. Make DrushConfig extend ConfigOverlay, and add all the accessors you want.  DI will happen as it currently does, and all of the accessors will be available.

weitzman:
Good idea. I've updated the PR. A few thoughts.

1. Should the new class be called Config or DrushConfig. I think Config is slightly better.
1. I changed the Doxygen for Drupal::config() to return DrushConfig so that the new accessors show up for completion. OK?
1. $this->getConfig() is type hinted for ConfigInterface so they new accessors are not shown in auto-complete. Just accept that?
1. Commands wishing to use this now have to become ConfigAware. Seems OK.

greg-1-anderson:
>1. Should the new class be called Config, ConfigOverlay, or DrushConfig?

DrushConfig seems best to me.

>2. I changed the Doxygen for Drupal::config() to return DrushConfig so that the new accessors show up for completion. OK?

Sounds great.

>3. $this->getConfig() is type hinted for ConfigInterface so the new accessors are not shown in IDE auto-complete. Just accept that? We have same problem with the DrupalBoot classes that extend a simpler BaseBoot.

Since there is only one kind of Config object in all of Drush, we could create `interface DrushConfigAwareInterface extends ConfigAwareInterface`, and typehint getConfig() to DrushConfig.  I don't think we would need to make a `DrushConfigAwareTrait` if we did this, but maybe we would.  Not sure if it's worth doing or not. Doesn't seem like so much extra overhead, and the IDE completion is valuable.

I'd rather keep BaseBoot as it is, though. I still have the aspirational goal of supporting multiple boot classes in the bootstrap manager. There are fewer clients who will need to call methods of the boot object, and these should only be calling methods in the interface. Otherwise the generalization won't work, e.g. if we need a new boot for Drupal 9.

>4. Commands wishing to use this now have to become ConfigAware. Seems OK.

Yeah, just use where needed. I think it's better to make commands declare what they use, and only add interfaces to DrushCommands that are required for all commands.

weitzman:
I made a made a DrushConfigAwareInterface and changed a bunch of classes to use it. I did not add a new Trait yet.

Unfortunately, this hasn't helped with IDE autocomplete of `$this->getConfig()->`. I think the IDE is using the @return from \Robo\Common\ConfigAwareTrait::getConfig which points to regular Config, not DrushConfig. So my reading is that we do need to create a new DrushConfigAwareTrait. I think [this SO answer](https://stackoverflow.com/a/37687295) shows how I would reuse all but one method from ConfigAwareTrait.

greg-1-anderson:
If the IDE is using the implementation in the trait over the declaration in the interface, then maybe we don't need the interface. I'd try both first to see how the IDE reacts, then try taking out the interface and see if it still works. Kind of a bummer that it does not respect the annotation in the interface, as it's a little less convenient to make a trait that overrides another trait (as your link above shows).

If we added ConfigAwareTrait to DrushCommand, then we could simply override `getConfig()` in DrushCommand, as PHP trait methods can be overridden by the command that `use`s them. Despite this convenience, I don't think that we should add ConfigAwareTrait to DrushCommand. I'm willing to be flexible on this point if you think most commands are going to need it.

weitzman:
I ran the experiment. Here is the autocomplete success rate.

1. 👎  Just the interface
1. 👍  Both interface and trait: 
1. 👍  Just the trait

I'm worried about the developer complexity that comes along with conditional rules like "When you need to access config, use DrushConfigAwareTrait but dont implement any interface. Or, implement DrushConfigAwareTrait with ConfigAwareInterface. Or, we make a "useless" DrushConfigAwareInterface just to match the trait.

All in all, I'm leaning toward your proposal to implement ConfigAwareInterface in DrushCommands. Its terrific from a DX perspective, with relatively little "bloat" for commands that dont need Config.

greg-1-anderson:
I'm fine with just adding DrushConfigAwareTrait to DrushCommands. Then derived commands don't need to worry about which trait to implement.

src/Commands/DrushCommands.php
```diff
@@ -47,6 +50,16 @@ protected function io()
     }
 
     /**
+     * Overrides same method in ConfigAwareTrait in order to provide a
+     * DrushConfig as return type. Helps with IDE completion.
+     *
+     * @return \Drush\Config\DrushConfig
```
weitzman: Pushed a commit which adds ConfigAware to DrushCommands. Everything works, except autocomplete! I feel like I'm learning via self flaggelation. 

Is there anything more I should be doing in in DrushCommands::getConfig or elsewhere.

greg-1-anderson: Maybe the IDE does not agree that you can override trait methods in classes (even though php itself seems to be of the mind that this is okay).  Do the use-trait-with-renamed-method trick, as if you were overriding one trait with another (move the parent getConfig out of the way) and see if that helps.

I'd say this is an IDE bug, but we should work around it if possible.

PR #3152, opened 2017-11-08
Use cbf to make tests/ PSR2 compliant.

weitzman:


tests/annotatedCommandTest.php
```diff
@@ -9,70 +9,72 @@
...
+
+        // In theory this is not necessary, but this test keeps failing.
+        // $this->drush('cc', array('drush'), $options);
+
+        // Make sure that modules can supply DCG Generators and they work.
+        $optionsExample['answers'] = json_encode([
+        'name' => 'foo',
+        'machine_name' => 'bar',
+        ]);
+        $optionsExample['directory'] = self::getSandbox();
```
weitzman: Anyone know if we can convince phpcs to preserve indentation on these array elements?

greg-1-anderson:
In general, I am against style-checking the tests. It's fine to run cbf on them to make them comply, but I usually don't make test style variations failures.

I know plenty of projects that do style-check the tests, though, so I won't object if you want to move forward here.

PR #3154, opened 2017-11-09
Add Prefllight logger. Use in RedispatchSiteLocal. Refs #3110

weitzman:


src/Preflight/Preflight.php
```diff
@@ -60,6 +65,22 @@ public function __construct(Environment $environment, $verify = null, $configLoc
...
+     */
+    public function logger()
+    {
+        return $this->logger;
+    }
+
+    /**
+     * @param PreflightLog $logger
+     */
+    public function setLogger($logger)
```
greg-1-anderson: You can typehint this too

src/Preflight/PreflightLog.php
```diff
@@ -0,0 +1,31 @@
...
+     */
+    public function setDebug($debug)
+    {
+        $this->debug = $debug;
+    }
+
+    public function log($message)
+    {
+        if ($this->getDebug()) {
+            file_put_contents('php://stderr', ' [preflight] ' . $message . "\n", FILE_APPEND);
```
greg-1-anderson: Maybe `fwrite` is clearer

src/Preflight/PreflightArgs.php
```diff
@@ -227,6 +231,11 @@ public function selectedSite($default = false)
         return $this->get(self::ROOT, $default);
     }
 
+    public function setDebug($value)
+    {
+        $this->set(self::DEBUG, $value);
```
greg-1-anderson: Sort of wondering if we need to set the debug flag in the pass through args too.

src/Preflight/Preflight.php
```diff
@@ -202,6 +223,11 @@ public function preflight($argv)
         $this->preflightArgs = $this->preflightArgs($argv);
         $this->prepareConfig($this->environment);
 
+        // Setup our early logger.
+        $logger = new PreflightLog();
+        $logger->setDebug($this->preflightArgs->get(PreflightArgs::DEBUG));
+        $this->setLogger($logger);
```
greg-1-anderson: Usually DI happens from outside the class. Not sure if we want to add another object instantiation to drush_main (or whatever we're calling it) tho

weitzman: Not sure I follow. This is Preflight and PreflightLog. Two different classes.

src/Preflight/PreflightArgs.php
weitzman: The --debug option is declared in Application.php so nothing more to do, if I understand your comment correctly.

weitzman:
I merged this, after implementing the feedback that I understand. We can do more in follow ups.

src/Preflight/Preflight.php
greg-1-anderson: Yes, `--debug` is defined as a global option; however, if you run `drush status --debug`, it no longer turns on debug mode for most of Drush. The value is only set for preflight.

src/Preflight/PreflightArgs.php
greg-1-anderson: Yes, `--debug` is defined as a global option; however, if you run `drush status --debug`, it no longer turns on debug mode for most of Drush. The value is only set for preflight.

weitzman:
Just pushed a couple commits which fix DI, fixes standard debug log, and uses the new PreflightLog in 3 places.

PR #3158, opened 2017-11-10
Validate legacy command file

Chi-teck:
Setting wrong file path in `drush generate dcf` causes a fatal error.
"PHP Fatal error:  require_once(): Failed opening required 'wrong file path'"

Throwing exceptions would allow console application fail gracefully.

This PR also handle the case when the provided file has no valid Drush command hook.


Chi-teck:
Throwing exception in question validator callback allows to type the answer again.
dcf-validation

PR #3159, opened 2017-11-10
Fixed call to legacy bootstrap function.

danepowell:
This commit removed `drush_bootstrap_max()` but it was still being called elsewhere.
https://github.com/drush-ops/drush/commit/440be17aceeacef9308db747bd17e874883cb2ef


weitzman:
Sorry for the oversight. And thanks for the fix.

PR #3166, opened 2017-11-14
Fix incorrect script name when Drupal uses a domain + subdir

Chi-teck:
This only happens when Drupal is installed in subdirectory and site URL is passed by means global uri option like follows.
`drush <CMD> --uri=http://localhost/subdirectory`

Because of wrong script name some Drupal global variables (_$base_path_ and _$base_url_) get wrong values.


Chi-teck:
The URLs generated by Drupal core Url class are still wrong though.

Chi-teck:
The simplest way to reproduce the issue is to put the following code to some command and call it with `uri=http://llocalhost/path/to/drupal/` option.
```
$this->io()->write((new Url('system.admin'))->toString());
```
Instead `/path/to/drupal/admin` it prints `/admin`.

weitzman:
tests are failing, unfortunately.

weitzman:
Still needed?

Chi-teck:
The bug still exists in `master`. I am having troubles with setting up SUT. This might be related to my local environment (PHP 7.2).

greg-1-anderson:
You must use Drupal 8.4.4 (release yesterday) if you are using PHP 7.2. Earlier versions did not support 7.2.

Chi-teck:
I am using the latest code form 8.5.x Git branch. Is Drush itself compatible with PHP 7.1 and PHP 7.2? It appears it is not tested against these PHP versions. 
https://github.com/drush-ops/drush/blob/master/.travis.yml#L11

hansfn:
Since this issue most likely is the cause for the missing subdir when using `drush uli`with uri option, I'm adding it here so other people can find it:

`drush uli --no-browser --uri=http://nonexisting.domain/folder

returns "http://nonexisting.domain/user/reset/1/[...]" without the subdir in the URL.

ao2:
Some other commands to reproduce the issue are in https://github.com/drush-ops/drush/issues/3463

@Chi-teck can you rebase your branch on the latest drush version so that we can validate your changes and go forward with this issue?

Thanks,
   Antonio

weitzman:
Github updated the PR so its up to date with master. Still is failing tests tho.

ao2:
Thanks @weitzman

gitressa:
It works for sub-folders if there is a trailing slash:

```
$ drush uli --uri=http://nonexisting.domain/folder
http://nonexisting.domain/user/reset/1/153...uPBiz5lA/login
```

```
$ drush uli --uri=http://nonexisting.domain/folder/
http://nonexisting.domain/folder/user/reset/1/1539196592/tMD_...If7mKyRvQ/login
```

Issue #3163, opened 2017-11-13
symfony http-foundation changes and drush_print()


douggreen:
Symfony's latest release checks for headers_sent() before creating a session. PHP print's and thus drush_print() cause this check to fail, and exceptions to be thrown.  The Symfony commit that causes this change is https://github.com/symfony/http-foundation/commit/ef7a168bd5eaea517fb36f60c477a4460f722644

I have errors on site-install. The #drush slack channel also mentions errors in ```drush updb``` from this.

```
Starting Drupal installation. This takes a while. Consider using the [ok]
--notify global option.
RuntimeException: Failed to set the session handler because headers  [error]
have already been sent by
"/path/to/drupal/vendor/drush/drush/includes/output.inc"
at line 41. in
/path/to/drupal/vendor/symfony/http-foundation/Session/Storage/NativeSessionStorage.php:395
Stack trace:
#0
/path/to/drupal/vendor/symfony/http-foundation/Session/Storage/NativeSessionStorage.php(120):
Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage->setSaveHandler(Object(Drupal\Core\Session\WriteSafeSessionHandler))
#1
/path/to/drupaldocroot/core/lib/Drupal/Core/Session/SessionManager.php(91):
Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage->__construct(Array,
Object(Drupal\Core\Session\WriteSafeSessionHandler),
Object(Drupal\Core\Session\MetadataBag))
#2
/path/to/drupaldocroot/core/lib/Drupal/Component/DependencyInjection/Container.php(284):
Drupal\Core\Session\SessionManager->__construct(Object(Symfony\Component\HttpFoundation\RequestStack),
Object(Drupal\Core\Database\Driver\mysql\Connection),
Object(Drupal\Core\Session\MetadataBag),
Object(Drupal\Core\Session\SessionConfiguration),
Object(Drupal\Core\Session\WriteSafeSessionHandler))
#3
/path/to/drupaldocroot/core/lib/Drupal/Component/DependencyInjection/Container.php(177):
Drupal\Component\DependencyInjection\Container->createService(Array,
'session_manager')
#4
/path/to/drupaldocroot/core/lib/Drupal/Component/DependencyInjection/Container.php(494):
Drupal\Component\DependencyInjection\Container->get('session_manager',
1)
#5
[...]

PR #3169, opened 2017-11-14
[#3163]: always fwrite in drush_print, instead of print.

douggreen:
See https://github.com/drush-ops/drush/issues/3163


greg-1-anderson:
This looks good, but:

- We always fix the `master` branch first, then back-port
- We should be sure that there are no other occurrences of `print` in the codebase

PR #3170, opened 2017-11-14
#3163: always fwrite in drush_print, instead of print.

douggreen:
See #3163. Replaces https://github.com/drush-ops/drush/pull/3169

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -29,7 +29,7 @@ use Drush\Utils\StringUtils;
  * @deprecated
  *   Use $this->output()->writeln() in a command method.
  */
-function drush_print($message = '', $indent = 0, $handle = NULL, $newline = TRUE) {
+function drush_print($message = '', $indent = 0, $handle = STDOUT, $newline = TRUE) {
   $msg = str_repeat(' ', $indent) . (string)$message;
   if ($newline) {
     $msg .= "\n";
@@ -37,14 +37,7 @@ function drush_print($message = '', $indent = 0, $handle = NULL, $newline = TRUE
   if (($charset = drush_get_option('output_charset')) && function_exists('iconv')) {
     $msg = iconv('UTF-8', $charset, $msg);
   }
-  if (isset($handle)) {
-    fwrite($handle, $msg);
-  }
-  else {
-    print $msg;
-    //$output = Drush::service('output');
-    //$output->write($msg);
-  }
+  fwrite($handle, $msg);
 }
 
 /**


douggreen:
Per the question on the original PR, this is the only reference to print.
```
$ grep -r 'print ' .
./.git/hooks/pre-commit.sample:# If there are whitespace errors, print the offending file names and fail.
./.git/hooks/pre-rebase.sample:					print STDERR $msg;
./.git/hooks/pre-rebase.sample:				print STDERR " $elem->[1]\n";
./.git/hooks/prepare-commit-msg.sample:#      print "\n" . `git diff --cached --name-status -r`
Binary file ./.git/objects/pack/pack-f23e288cd88cdeb3961d950f20134613a96dc5e5.pack matches
./.travis.yml:  # - echo "sendmail_path='true'" >> `php --ini | grep "Loaded Configuration" | awk '{print $4}'`
./docs/cron.md:By default, Drush will print a success message when the run of cron is completed. The --quiet flag will suppress these and other progress messages, again avoiding an unnecessary email message.
./docs/io.md:- If you need to ask the user a question, or print non-object content, use the io() system. 
./examples/example.bashrc:#       ddd @dev         - print the path to the root directory of @dev
./includes/backend.inc: * Output produced by the called script will be printed if we didn't print it
./includes/backend.inc:  // not have processed the log entries yet, and must print them here.
./includes/backend.inc:            // the backend packets are removed, then print it.
./includes/drush.inc: *   If you want your log messages to print to screen without the user entering
./src/Boot/DrupalBoot6.php:            // n.b. we want D5 and earlier to match here, if possible, so that we can print a 'not supported'
./src/Commands/core/PhpCommands.php:     * @usage drush php:eval '$node = node_load(1); print $node->title;'
./src/Commands/core/UpdateDBCommands.php:            // types, but for now we'll just print the message.
./src/Drupal/Commands/core/StateCommands.php:     * @usage php -r "print json_encode(array(\'drupal\', \'simpletest\'));"  | drush state-set --input-format=json foo.name -
./src/Log/Logger.php:                    // print nothing. exit cleanly.
./src/Log/Logger.php:                    // print nothing unless --debug AND --verbose. exit cleanly.
./src/Log/Logger.php:                    // print nothing. exit cleanly.
./src/Sql/SqlBase.php:     * If you don't want to query results to print during --debug then
./src/Sql/SqlBase.php:     * If you don't want to query results to print during --debug then
./src/Sql/SqlBase.php:     * A string to add to the command when queries should not print their results.
./src/Sql/SqlSqlsrv.php:        $query = "if db_id('$database') IS NOT NULL print 1";
./tests/UnishTestCase.php:            print $chars[($counter++ % 4)] . "\033[1D";
./tests/WatchdogTest.php:        // not print it completely in the listing unless --full is given.
```

weitzman:
A Drush8 patch for this might have more print statements to change.

greg-1-anderson:
I tried to re-open #3169 as the back-port of this issue, but the branch was already deleted.

douggreen:
Should I create a PR against 8.x?

PR #3171, opened 2017-11-14
#3163: always fwrite in drush_print, instead of print. (backport)

douggreen:
Backport of https://github.com/drush-ops/drush/pull/3170. There are additional print statements in the 8.x branch. I am comparing them all against master, to see how it was handled there.

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -26,13 +26,16 @@ use Drush\Log\LogLevel;
  * @param $newline
  *    Add a "\n" to the end of the output.  Defaults to TRUE.
  */
-function drush_print($message = '', $indent = 0, $handle = STDOUT, $newline = TRUE) {
+function drush_print($message = '', $indent = 0, $handle = NULL, $newline = TRUE) {
   $msg = str_repeat(' ', $indent) . (string)$message;
   if ($newline) {
     $msg .= "\n";
   }
   if (($charset = drush_get_option('output_charset')) && function_exists('iconv')) {
     $msg = iconv('UTF-8', $charset, $msg);
+  }
+  if (!$handle) {
+    $handle = STDOUT;
   }
   fwrite($handle, $msg);
 }
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -26,7 +26,7 @@ use Drush\Log\LogLevel;
  * @param $newline
  *    Add a "\n" to the end of the output.  Defaults to TRUE.
  */
-function drush_print($message = '', $indent = 0, $handle = NULL, $newline = TRUE) {
+function drush_print($message = '', $indent = 0, $handle = STDOUT, $newline = TRUE) {
   $msg = str_repeat(' ', $indent) . (string)$message;
   if ($newline) {
     $msg .= "\n";
@@ -34,12 +34,7 @@ function drush_print($message = '', $indent = 0, $handle = NULL, $newline = TRUE
   if (($charset = drush_get_option('output_charset')) && function_exists('iconv')) {
     $msg = iconv('UTF-8', $charset, $msg);
   }
-  if (isset($handle)) {
-    fwrite($handle, $msg);
-  }
-  else {
-    print $msg;
-  }
+  fwrite($handle, $msg);
 }
 
 /**
--- a/commands/core/outputformat/html.tpl.php
+++ b/commands/core/outputformat/html.tpl.php
@@ -4,25 +4,25 @@
   <h3>Global Options (see `drush topic core-global-options` for the full list)</h3>
   <table><?php
     foreach ($global_options_rows as $key => $row) {
-      print '<tr>';
+      drush_print('<tr>');
       foreach ($row as $value) {
-        print  "<td>" . htmlspecialchars($value) . "</td>\n";
+        drush_print("<td>" . htmlspecialchars($value) . "</td>\n");
       }
-      print "</tr>\n";
+      drush_print("</tr>\n");
     } ?>
   </table>
   <h3>Command list</h3>
   <table><?php
     foreach ($input as $key => $command) {
-      print "  <tr><td><a href=\"#$key\">$key</a></td><td>" . $command['description'] . "</td></tr>\n";
+      drush_print("  <tr><td><a href=\"#$key\">$key</a></td><td>" . $command['description'] . "</td></tr>\n");
     } ?>
   </table>
 <h3>Command detail</h3>
 <dl><?php
       foreach ($input as $key => $command) {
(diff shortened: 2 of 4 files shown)


includes/output.inc
```diff
@@ -26,20 +26,15 @@ use Drush\Log\LogLevel;
  * @param $newline
  *    Add a "\n" to the end of the output.  Defaults to TRUE.
  */
-function drush_print($message = '', $indent = 0, $handle = NULL, $newline = TRUE) {
+function drush_print($message = '', $indent = 0, $handle = STDERR, $newline = TRUE) {
```
greg-1-anderson: I think `drush_print` should default to STDOUT. Switching to STDERR would be a backwards compatibility break.

greg-1-anderson:
`master` completely replaced a lot of Drush 8.x code. Just use `drush_print`, or call `fwrite` directly if it is not feasible or appropriate to call `drush_print`.

includes/output.inc
douggreen: Opps, sorry, I didn't intend to do that. This is what happens when you can't just ```git pick``` because branches have changed!

douggreen:
I'm replacing print with drush_print(). But wholesale changes like this have me nervous. Do tests run that will catch issues?

greg-1-anderson:
There are a lot of tests, but there is no guarentee that they will catch every regression. In instances like this, we can merge the PR to the 8.x branch and wait a while before tagging a release, to allow some time for breaking changes to be discovered.

We will also review your PR to see if anything is revealed through manual inspection.

Thanks for diving in here.

greg-1-anderson:
The transformations from `print` to `drush_print` here look safe.

douggreen:
Please wait to merge this. I couldn't test the master branch but I can test this branch. And I'm getting some local errors based on the way I'm assigning stdout here. We may need to change the parameter default back to NULL and do it in code.

douggreen:
I had thought the parameter assignment would just work, and if it didn't, that tests would pick it up.

douggreen:
ahh, I see why it doesn't work as a parameter default!
```
define('STDOUT', fopen('php://stdout', 'w'));
```

douggreen:
We don't want that.

greg-1-anderson:
Maybe:
```
if (!isset($handle)) {
    $handle = STDOUT;
}
```
I didn't look at the PR on `master` before Moshe merged it

greg-1-anderson:
Looks like this needs to be fixed in `master` too (STDOUT default parameter value).

douggreen:
See https://github.com/drush-ops/drush/pull/3172

greg-1-anderson:
The [PHP I/O streams documentation](http://php.net/manual/en/features.commandline.io-streams.php) says that STDOUT is an already-opened stream. Surely the fopen is called once, when the define is processed, and not every time the constant is referenced, right?

I tried:
```
php > $x = STDOUT;
PHP Notice:  Use of undefined constant STDOUT - assumed 'STDOUT' in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0

Notice: Use of undefined constant STDOUT - assumed 'STDOUT' in php shell code on line 1

Call Stack:
    4.7210     352768   1. {main}() php shell code:0

php > function x() { static $x = 0; $x = $x + 1; return $x; }
php > print x();
1
php > print x();
2
php > print x();
3
php > define('foo', x());
php > print foo;
4
php > print foo;
4
```
So, although STDOUT isn't available in `php -a`, it seems that there should be just one STDOUT handle. I'm therefore unsure why you'd be running out of file handles.

douggreen:
Ok, I've tracked down why this is a problem. It's not the number of handles. It's backwards compatibility. drush_print_prompt() has a handle with a default of NULL and then it passes that to drush_print(). Since it passes a default, the file descriptor is NULL and and not STDOUT. We could fix all internal calls, sure, but I think for backwards compatibility with drush_print() we shouldn't change the default, and we should handle it in the function.

weitzman:
I think this is fixed now?

greg-1-anderson:
I think it would be beneficial to continue here. Even if Symfony is not checking the headers, we are at risk for any other code that might do so in the future. I therefore think it is better to use fwrite instead of print for any cli tool that may call library code intended to run in a browser, just as a matter of simple defensiveness (e.g. a contrib module with an init hook could break, etc. etc.).

douggreen:
I think this is the backport fix we want. I'm not sure why one of the builds failed.

m4olivei:
I see three themes in the error messages.

First example: https://travis-ci.org/drush-ops/drush/jobs/302251693
```
1) Unish\backendCase::testTarget
Failed asserting that '' starts with " Drush Version ".

/home/travis/build/drush-ops/drush/tests/backendTest.php:92
/home/travis/build/drush-ops/drush/tests/Unish/CommandUnishTestCase.php:332
/home/travis/build/drush-ops/drush/vendor/phpunit/phpunit/phpunit:52
```
Second example: https://travis-ci.org/drush-ops/drush/jobs/302251694
```
1) Unish\QueueCase::testQueue
Unexpected exit code 1 (expected 0) for command:
/home/travis/build/drush-ops/drush/drush --root=/tmp/drush-sandbox/web --uri=dev --nocolor queue-run aggregator_feeds --yes

Command stderr:
SQLSTATE[HY000]: General error: 1366 Incorrect string value:             [error]
'\xF0\x9F\x9B\xA0 I...' for column 'description' at row 1
Processed 0 items from the aggregator_feeds queue in 0.09 sec.         [warning]

/home/travis/build/drush-ops/drush/tests/Unish/CommandUnishTestCase.php:196
/home/travis/build/drush-ops/drush/tests/Unish/CommandUnishTestCase.php:303
/home/travis/build/drush-ops/drush/tests/queueTest.php:43
/home/travis/build/drush-ops/drush/tests/Unish/CommandUnishTestCase.php:332
/home/travis/build/drush-ops/drush/vendor/phpunit/phpunit/phpunit:52
```
Third example: https://travis-ci.org/drush-ops/drush/jobs/302251700
```
1) Unish\userCase::testUserLogin
Login returned a reset URL for uid 1 by default
Failed asserting that '' contains "/user/reset/1".

/home/travis/build/drush-ops/drush/tests/userTest.php:94
/home/travis/build/drush-ops/drush/tests/Unish/CommandUnishTestCase.php:332
/home/travis/build/drush-ops/drush/vendor/phpunit/phpunit/phpunit:52
```

weitzman:
Pretty sure we fixed this in https://github.com/drush-ops/drush/commit/a97c1d1f4df08e8ff2c13230f1f1c9a2599babef

PR #3172, opened 2017-11-15
#3163: don't use STDOUT as a function parameter default, because the …

douggreen:
…define actually calls fopen()

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -29,13 +29,16 @@ use Drush\Utils\StringUtils;
  * @deprecated
  *   Use $this->output()->writeln() in a command method.
  */
-function drush_print($message = '', $indent = 0, $handle = STDOUT, $newline = TRUE) {
+function drush_print($message = '', $indent = 0, $handle = NULL, $newline = TRUE) {
   $msg = str_repeat(' ', $indent) . (string)$message;
   if ($newline) {
     $msg .= "\n";
   }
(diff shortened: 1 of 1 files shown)


douggreen:
See https://github.com/drush-ops/drush/pull/3171

douggreen:
Sorry for the extra work. I'd hoped tests would of caught that. We now are back to the code that I'd originally written. The code original merged just looked cleaner, but it was alas, untested.

greg-1-anderson:
I just tested HEAD of `master` without this PR (using STDOUT as a default parameter) and it seems to work.

I'm still +1 on merging this once the tests pass. Maybe works in newer versions of php, but not necessarily older?

douggreen:
```
```
Yea, maybe. My PHP is not cleaning up the file pointer when it's opened as part of the function parameters, but does when it's opened in the function.

The original solution looked cleaner. But I don't like calling a function outside a function. It's dangerous. This is a bit different. It's calling a function in the function argument, but it still seems dangerous.

I get these errors, not on every drush command, but on the site-install.
```
opendir(/Users/douggreen/src/drush/includes/../vendor/phpunit/php-code-coverage/src/CodeCoverage/Report/HTML/Renderer/Template):[warning]
failed to open dir: Too many open files filesystem.inc:685
```

douggreen:
Odd, I'm now getting that same error on the old solution. :(

PR #3174, opened 2017-11-15
Escape SSH Command For LOCAL And Host OS

fractile81:
When escaping the remote SSH command, escape it first for the target/LOCAL OS, then escape it for the host OS.

In the instance of Windows, using just "LOCAL" to escape the SSH command will fall-through and escape using the Linux function.  This causes a "File or directory not found" error when trying to use a remote alias from a Windows machine.  Specifically, Windows has a problem with the (Open)SSH call having the command wrapped in single-quotes instead of double-quotes.

By double-escaping, the command is escaped for the target OS and then the host OS.  Escaping using the "raw" option for the target keeps it from surrounding the command with single-quotes, while escaping again without specifying an OS will add single- or double-quotes based on the host OS.


includes/backend.inc
```diff
@@ -1173,7 +1173,16 @@ function _drush_backend_generate_command($site_record, $command, $args = [], $co
...
     $ssh_cmd[] = $username . drush_escapeshellarg($hostname, "LOCAL");
-    $ssh_cmd[] = drush_escapeshellarg($command . ' 2>&1', "LOCAL");
+
+    // Escape the remote command for the LOCAL OS, then add in the proper quote style based on the host OS
+    $remote_cmd = drush_escapeshellarg($command . ' 2>&1', "LOCAL", TRUE);
+    if (drush_is_windows()) {
+      $remote_cmd = '"' . $remote_cmd . '"';
+    }
+    else {
+      $remote_cmd = "'" . $remote_cmd . "'";
```
greg-1-anderson: I don't think that you can reliably just wrap an escaped command in single-quotes & trust that it will correctly escape an escaped command,  e.g. if $remote_cmd contains a single quote.

fractile81: $remote_cmd is still escaped two lines above that with $raw = TRUE.  This is just doing what the escape functions do when $raw is FALSE, while still escaping for LOCAL.

greg-1-anderson: If you escape `$foo` to `'$foo'`, then wrapping that in single quotes will give you `''$foo''`, which is not correct.

fractile81: I agree, that would be bad.

With this code change, the call to `drush_escapeshellarg($arg, $os, $raw)` (line 1178) has `$raw` set to `TRUE`.  Both the `_drush_escapeshellarg_linux($arg, $raw=FALSE)` and `_drush_escapeshellarg_windows($arg, $raw=FALSE)` functions, which are used by this parent function, are passed `$raw`.  From what I'm seeing, having `$raw = TRUE`  ensures `$foo` remains `$foo` after line 1178 has completed execution.

If `'$foo'` is passed as $arg (single-quotes included) on line 1178, then the Linux escape will make it `'\''$foo'\''`, as you mentioned, and then this PR will make it `''\''$foo'\'''`.  Similar-ish things will happen with Windows and double-quotes.  Is this the edge-case you're talking about?  That would imply drush is being called in a manner similar to: `drush @remote-host 'command'`.

Is this a supported use case?  If so, would stripping existing wrapped quotes from `$arg` before escaping take care of this concern?

greg-1-anderson: Looking at the `$raw` handling, it seems that your code block above passes `$raw` as True to avoid wrapping the result in `"` or `'`, and then immediately puts back on the quote characters that it asked to be omitted.

I don't remember why we have the `$raw` parameter at all. It seems that in the case of the code above, you should let `$raw` stay FALSE. Then you should get the result you expect without having to manually re-add the quotes yourself.

Note that it's still not entirely clear to me which exact character escape combinations this PR is intended to fix. The existing escaping code appears to work really well for Linux-to-Linux commands, so it would be a good idea to ensure that code path passes data through the same way it does today. It would also be really helpful if you could add failing test cases (that is, tests that fail without this PR, and pass with it).

fractile81: This is a Windows-to-Linux issue, something I'm guessing the tests don't cover.  I discovered all of this when using the Windows command line and trying to use an alias for a remote server at Acquia.

The original escape happens for the LOCAL, or target, OS.  The escape function assumes that if `$os` doesn't start with "WIN" that it should escape `$arg` as Linux.  Since line 1178 has `$os = LOCAL` (matching the original code), the whole command is escaped as Linux.  With `$raw = TRUE`, we need to add wrapping quotes back in, which is what the rest of the PR does.

With (Open)SSH on Windows command line/PowerShell, the SSH call in `proc_open()` fails if single quotes are used.  However, the same SSH call works when double-quotes are used.  Setting `$raw = TRUE` and wrapping with quotes based on the host OS is what this PR is trying to address.

Ultimately, the problem seems to be the use of `$os = LOCAL` when escaping the SSH command.  If the escaping only needs to happen for the host machine, then `$os` shouldn't be set and this PR would simply change the one-line function call to exclude LOCAL.  If the command needs to be escaped for the LOCAL OS, which the existing code does, then this PR takes care of escaping the SSH command for the LOCAL OS and wrapping the command with quotes that are appropriate for the host OS.  This way the SSH command is escaped for the LOCAL OS, but wrapped with quotes based on the host OS.

greg-1-anderson: I think that perhaps the code needs to do what you said in the issue description -- first call `drush_escapeshellarg` for the target platform, and then, if the local platform is different, call `drush_escapeshellarg` for the local platform.

I think that setting `$raw` and manually wrapping in quotes isn't going to work. Make sure that you can escape complex commandline arguments, e.g.
```
drush ev '$item = menu_get_item(); var_export($item);'
```
> The escape function assumes that if $os doesn't start with "WIN" that it should 
> escape $arg as Linux.

That sounds like a bug. If $os is "LOCAL", then the local OS should be looked up and used to determine the escape mechanism.

fractile81: I've updated the PR to better align with convention.

I'm not entirely sure how "LOCAL" is meant to be used, but `_drush_get_os()` makes it sound like it's meant to be used as documentation-by-code.  However, `drush_escapeshellarg()` doesn't take "LOCAL" into account in any way.  One of my iterations on this locally was to place `$os = strtoupper($os) == 'LOCAL' ? NULL : $os;` at the top of this function, but I decided that sort of a change would have a far wider impact than what this PR does.

greg-1-anderson:
It seems that the upshot is that `drush_escapeshellarg()` does not support LOCAL, and the backwards-compatible solution is to not use that constant with that function. I don't know why there is a LOCAL constant instead of just `drush_os()` -- or I should say I don't remember, because I'm pretty sure this is my doing.

In any event, the current state of this PR looks like it could plausibly be correct. It just needs some additional tests to show that the expected escaping is being done for Windows-to-Linux remote calls.

fractile81:
Would the best route for these tests be to extract the updated code so that I can control what environment it believes it's in (the function relies on `PHP_OS`)?  AFAIK, there are no Windows environments/tests setup in TravisCI.  I realize testing it in this way is clunky, ergo asking if there are any better suggestions?

greg-1-anderson:
Refactoring so that PHP_OS is a parameter and not hardcoded is the way to go if that change would not require any backwards compatibility breaking adjustments.

Windows testing can be done on Appveyor; some of Drush's dependencies to Appveyor testing, but it would be a very large effort to get that going for Drush. It would be great to have, but should be a separate follow-on PR to this if anyone was inclined to attempt it.

Ideally, both the Windows and Linux code paths should be tested with unit tests that are independent of the platform the tests are run on, so that both are tested under both Travis and (aspirationally, someday) Appveyor.

fractile81:
I've updated the PR so that LOCAL defaults to the OS when escaping the SSH command.  I also updated the build function to consistently use either the remote or local OS, with variables to match.

I've been racking my brain on how to add tests for this properly.  I've looked through a number of the test cases to see if I could find a good example, but haven't found anything.  The closest I came was to BackendTest.testOrigin(), which built out a simulated SSH command.  However, I can't figure out how to push an alternate OS into this example without making big changes and/or adding --local-os as a CLI option (something I don't think any of us wants).

Any further suggestions so I can get this change in?

greg-1-anderson:
I think that you should back out all of these changes and instead change `drush_is_windows` to:
```
function drush_is_windows($os = NULL) {
  return strtoupper(substr(_drush_get_os($os), 0, 3)) === 'WIN';
}
```
All of your changes seem to be dancing around this one bug. Try that and see if it fixes your use-case.

weitzman:
What does ther new code solve exactly?

Also, does this introduces a circular loop where drush_is_windows() calls _drush_get_os and vice versa?

fractile81:
There is a use case in Windows where the OS "LOCAL" is used in `_drush_backend_generate_command();`.  As it exists today, "LOCAL" will resolve to a *nix OS in `drush_escapeshellarg();` since the check in `drush_is_windows();` fails to match against "LOCAL".  When using drush for a remote SSH call on the Windows CLI, the wrong quote-type is used and the call to `ssh.exe` (manually installed) fails.

It is indeed possible for this to recurse.  I'll have to take a closer look with test arguments to backup what I think is happening; I'll post what I find.

fractile81:
Follow-up on the recursion concern.

The only time `_drush_get_os();` will call `drush_is_windows();` is when the "RSNYC" OS is passed.  This is a special "OS" used to choose between using "CWRSYNC" and the local OS, per the code.  This branch will call `_drush_get_os("LOCAL");` and get `PHP_OS`.  The resulting value of `PHP_OS` is the passed to `drush_is_windows();`.

Since `PHP_OS` shouldn't ever match "RSYNC" (built-in assumption to `_drush_get_os();`), the deepest this should go is `_drush_get_os("RSYNC")` > `_drush_get_os("LOCAL")` +  (`drush_is_windows(PHP_OS)` > `_drush_get_os(PHP_OS)`).

I've tried a few manual tests to make sure nothing went out of whack, and it appears to be working correctly.

weitzman:
Thanks for the research. 

I dislike the arcane logic in _drush_get_os(), and would rather not use it here. We are taking a simple function `drush_is_windows()` and making it more complex. Maybe the caller can handle the complexity?

fractile81:
The issue I originally presented was due to `drush_escapeshellarg();` receiving "LOCAL" as the OS.  The existing logic doesn't translate "LOCAL" into `PHP_OS`.  I'd be fine with moving the `_drush_get_os();` call out into `drush_escapeshellarg();`.

Circumventing the use of `_drush_get_os();` is where I believe this error is coming from in the first place.  On the one hand, moving the call up a level isn't a problem.  On the other hand, the Windows-checking function should be able to take the special case OSes and be able to boil them down to something it can test against.  The other `drush_is_X();` OS checks end up calling `_drush_get_os();` eventually, so adding it into the Windows check made sense.

All of this considered, would you still like me to move `_drush_get_os();` out of `drush_is_windows();`?

greg-1-anderson:
Well, I clearly made a mistake in https://github.com/drush-ops/drush/pull/3174#issuecomment-347966568. I don't know if I meant something other than `_drush_get_os()`, but note that `_drush_get_os()` can call `drush_is_windows()`, so it's wrong wrong wrong to have each call the other, even if this HAPPENS to not cause an infinite loop as currently written.

It's been a long time since I've done anything with Windows escaping. It appears that the RSYNC and CWRSYNC special OS types are not even used in the Drush 9 implementation of rsync. The 'LOCAL' special OS type is only used in backend.inc. The comment in _drush_get_os says this is only for documentation; in theory we can just use PHP_OS in backend.inc.  After that, _drush_get_os is reduced to just testing for mingw. Since I don't even know if that works any more (last used around the Drush 5 era, AFAIK), I'd be +1 on dropping support for that in Drush 9.

So, we could just punt _drush_get_os() and use PHP_OS everywhere it appears, I think. That might sidestep the original `drush_escapeshellarg()` bug completely.

Sorry for the big detour I caused here.

includes/environment.inc
```diff
@@ -192,28 +192,13 @@ function drush_bit_bucket() {
...
-  // "CWRSYNC" when cwrsync is installed, or default to the local OS otherwise.
-  if (strtoupper($os) == "RSYNC") {
-    $os = _drush_get_os("LOCAL");
-    // For now we assume that cwrsync is always installed on Windows, and never installed son any other platform.
-    return drush_is_windows($os) ? "CWRSYNC" : $os;
-  }
   // We allow "LOCAL" to document, in instances where some parameters are being escaped
   // for use on a remote machine, that one particular parameter will always be used on
   // the local machine (c.f. drush_backend_invoke).
   if (isset($os) && ($os != "LOCAL")) {
```
greg-1-anderson: If we are going to keep `_drush_get_os`, I'd prefer to keep MINGW and delete LOCAL. At this point, though, I think we should just replace all calls to `_drush_get_os()` to PHP_OS. Doing this requires removing LOCAL from backend.inc though.

fractile81: I think the abstraction can still be important enough to justify keeping `_drush_get_os();`.

I just pushed an update that removed everything but the LOCAL check and the return of PHP_OS.  I'll add the MINGW code back in.

includes/environment.inc
```diff
@@ -246,7 +236,7 @@ function drush_is_local_host($host) {
  * Determine whether current OS is a Windows variant.
  */
 function drush_is_windows($os = NULL) {
-  return strtoupper(substr($os ?: PHP_OS, 0, 3)) === 'WIN';
+  return strtoupper(substr(_drush_get_os($os), 0, 3)) === 'WIN';
```
greg-1-anderson: If we stop using `LOCAL` in backend.inc, then this can stay as it used to be

fractile81: I was concerned that would be a big deal, but it looks like `_drush_backend_generate_command()` is the only function using LOCAL anymore.  My only lingering thought remains in keeping the `_drush_get_os()` abstraction to remain consistent with the other OS checks.

I'll go back and remove LOCAL and update `_drush_get_os()` one more time.

fractile81:
Keeping MINGW proved problematic as `_drush_test_os()` called `_drush_get_os()`.  I tried keeping the check in, but I don't understand its use case well enough to know how to properly test for it without creating an infinite loop.

* `_drush_get_os()` has been stripped down to only return `$os ?: PHP_OS`
* All instances of "LOCAL" have been removed as a special-case OS name
* `drush_is_windows()` still calls `_drush_get_os()` for consistency

greg-1-anderson:
Looks great. Thanks for working through this and simplifying.

Issue #3155, opened 2017-11-09
drush 9 site-install "db-su" option not taken into account


gaelg:
The problem is in the title :)

Here's a log:
```
drush si config_installer --db-url=mysql://normal_user:normal_pw@localhost/my_table --db-su=su_user --db-su-pw=su_pw --yes -vvv
 [bootstrap] Bootstrap to root [0.21 sec, 9.66 MB]
 [bootstrap] Drush bootstrap phase 1 [0.21 sec, 9.67 MB]
 [bootstrap] Try to validate bootstrap phase 1 [0.21 sec, 9.67 MB]
 [bootstrap] Try to validate bootstrap phase 1 [0.21 sec, 9.67 MB]
 [bootstrap] Try to bootstrap at phase 1 [0.22 sec, 9.87 MB]
 [bootstrap] Drush bootstrap phase: bootstrapDrupalRoot() [0.22 sec, 9.87 MB]
 [bootstrap] Change working directory to /doc-root/web [0.22 sec, 9.87 MB]
 [bootstrap] Initialized Drupal 8.4.0 root directory at /doc-root/web [0.22 sec, 9.87 MB]
 [bootstrap] Drush bootstrap phase: bootstrapDrupalSite() [0.23 sec, 10.7 MB]
 [bootstrap] Initialized Drupal site default at sites/default [0.23 sec, 10.7 MB]
 [bootstrap] Drush bootstrap phase: bootstrapDrupalConfiguration() [0.23 sec, 10.7 MB]
 [debug] Add service modifier [0.24 sec, 11.09 MB]
 [info] sql-query: SELECT 1; [0.24 sec, 11.25 MB]
 [info] Executing: mysql --defaults-file=/tmp/drush_xpp7D9 --database=my_table --host=localhost --silent  < /tmp/drush_tnPkSW > /dev/null [0.24 sec, 11.25 MB]

 // You are about to CREATE the 'my_table' database. Do you want to continue?: yes.                                     

 [info] Sites directory sites/default already exists - proceeding. [0.26 sec, 11.43 MB]
 [info] sql-query: SELECT 1; [0.26 sec, 11.43 MB]
 [info] Executing: mysql --defaults-file=/tmp/drush_TSCviz --database=my_table --host=localhost --silent  < /tmp/drush_FnGitm > /dev/null [0.26 sec, 11.43 MB]
[...]

PR #3175, opened 2017-11-16
Fix drush si --db-su option

gaelg:
This is a fix for #3155 The problem comes from Preflight::remapOptions() which remap the --db-su option. There might be a better way to fix this, but at least it works.


weitzman:
Looks good. I think config object is injected so please use $this->getConfig()->get() instead.

weitzman:
Merged locally and pushed

PR #3178, opened 2017-11-16
Fixes #3164: Introduce artwork:show command, et. al.

### LINKED ISSUE #3164: Sandwich example is not inclusive and should be replaced.
It may seem like a cute easter egg, but the sandwich example relies on a sexist trope with a [good bit of history implying women are subservient and belong in the kitchen](https://en.wikipedia.org/wiki/Make_me_a_sandwich) — seeing as drush is such a standard tool that many developers use, and that Drupal's foundation and expansion has been accompanied by a [**huge** community of female developers](https://groups.drupal.org/women-drupal), the joke isn't worth it.

This isn't about safe spaces or censorship. This is about not forcing women developers to see something fostered by a "men's club" culture that is still very real and prevalent in the industry.

greg-1-anderson:
This example command demonstrates how to:

- Define a new Drush command
- Use `@hook interact` to query for missing arguments
- Use `@hook validate` to check for invalid data
- Use `@hook on-event` to define new hooks to collect data from other command files

At the moment, we only have one piece of artwork: a tasty sandwich. We could use more art.


examples/Commands/ArtCommands.php
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * Commandfiles may also add topics.  These will appear in
+     * the list of topics when `drush topic` is executed.
+     * To view the topic below, run `drush --include=/full/path/to/examples topic`
+     */
+
+    /**
+     * Ruminations on the true meaning and philosophy of sandwiches.
```
greggles: Here's a leftover sandwich mention.

PR #3183, opened 2017-11-19
Remove legacy.inc file

weitzman:


greg-1-anderson:
Oh, looks like `drush_convert_db_from_db_url()` at a minimum is still being used.

PR #3184, opened 2017-11-22
Simplified site aliases. Fixes #3107

### LINKED ISSUE #3107: Rename name.alias.yml to name.site.yml and drop support for group.aliases.yml
The docs (https://raw.githubusercontent.com/drush-ops/drush/master/examples/example.aliases.yml) have no example for a "aliases.yml". How should this fill look like? I tried the following.

```
stage:
  host: stage.example.com
prod:
  host: example.com
```

Does not work.

```
sites:
  stage:
    host: stage.example.com
  prod:
    host: example.com
```

Does not work.

```
sites:
  my_project:
    stage:
      host: stage.example.com
    prod:
      host: example.com
```

Works.

This means the resulting site-aliases are quite verbose. Because its always site+environment. I use this in combination with drupal-composer/drupal-project and never have global available site-aliases. Is it still possible to use short site aliases? example.aliases.yml has ``@stage`` as an example but i was not able to produce such an site-alias.

---

PS: I did some debugging. ``\Drush\SiteAlias\SiteAliasFileLoader::fetchAliasRecordFromSiteAliasData`` in \Drush\SiteAlias\SiteAliasFileLoader::fetchAliasRecordFromGroupAliasData returns ``false`` if the group is empty. It this the root cause? Thanks!

greg-1-anderson:
First pass. Mostly done.

- Ad-hoc usage is pretty consistent
- Isolation tests are passing
- Functional test fixtures (e.g. sut aliases) may still need adjusting
- There may be a little more site alias code that can be deleted


greg-1-anderson:
Isolation tests are failing because a list needs to be sorted before the assertion.

src/SiteAlias/SiteAliasFileDiscovery.php
```diff
@@ -4,19 +4,21 @@
...
- * Otherwise, Drush will first search for the provided alias name in a
- * single-alias alias file. If no such file can be found, then it will try
- * all sitenames in all group alias files.
+ * Drush aliases that contain both a site name and an environment
+ * (e.g. @site.env) will cause Drush to find the file named after
+ * the respective site name and retreive the specified environment
+ * record.
+ *
+ * Sites may also define a special alias file self.site.yml, which
+ * may be stored in the drush/site-aliases directory relative to either
```
weitzman: Looks like this dir is now 'sites', not 'site-aliases'.

greg-1-anderson:
I increased the search depth of site aliases to `<=1` so that it would be possible to store related alias files together in a directory (e.g. ~/.drush/sites/pantheon/example.site.yml).

Currently, the behavior of the site:alias command is as follows:

| Command | Result |
|-----------|---------------------|
| `drush sa`    | Show all sites from all site alias files that can be found |
| `drush sa @example.dev` | Show just the dev environment of the example site |
| `drush sa @example`         | Show all of the environments of the example site |

In the last example, environments defined in the file `self.site.yml` are also included for the first local site found in `example.site.yml`.

The next question is, should there be some way to run `drush sa` with an additional argument or option to show all of your aliases in a particular directory (e.g. all of your aliases for a certain isp)?
```
drush sa --group=dirname
drush sa --filter=pathregex
drush sa /regex/
```
Or perhaps some other notation?

greg-1-anderson:
All green! We could merge here and solve ^^ as a follow-on issue, or fix it up prior to merging as desired.

weitzman:
I don't feel the need to see aliases in a particular directory. Thats
getting real close to `more /path/to/site.yml`. So, lets merge this now.

PR #3185, opened 2017-11-22
Backport of #3174, SSH command escaping

fractile81:
A backport (see #3174) of SSH command escaping that properly handles "LOCAL" as a reference to the host OS, as opposed to the remote OS.

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -1130,12 +1130,14 @@ function _drush_backend_generate_command($site_record, $command, $args = array()
   );
   $backend_options += array(
     '#tty' => FALSE,
+    '#local-os' => 'LOCAL',
   );
 
   $hostname = $site_record['remote-host'];
   $username = $site_record['remote-user'];
   $ssh_options = $site_record['ssh-options'];
-  $os = drush_os($site_record);
+  $remote_os = drush_os($site_record);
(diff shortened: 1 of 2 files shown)


fractile81:
Back-porting this fix wasn't as straight forward.  LOCAL is used through out the 8.x branch, and `drush_is_windows()` may not have access to `_drush_get_os()`.  The simple solution was to move a check for LOCAL into the Windows check.  Anything more extensive would likely require more refactoring than this warrants.

includes/backend.inc
```diff
@@ -1152,15 +1150,15 @@ function _drush_backend_generate_command($site_record, $command, $args = array()
 
   $cmd[] = $command;
   foreach ($args as $arg) {
-    $cmd[] = drush_escapeshellarg($arg, $remote_os);
+    $cmd[] = drush_escapeshellarg($arg, $os);
```
greg-1-anderson: I think this change breaks backwards compatibility.

fractile81: I reverted a number of changes I had started with.  The final result should only be a change in `drush_is_windows()`.  I can try creating a new branch/PR to clean this up.

greg-1-anderson:
This solution looks reasonable. Thanks.

PR #3189, opened 2017-11-24
Check to see if Drush extensions are compatible with the current version.

greg-1-anderson:
Still needs to check the Drush version in a meaningful way.


greg-1-anderson:
Ideally we'd factor out some of this code into a separate class, maybe do some DI. This should be good enough for rc1, I think, though. Please review and LMK if you concur.

weitzman:
Code looks good.

Extra section needs to be added to command authoring docs and DrushAliasFile generator. These can be a followup.

greg-1-anderson:
Great. I'll merge this in now & update documentation as soon as I get a chance.

PR #3191, opened 2017-11-26
Add a composer.json file to the output of the 'drush gen dcf' command.

greg-1-anderson:
- Defines a composer.json that the dcf generator writes.
- Improves error reporting for bad composer.json content (still experimenting with right levels)
- Adds some documentation on how to stipulate multiple drush.services.yml files.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -6,10 +6,37 @@ Creating a new Drush command or porting a legacy command is easy. Follow the ste
 1. Run `drush generate drush-command-file`.
 1. Drush will prompt for the machine name of the module that should "own" the file.
     1. (optional) Drush will also prompt for the path to a legacy command file to port. See [tips on porting command to Drush 9](https://weitzman.github.io/blog/port-to-drush9)
-1. Drush will then report that it created a commandfile and a drush.services.yml file. Edit those 2 files as needed.
+    1. The module selected must already exist and be enabled. Use `drush generate module-standard` to create a new module.
+1. Drush will then report that it created a commandfile, a drush.services.yml file and a composer.json file. Edit those files as needed.
 1. Use the classes for the core Drush commands at /src/Drupal/Commands as inspiration and documentation.
 1. See the [dependency injection docs](dependency-injection.md) for interfaces you can implement to gain access to Drush config, Drupal site aliases, etc.
 1. Once your two files are ready, run `drush cr` to get your command recognized by the Drupal container.
+
+Providing Multiple Services File
+================================
+
+A module's composer.json file stipulates the filename where the Drush services (e.g. the Drush command files) are defined. The default services file is `drush.services.yml`, which is defined in the extra section of the composer.json file as follows:
+```
+  "extra": {
+    "drush": {
+      "services": {
+        "drush.services.yml": "^9"
+      }
+    }
+  }
+```
+If for some reason you need to load different services for different versions of Drush, simply define multiple services files in the `services` section. For example:
+```
+  "extra": {
+    "drush": {
+      "services": {
+        "drush-9-99.services.yml": "^9.99",
+        "drush.services.yml": "^9"
+      }
+    }
+  }
+```
+In this example, the file `drush-9-99.services.yml` loads commandfile classes that require features only available in Drush 9.99 and later, and drush.services.yml loads an older commandfile implementation for earlier versions of Drush.
 
 Global Drush Commands
 ==============================
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -12,7 +12,7 @@ Creating a new Drush command or porting a legacy command is easy. Follow the ste
 1. See the [dependency injection docs](dependency-injection.md) for interfaces you can implement to gain access to Drush config, Drupal site aliases, etc.
 1. Once your two files are ready, run `drush cr` to get your command recognized by the Drupal container.
 
-Providing Multiple Services File
+Specifying the Services File
 ================================
 
 A module's composer.json file stipulates the filename where the Drush services (e.g. the Drush command files) are defined. The default services file is `drush.services.yml`, which is defined in the extra section of the composer.json file as follows:
@@ -25,7 +25,7 @@ A module's composer.json file stipulates the filename where the Drush services (
     }
   }
 ```
-If for some reason you need to load different services for different versions of Drush, simply define multiple services files in the `services` section. For example:
+If for some reason you need to load different services for different versions of Drush, simply define multiple services files in the `services` section. The first one found will be used. For example:
 ```
   "extra": {
     "drush": {
@@ -37,6 +37,10 @@ If for some reason you need to load different services for different versions of
   }
 ```
 In this example, the file `drush-9-99.services.yml` loads commandfile classes that require features only available in Drush 9.99 and later, and drush.services.yml loads an older commandfile implementation for earlier versions of Drush.
+
+It is also possible to use [version ranges](https://getcomposer.org/doc/articles/versions.md#version-range) to exactly specify which version of Drush the services file should be used with (e.g. `"drush.services.yml": ">=9 <9.99"`).
+
+In Drush 9, the default services file, `drush.services.yml`, will be used in instances where there is no `services` section in the Drush extras of the project's composer.json file. In Drush 10, however, the services section must exist, and must name the services file to be used. If a future Drush extension is written such that it only works with Drush 10 and later, then its entry would read `"drush.services.yml": "^10"`, and Drush 9 would not load the extension's commands. It is all the same recommended that Drush 9 extensions explicitly declare their services file with an appropriate version constraint.
 
 Global Drush Commands
 ==============================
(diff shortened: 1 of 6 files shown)


src/Drupal/DrupalKernel.php
```diff
@@ -161,21 +165,35 @@ protected function findModuleDrushServiceProvider($module, $dir)
     protected function findModuleDrushServiceProviderFromComposer($dir)
     {
         $composerJsonPath = "$dir/composer.json";
+        drush_log(dt("Looking for $composerJsonPath"), LogLevel::DEBUG);
```
greg-1-anderson: Useful for debugging, but to be removed.

src/Drupal/DrupalKernel.php
```diff
@@ -161,21 +165,35 @@ protected function findModuleDrushServiceProvider($module, $dir)
...
         if (!file_exists($composerJsonPath)) {
             return false;
         }
         $composerJsonContents = file_get_contents($composerJsonPath);
+        var_export($composerJsonContents);
```
greg-1-anderson: Also to be removed.

src/Drupal/DrupalKernel.php
```diff
@@ -161,21 +165,35 @@ protected function findModuleDrushServiceProvider($module, $dir)
...
+        return $info['extra']['drush']['services'];
+    }
+
+    protected function findAppropriateServicesFile($module, $services, $dir)
+    {
+        $version = Drush::getVersion();
+        foreach ($services as $serviceYmlPath => $versionConstraint) {
             $version = preg_replace('#-dev.*#', '', $version);
             if (Semver::satisfies($version, $versionConstraint)) {
+                drush_log(dt('Found {services} for {module} Drush commands', ['module' => $module, 'services' => $serviceYmlPath]), LogLevel::DEBUG);
```
greg-1-anderson: Probably keep.

src/Drupal/DrupalKernel.php
```diff
@@ -161,21 +165,35 @@ protected function findModuleDrushServiceProvider($module, $dir)
...
                 return $dir . '/' . $serviceYmlPath;
             }
         }
+        drush_log(dt('{module} has Drush commands, but none of {constraints} match the current Drush version "{version}"', ['module' => $module, 'constraints' => implode(',', $services), 'version' => $version]), LogLevel::DEBUG);
```
greg-1-anderson: Definitely keep.

src/Drupal/DrupalKernel.php
```diff
@@ -161,21 +165,35 @@ protected function findModuleDrushServiceProvider($module, $dir)
...
         }
         $composerJsonContents = file_get_contents($composerJsonPath);
+        var_export($composerJsonContents);
         $info = json_decode($composerJsonContents, true);
+        if (!$info) {
+            drush_log(dt('Invalid json in {composer}', ['composer' => $composerJsonPath]), LogLevel::WARNING);
+            return false;
+        }
         if (!isset($info['extra']['drush']['services'])) {
+            drush_log("No extra.drush.services in $composerJsonPath");
```
greg-1-anderson: Probably remove.

greg-1-anderson:
How would I test for and edit an existing `composer.json` file in the dcf generator?

weitzman:
You can check for a file with is_file().

Existing generator for a composer.json https://github.com/Chi-teck/drupal-code-generator/blob/master/src/Command/Drupal_8/Composer.php

As for editing composer.json, I think you would do that at end of interact() instead of the usual call to something like `$this->setFile('src/Commands/' . $vars['class'] . '.php', 'drush-command-file.twig', $vars);`

greg-1-anderson:
Yes, my question should have been more along the lines of the fact that `setFile` replaces whatever is already there with the expansion of a twig template. I suppose my twig template could simply have substitutions for re-inserting the former contents of the composer.json file.

As currently written, this warns the user that the default fallback of loading drush.services.yml is temporary, and will be changed (presumably prior to 9.0.0 stable) to always require the entry in composer.json. We could consider keeping the default fallback indefinitely, though, as it probably would not do any harm to do so. I'd prefer that everyone name and version it explicitly in composer.json, but there are already a number of Drush 9 extensions that would have to change if we required this.

greg-1-anderson:
@Chi-teck: The existing `setFile` looks like this:
```
  protected function setFile($path, $template, array $vars) {
    $this->files[$path] = [
      'content' => $this->render($template, $vars),
      'action' => 'replace',
    ];
  }
```
What I want for my purpose is a variant of this that allows me to bypass the template entirely. Perhaps:
```
  protected function setFileContent($path, $content) {
    $this->files[$path] = [
      'content' => $content,
      'action' => 'replace',
    ];
  }
```
The way I'd use this would be to read in the existing composer.json file, decode_json it, add my own entries, and then encode_json it again. Maybe even the following would be useful:
```
  protected function setFileJson($path, $data) {
    $this->setFileContent($path, encode_json($data));
  }
```
It would work to add this directly to Drush. Should we put it in the BaseGenerator class instead?

greg-1-anderson:
Also, the existing Drush CommandFile generator uses the `interact` method to add its file data; however, `interact` is not called at all in `--no-interact` mode, so perhaps this is not ideal.

greg-1-anderson:
Implemented editing an existing composer.json file.

Chi-teck:
`$files` is not a private property and generally it is considered as primary way for specifying files to be created. `setFile()` and `setServicesFile()` methods are just helpers that make the code a bit less verbose in certain cases. They never meant to hide or encapsulate `$files` property. Many DCG generators still set `$files` [directly](https://github.com/Chi-teck/drupal-code-generator/blob/master/src/Command/Drupal_8/Plugin/CKEditor.php#L53).

greg-1-anderson:
@Chi-teck OK, maybe I'll just remove the `_`s from my new methods then. Thanks.

docs/commands.md
```diff
@@ -6,11 +6,38 @@ Creating a new Drush command or porting a legacy command is easy. Follow the ste
...
+  "extra": {
+    "drush": {
+      "services": {
+        "drush-9-99.services.yml": "^9.99",
+        "drush.services.yml": "^9"
+      }
+    }
+  }
+```
+In this example, the file `drush-9-99.services.yml` loads commandfile classes that require features only available in Drush 9.99 and later, and drush.services.yml loads an older commandfile implementation for earlier versions of Drush.
```
weitzman: This begs the question - do both files get loaded if weare using drush 9.99 or "first file wins". Lets be explicit about that.

greg-1-anderson: First one wins. I'll clarify the comment.

Chi-teck:
> $files is not a private property and generally considered as primary way for specifying files to be created.

Eventually I deprecated `$files` property and related methods in favor of new `$assets` property which holds a list of `Asset` objects. This enables using [chain syntax](https://github.com/Chi-teck/drupal-code-generator/blob/master/src/Command/Drupal_8/Service/EventSubscriber.php#L28) and variable substitutions in file path. Besides this template rendering now happens in _execute_ phase not in _interact_ as it did before which I think is more logical.

The old way to set files should still work.

Chi-teck:
Adding simple files can be done as follows.
```
$this->addFile()
  ->path($path)
  ->content($content);
```

weitzman:
So, omitting the extra section is the equivalent of entering this into a module's composer.json?

```
"extra": {
     "drush": {
       "services": {
         "drush.services.yml": "*"
       }
     }
   }
```

If so, is the idea that we will ask implementers to use a different filename (drush10.services.yml) when the API changes?

greg-1-anderson:
The general idea is that Drush 10 (and Drush 8, if we backport this mechanism) will not have a default services section, so those versions of Drush will not load any extensions at all if composer.json is empty.  Drush extensions written for Drush 10, then, must have a `"drush.services.yml": "^10"` entry to load at all. If the extension is somehow compatible with both versions of Drush, then it would say `"drush.services.yml": "^9|^10"`. If the version constraint does not include 9, then Drush 9 will not load the services, regardless of what the filename is. Folks can choose different filenames if they want, e.g. if they support multiple versions of Drush by way of separate services files that inject different classes based on the Drush version.

The warning message currently tells folks that they must add the Drush services entry to composer.json, as it will be required in a future version of Drush 9. We have several options:

1. Require the services entry from 9.0.0-stable (breaking about a dozen Drush 9 extensions, until they are fixed).
1. Leave this message as it is, and require it later, after we have submitted patches to all of the existing Drush 9 extensions.
1. Just take out the message and allow the implicit entry to stand for Drush 9.

The last option is sufficient, but may be slightly confusing. Consistently requiring the entry would be more clear.

weitzman:
That strategy sounds good to me. Thanks for elaborating.

Feel free to merge this when you are satisfied. Note https://github.com/drush-ops/drush/pull/3191#issuecomment-347148891

greg-1-anderson:
@Chi-teck Tried to use `addFile` as you suggested above and got:
```
Call to undefined method Drush\Commands\generate\Generators\Drush\DrushCommandFile::addFile()
```
Is this feature available in a stable release yet?

Chi-teck:
@greg-1-anderson not yet, the release should happen in a few days, because this needs more testing.

For now I propose using old syntax. Later  I can submit a PR to remove all deprecated code from Drush generators.

FGM:
Just noticed an issue with this, albeit non blocking : in projects, the custom module code, unlike the contrib code, is normally committed to the project repo and /not/ brought in by Composer, so that custom logic does not have a composer.json at all, and needs to have one added just for the section, which is rather weird-looking.

greg-1-anderson:
@FGM Yes, that's a good point. However, in the balance of things, it doesn't bother me too much. The alternative would be to create even another metadata file, and I think I like that even less than vestigial composer.json files in custom modules.

Issue #3193, opened 2017-11-28
Allow updating modules with new service dependencies


claudiu-cristea:
See https://www.drupal.org/project/drupal/issues/2863986 for a description of the problem.

If I'm applying the patch from https://www.drupal.org/project/drupal/issues/2863986#comment-12361105, the problem is fixed with `update.php`, but not with Drush.

So, somehow we need to alter in the same way and remove the services. But I'm totally lost with Drush architecture. I have no idea how to hook into `\Drush\Runtime\Runtime` or `\Drush\Runtime\DependencyInjection::initContainer` based on the command (we have to check if the `updatedb` command is about to be run).

Potentially related to #2611

pfrenssen:
I am looking into this. We need to be able to use the `UpdateKernel` instead of the standard `DrupalKernel` when running `updatedb`.

Ideally we should be able to put information about which kernel to use in the annotation of the command and pass this on. The annotation data is available up to `BootstrapHook::initialize()` which then strips out the bootstrap phase from the annotation data and only passes this on to `BootstrapManager::bootstrapToPhase()`. I think it would make sense to pass on the kernel to use as an additional option to `::bootstrapToPhase()`.

At this time I haven't put much thought into the compatibility between the different Drupal versions. Drupal 7 and older don't need it but Drupal 9 and later will probably still have it in one form or another. I want to avoid hardcoding the full FQN `\Drush\Drupal\UpdateKernel` in the annotation of the command, since this is the responsibility of the core version to resolve (in our case this will be handled by `DrupalBoot8::bootstrapDrupalConfiguration()`.

 Our immediate use case is to be able to bootstrap the `UpdateKernel` so I am thinking of supporting two constants right now: `const KERNEL_DRUPAL = 'drupal'` and `const KERNEL_UPDATE = 'update'` so we can have an annotation like this:

```
@kernel: update
```

weitzman:
I looked into this a little bit but don't have much guidance. Perhaps @greg-1-anderson could chime in.

I will say that this is a very infrequent need so if we end up special casing updatedb thats fine for now. Drupal doesn't have an abundance of kernels.

pfrenssen:
OK thanks that's good to know. I'll have a go at this.

weitzman:
A couple ideas:

1. Build a new UpdateKernel() at start of updatedb and then overwrite the DrupalKernal by making a new \Drush\Boot\DrupalBoot8::setKernel method.
[...]

PR #3196, opened 2017-11-28
Initial attempt at using UpdateKernel for the updatedb command

pfrenssen:
A possible approach for #3193.


src/Drupal/Commands/core/BatchCommands.php
```diff
@@ -13,6 +13,7 @@ class BatchCommands extends DrushCommands
      * @aliases batch-process
      * @param $batch_id The batch id that will be processed.
      * @hidden
+     * @kernel update
```
pfrenssen: This is obviously not correct, but it is an attempt at getting the batch process to use the `UpdateKernel`. This is not working unfortunately, for a reason I don't understand yet the `batch-process` command is not found by the Symfony Console `Application`. It throws a `CommandNotFoundException` with message `Command "batch-process" is not defined.`. This is caught in `\Drush\Application::bootstrapAndFind()` which will then proceed to raise the bootstrap level to maximum, booting the wrong kernel.

src/Boot/BootstrapHook.php
```diff
@@ -24,7 +24,8 @@ public function initialize(InputInterface $input, AnnotationData $annotationData
     {
         // Get the @bootstrap annotation. If there isn't one, then assume NONE.
         $phase = $annotationData->get('bootstrap', 'none');
-        $bootstrap_successful = $this->bootstrapManager->bootstrapToPhase($phase);
+        $kernel = $annotationData->get('kernel', 'drupal');
+        $bootstrap_successful = $this->bootstrapManager->bootstrapToPhase($phase, $kernel);
```
greg-1-anderson: Bootstrap manager should not receive Drupal-specific data. Pass in the annotation data here, and let the Drupal 8 bootstrap object look up the Kernel from the annotations.

src/Drupal/Commands/core/BatchCommands.php
greg-1-anderson: I think the best way to be successful here would be to factor out the batch process execution code into a separate class, and use that class from the batch process command and the updatedb command.

src/Boot/DrupalBoot8.php
```diff
@@ -224,4 +225,22 @@ public function terminate()
...
+  /**
+   * Returns the factory method that can be used to retrieve the given kernel.
+   *
+   * @param string $kernel
+   *   The kernel to retrieve.
+   *
+   * @return callable
+   *   The factory method.
+   */
+    protected function getKernelFactory($kernel)
```
greg-1-anderson: This should probably be a method of Drush\Boot\Kernels.

src/Drupal/UpdateKernel.php
```diff
@@ -0,0 +1,12 @@
+<?php
+namespace Drush\Drupal;
+
+use Drupal\Core\Update\UpdateKernel as DrupalUpdateKernel;
+
+/**
+ * Overridden version of UpdateKernel adapted to the needs of Drush.
+ */
+class UpdateKernel extends DrupalUpdateKernel
```
greg-1-anderson: We should probably just use `Drupal\Core\Update\UpdateKernel` directly for the updatedb command. This means that commands that bootstrap with the UpdateKernel cannot use Drush commands. If we are okay with this limitation, then we wouldn't need the DrupalKernelTrait.

The one way in which this could go really wrong would be if updatedb did a cache rebuild at the end (in the Drupal code), and that cache rebuild cached all of the services, excluding the Drush services. We could get around this by using `drush_invoke_process` at the end of updatedb (in Drush code) to call the cache:rebuild command. It's likely that we already do this.

src/Boot/DrupalBoot8.php
pfrenssen: Ok! I put this in `DrupalBoot8` because these factory methods are Drupal 8-specific and might be different for Drupal 9 in the future. I agree that for the moment we only have one set of kernels, so we can put the code in `Kernels` instead.

src/Boot/BootstrapHook.php
pfrenssen: Done.

src/Boot/DrupalBoot8.php
pfrenssen: Done.

src/Drupal/UpdateKernel.php
pfrenssen: Already reverted to the vanilla `UpdateKernel`. Still need to see where this will go wrong.

src/Boot/DrupalBoot8.php
greg-1-anderson: You make a good point about Drupal 9. I like the separation provided by the `Kernels` class, though; I suppose if this needs to change for Drupal 9, then we'd just make a parallel subclass for Drupal 9 and instantiate it from DrupalBoot9.

src/Drupal/UpdateKernel.php
pfrenssen: It appears we still need those overrides. `drush_invoke_process()` relies on the `DrushCommand` services being available on the container. They get added by the `DrushServiceModifier`.

pfrenssen:
I have it working now, I didn't even need to refactor the batch process code into a service. Because the batch code is procedural and in the global namespace I could just call `drush_batch_command()` directly.

The process works fine now. The updates are executed successfully, and the cache is cleared at the end. I do get some strange output, some empty notices and a line that mentions `Array Array`. I am not sure if this is related to this PR, but am going to look into that right now.

```
$ drush updb -y
 --------------------- -------------------------------------- ------------- ----------------------------------------------------------------------- 
  Module                Update ID                              Type          Description                                                            
 --------------------- -------------------------------------- ------------- ----------------------------------------------------------------------- 
  joinup_subscription   install_flag                           post-update   Install the 'flag' module.                                             
  joinup_user           add_subscribe_discussion_permissions   post-update   Adds permissions to subscribe to discussions for authenticated users.  
 --------------------- -------------------------------------- ------------- ----------------------------------------------------------------------- 

 // Do you wish to run the specified pending updates?: yes.                                                             

 [ok] Post updating joinup_subscription
 [ok] Post updating joinup_user
 [notice] 
 [notice] 
 [notice] 
 [notice] 
Array	Array
 [success] Finished performing updates.
```

pfrenssen:
I'm getting those notices and the `Array Array` output also on the master branch so it is not this PR that causes it.

Going to start working on a test for this.

pfrenssen:
I'm working on the test in #3206.

pfrenssen:
Closing in favor of #3206 which starts with a failing test and is rebased to clean up the long string of commits.

PR #3198, opened 2017-11-28
Test consolidation/config PR

greg-1-anderson:


greg-1-anderson:
Our sut building doesn't work well with this construct. Just going to skip this rather than fix it.

PR #3201, opened 2017-11-29
The debug logs come out as notice

pguillard:
I guess this was already planned :
// TODO: RfcLogLevel::DEBUG should be 'debug' rather than 'notice'?

I get too much logs on my production servers.


pguillard:
[the_debug_logs_come_out_as_notice.patch.txt](https://github.com/drush-ops/drush/files/1513871/the_debug_logs_come_out_as_notice.patch.txt)

Issue #3202, opened 2017-11-29
Allow drush to use environment variables in alias files


legovaer:
For Drush < 9 it was rather easy to inject environment variables into an alias file as it was written in PHP. Now that we're moving towards the yaml format for alias files in Drush 9.x it is more complex.

It would be nice to allow users to inject environment variables in their drush files the [same way as Symfony](https://symfony.com/doc/current/configuration/external_parameters.html) allows users it to do.

The format will be `%env(VARIABLE)%`.

weitzman:
You can override values by env variable but thats not documented yet - #3009 

As a workaround, see https://github.com/drush-ops/drush/blob/master/examples/Commands/SiteAliasAlterCommands.php where you can use arbitary php to override alias pieces.

Neither of those approaches is better than your proposal so leaving this open, though it may move to https://github.com/consolidation/config

greg-1-anderson:
See https://github.com/grasmash/yaml-expander/issues/9

legovaer:
Oh cool. I'll go ahead and try testing with [yaml-expander](https://github.com/grasmash/yaml-expander). I hope I get it to work with Drush :)

PR #3203, opened 2017-11-29
#3202 allow environment variables in Drush 9.x

legovaer:
This PR will allow users to have environment variables in their drush aliases as described in #3202


greg-1-anderson:
As implemented, this only allows environment substitution in code that uses the legacy site alias API. Modern config API calls miss out on this feature.

I think we should contribute via https://github.com/grasmash/yaml-expander/issues/9.

Issue #3199, opened 2017-11-29
Make drush_scan_directory() ignored directories configurable


ademarco:
Work done in #2062 excluded `node_modules` and `bower_components` in `drush_scan_directory()`:

```php
$nomask = array_merge($nomask, array('node_modules', 'bower_components'));
```

We could make this configurable with a global Drush option that defaults to the values above, like:

```php
$options['ignored-directories'] = ['node_modules', 'bower_components'];
```

Would this be something interesting? I could work on it if so.

drupol:
:+1:

greg-1-anderson:
What are the use cases for this?  In Drush 9, we simply removed deep scanning for command files and aliases. It seems that limiting scan depth (via option) would be a better place to start than customizing the exclusions. Even better would be to prevent overlap between the places Drush is scanning, and the places where you have large directories.

What are the large directories that you want to exclude, and where are they located in your Drupal site's codebase?

ademarco:
So the use-case is pretty much what described here: https://github.com/nuvoleweb/drupal-component-scaffold

We want to develop Drupal modules and themes as any other modern PHP library is developed, meaning:

1. Clone module repository
2. Rune composer install: this downloads all development dependencies which, in this case, is Drupal core, required modules etc. and build a fully functional test Drupal site within the module's directory
3. Finally: symlink repository root, (i.e. the module/theme directory) in the right place within the built target site, and here is where recursion happens.

The patch above would allow us to exclude the build directory avoiding Drush to recurse following the symlink to the very same repository root.

The approach described above has several benefits, like:

- Lean travis (or any other) CI configuration files
- Allow all module's maintainers / contributors to easily bootstrap the same development environment used on CI lowering the barrier to work with tests etc.

For a working example of such an approach check the [UI Patterns module](https://www.drupal.org/project/ui_patterns), especially its [travis configuration file](https://github.com/nuvoleweb/ui_patterns/blob/8.x-1.x/.travis.yml) and [composer.json](https://github.com/nuvoleweb/ui_patterns/blob/8.x-1.x/composer.json).

weitzman:
[...]

PR #3204, opened 2017-11-30
Issue #3199: Add ignored-directories option.

ademarco:
Ref. #3199


greg-1-anderson:
c.f. https://github.com/drush-ops/drush/issues/3199#issuecomment-348196845

The function you have modified here is deprecated; Drush 9 will be moving to symfony/finder for scanning for files, so this implementation will be less viable as time goes along. `drush_get_option_list` might even be obsolete; I'm not sure that we populate the old contexts any more.

ademarco:
Ok thanks! I'll open a PR against 8.x.

PR #3206, opened 2017-12-01
Allow updating modules with new service dependencies

pfrenssen:
This is a continuation of #3196, fixing #3193.

The PR in #3196 is now functional but is rather messy. This is a fresh one that starts with the test coverage.


weitzman:
Hi @pfrenssen. Is this ready for review?

pfrenssen:
Not yet, I'm attempting to write a test, but I haven't managed to find a minimal use case yet that proves the failure.

I have 4 hours left to work on this today, I hope to get there in time.

weitzman:
Thanks for all the work here. 

I’d love to hear from @berdir and others about just removing —entity-updates option from updatedb. Apparently it is now just a “developer tool”. Folks could still run entity-updates command separately. This would make the code simpler.

pfrenssen:
Yeah that's fine, it was because `--entity-updates` was running in our distribution that this bug was exposed to us. I am very close now to having a working test case that will hopefully work even without the entity updates.

I have to do a bit of catching up on whether we will be needing to do entity updates in post-update hooks.

pfrenssen:
OK the test case is giving back the correct failure now:

```
  [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]   
  The service "woot.depending_service" has a dependency on a non-existent ser  
  vice "devel.dumper".
```

I will finalize the test case: remove the debugging statements and add some missing documentation. I'll then rebase this so it is one single, clean commit.

Next up is applying the fix from #3196, which will hopefully turn this green again and prove that the bug is fixed.

pfrenssen:
Failing test is ready. I will now rebase the fix from #3196 on top of this.

pfrenssen:
@weitzman @greg-1-anderson I'm done, this is ready for review!

tests/resources/codebase/composer-highest.json
```diff
@@ -37,6 +37,10 @@
       "web/themes/contrib/{$name}": ["type:drupal-theme"],
       "drush/contrib/{$name}": ["type:drupal-drush"]
     },
-    "patches": {}
+    "patches": {
+      "drupal/core": {
+        "Allow updating modules with new service dependencies https://www.drupal.org/project/drupal/issues/2863986": "https://www.drupal.org/files/issues/2863986-2-49.patch"
```
pfrenssen: This currently depends on the patch from the core issue for the test coverage.

src/Boot/BootstrapManager.php
```diff
@@ -377,14 +380,20 @@ public function bootstrapValidate($phase)
...
      *
      * @return bool
      *   TRUE if the specified bootstrap phase has completed.
+     *
+     * @throws \Exception
+     *   Thrown when an unknown bootstrap phase is passed in the annotation
+     *   data.
      */
-    public function bootstrapToPhase($bootstrapPhase)
+    public function bootstrapToPhase(AnnotationData $annotationData)
```
weitzman: The signature here feels less usable/convenient. One has to build an AnnotationData.

pfrenssen:
> The other option that I mentioned earlier is to remove --entity-updates from updatedb.

I think that's a good idea, but it won't help with the underlying issue in this case. It turns out that the problem wasn't with the entity updates but with not having a compiled dependency injection container available. This is something that happens when running entity updates, but it might happen for other reasons outside of the control of Drush as well.

pfrenssen:
I tried to convert the `AnnotationData` into a regular array but I made a mistake somewhere and this causes a failure in `AnnotatedCommandCase::testExecute()`. I reverted it for now, I'll have a look at it again later.

pfrenssen:
I am going to look into the test failures. I am going to get rid of the `AnnotationData` -> array conversion, it seems to cause problems in other tests and it's probably not so terrible to keep it in after all. There is only one call to `BootstrapManager::bootstrapToPhase()` and it is done in `BootstrapHook::initialize()` and this actually has the full `AnnotationData` object available, so it doesn't need to build it.

I'm also thinking that future wise if anyone is going to add code that passes annotation data to the method, then they will probably also have the `AnnotationData` object readily available, it seems unlikely that people will come up with their own way to parse the annotations.

For reference here is a link to the original commit that implemented the `AnnotationData` to array conversion: https://github.com/drush-ops/drush/pull/3206/commits/e951152575a264131fba389e2111975ad34857c6 - in case we want to get back to this approach after all.

pfrenssen:
OK it's green again, @weitzman do you care a lot about getting rid of the `AnnotationData` in `BootstrapManager::bootstrapToPhase()`? This method is only called once in the code base, and it has the object readily available at this point. I think also that this class better represents the annotation data than a simple array.

pfrenssen:
Ah wait I have a good solution that will be convenient, let's change the signature to:

```
public function bootstrapToPhase($bootstrapPhase, AnnotationData $annotationData)
```

weitzman:
Thanks for updating the signature. Lets get some feedback from @greg-1-anderson

claudiu-cristea:
Fixes #3193

claudiu-cristea:
@weitzman, @greg-1-anderson I tested this PR manually and works well. Also I see it has good test coverage. Together with the d.o patch (https://www.drupal.org/project/drupal/issues/2863986) is doing the job. What else?

PS: I see that the PR doesn't apply. Would you like me to create a new PR based on fresh `master`?

pfrenssen:
I can reroll this.

pfrenssen:
There was a conflict after #3236 went in. Merged latest master.

pfrenssen:
There was a merge conflict due to https://github.com/drush-ops/drush/commit/f709c11e70370e23bf21cad68bfa795ecc99b899 - rebased on latest master.

weitzman:
@pfrenssen Feel free to merge this once its green. Seems like there is a coding standards problem right now.

pfrenssen:
Fixed the coding standards violations and rebased on the latest master. It's all green now so I'm merging. We still have the patch from the upstream issue in the composer manifest of the D8 test suite so this is something that will fail once it gets committed upstream and a new release of Drupal core is made. But we can deal with it then, it will just be a matter of removing the patch.

PR #3211, opened 2017-12-06
Remove deprecated code from Drush generators

Chi-teck:
This might require manual testing.


weitzman:
At first glance the the test fail looks unrelated.

PR #3213, opened 2017-12-06
Enable calling module and theme generators by their aliases

Chi-teck:
Currently it's not possible to run `drush generate module` or `drush generate theme` because Symfony console application cannot identify these commands by aliases (see https://github.com/symfony/symfony/issues/25355). As a workaround I propose we identify `generate module` command explicitly because it is potentially one of the most used generators.

For `generate theme` it is not a problem any longer as it has been renamed (d8:theme:standard => d8:theme) in DCG 1.21.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
   "require": {
     "php": ">=5.6.0",
     "ext-dom": "*",
-    "chi-teck/drupal-code-generator": "^1.17.3",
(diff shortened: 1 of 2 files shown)


weitzman:
I'm retesting Travis now ... Needs a tiny fix for https://circleci.com/gh/drush-ops/drush/1232?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #3217, opened 2017-12-06
fix password array key

### LINKED ISSUE #3216: drush9 user:information does not print password
Because the infoArray method has the wrong key: https://github.com/drush-ops/drush/blob/master/src/Drupal/Commands/core/UserCommands.php#L331

It should match the key set in the annotation: https://github.com/drush-ops/drush/blob/master/src/Drupal/Commands/core/UserCommands.php#L34

Example: `drush uinf admin --fields=name,pass` only shows name.

PR coming.

richardbporter:
Resolves #3216


greg-1-anderson:
AnnotatedCommandCase seems to be failing in multiple PRs with seemingly unrelated changes -- and yet, master is passing. I'm not sure why this is.

PR #3218, opened 2017-12-07
Fix router.php to work with update.php

alexpott:
Whilst working with drush runserver for a test environment I discovered that it doesn't support update.php. Drupal core's script has the same problem see: https://www.drupal.org/project/drupal/issues/2929198#comment-12374814

Steps to test
1. Install standard Drupal
2. Start up server using drush runserver --default-server=builtin 8080
3. Visit http://localhost:8080/update.php
4. Click on the continue button
5. You should see the next page either the review or a message saying no updates. Without this patch you get a 404.
6. Ensure that admin/structure/types/manage/article/fields/node.article.body is not broken since this is why with the script in the first place - see https://bugs.php.net/bug.php?id=61286
7. Ensure that http://localhost:8888/update.php/selection/blahd.blah
falls back to an update screen since this is the behaviour on Apache.

### DIFF
--- a/misc/d8-rs-router.php
+++ b/misc/d8-rs-router.php
@@ -26,8 +26,34 @@ $base_url = runserver_env('RUNSERVER_BASE_URL');
 // contain multiple dots (such as config entity IDs) in the path. Since this is
 // a virtual resource, served by index.php set the script name explicitly.
 // See https://github.com/drush-ops/drush/issues/2033 for more information.
-$_SERVER['SCRIPT_NAME'] = '/index.php';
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks. Would be great to get help from someone in running through those testing steps.

chrfritsch:
I ran through the testing steps and can confirm that it works with the patch.

I also applied the patch to commands/runserver/d8-rs-router.php in drush 8, and that works, too.

alexpott:
This broke image styles - sorry.

New PR to fix that https://github.com/drush-ops/drush/pull/3264

PR #3220, opened 2017-12-07
Drop support for Symfony 2 in Drush9's composer.json.

greg-1-anderson:
**WHAT**

Remove Symfony 2 from Drush9's composer.json

**WHY**

Some of our dependencies are going to be dropping Symfony 2 support. Dropping this now means we won't need to jump up to Drush 10 just to take the next major version of some component that is no longer compatible with Symfony 2, which is already obsolete anyway.

**IMPACT**

Site-local installations of Drush 9 will no longer be possible with Drupal 8.3.x (which has reached EOL). Note that this configuration currently works, but is not officially supported. Note also that global installations of Drush 9 will continue to work with Drupal 8.3.x, although this will continue to be an unsupported configuration.

It is our expectation that folks will use Drush 8 for these versions of Drupal.


weitzman:
I've got some cold feet on this, based on how much Drupal 8.3 is out in the wild, and the lack of tangible benefit to dropping Symfony2. 

I propose we add a message at end of Composer install for all users of Symfony 2 thats similar to `Package guzzle/inflection is abandoned, you should avoid using it. Use guzzle/guzzle instead.`. Insted, we state that Symfony2 support for Drush is ending soon, and user should update to Drupal 8.4+

Chi-teck:
http://docs.drush.org/en/master/install/#drupal-compatibility
Per the documentation Drush 9 does not support Drupal 8.3. Current support state can be called "happens to work". From this point it should be eligible to drop this unofficial support of Drupal 8.3 in any Drush 9 minor release when the need arises.

Developers also can simply stick to particular Drush 9 release that support Drupal 8.3 (say 9.0-rc1).

greg-1-anderson:
We cannot mark Symfony 2 as abandoned for two reasons:

1. Only the owner of the project can do that
2. Abandoned status is per project, so it's not possible to abandon Symfony 2 without also abandoning Symfony 3 and 4.

Instead of abandoning a project, we could instead use a post-install command to check for Symfony 2, and pring a message then. Unfortunately, these scripts only run from the top-level composer.json file. This means that we could print a message after `composer install` of Drush, but that message would not be printed if you required Drush as a dependency of a Drupal 8.3.x project.

Perhaps the right thing to do would be to modify the output of `drush status` in table format only to include the support status of the current version of Drupal.

Recommended values:

- **Deprecated** Works and supported for now, but definitely will not be supported in next major version of Drush (currently not printed for any version of Drupal)
- **Unsupported** "Happens to work", but not supported, and may stop working at any time (print this for Drupal 8.3.x)
- **Obsolete** Has known operational problems; use not recommended (print this for Drupal 8.2.x and earlier)

Seems too noisy to print a warning on every command, but folks run `drush status` very frequently, so perhaps the would be a good place for this warning.

We might want ton just hard fail and bail out of all commands for obsolete versions, though.

weitzman:
OK, looks like we aren't going to drop Symfony2 in the next few months. I'm going to close this issue. We can revisit later.

PR #3231, opened 2017-12-11
Fix 8.x symfony var dumper conflicts

solody:


weitzman:
We need testing before we support symfony4.

PR #3232, opened 2017-12-11
Next cache dir candidates are not tested if first fails

### LINKED ISSUE #3221: Drush no longers works without write access to the users homefolder
The refactoring in 9.0.0-rc (issue is not present in 9.0.0-beta9) causes drush to fail with the following error:
>  [error]  Directory /home/ubuntu exists, but is not writable. Please check directory permissions.

This is especially problematic for automated deployment, where the deployment might run as a system user that lacks a home directory.

claudiu-cristea:
Fixes #3221 

After `drush_mkdir()` was deprecated, it always returns `TRUE`, even the cache dir candidate has failed to be created. For this reason, the next candidate never get a chance to be tested or used.

### DIFF
--- a/src/Config/DrushConfig.php
+++ b/src/Config/DrushConfig.php
@@ -52,6 +52,6 @@ class DrushConfig extends ConfigOverlay
                 // Do nothing. Jump to the next candidate.
             }
         }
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks for the fix.

PR #3235, opened 2017-12-12
#3110: Prevent infinite loops in RedispatchToSiteLocal.

grasmash:
This is a partial resolution to #3110.

### DIFF
--- a/src/SiteAlias/AliasRecord.php
+++ b/src/SiteAlias/AliasRecord.php
@@ -4,6 +4,7 @@ namespace Drush\SiteAlias;
 use Consolidation\Config\Config;
 use Consolidation\Config\ConfigInterface;
 use Consolidation\Config\Util\ArrayUtil;
+use Drush\Utils\FsUtils;
 
 /**
  * An alias record is a configuration record containing well-known items.
@@ -116,7 +117,9 @@ class AliasRecord extends Config
      */
     public function root()
     {
-        return realpath($this->get('root'));
+        $root = FsUtils::realpath($this->get('root'));
+
+        return $root;
     }
 
     /**
--- a/src/SiteAlias/AliasRecord.php
+++ b/src/SiteAlias/AliasRecord.php
@@ -116,7 +116,7 @@ class AliasRecord extends Config
      */
     public function root()
     {
-        return $this->get('root');
+        return realpath($this->get('root'));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


grasmash:
This is a partial resolution to #3110.

greg-1-anderson:
The `realpath` call from a few lines below is like a cancer spreading through the code. First it moves up here; now it also seems that there is an edge case where the `if` statement a few lines above might also benefit from a call from realpath. There is also the very real possibility that there are other places in the code that are not quite right, and would also need a realpath call. This is not quite right.

It's not your fault, but we shouldn't continue like this. Could you find the original place in the code where `vendor` and `siteLocalDrush` are first calculated, and insert the realpath in at that point?

We must be careful with `realpath`, as it will return `false` if the provided path does not exist at all, and in general is damaging for any paths that may eventually be used in a remote call. Both `vendor` and `siteLocalDrush` should only be calculated for local paths, though, so I think in this case we should simplify.

grasmash:
@greg-1-anderson I've refactored the PR as per your suggestion. There may be [more paths](https://github.com/drush-ops/drush/blob/master/src/Config/Environment.php#L38) that require `realpath()` wrappers, but I'm only adding those that I know are required.

src/SiteAlias/AliasRecord.php
```diff
@@ -116,7 +117,9 @@ public function hasRoot()
      */
     public function root()
     {
-        return $this->get('root');
+        $root = FsUtils::realpath($this->get('root'));
```
greg-1-anderson: This is unsafe. 'root' might refer to a remote site, and realpath will return `false` if the remote root does not also exist locally. Keep the realpath in the localRoot function.

greg-1-anderson: Oh, I see what you did here. 😄

src/Utils/FsUtils.php
```diff
@@ -60,4 +60,23 @@ public static function prepareBackupDir($subdir = null)
...
+     *
+     * @param string $path
+     *   The path being checked.
+     *
+     * @return string
+     *   The canonicalized absolute pathname.
+     */
+    public static function realpath($path)
+    {
+        if (file_exists($path)) {
```
greg-1-anderson: You should probably just call `realpath` and return the result if it is not false, and return `$path` otherwise. `realpath` and `file_exists` may have slightly different behaviors (e.g. if $path points to a symlink that points to a file that does not exist)

grasmash: Good point.

PR #3237, opened 2017-12-12
Fixed failing updb command.

RoSk0:
It was failing with "Error: Cannot use object of type
Drupal\Core\StringTranslation\TranslatableMarkup as array in
Drush\Commands\core\UpdateDBCommands->updateFinished() (line 357 of
vendor/drush/drush/src/Commands/core/UpdateDBCommands.php).".

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -352,10 +352,8 @@ class UpdateDBCommands extends DrushCommands
         unset($results['#abort']);
         foreach ($results as $module => $updates) {
             foreach ($updates as $number => $update) {
-                if (empty($update['#abort'])) {
-                    foreach ($update['results'] as $result) {
-                        $this->logger()->notice(strip_tags($result['query']));
-                    }
+                if (empty($update['#abort']) && $update['results']['success']) {
+                    $this->logger()->notice(strip_tags($update['results']['query']));
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


tests/resources/modules/d8/woot/woot.install
```diff
@@ -1,8 +1,15 @@
...
 function woot_update_8101() {
-  throw new \Exception('8101 error');
+  return t('woot_update_8101');
+}
+
+/**
+ * Failing update.
+ */
+function woot_update_8102() {
+  throw new \Exception('8102 error');
```
weitzman: Any reason you added a new update?

src/Commands/core/UpdateDBCommands.php
```diff
@@ -352,10 +352,8 @@ public function updateFinished($success, $results, $operations)
         unset($results['#abort']);
         foreach ($results as $module => $updates) {
             foreach ($updates as $number => $update) {
-                if (empty($update['#abort'])) {
-                    foreach ($update['results'] as $result) {
-                        $this->logger()->notice(strip_tags($result['query']));
-                    }
+                if (empty($update['#abort']) && $update['results']['success']) {
+                    $this->logger()->notice(strip_tags($update['results']['query']));
```
weitzman: Looks good.

tests/resources/modules/d8/woot/woot.install
RoSk0: Yes, that was done on purpose to test the change I have made to src/Commands/core/UpdateDBCommands.php .

Without that change it will fail with exception from description.

RoSk0: I created a new pull request https://github.com/drush-ops/drush/pull/3238 fix test changes only to prove the bug.

PR #3241, opened 2017-12-13
Use uri from an alias when that alias has no root.

weitzman:
I have a local site alias that defines a uri but no root. A root is not needed anymore because that gets located correctly based on me using vendor/bin/drush.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -154,6 +154,9 @@ class Application extends SymfonyApplication implements LoggerAwareInterface, Co
             return;
         }
         $selfAliasRecord = $this->aliasManager->getSelf();
+        if (!$selfAliasRecord->hasRoot() && !$this->bootstrapManager()->drupalFinder()->getDrupalRoot()) {
+            return;
+        }
         $uri = $selfAliasRecord->uri();
 
         if (empty($uri)) {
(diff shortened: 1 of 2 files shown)


src/Preflight/Preflight.php
```diff
@@ -263,7 +263,9 @@ public function preflight($argv)
 
         // Process the selected alias. This might change the selected site,
         // so we will add new site-wide config location for the new root.
-        $root = $this->setSelectedSite($selfAliasRecord->localRoot());
+        if ($localRoot = $selfAliasRecord->localRoot()) {
```
greg-1-anderson: Minor style point: I'd rather see this `if` inside of the `setSelectedSite()` method rather than here.

src/Application.php
```diff
@@ -154,9 +154,6 @@ public function refineUriSelection($cwd)
             return;
         }
         $selfAliasRecord = $this->aliasManager->getSelf();
-        if (!$selfAliasRecord->hasRoot()) {
```
greg-1-anderson: Do we maybe need to still exist if there is neither a site record root nor a preflight-discovered root? I'm unsure of the impact; I'm fine with just removing this if there are no problems in that scenario.

PR #3247, opened 2017-12-15
Fix router.php to work with update.php

chrfritsch:
Accordingly to https://github.com/drush-ops/drush/pull/3218, I would like to see that fix in drush 8 as well.

CC: @alexpott


alexpott:
We need to merge in the changes from https://github.com/drush-ops/drush/pull/3264

weitzman:
Failing CI check

weitzman:
I retested and we are green now. Merged, Thanks.

PR #3249, opened 2017-12-15
Fix #2563. Add diff preview to config-export

### LINKED ISSUE #2563: Feature request add --preview=<format> option to cex - config-export
Hi !
Currently the `drush cim` command has the following option:
`--preview=<list>                          Format for displaying proposed changes. Recognized values: list, diff. Defaults to list.`

This is useful to get a diff of what will be imported. It would be helpful to have this same functionality in the `drush cex` so that a diff of what will be export can be seen. I know that I can do this with git after the configuration as be exported. Or is this not possible without a previous export? Maybe a temporary export to a tmp directory would make this possible.

Thanks for your consideration of this new feature.
Frederick

weitzman:

### DIFF
--- a/src/Drupal/Commands/config/ConfigCommands.php
+++ b/src/Drupal/Commands/config/ConfigCommands.php
@@ -317,7 +317,7 @@ class ConfigCommands extends DrushCommands
      */
     public function getChanges($target_storage)
     {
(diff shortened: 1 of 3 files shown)


bircher:
+1

PR #3250, opened 2017-12-15
Fix #3214. Canonicalize config paths before comparing.

### LINKED ISSUE #3214: Do not rely on file path for cim and cex
Several issues have been reported on drupal.org with projects that use [Config Filter](https://www.drupal.org/project/config_filter).
Config Filter decorates the `config.storage.sync` service to enhance cores configuration management.
This works with drush since #2471 but it seems that with the global drush it has stopped working.

Issues on drupal.org:
- [Config split not being picked up in settings.php on Pantheon](https://www.drupal.org/project/config_split/issues/2920397)
- [Drush CEX doesn't include splits anymore](https://www.drupal.org/project/config_split/issues/2916765)
- [Drupal console and drush doesn't respect config ignore filters](https://www.drupal.org/project/config_ignore/issues/2916654)

@greg-1-anderson mentioned on slack that this could be the result of using the Path component.

Proposed solution: Pass around the argument to the command or the config storage object rather than the file path to the sync storage between method calls for config import and export.
Ideally we could also add a test that asserts the usage of the core service when no argument is passed to the commands.

weitzman:


bircher:
That looks like it could solve the problem!
+1

Issue #3251, opened 2017-12-15
Add command to export .po files


Sutharsan:
I want to deploy custom translations (translations that are set in the admin interface to override community translations). I want to deploy them by exporting and importing custom translations. This issue pairs with my proposal in #2164.

```
Exports to a gettext translation file.

Examples:
 locale-export nl > drupal.po
 locale-export nl --type=customized > custom-translations.po
 locale-export --template > drupal.pot

Arguments:
 langcode		The language code of the exported translations. Use 'template' to
 			export a translation template that contains only source strings.

Options:
 --types		The type of translations to be exported. Defaults to all below options.
 			Options:
 			- customized: Export only custom translations.
			- not-customized: Export only non-custom translations.
			- not-translated: Export untranslated strings too.
--template		Export a gettext template file. 
```

TessaBakker:
1. I would change --type='' to --include=[non-custom, custom, untranslated], because this is how the export options are at admin/config/regional/translate/export.

2. Also I'm not a fan of using 'language' while you have to enter a language code, suggestion: langcode instead of language.

Sutharsan:
I agree with changing 'language' to 'langcode'.

I like to keep the name of the 'type' option the same as with the import command. Not sure if 'include' makes things clearer.

'untranslated' does not require a language code. Either, as I propose, we use 'template' as langcode, or make the langcode an optional argument (or are arguments always required?)

TessaBakker:
Include is a reserved option and type is better as suggested. Because you can use multiple types if needed, I changed it to types.
The type 'not_translated' is needed to get all untranslated strings of a language.

I think the command #3253 is ready to be reviewed.

Sutharsan:
Instead of 'not_translated' I proposed a special "language" 'template', perhaps you overlooked that. The example: `locale-export template > drupal.pot` It is a special, and probably rarely used use case. Most developers will use Potx module and target specific modules or directories. Because of the special nature, I do suggest to compromise the 99% use case as little as possible.
As a bonus, the code is probably simpler when using langcode 'template' instead of type 'not_translated'.

Sutharsan:
I propose to export *all* types by default, that is what core does, and I think it is a sensible default because it exports all strings.

Sutharsan:
Updated the initial proposal with 'langcode' and 'types' as per your suggestions.

Sutharsan:
[...]

PR #3253, opened 2017-12-16
#3251: Add command to export .po files

TessaBakker:
See #3251 for more information.


src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,83 @@ public function update($options = ['langcodes' => self::REQ])
...
+            $language = null;
+        }
+
+        $includes_allowed = [
+            'non_customized',
+            'customized',
+            'not_translated',
+        ];
+
+        foreach ($options['include'] as $include) {
```
Sutharsan: Separate this code out as an input validation.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,83 @@ public function update($options = ['langcodes' => self::REQ])
...
+            if (!in_array($include, $includes_allowed)) {
+                throw new \Exception(dt('Include must contain one of: @types.', ['@types' => implode(', ', $includes_allowed)]));
+            }
+        }
+
+        $content_options = is_array($options['include']) ? $options['include'] : [$options['include']];
+        $reader = new PoDatabaseReader();
+        $language_name = '';
+        if ($language != null) {
```
Sutharsan: `if ($language)` will do the same.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,83 @@ public function update($options = ['langcodes' => self::REQ])
...
+            'not_translated',
+        ];
+
+        foreach ($options['include'] as $include) {
+            if (!in_array($include, $includes_allowed)) {
+                throw new \Exception(dt('Include must contain one of: @types.', ['@types' => implode(', ', $includes_allowed)]));
+            }
+        }
+
+        $content_options = is_array($options['include']) ? $options['include'] : [$options['include']];
```
Sutharsan: You don't trust here that $options['include'] is an array, but at line 171 you do rely on it being an array. Do it once, and early.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,83 @@ public function update($options = ['langcodes' => self::REQ])
...
+     *
+     * @throws \Exception
+     * @return string
+     *
+     * @aliases locale-export
+     * @drupal-dependencies locale
+     */
+    public function export($langcode = null, $options = ['include' => ['customized']])
+    {
+        // If template is required, language code is not given.
```
Sutharsan: Not clear why this code exists. Use locale_translatable_language_list() to check for valid languages. Use a (reusable) method for language validation.

Sutharsan:
I did not know I had to complete a review after entering comments... So this review contains some outdated comments.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,89 @@ public function update($options = ['langcodes' => self::REQ])
...
+
+        if ($language != null) {
+            $options['types'] = explode(',', $options['types']);
+            $allowed_types = [
+                'not-customized',
+                'customized',
+                'not-translated',
+            ];
+
+            if (empty(reset($options['types']))) {
```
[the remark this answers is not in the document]
TessaBakker: That was my first idea also, but by default we have $options[...][0] = 0 this is because of the default state of 'types' with only --types you'll get $options['types'][0] = 1;

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,89 @@ public function update($options = ['langcodes' => self::REQ])
...
+            }
+
+            $options['types'] = array_map(function ($value) {
+                return str_replace('-', '_', $value);
+            }, $options['types']);
+
+            $reader->setLangcode($language->getId());
+            $reader->setOptions(array_fill_keys($options['types'], true));
+            $languages = \Drupal::languageManager()->getLanguages();
+            $language_name = isset($languages[$language->getId()]) ? $languages[$language->getId()]->getName() : '';
```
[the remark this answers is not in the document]
TessaBakker: True, don't know why this is there, core has it also, maybe an old "feature"? .. will look into that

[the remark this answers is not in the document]
TessaBakker: Changed

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,76 @@ public function update($options = ['langcodes' => self::REQ])
...
+            if (array_diff($options['types'], $allowed_types)) {
+                throw new \Exception(dt('Types must contain one of: @types.', ['@types' => implode(', ', $allowed_types)]));
+            }
+
+            $reader->setLangcode($language->getId());
+            $reader->setOptions(array_fill_keys($options['types'], true));
+            $languages = \Drupal::languageManager()->getLanguages();
+            $language_name = isset($languages[$language->getId()]) ? $languages[$language->getId()]->getName() : '';
+        }
+
```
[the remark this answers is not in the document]
TessaBakker: Changed

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,83 @@ public function update($options = ['langcodes' => self::REQ])
...
+        $content_options = is_array($options['include']) ? $options['include'] : [$options['include']];
+        $reader = new PoDatabaseReader();
+        $language_name = '';
+        if ($language != null) {
+            $reader->setLangcode($language->getId());
+            $reader->setOptions($content_options);
+            $languages = \Drupal::languageManager()->getLanguages();
+            $language_name = isset($languages[$language->getId()]) ? $languages[$language->getId()]->getName() : '';
```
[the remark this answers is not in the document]
TessaBakker: Changed

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,83 @@ public function update($options = ['langcodes' => self::REQ])
...
+    public function export($langcode = null, $options = ['include' => ['customized']])
+    {
+        // If template is required, language code is not given.
+        if ($langcode != LanguageInterface::LANGCODE_SYSTEM) {
+            $language = \Drupal::languageManager()->getLanguage($langcode);
+        } else {
+            $language = null;
+        }
+
+        $includes_allowed = [
```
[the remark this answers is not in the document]
TessaBakker: Changed

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +131,76 @@ public function update($options = ['langcodes' => self::REQ])
...
+     * @aliases locale-export
+     * @see \Drupal\locale\Form\ExportForm::submitForm
+     *
+     * @throws \Exception
+     * @return string
+     */
+    public function export($langcode = null, $options = ['types' => 'customized'])
+    {
+        // If template is required, language code is not given.
+        if ($langcode != LanguageInterface::LANGCODE_SYSTEM) {
```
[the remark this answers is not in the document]
TessaBakker: Changed

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -144,76 +145,121 @@ public function update($options = ['langcodes' => self::REQ])
      *   Export the Dutch non-customized translations.
      * @usage drush locale:export --langcode=nl --types=customized,not-customized > nl.po
      *   Export the Dutch customized and not customized translations.
-     * @usage drush locale:export > drupal.pot
+     * @usage drush locale:export --template > drupal.pot
```
Sutharsan: Add an @option description for 'template' too.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -144,76 +145,121 @@ public function update($options = ['langcodes' => self::REQ])
...
      *   Export the template file to translate.
      * @aliases locale-export
      * @see \Drupal\locale\Form\ExportForm::submitForm
      *
      * @throws \Exception
      */
-    public function export($options = ['langcode' => self::OPT, 'types' => self::OPT])
+    public function export($options = ['template' => false, 'langcode' => self::OPT, 'types' => self::OPT])
```
Sutharsan: Minor: Order by importance / priority / frequency of use. Move 'template' to the end.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -144,76 +145,121 @@ public function update($options = ['langcodes' => self::REQ])
...
     {
+        $language = null;
+        $poreader_options = null;
+        $template = (bool)$options['template'];
```
Sutharsan: Perhaps name `$useTemplate` or `$use_template`

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -144,76 +145,121 @@ public function update($options = ['langcodes' => self::REQ])
...
-                    '@langcode' => $options['langcode'],
-                ]));
-                return;
-            }
-        } else {
-            $language = null;
+        if ($options['langcode'] != null) {
+            $language = $this->getLanguage($options['langcode']);
+        } elseif ($template === false) {
+            throw new \Exception(dt('Set --langcode or --template, see help for more information.'));
```
Sutharsan: Does Drush have exception types? If so, can we use it instead of the generic `\Exception`

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -144,76 +145,121 @@ public function update($options = ['langcodes' => self::REQ])
...
-                $this->logger()->warning(dt('Language code @langcode is not configured.', [
-                    '@langcode' => $options['langcode'],
-                ]));
-                return;
-            }
-        } else {
-            $language = null;
+        if ($options['langcode'] != null) {
+            $language = $this->getLanguage($options['langcode']);
+        } elseif ($template === false) {
```
Sutharsan: `if ($options['langcode'] != null) {`
Move this logic into ::getLanguage() (or delete it) and base the error message on the value of $language.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -144,76 +145,121 @@ public function update($options = ['langcodes' => self::REQ])
...
-            if (empty(reset($options['types']))) {
-                $options['types'] = $allowed_types;
-            } elseif (array_diff($options['types'], $allowed_types)) {
-                throw new \Exception(dt('Types must contain one of: @types.', [
-                    '@types' => implode(', ', $allowed_types),
-                ]));
-            }
+        if ($language) {
+            $poreader_options = $this->getReaderOptions($options['types']);
+        }
```
Sutharsan: Do we have to call this conditionally? The performance consequence is minimal, the DX impact is significant.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +133,168 @@ public function update($options = ['langcodes' => self::REQ])
...
+
+    /**
+     * Exports to a gettext translation file.
+     *
+     * @command locale:export
+     * @drupal-dependencies locale
+     * @option template To export the template file.
+     * @option langcode The language code of the exported translations.
+     * @option types String types to include, defaults to all types.
+     *      Types: 'not-customized', 'customized', 'not-translated'.
```
TessaBakker: 2 spaces instead of 4

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +133,168 @@ public function update($options = ['langcodes' => self::REQ])
...
+     * @usage drush locale:export --langcode=nl > nl.po
+     *   Export the Dutch translations with all types.
+     * @usage drush locale:export --langcode=nl --types=customized,not-customized > nl.po
+     *   Export the Dutch customized and not customized translations.
+     * @usage drush locale:export --template > drupal.pot
+     *   Export the template file to translate.
+     * @aliases locale-export
+     * @see \Drupal\locale\Form\ExportForm::submitForm
```
Sutharsan: Nit-pick: Add empty line before `@see ...`. Makes it more readable as a separate subject. Not sure if any code style addresses this.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +133,168 @@ public function update($options = ['langcodes' => self::REQ])
...
+    /**
+     * Get PODatabaseReader options.
+     *
+     * @param array $types
+     * @return array
+     * @throws \Exception
+     */
+    private function getReaderOptions(array $types = [])
+    {
+        $allowed_types = [
```
Sutharsan: Either $allowed_types or $types can use some documentation. Allowed types 'key' is type accepted by Reader, 'value' is drush command option input. Without comments, I have to read and interpret the relatively complex last two lines of this method.

TessaBakker:
With the latest commit I changed the code a little bit too much, if there are no types defined by the user, the exported PO file will be a template file for the given language. It's a little bit different than the option --template, because template doesn't show a project name and language information in the template, it's a template for any (new) language. 
But with no defaults for types, it is becoming a little bit unclear what to expect.

**Proposal**
Drop --types and allow types (export options) to be included or excluded.

Pros:
Very clear what will be exported or not
Easier defaults for all and none

Cons:
Little bit more code

**Examples**
```
# Exporting everything
drush locale:export --langcode=nl
drush locale:export --langcode=nl --include

# Exporting 'not-translated'
drush locale:export --langcode=nl --include=not-translated
drush locale:export --langcode=nl --exclude=not-customized,customized

# Exporting language template
drush locale:export --langcode=nl --exclude
```
If exclude is defined, include can't.

**Alternative**
Make every export option a boolean (default: true), but then the commands will become very long to export one type or none. On the other side, the code will be very clean. (My favourite solution, but probable not every developer.)

Other ideas are welcome.

Sutharsan:
It is not clear to me what problem you are trying to solve. Exporting a template with this command is imho an exception. Potx is the go-to for templates. I suggest to drop the template export function.

Lets first focus on the UX of the Drush command (line). Ease of code should be second (not first).
I’m not convinced that include/exclude is an improvement for UX.

Sutharsan:
Cons:
- Negative and double negative logic is inherently difficult (e.g. --exclude=not-translated) and should be avoided. Human brains don't know the word "not", the logic always needs to be translated to positive by reversing, and that is a risk. See next point.
- This will cause confusion: `--include=customized` is not equal to `--exclude=not-customized`
- There is no benefit in replacing `--template` by `--exclude`. It only complicates the use of`--exclude` because it now has two use cases.

TessaBakker:
After some thinking, there was a third solution, so I added that option as an example in the command.

TessaBakker:
Hi @greg-1-anderson,

I think this PR is ready to be merged, unless you have some points for improvement.

When it's merged I will start on the locale:import command.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +133,183 @@ public function update($options = ['langcodes' => self::REQ])
...
+    public function export($langcode = null, $options = ['template' => false, 'types' => self::OPT])
+    {
+        $language = $this->getTranslatableLanguage($langcode);
+        $poreader_options = [];
+
+        if (!(bool)$options['template']) {
+            $poreader_options = $this->convertTypesToPoDbReaderOptions(StringUtils::csvToArray($options['types']));
+        }
+
+        $file_uri = drush_save_data_to_temp_file('temporary://', 'po_');
```
weitzman: WHat are we doing here. Writing the string temporary:// as file contents?

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +133,183 @@ public function update($options = ['langcodes' => self::REQ])
...
+     * @return LanguageInterface|null
+     * @throws \Exception
+     */
+    private function getTranslatableLanguage($langcode)
+    {
+        if (!$langcode) {
+            return null;
+        }
+
+        $language = \Drupal::languageManager()->getLanguage($langcode);
```
weitzman: You should be able to use dependency injection to get the languageManager(). Also Drupal::config() later.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +157,185 @@ public function update($options = ['langcodes' => self::REQ])
 
         drush_backend_batch_process();
     }
+
+    /**
+     * Exports to a gettext translation file.
+     *
+     * @see \Drupal\locale\Form\ExportForm::submitForm
```
weitzman: Lets comment this out because its leads to an error during docs generation  https://github.com/FriendsOfPHP/Sami/issues

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +157,185 @@ public function update($options = ['langcodes' => self::REQ])
...
+     *
+     * @throws \Exception
+     *
+     * @command locale:export
+     * @drupal-dependencies locale
```
weitzman: this should be `@validate-module-enabled` now

Sutharsan:
All passed :)

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -336,4 +398,76 @@ private function isTranslatable(LanguageInterface $language)
...
+            $reader->setOptions($options);
+        }
+
+        $reader_item = $reader->readItem();
+        if (empty($reader_item)) {
+            return false;
+        }
+
+        $header = $reader->getHeader();
+        $header->setProjectName(drush_get_context('DRUSH_DRUPAL_SITE'));
```
weitzman: We are trying to get rid of calls to drush_get_context(). Can you do something like Drupal::config()->get('system.site_name')? Or Drush::uri()

I'll try to more of a review soon.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -155,6 +159,64 @@ public function update($options = ['langcodes' => self::REQ])
...
+     *   Export the Dutch template file to translate.
+     * @aliases locale-export
+     * @validate-module-enabled locale
+     */
+    public function export($langcode = null, $options = ['template' => false, 'types' => self::OPT])
+    {
+        $language = $this->getTranslatableLanguage($langcode);
+        $poreader_options = [];
+
+        if (!(bool)$options['template']) {
```
weitzman: The default value of --template is `false` which tells AnnotatedCommand that no string values are allowed, only bool. We aren't great anymore with options that can be bool or string. I think its possible, but awkward.

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -155,6 +159,64 @@ public function update($options = ['langcodes' => self::REQ])
...
+            $poreader_options = $this->convertTypesToPoDbReaderOptions(StringUtils::csvToArray($options['types']));
+        }
+
+        $file_uri = drush_tempnam('drush_', null, '.po');
+        if ($this->writePoFile($file_uri, $language, $poreader_options)) {
+            $this->output()->writeln(file_get_contents($file_uri));
+        } else {
+            $this->logger()->notice(dt('Nothing to export.'));
```
weitzman: We often use ->success() for this instead of ->notice(). Not a big deal either way.

src/Drupal/Commands/core/LocaleCommands.php
Sutharsan: 'template' is not used as string. I think the explicit type casting causes the confusion. The casting is not needed, now removed.

Sutharsan:
Tests 'test_71_mysql' and 'test_56_lowest' failed with:
>Failed asserting that '[warning] One or more of your dependencies has an outstanding security update.

I don't think this that related to this issue.

greg-1-anderson:
Rebase with master, and that test failure should go away.

TessaBakker:
Tested it locally and seems to work :) this PR can be merged :)

tests/functional/LocaleTest.php
```diff
@@ -11,6 +11,41 @@
...
+
+        // Export template file.
+        $this->drush('locale:export', ['nl'], ['template' => null]);
+        $this->assertGettextTranslation('Devel', '');
+        $this->assertGettextTranslation('Something untranslated', '');
+    }
+
     public function testLocaleImport()
     {
         $this->setUpDrupal(1, true);
```
weitzman: We should do 1 setUpDrupal() per class. See other tests for examples ... Or even better, see if this can be an Integration test instead of a Functional test.

Sutharsan: @weitzman, can you explain which groups of tests you refer to with "Integration test"? Which test(s) can serve as an example?

weitzman: See files in https://github.com/drush-ops/drush/tree/master/tests/integration

Sutharsan:
I added a `::setup()`.

I did try to convert the functional test to integration test, but integration tests don't seem to allow the `drush sql:query` command which this these tests heavily relies on. How important is this to you?

src/Drupal/Commands/core/LocaleCommands.php
greg-1-anderson: We have put off making specific exception types until we have a need to catch exceptions based on the type that is thrown. So far we have not had that need.

In this specific case, I would be -1 on making a Locale-specific exception to throw here. If we were going to use something other than \Exception, I'd want something like CommandFailedException, as the intention here is to stop the execution of the command and display an error message (with corresponding failure code) to the user. If something like that were to be defined, we'd probably want to put it in AnnotatedCommand.

If there are other exceptions that might be caught separately, then it would be okay to start making specific types. However, I think in most cases, a command implementation is only going to want to throw CommandFailedExceptions.

Sutharsan:
What needs to be done on this PR to get it accepted?

soniCaH:
Any news? 👍

weitzman:
I took a look at this and it LGTM. Anyone available to fix the conflict? I will then merge, as long as its green.

Sutharsan:
Hi Moshe, I will do this later today.

Erik

tests/functional/LocaleTest.php
```diff
@@ -10,46 +10,81 @@
...
 {
+    /**
+     * File name of Gettext PO source file.
+     *
+     * @var string
+     */
+    protected $sourceFile;
 
-    public function testLocaleImport()
+    public function setUp()
```
weitzman, line 20: I just noticed that setUp() is now installing Drupal twice, once for each test method. See https://github.com/drush-ops/drush/blob/50ae39d721fd45727349ab5915453c61925764a4/tests/functional/RsyncTest.php#L21-L23 for a simple way to avoid that (just put your logic inside a `getSites()` check.

PR #3258, opened 2017-12-18
WIP Fix #3071. Add sql:fetch command

### LINKED ISSUE #3071: Add sql-fetch command
That would be useful for scripting to be able to retrieve data from database in different formats. As I understood this happens automatically if a command returns _RowsOfFields_. With current implementation of `drush sql-query` command this is not possible because it runs queries through shell. And besides it can be used for any kind of SQL queries not just SELECTs.

weitzman:
Per #3071, make it easy to get data out of the Drupal DB in any format.

The command here can be finished easily but the output formatter system has no way to handle variable field input AFAICT. We ran into this limitation once before, I think.


weitzman:
Arbitrary columns is also needed in `migrate-messages` command which shows rows with keys of dynamic name and number

weitzman:
One option that works today - https://stackoverflow.com/questions/39744888/how-to-select-keys-and-values-from-mysql-database-to-a-json-format

greg-1-anderson:
It's probably a mistake to use RowsOfFields for sql dumps. The library does a lot of processing on the rows, so this would bog down if someone used it to fetch a large database. It's probably better to just return the data as a string and let sql functions such as the one mentioned above do the formatting.

weitzman:
OK, lets close this with the noted workaround.

PR #3259, opened 2017-12-19
Replace DRUSH_NOCOLOR with OutputInterface::isDecorated

webflo:
I did some debugging related to #2214 and noticed that DRUSH_NOCOLOR is mostly unused now. 

---

Disclaimer: This is very low-level code in Drush, not sure if this is the right thing.


greg-1-anderson:
Thanks

Issue #3127, opened 2017-11-01
Non-interactive commands on remote host fail


danepowell:
Non-interactive Drush commands run against a remote site alias will fail with a fairly obscure error.

In my example, `@blted91.local` is a site inside of a DrupalVM:
```yaml
local:
    root: /var/www/blted91/docroot
    uri: 'http://local.blted91.com'
    host: local.blted91.com
    user: vagrant
    ssh:
      options: '-o PasswordAuthentication=no -i /home/dane/.vagrant.d/insecure_private_key'
```

Running `drush @blted91.local status --no-interaction` produces this:
```
Drupal version   : 8.4.0                                                 
 Site URI         : http://local.blted91.com                              
 DB driver        : mysql                                                 
 DB hostname      : localhost                                             
 DB port          : 3306                                                  
 DB username      : drupal                                                
 DB name          : drupal                                                
 Database         : Connected                                             
 Drupal bootstrap : Successful                                            
 Default theme    : bartik                                                
 Admin theme      : seven                                                 
 PHP binary       : /usr/bin/php7.1                                       
 PHP config       : /etc/php/7.1/cli/php.ini                              
 PHP OS           : Linux                                                 
 Drush script     : /var/www/blted91/vendor/drush/drush/drush             
 Drush version    : 9.0.0-beta8                                           
 Drush temp       : /tmp                                                  
 Drush configs    : /var/www/blted91/vendor/drush/drush/drush.yml         
                    /var/www/blted91/drush/drush.yml                      
 Drush aliases    : /var/www/blted91/drush/site-aliases/blted91.alias.yml 
 Install profile  : lightning                                             
 Drupal root      : /var/www/blted91/docroot                              
 Site path        : sites/default                                         
 Files, Public    : sites/default/files                                   
 Files, Private   : /var/www/blted91/files-private                        
 Files, Temp      : /tmp                                                  
 [error]  The command could not be executed successfully (returned:  Drupal version   : 8.4.0                                                 
 Site URI         : http://local.blted91.com                              
 DB driver        : mysql                                                 
 DB hostname      : localhost                                             
 DB port          : 3306                                                  
 DB username      : drupal                                                
 DB name          : drupal                                                
 Database         : Connected                                             
 Drupal bootstrap : Successful                                            
 Default theme    : bartik                                                
 Admin theme      : seven                                                 
 PHP binary       : /usr/bin/php7.1                                       
 PHP config       : /etc/php/7.1/cli/php.ini                              
[...]

PR #3260, opened 2017-12-19
If drush_backend_parse_output() returns a string, it means the command could have run successfully.

jonpugh:
I'm running into a very strange issue with a site hosted on Aegir (Running Drush 8). It only happens with a Drupal 8 platform based on the standard drupal composer project, which itself includes Drush.
 
In aegir, the `drush provision-verify` command uses `drush_invoke_process()` to run `drush provision-save`. When I do this on the Drupal 8 platform, `drush_invoke_process()` throws the error "The command could not be executed successfully".

If I remove the `vendor/drush/drush` folder from the Drupal 8 platform, the command works as expected!

What happens is that `drush_backend_parse_output()` returns the output of the `drush provision-verify` command as a string when it bootstraps the Drupal8/drush9 platform.

I have no idea how this could be happening, but I figured `_drush_backend_invoke()` should handle that string instead of an array.

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -1049,6 +1049,9 @@ function _drush_backend_invoke($cmds, $common_backend_options = array(), $contex
             $ret['concurrent'][] = $values;
           }
         }
+        elseif (is_string($values)) {
+          $ret['output'] = $values;
+        }
         else {
           $ret = drush_set_error('DRUSH_FRAMEWORK_ERROR', dt("The command could not be executed successfully (returned: !return, code: !code)", array("!return" => $proc['output'], "!code" =>  $proc['code'])));
         }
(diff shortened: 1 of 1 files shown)


jonpugh:
Some others who have seen the issue:

https://www.drupal.org/node/2821807
https://www.drupal.org/project/provision/issues/2796555

weitzman:
I'd love to do some root cause analysis here. drush_invoke_process() is supposed to return an array.

greg-1-anderson:
Maybe the remote call is getting switched to interactive mode instead of backend mode?

I second that we should identify the root cause before just fixing the symptom. It could be that we proceed as suggested here, but we should know why first, lest we mask some other problem.

weitzman:
I wonder if this is another instance of #3127

greg-1-anderson:
I think that the problem here is that Drush is going through the wrong code path here. I don't think this function should be fixed like this, because we shouldn't be here unless the remote end was asked to produce backend json output.

weitzman:
no activity

PR #3261, opened 2017-12-19
Allow --uri to be specified on CLI with alias.

grasmash:
Using `--uri` does not work when executing commands against a drush alias in Drush 9.

`drush @blted82.local --uri=site2 status` executes against `site1`.

This PR allows `drush @blted82.local --uri=site2 status` to execute against site2.

### DIFF
--- a/src/Commands/sql/SqlSyncCommands.php
+++ b/src/Commands/sql/SqlSyncCommands.php
@@ -79,6 +79,12 @@ class SqlSyncCommands extends DrushCommands implements SiteAliasManagerAwareInte
         if (!$targetRecord = $manager->get($target)) {
             throw new \Exception(dt('Error: no alias record could be found for target !target', ['!target' => $target]));
         }
+        if (!$sourceRecord->uri() && $manager->getSelf()->uri()) {
+            $sourceRecord->setUri($manager->getSelf()->uri());
+        }
+        if (!$targetRecord->uri() && $manager->getSelf()->uri()) {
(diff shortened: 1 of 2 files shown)


weitzman:
This doesn't look right to me. For one thing, we have a number of "compound" commands that do similar things to sql-sync (i.e. config-pull). Not sure why sql-sync would be so special. Hopefully @greg-1-anderson can review this.

greg-1-anderson:
So, I think the issue here is that while `drush @foo sql-sync @self @remote` works, the desire is to have `drush sql-sync @foo @remote` also work if `uri` is correctly derived for `@self`, but `@foo` is missing its uri, and the uri for `@foo` should be the same as for `@self`.

My thoughts on this are:

1. This use case seems too complex to support.
2. Copying the uri from `@self` is not always going to be correct.

Rather than try to guess what the value of uri should be, I think we should rely on explicitly setting the uri in a way that associates it with the alias record (e.g. by defining it in the alias record). Also, we already support:

`drush sql-sync /path/to/local/drupal#uri @remote`

I'm not sure if, in this context, uri must be a simple folder name, or if `https://example.com` also works here. I suspect not, but did not check. We could consider supporting the later form, and could also consider supporting `@foo#https://example.com`, which would be a way to reference a different multisite of `@foo` without making a new alias.

I'm not sure that I'm sold on the complexity of supporting multisites that do not have their own alias defined, but if we do that, it should be limited in implementation to the site alias code; we don't want to special-case sql-sync.

PR #3265, opened 2017-12-20
WIP - Support global generators

richardbporter:
This got it working for me with a few constraints:

- Generators have to be placed in the Commands directory below a global command file path.
-- Ex. ../drush/contrib/foo/Commands/
- The file holding the generator has to end in *Generators.php.
-- Ex. ../drush/contrib/foo/Commands/FooGenerators.php

This isn't ideal but it does work. The class needs to be loaded before calling $discovery-getGenerators(). I tend to put commands and such in a src/ subdirectory - not sure if other developers do the same. The filename is also strange considering one generator exists in one file.

Maybe a follow up can be to improve the generator autoloading. I wasn't sure how to handle that. Either modifying CommandFileDiscovery or creating a new similar class (GeneratorFileDiscovery).


weitzman:
The code looks good. If we could add some docs and testing, that would be ideal.

richardbporter:
Sure. To test the discovery? I'm looking for examples but I can only find generate-related tests in https://github.com/drush-ops/drush/blob/master/tests/AnnotatedCommandTest.php.

weitzman:
Yes, to test the discovery.

richardbporter:
I think I've address the issues in my original post. I thought it made more sense to put these in a "Generators" directory.

weitzman:
Thanks. Code looks good. See the test failure.

weitzman:
I'm a little wary that we want to test the discovery but not the generation. Do we have any notion that global generators dont work?

richardbporter:
I wouldn't think so but I reverted that commit anyway.

weitzman:
Merged - many thanks.

PR #3272, opened 2017-12-23
WIP: Integrate composer/xdebug-handler

webflo:
[composer/xdebug-handler](https://github.com/composer/xdebug-handler) is a relatively new package. It has no stable release yet. The package provides the xdebug flag composer (COMPOSER_ALLOW_XEBUG) in a general fashion. It works seamlessly with symfony console.

This PR introduces the same feature for drush with DRUSH_ALLOW_XEBUG.

Some non-scientific benchmarks on macOS with PHP 7.1

 ## Before (current default)

```sh
$ export DRUSH_ALLOW_XDEBUG=1 && time ~/Projects/drush9/drush cr
~/Projects/drush9/drush cr  3,72s user 0,40s system 86% cpu 4,773 total
```

**4,773s**

## After (new default)

```sh
$ export DRUSH_ALLOW_XDEBUG=0 && time ~/Projects/drush9/drush cr
~/Projects/drush9/drush cr  1,12s user 0,35s system 70% cpu 2,091 total
```

**2,091s**

---

This is a big BC break. I am not sure if this improves the DX in general, because i often want to use xdebug in conjunction with php-eval or while debugging update hooks. 

Lets discuss :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
     "ext-dom": "*",
     "chi-teck/drupal-code-generator": "^1.21.0",
     "composer/semver": "^1.4",
+    "composer/xdebug-handler": "dev-master",
     "consolidation/annotated-command": "^2.8.1",
     "consolidation/config": "^1.0.9",
     "consolidation/output-formatters": "^3.1.12",
--- a/drush.php
+++ b/drush.php
@@ -1,5 +1,6 @@
 <?php
 
+use Composer\XdebugHandler\XdebugHandler;
 use Drush\Config\Environment;
 use Drush\Preflight\Preflight;
 use Drush\Runtime\Runtime;
@@ -52,6 +53,10 @@ if (file_exists($autoloadFile = __DIR__ . '/vendor/autoload.php')
 } else {
     throw new \Exception("Could not locate autoload.php. cwd is $cwd; __DIR__ is " . __DIR__);
 }
+
+$xdebug = new XdebugHandler('drush', '--ansi');
+$xdebug->check();
+unset($xdebug);
 
 // Set up environment
 $environment = new Environment(Path::getHomeDirectory(), $cwd, $autoloadFile);


greg-1-anderson:
Looks pretty interesting. Why is this a BC break?

weitzman:
I'm interested too. We could start by adding it to the launcher?

webflo:
BC break because it functions different than before. I think people use drush often do debug something in Drupal? But of course this it not and API change. I think some new docs about drush and xdebug should be sufficient.

webflo:
@weitzman This is the PR for the launcher. https://github.com/drush-ops/drush-launcher/pull/41

weitzman:
Just tested this - got similarly impressive results. I think we need a stable release before we add this. Or else Composer users have minimum-stability hassles. here are some docs:


> Drush, like Composer automatically disables Xdebug by default. This improves performance substantially. You may override this feature by setting an environment variable - `DRUSH_ALLOW_XDEBUG=1 drush [command]`. See https://github.com/drush-ops/drush/pull/3272.

weitzman:
Still no releases we can build on :(

webflo:
New release for drush-launcher: https://github.com/drush-ops/drush-launcher/releases/tag/0.6.0

weitzman:
Unlike the Launcher, I think Drush can pass its Logger using https://github.com/composer/xdebug-handler/blob/master/README.md#output. The PreflightLog is not a real Logger so I suspect we have to do the xdebug check after this line https://github.com/drush-ops/drush/blob/a7814d2afcf13236455f509679c74f77fc933700/src/Runtime/Runtime.php#L78. 

Also worth setting DRUSH_ as the env variable prefix. Launcher could do that as well. Scroll up on the README to learn more.

weitzman:
Lets stop here for now. This package is integrated into drush-launcher but not drush itself.

Issue #3009, opened 2017-10-04
Document setting Drush config values via environment variable


weitzman:
The logic for the default value for this config is in drush_find_tmp(). Not changing any logic here, using config system.

greg-1-anderson:
The standard is for a `.env` file with environment variables. Maybe we should support that in consolidation/config?

weitzman:
I have used .env files on a couple project and find them very convenient. Sounds great. 

I think this change in Drush could happen in parallel with a consolidation/config change, right?

greg-1-anderson:
For new designs it is often preferable to implement in Drush first, and then refactor / extract to external components. In cases like this where the new behavior is already incremental to functionality that has already been extracted, I'd say that it is best & just as easy to implement directly in consolidation/config first.

greg-1-anderson:
We need a way to map from a config element to an environment variable. Maybe convert `.` to `_` and then uppercase and prepend `DRUSH_`. That's a start. With that rule, `drush.alias-path` would become `DRUSH_DRUSH_ALIAS-PATH`.

The two obvious flaws with this proposal:

- DRUSH_DRUSH is sub-optimal, but not all config items start with `drush.`, and all environment variables should.
- Having a `-` instead of a `_` is inconsistent; however, if we converted it to `DRUSH_DRUSH_ALIAS_PATH`, I'm not sure how we would know whether to convert that back to `drush.alias-path` instead of `drush.alias.path`.

Terminus [searches both $_SERVER and $_ENV for matches](https://github.com/pantheon-systems/terminus/blob/master/src/Config/EnvConfig.php#L21), and [overrides `set` to convert the key to a constant](https://github.com/pantheon-systems/terminus/blame/master/src/Config/TerminusConfig.php#L150) if applicable. Config is read via [DotEnvConfig](https://github.com/pantheon-systems/terminus/blob/master/src/Config/DotEnvConfig.php). There is a [yml file listing all valid key constants](https://github.com/pantheon-systems/terminus/blob/master/config/constants.yml).

Terminus has neither hierarchy nor `-`s in their config symbols; they would just use `alias_path` instead of `drush.alias-path`. Not sure I'd be happy to do likewise.

Maybe rather than reading all of the env variables and applying them as a config, we could instead change the implementation of default handling to check the environment variable before returning the default value. Then the lossy transformation to `DRUSH_DRUSH_ALIAS_PATH` would be okay, as we would never need to go the other way.

weitzman:
[...]

PR #3278, opened 2017-12-29
Fixes #3277: Site-specific drush.yml file not loaded.

### LINKED ISSUE #3277: Site-specific drush.yml file not loaded
Drush does not appear to load `docroot/sites/default/drush.yml`, although the documentation indicates that this should work.

Adding the following to `drush/drush.yml` has no effect:
```
drush:
  paths:
    config:
      - ../docroot/sites/default/drush.yml
```

Executing `drush status --config=docroot/sites/default/drush.yml` also does not load it.

grasmash:


greg-1-anderson:
`uri` is often `https://example.com` rather than a directory name. Not sure how many existing places in Drush already make the assumption that `uri` is the directory name, but some modules need `uri` to be set to an fqdn in order to work correctly. Drush allows Drupal to do the conversion from fqdn to directory name during bootstrapping, but we may need to do some level of transformation ourselves prior to that point to support features such as this.

I don't fully remember what Drush 8 does, but I think that sitealias.inc had some provision for this.

grasmash:
https://github.com/drush-ops/drush/blob/8.x/includes/sitealias.inc#L1593

greg-1-anderson:
Yeah, that's pretty minimal.

src/Config/ConfigLocator.php
```diff
@@ -256,29 +258,116 @@ public function addDrushConfig($drushProjectDir)
...
+        }
+
+        // There might not be a site directory.
+        $site_dir = "$drupalRoot/sites/$uri";
+        if (!is_dir($site_dir)) {
+            return;
+        }
+
+        // We might have already processed this site.
+        if (in_array($uri, $this->siteUris)) {
```
greg-1-anderson: Maybe we should cache the directory rather than the uri?

src/Config/ConfigLocator.php
```diff
@@ -256,29 +258,116 @@ public function addDrushConfig($drushProjectDir)
...
+            $hostname = $this->convertUriToHostname($uri);
+
+            // If $hostname matches a sites.php mappings, use dir from mapping.
+            if ($dir_name = $this->lookupSiteDirFromHostname($hostname, $drupalRoot)) {
+                $uri = $dir_name;
+            }
+        }
+
+        // There might not be a site directory.
+        $site_dir = "$drupalRoot/sites/$uri";
```
greg-1-anderson: Maybe site dir lookup should be a method of the alias record class (bring along convertUriToHostname, lookupSiteDirFromHostname).

greg-1-anderson: n.b. I'm not sure about this - considering it. Maybe factoring out into a utility class would be better. I don't think that I want the alias record code to know about Drupal site directory conventions. Ideally the site alias classes will be factored out into a separate consolidation project.

src/Preflight/Preflight.php
```diff
@@ -276,7 +276,9 @@ public function preflight($argv)
 
         // If we did not redispatch, then add the site-wide config for the
         // new root (if the root did in fact change) and continue.
-        $this->configLocator->addSitewideConfig($root);
+        $this->configLocator->addDrupalConfig($root);
+        // Set multisite config using uri from alias.
+        $this->configLocator->addSiteConfig($root, $selfAliasRecord);
```
greg-1-anderson: Is it necessary to pass in `$root` here? i.e. will $root ever be different than the value held by the self alias record? If it is, wouldn't the value in the self alias be more correct?

src/Config/ConfigLocator.php
```diff
@@ -256,29 +259,92 @@ public function addDrushConfig($drushProjectDir)
...
+    /**
+     * Lookup a site's directory via the sites.php file given a hostname.
+     *
+     * @param $hostname
+     *   The hostname of a site. May be converted from URI.
+     *
+     * @return $drupalRoot
+     *   The directory associated with that hostname.
+     */
+    public function lookupSiteDirFromHostname($hostname, $drupalRoot) {
```
grasmash: Not sure where this belongs. I can move to StringUtils, though it's not strictly doing string manipulation.

src/Commands/core/SiteInstallCommands.php
```diff
@@ -276,7 +276,7 @@ public function pre(CommandData $commandData)
             // a fallback name when '--sites-subdir' is not specified, but
             // only if the uri and the folder name match, and only if
             // the sites directory has already been created.
-            $dir = $this->getSitesSubdirFromUri($root, $aliasRecord->get('uri'));
+            $dir = StringUtils::lookupSiteDirFromUri($aliasRecord->get('uri'), $root);
```
grasmash: Could we fall back to 'default' if the URI does not have a matching entry in sites.php?

I have a single site in `sites/default` and I have `drush.yml` file that sets 'options.uri: http://example.com`, I expect `drush si` to install the site even if I don't have a `sites.php` file.

weitzman:
Drupal 8 requires fqdn for lookup. We have moved drush toward assuming that
also. I think it is unwise to accept anything else.

grasmash:
Yes, I agree. My previous comment referred to setting the site directory value to default, not the URI. 

Here is what I am proposing. If someone passes a fully qualified domain name as the URI value and that value does not have a corresponding directory mapping in sites.php, drush will assume that the site directory that it refers to is “default”.

grasmash:
Tbh, this detail is incidental. I’m fine with leaving that as is. 

Any issues with the PR as it is?

src/Preflight/Preflight.php
```diff
@@ -249,7 +249,7 @@ public function preflight($argv)
         // Extend configuration and alias files to include files in
         // target site.
         $root = $this->findSelectedSite();
-        $this->configLocator->addSitewideConfig($root);
+        $this->configLocator->addDrupalConfig($root);
```
weitzman: In Drush8, we don't load any site specific config until the Site phase of the bootstrap. I'd like to keep consistent with that in Drush9. Since this change is in Preflight (i.e. happens before bootstrap), I dont think its properly placed.

greg-1-anderson: I think that moving all config processing to preflight is an improvement. When config was php, delaying loading site-specific config until site bootstrap has the benefit of allowing folks to use Drupal APIs in their config code. With yaml config, I'm not sure that there is a benefit to postponing config loading. Loading config in preflight might allow folks to use `--uri=foo` and set URI to `https://foo.example.com` in their drush.yml file, to give one example.

grasmash: That is exactly my use case.

weitzman: I think we are working around the problem that config is static after its loaded. It would be preferable IMO if uri were discovered during SITE phase and then used for link building and such during command execution.

We already do ROOT discovery during Preflight. This PR effectively moves SITE discovery there too. It begs the question of how much do we need the progressive Bootstrap that we do. Maybe NONE and FULL is all we need? We don't need to figure this out now but it would be good to get some clarity on future directions.

src/Config/ConfigLocator.php
```diff
@@ -101,7 +104,7 @@ public function __construct($envPrefix = '')
         }
         $this->config->addPlaceholder(self::USER_CONTEXT);
         $this->config->addPlaceholder(self::DRUPAL_CONTEXT);
-        $this->config->addPlaceholder(self::SITE_CONTEXT); // not implemented yet (multisite)
```
greg-1-anderson: n.b. "not implemented yet (mulitsite)" comment.

src/Preflight/Preflight.php
greg-1-anderson: I think that site selection is already happening in preflight. Note the comment removed in this PR: "// not implemented yet (multisite)" (see my comment https://github.com/drush-ops/drush/pull/3278#pullrequestreview-86221163). I think this PR is essentially implementing a TODO. I still don't see any practical benefit to postponing uri discovery to the SITE phase, and as previously mentioned, there are some disadvantages to doing that.

Regarding progressive bootstrap, for command discovery we only need NONE (built-in and global commands) and FULL (module commands).  Built-in and global commands can also declare the bootstrap level they want, but it is not possible for them to declare their bootstrap level, because we already have to bootstrap to FULL in order to find them.

src/Config/ConfigLocator.php
grasmash: This was removed intentionally, as this PR adds site support.

greg-1-anderson: Yeah I made that note for @weitzman to point out that this PR is implementing a TODO. Then I forgot to save the comment that referenced it.

greg-1-anderson: Oh, it did get saved; see above.

src/Preflight/Preflight.php
weitzman: I could be wrong, but I think this moves the inclusion of sites.php to Preflight from SITE phase. That might be safe, but I'm not sure what people are doing there.

grasmash: Actually, I think that prior to this PR, `sites.php` was never included by Drush9.

Yes it is now included during preflight if a uri is specified. Given that `sites.php` should only ever contain a static PHP array, this does not seem problematic. 

From a conceptual standpoint, `sites.php` really contains data that is above the "site" context. It contains data about _all_ sites belonging to the application, which is used to determine the appropriate site context.

weitzman: I agree this is pretty safe and and agree with your conceptual point.

FYI, Drush is relying on Drupal to include sites.php. This happens during the SITE phase I have been mentioning. See https://github.com/drush-ops/drush/blob/3fc8e7156ea08081eadedc4f61fe83315fa1f883/src/Boot/DrupalBoot8.php#L73-L83. I dont love duplicating that logic in Drush, but I may tolerate it.

greg-1-anderson: Ah, okay, I see now.

We really must include drush.yml files from the sites directory during preflight. In order to do this, we need to identify the sites directory, and in order to identify the sites directory (the same way Drupal does) we need to include sites.php.

I'm not fond of putting `lookupSiteDirFromUri` in StringUtil. I would also prefer to keep Drupal-specific code out of preflight. I'd also like to continue to rely on `DrupalKernel::findSitePath` to process sites.php.

Some of these desires are mutually exclusive. I'm not sure what the best compromise is. To whit:

1. If we do not include drush.yml in the preflight (or at least at the beginning of the bootstrap site phase), then we limit our ability to use `--uri=dir` and set a better uri in drush.yml.
1. If we do not include sites.php at all, then we limit our ability to use `--uri=https://example.com` and still find the right sites directory, and finding the right sites directory is necessary to include drush.yml soon enough to improve the uri.

I think it would be worthwhile to try harder to use `DrupalKernel::findSitePath` to find the site path and load drush.yml. Can this happen at the beginning of the site bootstrap phase, before we've done the site bootstrap, so that we can then set the uri and continue with the site bootstrap? Or do we need the uri to make the request object needed to call `DrupalKernel::findSitePath`?  Maybe the load order of drush.yml during the site bootstrap phase needs to vary depending on whether uri is a directory name or contains a protocol? Having a conditional order of events in bootstrap isn't great either, but I'd strongly prefer to keep the Drupal-specific steps in bootstrap, if that's possible. Breaking that rule makes the Drush preflight / bootstrap harder to understand.

weitzman: The code I linked to already runs during SITE VALIDATE. We could certainly add a call to \Drush\Config\ConfigLocator::addConfigPaths during SITE phase. I'm not sure how to get access to the ConfigLocator instance though. Also, this would not satisfy the desire to set uri from a site specific config file. I honestly doubt that we can/should support that right now.

greg-1-anderson: By the current design, the config locator is not available during bootstrap.

Does it not currently work to set `$options['uri']` from a site-specific drushrc.php file in Drush 8 with Drupal 8?

weitzman: I dont know. If so, we just accept that as a regression with known workarounds (use an alias, use site-set, specify uri on the request, ...).

greg-1-anderson: Yeah I agree. While I'd like to see this feature work, I think it's enough of an edge case that I'd like the solution to be clean (keep Drupal-specific code in the bootstrap phase).

weitzman: I have found one more workaround. If you set the environment variable `DRUSH_OPTIONS_URI=http://example.com` then your custom uri will be used. No need to fuss with Drush config or aliases.

So, how to set that env variable? If you have many sites on same host, or don't control the server, this can be challenging. 
1. Anyone used [direnv](https://github.com/direnv/direnv)?
1. I'm using PHPEnv successfully on Drupal projects but at this time there is no way to load custom code early enough in a Drush request for 'uri' to take effect. That happens at end of Runtime https://github.com/drush-ops/drush/blob/master/src/Runtime/Runtime.php#L88-L93

I will now writeup docs for #3009 as I now understand that system.

weitzman: As a follow-up, the drupal-project will soon support .env files and documents how to DRUSH_OPTIONS_URI so that site specific works. See https://github.com/drupal-composer/drupal-project/pull/351. You can implement that in your composer project even before this PR is merged.

weitzman:
So, the recommended way to get site specific config files to load is set an environment variable See https://github.com/drush-ops/drush/pull/3278#discussion_r173858140 and follow its links. This is now documented at https://github.com/drush-ops/drush/blob/master/examples/example.drush.yml#L26-L35

PR #3283, opened 2018-01-02
Allowing files by any name to be imported via -c option.

grasmash:


src/Config/ConfigLocator.php
```diff
@@ -478,4 +485,29 @@ public function setComposerRoot($selectedComposerRoot)
...
+     *   An array. The first row is an array of files, the second row is an
+     *   array of dirs.
+     */
+    protected function separateFilesAndDirs($paths) {
+        $files = [];
+        $dirs = [];
+        foreach ($paths as $path) {
+            if (file_exists($path)) {
+                if (is_dir($path)) {
+                    $dirs[] = realpath($path);
```
greg-1-anderson: If you pass in `$candidates` to this function then you could call `$this->identifyCandidates($dirs, $candidates);` here (or at the end of this function). This would allow you to keep `$dirs` private to this method and return only `$files`.

greg-1-anderson: This improvement would of course imply a renaming of this method.

greg-1-anderson:
LTGM - thanks for doing the extra complexity-reducing refactors.

PR #3290, opened 2018-01-03
Fixes #3289: Bash helpers don't work on Mac.

### LINKED ISSUE #3289: Bash helpers don't work on MacOS
I'm using Drush 8.1.15 on MacOS High Sierra. Well, technically I'm using Drush Launcher but I've run `drush init` from within a Composer project using Drush 8.1.15. I've also attempted this with 9.0.0-rc1 and the behavior is the same.

Drush init tries to install a bunch of helpers (Bash prompt and completion customizations), but these are broken on a Mac, for at least two reasons.

First, it appends file includes to `.bashrc`, which is fine for Linux environments, but doesn't work on Mac. Mac terminals are login shells by default, so they don't source `.bashrc`. These includes should instead be added to `.bash_profile`. A good example of the right switching logic to use here might be BLT: https://github.com/acquia/blt/blob/9.1.x/src/Robo/Commands/Blt/AliasCommand.php

Second, even if you add the correct includes to `.bash_profile`, the custom Git/Drush prompt doesn't work. The problem is that the `__git_ps1` function doesn't exist out of the box with recent versions of Git on Mac. Not sure if that's a recent change, or different from Linux, or what. If you manually download and source git-prompt.sh ([as described here](https://stackoverflow.com/questions/12870928/mac-bash-git-ps1-command-not-found)) it works fine. Seems like we should document that if it's the recommended workflow.

danepowell:
Fixes #3289

### DIFF
--- a/examples/example.prompt.sh
+++ b/examples/example.prompt.sh
@@ -2,7 +2,13 @@
 #
 # Example PS1 prompt.
 #
-# Use `drush init` to copy this to ~/.drush/drush.prompt.sh, and source it in ~/.bashrc
+# Use `drush init` to copy this to ~/.drush/drush.prompt.sh, and source it in
+# ~/.bashrc or ~/.bash_profile.
+#
+# Note that your Bash session must already have the __git_ps1 function available.
+# Typically this is provided by git-prompt.sh, see instructions for downloading
+# and including this file here:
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
LTGM thanks

Issue #3226, opened 2017-12-09
count(): Parameter must be an array or an object that implements Countable


Chi-teck:
This only happens on PHP 7.2
```
$ php drush.php st
count(): Parameter must be an array or an object that implements Countable   [warning]
Table.php:789
count(): Parameter must be an array or an object that implements Countable   [warning]
Table.php:789
count(): Parameter must be an array or an object that implements Countable   [warning]
Table.php:789
count(): Parameter must be an array or an object that implements Countable   [warning]
Table.php:789
count(): Parameter must be an array or an object that implements Countable   [warning]
Table.php:789
count(): Parameter must be an array or an object that implements Countable   [warning]
Table.php:789
count(): Parameter must be an array or an object that implements Countable   [warning]
Table.php:789
 PHP configuration      :  /etc/php/7.2/cli/php.ini 
 PHP OS                 :  Linux                    
 Drush script           :  /var/www/drush/drush.php 
 Drush version          :  8.1.16-dev               
 Drush temp directory   :  /tmp                     
 Drush configuration    :                           
 Drush alias files      :                             
```

https://github.com/pear/Console_Table/pull/14

anavarre:
Confirmed with Drush 9 RC1.

Chi-teck:
@anavarre Are you sure? Drush 9 has no dependency on `pear/console_table`.

anavarre:
BTW, is it on Drush to fix the issue? It seems to be it could be a core bug https://www.drupal.org/project/drupal/issues/2928846

Chi-teck:
It is a general PHP 7.2 issue. _count()_ function should be used to count uncountable values anymore.
http://php.net/manual/en/migration72.incompatible.php#migration72.incompatible.warn-on-non-countable-types.

This particular warning comes comes from Pear Console Table package. I referenced related issue in the ticket description. I am not sure what can be done here. That package does not seem actively maintained. Given that is just a single class, we could simply copy it Drush codebase.

Ayesh:
A backtrace says the call is in `includes/output.inc:303`, and it doesn't look like we can fix/workaround this from Drush's side. 

```php
function _updateRowsCols($rowdata = null) {
   // ...
  $this->_max_cols = max($this->_max_cols, count($rowdata));
  // ...
}
```

Console_Table class is doing it wrong because it relies on the `count()` to function as an `!empty()` as well. 
[...]

PR #3292, opened 2018-01-05
Fix count() calls on uncountable variables in \Console_Table class

Ayesh:
This is for #3226. 
In PHP 7.2, there is a notice raised because the PEAR `console_table` makes a `\count()` call on a variable that is not an `array`/`Countable` . Because maintaining a fixed version of this PEAR package is not worth the effort at this stage (Drush 9.x no longer uses this package), we avoid the notice by silencing PHP with the `@silent` operator.


Ayesh:
Some travis builds are failing because connectivity issues to getcomposer.org and Chrome download locations.

Issue #3308, opened 2018-01-15
Error occurring in failing update is not reported in the log


pfrenssen:
This has been split off from #3303.

@claudiu-cristea reported in #3303 that if an error occurs during a database update this is not logged correctly. It appears this is only working correctly for exceptions being thrown in `hook_update_N()`, but not when a post-update fails.

We should fix this and extend the test coverage for post-updates.

claudiu-cristea:
@pfrenssen,  I guess the problem is that messages logged in subsequent processes are not shown (have no explanation why for the moment). As `hook_update_N` are running first, they manage to run in the main process, while post updates are spawned in subsequent processes. Just a guess.

pfrenssen:
I did some tests on the branch which is exhibiting the problems (https://github.com/ec-europa/joinup-dev/pull/1045):

* The error which is being thrown is `Exception: The theme implementations may not be rendered until all modules are loaded. in Drupal\Core\Theme\ThemeManager->render() (line 139 of web/core/lib/Drupal/Core/Theme/ThemeManager.php)`. This is strange, why aren't the modules loaded?
* The error only occurs when running `hook_update_N()` updates together with a post-update. If the `hook_update_N()` updates are run first, and then the post-update separately the exception is not thrown and the update finishes successfully.

pfrenssen:
I have fixed the missing log output in #3310.

I have also provided a test case for the `The theme implementations may not be rendered until all modules are loaded` exception in #3311, but I didn't have the time yet to investigate this.

pfrenssen:
I found the cause why the modules are not being loaded, this has been introduced in https://github.com/drush-ops/drush/commit/978893e513ebaea30cf13e46268628cfb0840609

In this commit the container is being rebuilt after calling `drupal_flush_all_caches()`, but look what this does:

```
public function rebuildContainer() {
  // Empty module properties and for them to be reloaded from scratch.
  $this->moduleList = NULL;
  [...]
}
```

pfrenssen:
OK this took me a long time to find since I was completely on the wrong track. I thought it was caused by the kernel juggling that was introduced recently in the UpdateKernel PR, but it turns out it was this container rebuild. I have a green test now:

```
$ UNISH_NO_TIMEOUTS=1 UNISH_TMP=/home/pieter/v/drush-sut php unish.phpunit.php tests/UpdateDBTest.php --stop-on-failure -vvv
[...]

PR #3303, opened 2018-01-12
Improve the log of the updatedb command

pfrenssen:
This partially fixes #3296.

The output of the `updatedb` command can be quite confusing, since misleading messages are being output, and incorrect statuses are being communicated.

Here is some example output from the current update test suite:

```
 [notice] Executing woot_update_8101
 [ok] Performing woot_update_8101
 [notice] Executing woot_update_8102
 [error]  8102 error 
 [ok] Performing woot_update_8102
 [notice] woot_update_8101
 [error]  Failed batch process: woot_update_8102 
 [error]  Finished performing updates.
```

* There appears to be duplicated information about "executing" and "performing" each individual update. In fact when the log mentions that it is "Performing hook_update_x" the update has already finished.
* Update 8102 failed, so there should not be any `[ok]` associated with it.
* It shouldn't mention that the "batch process" failed. That the batch API is used behind the scenes is an implementation detail and not important to mention in the log output.

Edit: Note that the scope of this PR is solely to change the existing messages so they are less confusing. The order of the messages is being corrected in #3302, missing errors are in #3308.

PR #3302, opened 2018-01-12
Update messages are logged in the wrong order

pfrenssen:
Partially fixes #3296.

When performing multiple database updates in a row, the individual messages that are returned by the update hooks are output in the log after all the updates have finished. This might give the impression that the updates have been performed in parallel or out of order, and this might be misleading for system administrators that are investigating a failed update.

Currently the output might look like this:
```
 [notice] Executing woot_update_8101
 [ok] Performing woot_update_8101
 [notice] Executing woot_update_8102
 [error]  8102 error 
 [ok] Performing woot_update_8102
 [notice] woot_update_8101 <----- out of order
 [error]  Failed batch process: woot_update_8102 
 [error]  Finished performing updates. 
```

PR #3310, opened 2018-01-16
Post-update logging

pfrenssen:
Fix for #3308 

This is postponed on #3302 and #3303.


pfrenssen:
The tests show that practically no useful output is being logged during the execution of the post-updates.

This can be fixed by moving the post-update code into a batch operation inside Drush, similar to how it was done for `hook_update_N()` updates in `UpdateDBCommands::updateDoOne()`.

claudiu-cristea:
Discussed with @pfrenssen and I will review this PR tomorrow morning so it can be quickly ready for the 9 stable release.

src/Commands/core/UpdateDBCommands.php
```diff
@@ -210,17 +210,98 @@ public function updateDoOne($module, $number, $dependency_map, &$context)
...
+
+    /**
+     * Batch command that executes a single post-update.
+     *
+     * @param string $function
+     *   The post-update function to execute.
+     * @param array $context
+     *   The batch context.
+     */
+    public function updateDoOnePostUpdate($function, &$context)
```
claudiu-cristea: It seems that post updates have their own way of logging. Sad that we can not reuse the Drupal code.

claudiu-cristea:
Thank you. The exception message is now logged. Review done. @weitzman, this can be committed.

PR #3311, opened 2018-01-16
All modules should be loaded when performing post-updates

### LINKED ISSUE #3308: Error occurring in failing update is not reported in the log

pfrenssen:
Fixes #3308.

Postponed on #3310.


pfrenssen:
The test case replicates the error that was reported by @claudiu-cristea in #3308:

```
 [notice] Update started: woot_post_update_render
 [error]  The theme implementations may not be rendered until all modules are loaded. 
 [error]  Update failed: woot_post_update_render 
```

pfrenssen:
OK this is ready for a look. Note that it is blocked on #3310.

claudiu-cristea:
I tested this on complex project with a lot of updates and post-updates and is fixing the bug. We discovered this bug on the same project with the same list of updates/post-updates. I see there's also a test that proves the bug and the fix. @weitzman, @greg-1-anderson,  this could be committed  but not before #3310.

PR #3314, opened 2018-01-18
 Add private copy of ArgvInput to fix Console bug

weitzman:
A private copy of ArgvInput (v3.4.2) that has commented out the poor section of hasParameterOption(). Is it OK for other versions of Sumfony to get this file, including Symfony2 users? If not, how to filter to just right installations?

ArgvInput has private members so seems like copying is the only way (thats what LessStrictArgvInput does as well).

### DIFF
--- a/src/Preflight/PreflightArgs.php
+++ b/src/Preflight/PreflightArgs.php
@@ -4,8 +4,8 @@ namespace Drush\Preflight;
 use Consolidation\Config\Config;
 use Consolidation\Config\ConfigInterface;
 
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
I haven't tested or validated the specific changes made in the copied ArgvInput here, but I am :+1: on the technique and strategy here.

Once Symfony Console is fixed, we should just change our composer.json to reject buggy releases, either with a `conflict` that lists all of the bad versions (there are lots), or just by using a series of `^4.1.x|^3.4.x|...` (again, there are lots) to only allow good versions.

PR #3326, opened 2018-01-26
Fix argument remapping. Limit remapping to first argument.

### LINKED ISSUE #3295: Drush 9; Argument 'en' is always replaced with 'pm:enable'
If a Drush 9 commands allows arguments, the argument 'en' will be replaced with its full command name pm:enable. In this way it isn't possible to create an argument that accepts the language code 'en'.

Drush 8 doesn't have this issue.

Some test cases:
```Shell
$ drush @site_alias en 'en'
[error]  Unable to install modules pm:enable due to missing modules pm:enable.

$ drush @site_alias pm:enable 'en'
[error]  Unable to install modules pm:enable due to missing modules pm:enable.
```

Sutharsan:
Fixes #3295 by limiting the remapping to the first argument (properly).

### DIFF
--- a/tests/ArgsRemapperTest.php
+++ b/tests/ArgsRemapperTest.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Unish;
+
+use \Drush\Preflight\argsRemapper;
+
+/**
+ * Tests the Arguments Remapper.
+ *
+ * @group base
+ */
+class ArgsRemapperTest extends CommandUnishTestCase
+{
+
+    /**
+     * @covers argsRemapper::ArgsRemapper
+     * @dataProvider argsProvider
+     */
+    public function testCommandAliases($argv, $expected)
+    {
+        $remapOptions = [];
+        $remapCommandAliases = [
+            'en' => 'pm:enable'
+        ];
+        $sut = new ArgsRemapper($remapOptions, $remapCommandAliases);
+        $result = $sut->remap($argv);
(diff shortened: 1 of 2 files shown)


weitzman:
Thanks ... The old code had consistently wrong spelling or no? if yes, then I dont see how a spelling fix can eliminate the bug. I cant find the actual spelling inconsistency except for in Doxygen

greg-1-anderson:
If we're going to call this fixed, it would be helpful to have a failing test that passes when the fix is applied.

Sutharsan:
@weitzman not consistenly wrong, no. See line 50 (old code) that uses correct spelling ` $sawCommand = true;`. Therefore `$sawCommmand` (tripple m) does not get flagged.

Sutharsan:
Minimal unit test added.

tests/ArgsRemapperTest.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Unish;
+
+use \Drush\Preflight\argsRemapper;
```
weitzman: Capital A in class name, not 'a'.

Sutharsan:
Done

tests/ArgsRemapperTest.php
Sutharsan: Done

Sutharsan:
:) Thanks

PR #3327, opened 2018-01-26
Stop all processing of queue on SuspendQueueException

kasperg:
If a SuspendQueueException is thrown when processing an item the current
implementation will release the item and log an error. Processing of the
queue continues.

According to the documentation of the SuspendQueueException throwing 
this exception indicates that processing of the whole queue should be skipped.

This change breaks the queue item processing loop entirely after logging
the error.

The associated code comment is also updated as it does not make sense.
The queue run command only runs for a single queue so there is nothing
to skip to.

### DIFF
--- a/lib/Drush/Queue/Queue8.php
+++ b/lib/Drush/Queue/Queue8.php
@@ -65,9 +65,10 @@ class Queue8 extends QueueBase {
       }
       catch (SuspendQueueException $e) {
         // If the worker indicates there is a problem with the whole queue,
-        // release the item and skip to the next queue.
+        // release the item and stop further processing.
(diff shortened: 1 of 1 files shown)


weitzman:
I'm hoping someone can review this. I know that @damiankloip and @juampynr have done queue work for Drush in the past.

juampynr:
Yep! I will review it as soon as I get a chance.

damiankloip:
This change looks good to me. I just wonder if we should add a quick test case for this in `QueueTest` to cover this? I know we already cover something similar, re-queuing.

weitzman:
merged. a test would still be welcome.

this is already fixed in master branch.

Issue #3321, opened 2018-01-23
Drush sql-dump not working with 'extra' arguments with Postgres


kalpaitch:
To replicate: `sql-dump --skip-tables-list=cache_* --extra='-O -x'` with postgress.

The extra arguments only work with pg_dump utility command (we want to exclude dumping with ownership).

However, drush will also pass these extra arguments to psql when it does the select query to get a list of all the tables available to be dumped. Because these aren't options for psql it will return a failure and therefore just dump _all_ the tables, including the ones we want to skip.

If we're passing extra arguments to `sql-dump` it doesn't seem right to also pass them to `psql` command  that drush also runs, or at least if it does there should be some way to validate that these are actually valid arguments.

kalpaitch:
One of the main issues is in the main query constructor which states in the comments that drush_get_options() should be avoided, but then proceeds to use it in the method - https://github.com/drush-ops/drush/blob/8.x/lib/Drush/Sql/SqlBase.php#L167

kalpaitch:
https://github.com/drush-ops/drush/pull/3328

kalpaitch:
Related to - https://github.com/drush-ops/drush/issues/2079

kalpaitch:
"So we just need two more changes - line 437 of sql.drush.inc and lines 98-99 of Sqlpgsql.php" that would allow us to pass in the extras from the sql-dump command _only_ and override the query_extras just for that command.

weitzman:
See recent pg fixes and post here if still not working.

kalpaitch:
I can confirm this is still an issue with Drush 10.6.2.

kalpaitch:
This is the generated query, as you can see the `-O` should not be included in this query in `SqlPgsql::listTables()`:

```
psql -q --dbname=xxx --host=xxx --port=xxx --username=xxx  -O -x -w -c 'SELECT tablename FROM pg_tables WHERE schemaname="public";'
```

gitressa:
[Drush 10.6.2](https://github.com/drush-ops/drush/releases/tag/10.6.2) is from 15 Dec 2021, so perhaps you can try the latest version 11.1.1, and check if it works in that version?

kalpaitch:
> [Drush 10.6.2](https://github.com/drush-ops/drush/releases/tag/10.6.2) is from 15 Dec 2021, so perhaps you can try the latest version 11.1.1, and check if it works in that version?

Still same query generated. Same result.

[...]

PR #3328, opened 2018-01-26
#3321: Update SqlBase.php

kalpaitch:


kalpaitch:
```
$ ls -lah test*
-rw-rw-r-- 1 andrew andrew 78M Jan 26 13:21 test-dump.sql
$ cat test-dump.sql | grep 'OWNER'

```

weitzman:
Old, and not clear why this is needed. Note that sql:dump was just fixed in #2282

PR #3333, opened 2018-01-28
Add PHP 7.1 and PHP 7.2 to travis configuration

Chi-teck:


weitzman:
Lets start running the isolation tests on 7.2. The full suite is too long in duration as it is.

greg-1-anderson:
Let's continue at #3337

PR #3335, opened 2018-01-29
Make SiteInstallCommands::getSitesSubdirFromUri find dir from sites.php (#3334)

### LINKED ISSUE #3334: Make SiteInstallCommands::getSitesSubdirFromUri find dir from sites.php
The old drush_sitealias_local_site_path() function used to find the dir with the include of sites/sites.php
This is no more the case in SiteInstallCommands::getSitesSubdirFromUri, so I propose to fix that.
PR will follow

Sylry:
Fix #3334 issue


weitzman:
This seems non-intrusive and useful. I'll merge this soon unless someone objects.

PR #3337, opened 2018-01-30
Try testing PHP 7.2 on the isolation tests.

greg-1-anderson:
Replaces #3333

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ php:
 #  See master-fulltest branch for broader PHP version testing.
   - 5.6
   - 7.0
+  - 7.2
 
 # Cache Composer.
 cache:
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
PHP 7.1 is pretty similar to 7.0, so I did not add it here.

greg-1-anderson:
Maybe we should also skip composer-highest on php 5.6?

I kind of hate all of the `excludes`, but I don't think that naming all of the permutations to run is much of an improvement either.

greg-1-anderson:
Or, we could run all of the functional tests on one version of php only, and pick a third testing service (codeship?) to run them on. 😝

greg-1-anderson:
I didn't look for warnings in the test output, but the tests are all green, at least.

Codeship only provides 100 builds per month on the free plan. I looked at Shippable briefly, but couldn't figure out how to start a build on an org (drush-ops). It's authorized for drush-ops, but it doesn't show up in subscriptions.

Maybe what we have is okay.

weitzman:
PR looks good to me.

PR #3341, opened 2018-02-02
sql-sanitize to generate random user password (fixes #3086)

### LINKED ISSUE #3086: Drush 9: passwords should be scrambled by default when running `sql-sanitize`
(Reposted from Drupal slack #drush:)

Minor drush sql-sanitize suggestion:
https://github.com/drush-ops/drush/blob/master/src/Drupal/Commands/sql/SanitizeUserTableCommands.php#L89

The default for password is the literal string "password". This means that, under default options for `sql-sanitize`, given a sanitized database, anyone with access can log in as any user. This feels like a security risk, as in theory users with access to lower lifecycle environments could log in as admin and gain knowledge of a production system's settings; or, could log in as other users, and gain knowledge of their correspondance, their own user settings, etc (which should maybe also be sanitized).

Would it make sense to create a randomized password by default? Then not only would the users' passwords be scrubbed, but their accounts would become inaccessible (at least without direct DB access, at which point all bets are off.)

I'll provide a PR. Putting this here for documentation's sake.

CashWilliams:
Use StringUtils::generatePassword() to generate a random password when not provided.

### DIFF
--- a/src/Drupal/Commands/sql/SanitizeUserTableCommands.php
+++ b/src/Drupal/Commands/sql/SanitizeUserTableCommands.php
@@ -5,6 +5,7 @@ use Consolidation\AnnotatedCommand\CommandData;
 use Drupal\Core\Database\Database;
 use Drush\Commands\DrushCommands;
 use Drush\Sql\SqlBase;
+use Drush\Utils\StringUtils;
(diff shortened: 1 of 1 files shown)


weitzman:
I improved option description and added a test and merged this in #3344

Issue #2755, opened 2017-05-13
Drush commands within webroot loop infinitely running out of memory


mradcliffe:
I am setting up my Drupal 8 vm for 8.4.x core and contrib development and am trying to update and use drush 9 as it is required for Drupal 8.4.x.

* I completely removed drush:
   * `composer global remove drush/drush`
   * `rm -rf ~/.drush`
* I installed drush 9.0.0-beta1 after having first added minimum-stability to dev as it's needed.
   * `composer global require drush/drush:^9.0.0-beta1` installed dev-master `f46fe29`, which is the latest commit on master as of 2017.05.13.
* Running `drush status` in my home directory prints out basic PHP information.
* Running `drush status` in a Drupal 8.4.x webroot (e67f64e from a week ago) causes drush to loop infinitely trying to fork processes

The following line is repeated until OOM error occurs:
```
[info] Begin redispatch via drush_invoke_process(). Calling proc_open(env COLUMNS=91
$HOME/.composer/vendor/drush/drush/drush --root=/var/www/drupal8.dev/webroot --verbose  core-status 2>&1);
```

If I remove drush and use `composer require drush/drush:dev-master#990a4b321a1b299c63a1e1eb7b26fdaf0e8c4a50` then the error no longer occurs. So I think this is something to do with PR #2749 which introduced webflo/drupal-finder. I am not sure why it would have issues with the directory that I have checked out as I am using a standard git clone of drupal 8.4.x with a healthy amount of contrib. modules in modules/ that are there for porting/development work. webroot contains the following:

```
autoload.php
composer.json
composer.lock
core
example.gitignore
index.php
libraries
modules
profiles
README.txt
robots.txt
sites
themes
update.php
vendor
web.config
```

OOM error:
```
 [warning] exec(): Unable to fork [tput colors 2>&1] preflight.inc:615
 [warning] exec(): Unable to fork [tput colors 2>&1] preflight.inc:615
 [warning] exec(): Unable to fork [tput colors 2>&1] preflight.inc:615

mmap() failed: [12] Cannot allocate memory

mmap() failed: [12] Cannot allocate memory
PHP Fatal error:  Out of memory (allocated 8388608) (tried to allocate 20480
bytes) in
$HOME/.composer/vendor/phpdocumentor/reflection-docblock/src/phpDocumentor/Reflection/DocBlock.php
on line 113
PHP Stack trace:
PHP   1. {main}() $HOME/.composer/vendor/drush/drush/drush:0
[...]

PR #3342, opened 2018-02-02
Fix infinite loop of dispatching when drupal root in symlinked.

awm086:
Fix infinite loop of dispatching when drupal root lives in a symlinked directory.

### Description:
- when I have a d8 installation like /home/vagrant/sites/d8 where   /home/vagrant/sites -> /var/www/
- Path::isBasePath keeps returning false. 
- This causes constant dispatching and an infinite loop with messages . `[preflight] Redispatch to site-local Drush: /home/vagrant/sites/d8/vendor/drush/drush/drush.`

Fixes : #2755

### DIFF
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -268,7 +268,7 @@ class Preflight
         // a site-local Drush. If there is, we will redispatch to it.
         // NOTE: termination handlers have not been set yet, so it is okay
         // to exit early without taking special action.
-        $status = RedispatchToSiteLocal::redispatchIfSiteLocalDrush($argv, $root, $this->environment->vendorPath(), $this->logger())    ;
+        $status = RedispatchToSiteLocal::redispatchIfSiteLocalDrush($argv, realpath($root), $this->environment->vendorPath(), $this->logger())    ;
         if ($status !== false) {
             return $status;
         }


awm086:
@greg-1-anderson @weitzman  could you please review this?

weitzman:
I have been removing realpath() links, not adding them. Please say more about your setup, composer usage, etc. As a general principe, I'm ok with Drush not supporting every possible setup.

kmajzlik:
@weitzman - there is nothing special about configuration, see #2755 . 
Being in /var/www/html drush works.
Being in ~/www (symlink to /var/www/html) drush goes into some kind of loop. Doesnt matter if called Laucher or Drush directly. I tried it on multiple environments.
Drush 9, Drupal 8.

awm086:
@weitzman my set up is simple:
- d8 site via composer 
- drush master via composer and local to the project. I am executing the drush command via the vendor/drush 
- the only non-standard set up is that I have a symlink in the home directory to the site directory. `/home/vagrant/sites/d8 where /home/vagrant/sites -> /var/www/`. You can probably reproduce this easily. 
- Finally if symlinks are not supported then feel free to close this.

weitzman:
I set up a symlink as described and Drush bootstraps fine when I'm in ~/www. Do you happen to have any Drush config that sets options.root to `/home/vagrant/sites/d8`. If so try removing that.

```
vagrant@mass:~/www$ ls -al ~ 
lrwxrwxrwx  1 vagrant vagrant   19 Feb 17 02:51 www -> /var/www/mass.local
```

```
vagrant@mass:~/www$ ./vendor/bin/drush st   
 Drupal version   : 8.4.4                                            
 Site URI         : http://mass.local                                
 DB driver        : mysql                                            
 DB hostname      : 127.0.0.1                                        
 DB port          : 3306                                             
 DB username      : root                                             
 DB name          : drupal                                           
 Database         : Connected                                        
 Drupal bootstrap : Successful                                       
 Default theme    : mass_theme                                       
 Admin theme      : mass_admin_theme                                 
 PHP binary       : /usr/bin/php7.0                                  
 PHP config       : /etc/php/7.0/cli/php.ini                         
 PHP OS           : Linux                                            
 Drush script     : /home/vagrant/www/vendor/bin/drush               
 Drush version    : 9.1.0                                            
 Drush temp       : /tmp                                             
 Drush configs    : /var/www/mass.local/vendor/drush/drush/drush.yml 
                    /var/www/mass.local/drush/drush.yml              
 Install profile  : standard                                         
 Drupal root      : /var/www/mass.local/docroot                      
 Site path        : sites/default                                    
 Files, Public    : sites/default/files                              
 Files, Temp      : /tmp                           
```

awm086:
@weitzman Thanks. I tried a fresh install with drush 9.1.0 and I cannot reproduce the issue. I also tried 9.0.0 again and everything worked just fine ( I even tried a global drush install).   so the reason must be something with configuration. I am closing this PR until I can figure out the exact configs that resulted in the dipatch loop. thanks again.

PR #3353, opened 2018-02-05
Fixes #3351: site-install Drupal 7 with PHP 7.2

### LINKED ISSUE #3351: Drush 8.1.15 install fails with php 7.2.2
Howdy all. So, d7 doesnt work with php 7.2, or at least throws a lot of exceptions for deprecated each() and count() of non-countables. Except I cannot run a `drush si` using 8.1.15 for a d7 site on drupalci, thus we cant really fix these issues without a version of drush that will work. For some reason, the error messages I get are the following: 

`testbot:drupaltestbotpw@172.18.0.4/jenkins_drupal_d7_61404 --clean-url=0 --account-name=admin --account-pass=drupal --account-mail=admin@example.com`
You are about to CREATE the 'jenkins_drupal_d7_61404' database. Do you want to continue? (y/n): y
Starting Drupal installation. This takes a while. Consider using the --notify global option.                                                     [ok]
PHP Fatal error:  Uncaught Error: Call to undefined function cache_get() in /var/www/html/includes/module.inc:754
Stack trace:
#0 /var/www/html/includes/module.inc(954): module_implements('system_theme_in...')
#1 /var/www/html/modules/system/system.module(2511): module_invoke_all('system_theme_in...')
#2 /var/www/html/includes/theme.inc(798): _system_rebuild_theme_data()
#3 /var/www/html/includes/theme.maintenance.inc(57): list_themes()
#4 /var/www/html/includes/bootstrap.inc(2872): _drupal_maintenance_theme()
#5 /var/www/html/includes/errors.inc(179): drupal_maintenance_theme()
#6 /var/www/html/includes/bootstrap.inc(2609): _drupal_log_error(Array, true)
#7 [internal function]: _drupal_exception_handler(Object(Error))
#8 {main}
  thrown in /var/www/html/includes/module.inc on line 754

I *was* able to install the site manually, and then run the tests, but drupalci has to use drush si to create a testable site.

greg-1-anderson:
Always use 'fwrite' instead of 'print' in 'drush_print' to avoid undesired interaction with PHP's handling of the php headers.

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -36,6 +36,11 @@ function drush_print($message = '', $indent = 0, $handle = NULL, $newline = TRUE
   }
   if (!isset($handle)) {
     $handle = STDOUT;
+    // In the past, Drush would use `print` here; now that we are using
+    // fwrite (to avoid problems with php sending the http headers), we
+    // must explicitly capture the output, because ob_start() / ob_end()
+    // does not capture output written via fwrite to STDOUT.
+    drush_backend_output_collect($msg);
   }
   fwrite($handle, $msg);
 }
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -34,12 +34,10 @@ function drush_print($message = '', $indent = 0, $handle = NULL, $newline = TRUE
   if (($charset = drush_get_option('output_charset')) && function_exists('iconv')) {
     $msg = iconv('UTF-8', $charset, $msg);
   }
-  if (isset($handle)) {
-    fwrite($handle, $msg);
+  if (!isset($handle)) {
+    $handle = STDOUT;
   }
-  else {
-    print $msg;
-  }
+  fwrite($handle, $msg);
 }
 
 /**


greg-1-anderson:
A couple of tests fail because this change interferes with the way they were collecting output

ryanaslett:
I'll give this a whirl tomorrow morning

greg-1-anderson:
The test failures were happening because output emitted via `fwrite(STDOUT)` is not captured by `ob_start()` / `ob_end()`. I adjusted for this by explicitly capturing output in `drush_print()`.

The side-effect here is that if anyone is emitting output via `print` instead of `drush_print`, it has the potential to do the wrong thing. It has always been deprecated to call `print` instead of `drush_print`, though, so this is not a serious issue.

We have two options:

1. Leave this PR as-is.
1. Conditionalize this change on the php version, so that PHP 7.1 and lower continue to use the current implementation, and PHP 7.2 and onward use the implementation in this PR.

The later strategy would be better for strict maintenance of backwards compatibility; however, since we are only breaking non-conforming code here, I think it is probably okay to just make the change. I'm willing to go either way, though, depending on what folks think.

weitzman:
I'm fine with the PR as is.

Issue #3078, opened 2017-10-18
Env vars lead to backend "Could not open input file: env"


axel-rutz:
Drush version 8.1.15

Drush backend invoke with drush environment variable results in messed up drush invocation.
See bold code in command output.

Invocation code:

>     drush_invoke_process("@self", 'sql-sync', ["@$source", "@$destination"]);

Alias:

`$aliases['foo.default']['#env-vars']['DRUSH_PHP'] = 'php5.6';`

Command output:

> **Backend invoke: /usr/bin/php env DRUSH_PHP=php5.6 /opt/drush/drush.php** --php=/usr/bin/php  --backend=2 --verbose --debug                                                                           [command]
> --root=/foobar/docroot --uri=http://foobar.xxx  sql-sync '@live.default'
> '@self' 2>&1 [1.27 sec, 10.47 MB]
> /usr/bin/php env DRUSH_PHP=php5.6 /opt/drush/drush.php --php=/usr/bin/php  --backend=2 --verbose --debug                                                                                            [notice]
> --root=/foobar/docroot --uri=http://foobar.xxx  sql-sync '@live.default'
> '@self' 2>&1 [1.27 sec, 10.48 MB]
> Could not open input file: env
> The command could not be executed successfully (returned: Could not open input file: env                                                                                                             [error]
> , code: 1) [1.3 sec, 10.48 MB]
>

greg-1-anderson:
The env variables are mysteriously out-of-order relative to the path to the php binary. Not sure when that could have happened; this code hasn't changed in a long time.

How are you calling the code that runs the `drush_invoke_process`?

greg-1-anderson:
When I run `drush8 @localsite.dev status -d` it works fine. It targets the drush launcher, not php. You perhaps have set up your local path to php in some way other than just having it on the $PATH, I imagine?

axel-rutz:
Hmm, further strange experiences:

* every drush foo command in the cli runs fine
* sometimes when drush commands call others, the problem arises. i have a custom command that calls sql-sync which then calls core-rsync. the problem only arises in the 2nd call (to rsync). 

Spooky...

axel-rutz:
> You perhaps have set up your local path to php in some way other than just having it on the $PATH, I imagine?

Nothing like this. Just the above "#env-vars".

axel-rutz:
Quite notable: sql-sync does a correct invocation of "status" to get the local temp dir, but an incorrect one to rsync:

```
...

[...]

PR #3359, opened 2018-02-06
Environment variables are prepended to Drush command in the wrong order

itafroma:
`drush_build_drush_build_command()` adds environment variables in the wrong place within the command string. For example, if I have an alias like:

```php
$aliases['example'] = [
  '#env-vars' => [
    'FOO' => 'bar',
  ],
];
```

And Drush is loaded via Composer so that its path is something like:

```sh
~/.composer/vendor/drush-ops/drush/drush.php
```

`drush_build_drush_build_command()` will construct a command string that looks like:

```sh
/usr/local/bin/php  -c /etc/php/php.ini env FOO=bar --php=/usr/local/bin/php --php-options=' -c /etc/php/php.ini'
```

This will lead to Drush failing to execute any commands for that alias with the error:

> Could not open input file: env

Because the command trying to load a file named `env` into PHP.

The correct command string should be instead:

```sh
env FOO=bar /usr/local/bin/php  -c /etc/php/php.ini--php=/usr/local/bin/php --php-options=' -c /etc/php/php.ini'
```

Original issue: https://github.com/drush-ops/drush/issues/3078
h/t @axel-rutz for the solution, posted here: https://github.com/axel-rutz/drush/commit/c41235578ab3f9ff6b0dacae3d45a48b3dfae08e

### DIFF
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -135,10 +135,11 @@ function drush_build_drush_command($drush_path = NULL, $php = NULL, $os = NULL, 
 
   // Add environmental variables, if present
   if (!empty($environment_variables)) {
-    $prefix .= ' env';
+    $env = 'env';
     foreach ($environment_variables as $key=>$value) {
-      $prefix .= ' ' . drush_escapeshellarg($key, $os) . '=' . drush_escapeshellarg($value, $os);
+      $env .= ' ' . drush_escapeshellarg($key, $os) . '=' . drush_escapeshellarg($value, $os);
     }
+    $prefix = $env . ' ' . $prefix;
   }
 
   return trim($prefix . ' ' . drush_escapeshellarg($drush_path, $os) . $additional_options);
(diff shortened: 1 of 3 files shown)


axel-rutz:
Cool it works for you!

greg-1-anderson:
This used to work in the past. It would be super useful to have a simple test so that it does not break again in the future.

weitzman:
Looks like the test failed. Any ideas?

weitzman:
This feature is failing because it is not implemented in Drush9. Help wanted. https://github.com/drush-ops/drush/blob/88ea047f1aae9039e91336489a4a449b378f9808/examples/example.site.yml#L184

weitzman:
I merged the backport of this PR. Closing this as the feature does not yet exist in 9

itafroma:
Sorry about the failing test: was having trouble getting testing working locally and wanted to try a stab at a quick fix. Thanks for the backport merge!

PR #3361, opened 2018-02-06
(Backport) Environment variables are prepended to Drush command in the wrong order

itafroma:
This is a backport of #3359 to Drush 8.

### DIFF
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -498,10 +498,11 @@ function drush_build_drush_command($drush_path = NULL, $php = NULL, $os = NULL, 
 
   // Add environmental variables, if present
   if (!empty($environment_variables)) {
-    $prefix .= ' env';
+    $env = 'env';
     foreach ($environment_variables as $key=>$value) {
-      $prefix .= ' ' . drush_escapeshellarg($key, $os) . '=' . drush_escapeshellarg($value, $os);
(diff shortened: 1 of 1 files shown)


weitzman:
A test for this backport would be helpful. Thanks.

itafroma:
Just some notes after looking into adding tests:

- There already exists a test for `#env-vars` functionality: `Unish\saCase::testDispatchStrictOptions()`.
- However, it does not ever go through `drush_build_drush_command()`, so the problem does not get manifested during testing.
- In `drush_main()`, `drush_preflight_command_dispatch()` returns `true` for the two commands that test `#env-vars` (both use `unit-eval` as their command), so `$bootstrap->bootstrap_and_dispatch()`, which is where `drush_build_drush_command()` does get called, never executes.
- The first test in `Unish\saCase::testDispatchStrictOptions()`, which uses `ra-sync` as its test command, returns `false` for `drush_preflight_command_dispatch()` so `$bootstrap->bootstrap_and_dispatch()` _does_ get called as well as `drush_build_drush_command()`
- Even in this case, the problem does not appear in the tests because the Drush command being run for the test does not end with `.php`.
- I have not yet determined why `unit-eval` bypasses  `drush_build_drush_command()` but `ra-sync` does not.

PR #3365, opened 2018-02-07
Add entity:delete command

weitzman:


weitzman:
Future improvements could be a confirm prompt and a test. Merging for now.

PR #3367, opened 2018-02-07
Clarify namespaces in command authoring docs.

### LINKED ISSUE #3366: Drush global command documentation has an incorrect namespace
It says: `The file's namespace should be \Drush\Commands.`
Where in reality it should be `The file's namespace should be Drush\Commands.`.
Located here: http://docs.drush.org/en/master/commands/#global-drush-commands

richardbporter:
Resolves #3366 

The current docs aren't really incorrect but this might help clarify things for future use. @Skarlso see: http://php.net/manual/en/language.namespaces.rules.php

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -45,7 +45,7 @@ In Drush 9, the default services file, `drush.services.yml`, will be used in ins
 Global Drush Commands
 ==============================
 
-Commandfiles that don't ship inside Drupal modules are called 'global' commandfiles. See the examples/Commands folder for examples. In general, its better to use modules to carry your Drush commands. If you still prefer using a global commandfiles, please note:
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Sure I am +1 on this feature. It might also be helpful to show an example commandfile, similar to [the one in the annotated commands project](https://github.com/consolidation/annotated-command#example-annotated-command-class), but with the namespace included.

richardbporter:
Sure, I could add that. Do you think an entire class is necessary or just the beginning? I might just copy/paste one of the examples or the one generated via `drush generate dcf`.

greg-1-anderson:
Yeah, if folks just use `drush generate` or copy an example, then they'll have a better time, both with respect to this issue and other potential problems that might arise. I'm fine with this PR in its current form.

richardbporter:
Agreed. Less duplication.

PR #3378, opened 2018-02-12
Fixed wrong syntax (in drush.yml) when setting uri option

hansfn:


hansfn:
Updated patch based on comment https://github.com/drush-ops/drush/issues/3121#issuecomment-364802730 

I wasn't able to verify since I still not sure if sites/default/drush.yml is even loaded - ref #3277

hansfn:
A quick comment after looking at the commit the closed this issue:

The "Options" heading doesn't really make sense anymore.

weitzman:
Its trying to describe the uri option. Any suggestions on how to change or expand this document in general?

hansfn:
Not from the top of my head - sorry. Anyway, how about just changing "Options" to "Setting the uri option" if that is the only option you intend to cover? You don't have to reply - save that time for something else.

Issue #3383, opened 2018-02-13
Drush 9: Config import doesn't use labels provided.


seanbeaton:
According to the docs [here](https://drushcommands.com/drush-9x/config/config:import/) you should be able to specify a config directory for the config import/cim command, like you can for cex. This is also how it worked in drush 8.x. Unfortunately, it appears as though the label is completely ignored in ConfigImportCommands.php's import function, which is the variable provided to the import function both by the first command line argument and by the "Choose a source dialog" shown below:

```
$ drush cim

 Choose a source.:
  [0] Cancel
  [1] sync
  [2] dev
 >
```

To further compound issues, it appears that the --source command line argument does work to specify the directory from which you want to import config. 

I'm going to add a pull request to this issue shortly that fixes these issues.

seanbeaton:
Copy of the pull request text:

> This commit fixes the issue by looking for the label in ConfigImportCommands.php's import function. If the label exists, it uses the config directory specified in $config_directories in settings.php for that label (as described in [the documentation](https://drushcommands.com/drush-9x/config/config:import/)). If the source option is specified, it uses that but outputs a notice mentioning what happened.
> 
> This fix should retain all existing functionality, and allow users to once again use non-default directories by specifying them in the settings.php.

PR #3384, opened 2018-02-13
#3383 - fix label not being used for drush config:import command

seanbeaton:
This commit fixes the issue by looking for the label in ConfigImportCommands.php's import function. If the label exists, it uses the config directory specified in $config_directories in settings.php for that label (as described in [the documentation](https://drushcommands.com/drush-9x/config/config:import/)). If the source option is specified, it uses that but outputs a notice mentioning what happened.

This fix should retain all existing functionality, and allow users to once again use non-default directories by specifying them in the settings.php.


weitzman:
Lets do it like config:export does - https://github.com/drush-ops/drush/blob/a24b4dea65da5ae01e472d66b3a47dd6f38157b4/src/Drupal/Commands/config/ConfigExportCommands.php#L88

seanbeaton:
Cool - that makes a lot more sense. I've updated the pull request.

rbran100:
#3383

PR #3389, opened 2018-02-16
SQL commands are bootstrapping too much. Should bootstrap to database.

claudiu-cristea:
## Problem

Removing the tables from the database could run in trouble because, right now, the `sql:*` commands are doing a full Drupal bootstrap. In my project I have the case when I'm using `sql:drop` to drop all tables before doing a MySQL dump import. But the current DB content could be totally messed-up and, because we are fully bootstrapping, Drupal will try initialize the container, is crashing with exceptions regarding services (mostly services for modules not installed yet).

Also, it's clear that`sql:*` commands are operating at database level and they don't need the full Drupal to be bootstrapped.

## Proposal

1. Make all `sql:*` commands require DATABASE as maximum bootstrap level.
1. Remove the check for `key_value` table in Drupal database bootstrap validation. When running a SQL command, like dropping the tables, there's a big chance that the table doesn't exist making the bootstrap to fail. I have no idea what was the reason for checking if that table exists. I guess this was inherited from older Drush versions and it was a way to check if we're on a Drupal database. But at the database level we shouldn't be aware of that.

### DIFF
--- a/src/Boot/DrupalBoot8.php
+++ b/src/Boot/DrupalBoot8.php
@@ -127,17 +127,6 @@ class DrupalBoot8 extends DrupalBoot implements AutoloaderAwareInterface
         return true;
     }
 
-    public function bootstrapDrupalDatabaseValidate()
-    {
-        return parent::bootstrapDrupalDatabaseValidate() && $this->bootstrapDrupalDatabaseHasTable('key_value');
-    }
-
-    public function bootstrapDrupalDatabase()
-    {
-        // D8 omits this bootstrap level as nothing special needs to be done.
-        parent::bootstrapDrupalDatabase();
-    }
-
     public function bootstrapDrupalConfiguration(AnnotationData $annotationData = null)
     {
         // Default to the standard kernel.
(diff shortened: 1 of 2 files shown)


weitzman:
The sql commands bootstrap as far as they can. Thats what `max` means. Thats desireable behavior because otherwise commandfiles in modules couldn't hook into these commands.

For your case, I suggest importing the new DB and then running `drush cache-clear`. If that still creates errors then you are going to have do a custom workaround like renaming/dropping the key_value table temporarily.

claudiu-cristea:
> Thats desireable behavior because otherwise commandfiles in modules couldn't hook into these commands.

I don't understand. If I'm running `drush sql:drop` I'm not expecting any module to hook into. It's just `sql:drop` that needs to run and that is OK with only the database bootstrap. What I'm missing here?

weitzman:
Well, other people *are* expecting/hoping to hook into these command. For example, they may want to run some code before or after every sql:drop. They may want deny the drop via [a policy file](https://github.com/drush-ops/drush/blob/master/examples/Commands/PolicyCommands.php)

PR #3397, opened 2018-02-20
Support a new @max annotation in sql commands (fix #3389)

weitzman:
Less bootstrapping for sql commands.


greg-1-anderson:
Maybe `@bootstrap max configuration` would be more natural?

Issue #3398, opened 2018-02-20
Drush multisite installation with existing config


sgurlt:
Hey,

we are currently working on a multisite setup and just ran into a very strange bug.
I have setup a fresh Drupal 8.4.x installation with minimal setup, enabled views (for testing) and exported the config.

`drush site-install minimal`
`drush -y en views`
`drush site-install cex`

The config folder is located at `../config`

Afterwards I want to create the first multisite using drush. So within my drupal root dir:

`drush si --config-dir="../config" --sites-subdir="sub1.localhost" --db-url="mysql://root:root@mariadb/sub1"`

The installation process works quite well, but the issue we have is that the config is not getting imported:
```
You are about to create a sites/sub1.localhost/settings.php file and CREATE the 'sub1' database. Do you want to continue? (y/n): y
Starting Drupal installation. This takes a while. Consider using the --notify global option.                                                                                                                                                                                                                                                                            [ok]
Installation complete.  User name: admin  User password: H32RC2zWMb                                                                                                                                                                                                                                                                                                     [ok]
Do you want to update uuid key in system.site config? (y/n): y
There are no changes to import.                                                                                                                                                                                                                                                                                                                                         [ok]
Congratulations, you installed Drupal!
```

I also tried doing it with a absolut path to my config:

`drush si --config-dir="/var/www/html/config" --sites-subdir="sub2.localhost" --db-url="mysql://root:root@mariadb/sub2"`

But it gave me the same result.

Tested Drush Versions:
- 8.1.13
- 8.1.16


Maybe someone got an idea what could go wrong?

sgurlt:
Actually found the issue.

`/**
 * Command callback.
 */
function drush_core_site_install($profile = NULL) {
  $args = func_get_args();
  $form_options = array();

  if ($args) {
    // The first argument is the profile.
    $profile = array_shift($args);
    // Subsequent arguments are additional form values.
    foreach ($args as $arg) {
      list($key, $value) = explode('=', $arg, 2);

      // Allow for numeric and NULL values to be passed in.
      if (is_numeric($value)) {
        $value = intval($value);
      }
      elseif ($value == 'NULL') {
        $value = NULL;
      }

      $form_options[$key] = $value;
    }
  }

  // If the profile is not explicitly set, default to the 'minimal' for an issue-free config import.
  if (empty($profile) && drush_get_option('config-dir')) {
    $profile = 'minimal';
  }

  drush_include_engine('drupal', 'site_install');
[...]

PR #3400, opened 2018-02-20
Set URI if sites-subdir is set for config import

sgurlt:
Issue:
#3398


weitzman:
I can see how this is helpful. However, 8.x is in maintenance only and this PR is too disruptive. Folks are welcome to patch/fork their drush 8, or upgrade to drush9.

PR #3402, opened 2018-02-20
Add the migrate runner commands

### LINKED ISSUE #2926: Add migrate_run instead of migrate_tools
https://www.drupal.org/node/2905741 says "Add Migrate Tools to Drush (contrib)"

Instead add https://www.drupal.org/project/migrate_run

migrate_tools depends on migrate_plus which is an incredibly, incredibly broken module, full of bad ideas and missing understanding on how D8 works in general and how D8 migrate is architected.

claudiu-cristea:
Resolves #2926

### Should be fixed before merging:

- https://www.drupal.org/project/migrate_run/issues/2914429
- Migrate import runs dependent migrations multiple times.

Other improvements can be added after merging.


claudiu-cristea:
Just note here. I needed a namespace for some classes. I have no idea if my pick was correct: `Drush\Drupal\Migrate`.

weitzman:
One thing I did not consider earlier is documentation. Folks need to learn how to write and run migrations. Where should we point them to?

claudiu-cristea:
I moved the commands from `Drush\Commands\core` under `Drush\Drupal\Commands\core`. @weitzman, could you , please check the @todo from `\Drush\Drupal\Commands\core\MigrateRunnerCommands::getMigrationPluginManager`?

claudiu-cristea:
This is ready for review.

weitzman:
I added a todo about docs for Drupal => Drupal upgrade migrations.

claudiu-cristea:
Reworked the issue summary and added a first try for Migrate Plus in https://www.drupal.org/project/migrate_plus/issues/2951968

heddn:
> Add a doc section specifically telling folks how to get started with Drupal => Drupal upgrades. How to make use of the migrations that are in core.

I'm really happy with what I'm seeing going on here. I do want to be cautious about the D2D migration docs. Something like the following might be a good start.

Running `drush migrate:status` and other core drush migrate commands will not work out of the box for core migrations. Unless site owners have created a 'migrate' db entry in their settings(.local).php. This is because all Core the source plugins are all SQL based and require a defined connection. There is a fallback to a db key called 'migrate'. But that's kinda an evil hack and only supports a single DB at a time. If you want to support more than a single DB source, then install migrate_plus and use migrate groups instead.

^ is probably way to technical, but it is a start.

Eli-TW:
Question with respect to this approach - if migrate_plus alters the core commands rather than migrate_tools, that means people wanting to use other parts of migrate_plus (eg the prepareRow event, the JSON/XML source plugins) will be forced to use its commands rather than the Drush ones, whereas before it would have been possible to use migrate run with those source plugins. Is that correct? And is that a limitation that we have decided to accept?

heddn:
The prepare row event and the JSON/XML and CSV sources will not be effected by Drush core being the runner.

claudiu-cristea:
@Eli-TW this is a good question. The prepare row event is now provided also by this PR, so that will work also without `migrate_plus`. But the module also provides some plugins. For this reason, I think, it would be correct to consider moving the command alters in `migrate_tools` (?)

heddn:
@claudiu-cristea do we want to remove the prepare row event from drush core? Or bring it in separately? That would make this focused on the drush commands only then. And we can discuss the event in a follow-up.

claudiu-cristea:
@heddn That event is critical to migrate runner, so we cannot remove it from Drush core. We can probably rename it to avoid any confusion (DRUSH_PREPARE_ROW) or, better, move it to Drupal core.

heddn:
@claudiu-cristea idealy, moving to core would ideally happen. Not sure if it will, but I'd like it. That depends on commiter discretion.

I did a quick scan of the code here in this PR, what is using it inside of Drush? Why is MigratePrepareRowEvent critical?

claudiu-cristea:
@heddn to support the runner `--idlist` option. See https://github.com/drush-ops/drush/pull/3402/files#diff-d3fada9bc3f3bdd833a2400a4e4c57e3R400

claudiu-cristea:
There's Drupal core issue to add the event to the `migrate` core module. However, the issue scope is more seems broader. Probably adding only the event dispatcher could be split into a separate issue  https://www.drupal.org/project/drupal/issues/2488836

heddn:
Hmm, I wonder if we could bring the support for idlist in an implementation of the core prepare_row hook for now? With a TODO to a new issue to add it more better. Because ideally we'd have the event in core.

claudiu-cristea:
I split the event thing in https://www.drupal.org/project/drupal/issues/2952291

claudiu-cristea:
@heddn using the hook could be also an idea. So we avoid all the event headache.

heddn:
@claudiu-cristea I've re-booted the core issue. But it more than likely not going to land in 8.5, so let's proceed with a hook for the mean time?

claudiu-cristea:
As we've discussed in the chat and because implementing a hook doesn't work (because the migrate executable is not accessible from the hook implementation), we've decided to keep the Drush core event, even it, eventually, will be redundant with the `migrate_plus` prepare row event. Both will become obsolete after the event will land in Drupal core, in https://www.drupal.org/project/drupal/issues/2952291. In order to prevent any confusion, we renamed the event from `PREPARE_ROW` to `DRUSH_MIGRATE_PREPARE_ROW`.

weitzman:
Just want top say that I did a large migration with Migrate Run recently and worked really well. I am again happy to merge this. The source repo was deleted but the code here is recoverable via https://patch-diff.githubusercontent.com/raw/drush-ops/drush/pull/3402.patch

heddn:
There's been several new features added to migrate_tool's drush runner recently that might be considered useful. I'd be interested if using it as the basis might be another option.

heddn:
There's also pretty complete testing of the commands at this point too. I'd be more then happy to move those things over into drush core.

weitzman:
I'm interested in new runner features, for sure. And tests. What I'm not so interested in is migrations as config entities. Is that a built-in assumption for drush_tools?

heddn:
@weitzman I think it could be optional. If we wrote it so that the list of migrations be decorated or something. That way migrate_tools could provide config entities without having to support all the runner capabilities.

claudiu-cristea:
The `migrate_run` is only using Drupal core. Nothing more. For this reason, I think, it better suits Drush.

heddn:
In the area where we've agreed to differ is the fact that migrate_tools also finds migrate_plus config entities. But tools has added several new features in the past few months, namely support for syncing, progress bar, id-list, continue-on-failure. These are features it would be nice to maintain. If we have to drop config entities, I'd be inclined to do that and then figure out a way to re-add that support via some drush alter hook or something.

claudiu-cristea:
Then `migrate_tools` should implement command hooks that are adding migrations from configs to the discovery.

heddn:
I'm glad we can all agree :)

PR #3370, opened 2018-02-08
Add --result-file-auto to save to a date-based filename under ~/drush-backups directory

hugovk:
Fixes #3364.

Drush 9 has lost a Drush 8 feature:

* If `--result-file` is provided with no value, then a date-based filename will be created under ~/drush-backups directory.

That's a really useful option for automated backups, but Symfony Console makes it hard to work with options that can be both Boolean and a string.

This PR adds a new `--result-file-auto` Boolean option to resurrect this.

If both `--result-file` and `--result-file-auto` are given, `--result-file` takes precedence.

image

Does something like this look like a possible solution to #3364?

PR #3418, opened 2018-02-27
Fix #3417. sql-sync dump filename not automatically generated

### LINKED ISSUE #3417: 9.2.0 broke `sql-sync`
My test suites for acquia/blt began failing immediately after drush/drush 9.2.0 was released. 

Executing:
```
./vendor/bin/drush sql-sync @default.clone @self --structure-tables-key=lightweight --create-db --no-interaction --ansi
```

Outputs:
```
You will destroy data in drupal and replace with data from drupal3.
 [notice] Starting to create database on target.
Creating database drupal. Any existing database will be dropped!
 // Do you really want to continue?: yes.                                       
 [notice] Starting to dump database on source.
 [success] Database dump saved to 1.gz
 [notice] Starting to import dump file onto target database.
 [error]  File(s) not found: 1.gz 
 [error]  Failed to import 1.gz into target. 
```

weitzman:
One downside of this fix is that running this code with a prior version of Drush will cause your dump files to be names `auto` instead of date based backups. I think mismatched Drush versions are a lot less common now that we are local to the project.

### DIFF
--- a/.circleci/setup.sh
+++ b/.circleci/setup.sh
@@ -30,6 +30,7 @@ echo 'mbstring.http_output = pass' >> $HOME/php.ini
 echo 'memory_limit = -1' >> $HOME/php.ini
 echo 'sendmail_path = /bin/true' >> $HOME/php.ini
 echo 'date.timezone = "UTC"' >> $HOME/php.ini
+echo 'opcache.enable_cli = 0' >> $HOME/php.ini
 
 # Copy our php.ini configuration to the active php.ini file
 # We can't use `php -r 'print php_ini_loaded_file();` when there is no php.ini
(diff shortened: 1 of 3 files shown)


weitzman:
I will add that sql:sync should be working even without this fix so I'm not sure whats going on in the BLT test failure. Could you try with this one line change and see if sql:sync starts succeeding?

greg-1-anderson:
`'result-file' => true` should make backend invoke use `--result-file`; maybe some bug is causing it to insert `--result-file=1` instead?

The ` => NULL` trick is only for unish, if I recall correctly.

weitzman:
Could be ... But --result-file without a value is now invalid. That option requires a value. I just pushed a commit which documents this. This change went in at #3370.

grasmash:
I can verify that this does resolve the initial issue, and causes acquia/blt tests to pass.

PR #3422, opened 2018-02-28
Fixed pg_dump command. No extra password file needed for execution.

rackberg:
This pull request should make it possible again to dump the data from a PostgreSQL database.
The -w option of the pg_dump command is needed to prevent the command to ask for the database password. So a manually provided .pgpass file is not needed for this to work.


weitzman:
pretty sure this is a dupe issue.

weitzman:
See recent pg fixes, and then post here if this is still needed.

PR #3426, opened 2018-03-01
Do not allow drush to run a command in preflight successfully and again in the regular bootstrap

nickveenhof:
We discovered that drush uli in certain circumstances will execute twice.
I'll show it with drush -vd. I'm not sure if the bug also exists in Drush 9.

The server has a global drush 8.1.16 and the vendor folder of the project also has a drush executable.

`$ drush8 @myproject.dev uli -vd`

```
Using the Drush script found at /usr/local/src/drush8/drush.launcher using pcntl_exec
Drush preflight prepare loaded autoloader at                         [preflight]
/usr/local/src/drush8/vendor/autoload.php [0.01 sec, 2.45 MB]
Starting Drush preflight. [0.01 sec, 2.45 MB]                                                                                                                                                                                                                      [preflight]
Cache HIT cid: 8.1.16-commandfiles-0-5e6df3b528f4f0f3e00472493f0eb7de [0.01 sec, 2.73 MB]                                                                                                                                                                              [debug]
Scanning into /etc/drush for /.*aliases\.drush(8|)rc\.php$/ [0.04 sec, 5 MB]                                                                                                                                                                                           [debug]
...
Loaded alias @myproject.dev from file /etc/drush/myproject.aliases.drushrc.php [2.12 sec, 7.3 MB]                                                                                                                                                           [notice]
Begin redispatch via drush_invoke_process(). [2.12 sec, 7.31 MB]                                                                                                                                                                                                      [notice]
Backend invoke: env COLUMNS=270  /var/www/customprojects_D8/myproject/dev/vendor/drush/drush/drush.launcher  --root=/var/www/customprojects_D8/myproject/dev/docroot --verbose --debug  user-login 2>&1 [2.13 sec, 7.31 MB]                                [command]
Calling proc_open(env COLUMNS=270  /var/www/customprojects_D8/myproject/dev/vendor/drush/drush/drush.launcher  --root=/var/www/customprojects_D8/myproject/dev/docroot --verbose --debug  user-login 2>&1);
Drush preflight prepare loaded autoloader at                         [preflight]
/var/www/customprojects_D8/myproject/dev/vendor/autoload.php
[0.01 sec, 2.75 MB]
Starting Drush preflight. [0.01 sec, 2.75 MB]                                                                                                                                                                                                                      [preflight]
Cache HIT cid: 8.1.16-commandfiles-0-7d604cc7822369e48675695b14fcf8b3 [0.01 sec, 2.97 MB]                                                                                                                                                                              [debug]
[...]

PR #3428, opened 2018-03-01
Prevent 'drush @local uli' and related commands from executing twice.

greg-1-anderson:
Alternate implementation of #3426

### DIFF
--- a/includes/preflight.inc
+++ b/includes/preflight.inc
@@ -611,8 +611,15 @@ function drush_preflight_command_dispatch() {
     $local_drush = find_wrapper_or_launcher($root);
   }
   $is_local = drush_get_option('local');
+  $must_use_site_local = !empty($root) && !empty($local_drush) && empty($is_local);
+  // If the command sets the 'handle-remote-commands' flag, then we will short-circuit
+  // remote command dispatching and site-list command dispatching, and always let
+  // the command handler run on the local machine.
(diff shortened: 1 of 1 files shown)


nickveenhof:
That was insanely fast for such an old bug. Thanks @greg-1-anderson. Any timeline when a new drush  8.x release would be made that incorporates this fix?

greg-1-anderson:
We'll need some feedback (or at a minimum, lack thereof!) on the 8.x-dev branch first.

Spurlos:
Unfortunately these changes have broken ssh access using drush and aliases on some specific environments.

> drush @alias ssh
> A site alias is required. The way you call ssh command has changed to 'drush @alias ssh\'. [error]
> command terminated with exit code 1

I was able to ssh successfully to server from local development Ubuntu 18.04 machine, but the same command fails in Docker container that we use. The  base for the container is Alpine linux, which is missing pcntl php extension, which I think is the cause (untested, because alpine repos have no pcntl in them), but that is the only significant difference between local and container php cli extensions diff.

Reverting this commit fixes the issue. And looks like i'm not the only one affected: https://github.com/amazeeio/lagoon/issues/542

greg-1-anderson:
You'll need to do some additional diagnosis to determine which term is causing problems.

Does `drush @alias ssh` work for you in Drush 9?  The bootstrap is quite different on the master branch.

lmakarov:
@greg-1-anderson I submitted a bug report with more context ^

PR #3436, opened 2018-03-06
Fix #3423. Allow custom target dir for config:pull

### LINKED ISSUE #3423: drush config-pull @remote --target-dir=foo
Use case: As a site developer, i want to config-sync from a (fully installed) remote site to my local copy of the git repository. (Even if locally there is no bootstrapped drupal)

Something like:
`drush config-pull @remote --target-dir=foo`

weitzman:


axel-rutz:
Code looks straightforward what i wished for. 😍😍😍

Reusing known concepts of rsync path style, yeah!
Will test as soon as my tight schedule allows.

Just an idea: For the path example, we might use drupal-project-template's standard of `@self:../config/sync`.

weitzman:
Improved that example and merged. Thanks.

axel-rutz:
Thank you for this quick implementation.

(Not wanting to be greedy, but) has this a chance of a D8 backport? Or is that branch frozen?
(I don't like the situation, but with a plethora of legacy, i'll be stuck on that version for a handful of projects for some time still.)

weitzman:
HostPath doesn't exist for D8 so this will need a different implementation.
I dont think I'll work on that but others are welcome to.

axel-rutz:
OK cool. I'll scratch if the itching urges!

PR #3442, opened 2018-03-06
Update the ArtCommands URL

echernyavskiy:


greg-1-anderson:
It is no longer obvious that the art command was inspired by an xkcd comic, but I am +1 on the improvement here.

Issue #3444, opened 2018-03-07
pm-enable does not trigger translation import


Sutharsan:
In Drupal 8 with Locale module enabled, interface translations are imported when modules are enabled via the interface. But when enabled via Drush, this does not happen.
The same situation applies for enabling themes.

Proposed solution: Add a check for and execution of batch process as part of pm-enable.
Follow-up: option to suppress the auto-import.

Sutharsan:
Created pull request #3445 to execute the batch process.

weitzman:
Drush9 PR is in. Lets talk as needed in #3445

PR #3445, opened 2018-03-07
D8 Run batch process during pm-enable when needed.

Sutharsan:
Adds batch execution to to pm-enable command.


Sutharsan:
Fixes issue #3444.

weitzman:
could we add a test that changes language, then enables locale module and devel module? would that trigger the batch?

Sutharsan:
I'll give it a try

Sutharsan:
The down site of this test it that is depends on the availability of ftp.drupal.org to fetch the devel module translation from.

weitzman:
Is it feasible to put the translations in our repo (see /tests/resources) to avoid that downside?

Sutharsan:
I've been thinking about this. It has some implications but I think it can be done by:
- Change the source of translations to local files only
- Create and using virtual file system for the translation file
- Copy a translation file from local drush test assets to the vitual translation files directory
- Change the file name pattern to not include the module version (otherwise a (different) release version of the devel module would result in a different translation file name).

Sutharsan:
Converted the test to use a local source, and with that made it independent from an external source.

I expected Unish to use the vfStream for a virtual file system, but not. I used the drush-sut file system instead.

weitzman:
Alas tests are failing for this and the D9 PR. Some fails look unrelated but the test added here is failing.

Sutharsan:
It turns out that `drush php-eval` does not autoload `drush php-eval "\Drupal\language\Entity\ConfigurableLanguage::create(...)"` does not work. I've tried different variations with `require_once()` too, but in vain.

I've removed the test, as I see no way to add a language in the drupal sut. I propose to re-add this test once there is a drush `language-add`.

weitzman:
Why would that class not autoload? Is the command not bootstrapping properly? 

In any case, we really aren't adding features to 8.x anymore so i'm fine with skipping this for 8.x if we can't get a test working.

Sutharsan:
I don't know about the why.
Running `drush php-eval "\Drupal\language\Entity\ConfigurableLanguage::create(['id' => 'nl', 'label' => 'Dutch'])->save()"` in a drush8 + drupal8 environment _and_ in the unish test, both result in
> Fatal error: Uncaught Error: Class 'Drupal\language\Entity\ConfigurableLanguage' not found in /Users/erikstielstra/.drush/.versions/drush-8.x/commands/core/core.drush.inc(1165) : eval()'d code:1

While the same code does work as expected in Drush9 + Drupal8 and in tests.

weitzman:
We decided not to add any features to Drush 8.

PR #3447, opened 2018-03-07
Allow registering of command info alterer services

claudiu-cristea:
We are able to alter/replace commands in modules but we cannot intercept and alter the command info (annotations). This PR adds such capability.

How it works?

- Modules wanting to alter the command info should add a service class that implements `\Consolidation\AnnotatedCommand\CommandInfoAltererInterface`.
- The service should be added in `drush.services.yml` and should be tagged with the `drush.command_info_alterer` tag.
- Add the alter logic in the `alterCommandInfo()` of the alterer class.

See the testing `woot` module for details.

### DIFF
--- a/src/Boot/DrupalBoot8.php
+++ b/src/Boot/DrupalBoot8.php
@@ -192,6 +192,16 @@ class DrupalBoot8 extends DrupalBoot implements AutoloaderAwareInterface
         // The upshot is that the list of console commands is not available
         // until after $kernel->boot() is called.
         $container = \Drupal::getContainer();
+
+        // Set the command info alterers.
+        if ($container->has(DrushServiceModifier::DRUSH_COMMAND_INFO_ALTERER_SERVICES)) {
+            $serviceCommandInfoAltererlist = $container->get(DrushServiceModifier::DRUSH_COMMAND_INFO_ALTERER_SERVICES);
+            $commandFactory = Drush::commandFactory();
+            foreach ($serviceCommandInfoAltererlist->getCommandList() as $altererHandler) {
+                $commandFactory->addCommandInfoAlterer($altererHandler);
+            }
+        }
+
         $serviceCommandlist = $container->get(DrushServiceModifier::DRUSH_CONSOLE_SERVICES);
         if ($container->has(DrushServiceModifier::DRUSH_CONSOLE_SERVICES)) {
             foreach ($serviceCommandlist->getCommandList() as $command) {
--- a/src/Boot/DrupalBoot8.php
+++ b/src/Boot/DrupalBoot8.php
@@ -199,6 +199,7 @@ class DrupalBoot8 extends DrupalBoot implements AutoloaderAwareInterface
             $commandFactory = Drush::commandFactory();
             foreach ($serviceCommandInfoAltererlist->getCommandList() as $altererHandler) {
                 $commandFactory->addCommandInfoAlterer($altererHandler);
+                $this->logger->debug(dt('Commands are potentially altered in !class.', ['!class' => get_class($altererHandler)]));
             }
         }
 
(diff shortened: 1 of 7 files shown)


weitzman:
Woot woot! Lets see what @greg-1-anderson has to say.

greg-1-anderson:
The code looks good and I am generally +1 on this. Could you be a little more specific about what you are doing with this? Are you just renaming / re-aliasing existing commands, as the tests are doing?

src/Boot/DrupalBoot8.php
```diff
@@ -192,6 +192,16 @@ public function bootstrapDrupalFull()
         // The upshot is that the list of console commands is not available
         // until after $kernel->boot() is called.
         $container = \Drupal::getContainer();
+
+        // Set the command info alterers.
+        if ($container->has(DrushServiceModifier::DRUSH_COMMAND_INFO_ALTERER_SERVICES)) {
+            $serviceCommandInfoAltererlist = $container->get(DrushServiceModifier::DRUSH_COMMAND_INFO_ALTERER_SERVICES);
+            $commandFactory = Drush::commandFactory();
+            foreach ($serviceCommandInfoAltererlist->getCommandList() as $altererHandler) {
+                $commandFactory->addCommandInfoAlterer($altererHandler);
```
greg-1-anderson: We should probably write a debug log message if a module adds one of these.

greg-1-anderson:
My big concern here is that folks could get really confused if one module is altering commands in core or another module that is not related (in the same suite as) to the module doing the alteration.

Could you write some documentation on how this feature could be used, and also strongly encourage folks to debug log when altering commands, so that this behavior is discoverable to anyone who does get confused. It won't be possible for us to enforce logging, of course, but we can at least encourage it.

It would also be good if you could inject a  logger into the woot command and debug log there, just so that the test code serves as a good example of how to alter commands. It isn't necessary to have an assert that ensures the log is printed, as long as it's there for illustrative purposes.

claudiu-cristea:
@greg-1-anderson, we are preparing to add migrate commands to Drush core in #3402. But we want to coordinate with `migrate_tools`/`migrate_plus` so that `migrate_plus`-specific migrations will still work. We will achieve this by removing the Drush commands from `migrate_tools` and adding command alters/replacements in `migrate_plus`. Now, commands are easy to replace, using hooks. For example, in `migrate_plus` we swap the Drush core migrate status command in this way: `@hook replace-command migrate:status`. But also the annotations need some changes. For example, on the same command, we have different output structure in Drush core command than in `migrate_plus`. The `@field-labels` annotation need to be swapped. So, `migrate_plus` module has to add some alters to some command info, too.

greg-1-anderson:
Perhaps some of these operations should happen automatically in annotated command. I haven't checked yet to see if this is possible / easy, but if other annotations in a `@hook command-replace` always completely overwrote the annotations in the hooked command, would that reduce the number of things you need to alter? Would you ever find it undesirable if `@hook command-replace` overwrote overlapping annotations?

claudiu-cristea:
I saw that array annotations, such as @option, @usage are merged, not replaced. The others are kept from the original command.

tests/resources/modules/d8/woot/src/WootCommandInfoAlterer.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Consolidation\AnnotatedCommand\CommandInfoAltererInterface;
+use Consolidation\AnnotatedCommand\Parser\CommandInfo;
+use Drupal\Core\Logger\LoggerChannelFactoryInterface;
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
+
+class WootCommandInfoAlterer implements CommandInfoAltererInterface, LoggerAwareInterface
+{
+    use LoggerAwareTrait;
```
greg-1-anderson: You can drop the LoggerAwareInterface / LoggerAwareTrait here. Drush will only inject its logger into command classes that implement these interfaces -- not any object at all. Since you are instantiating this class via the Drupal service file (drush.services.yml), then Drupal will inject your logger factory, so the code you have below is sufficient.

claudiu-cristea: OK, removed the trait and interface.

greg-1-anderson:
LGTM thanks for contributing this - should be very useful.

PR #3453, opened 2018-03-10
Make it easier to run contrib module Unish tests

claudiu-cristea:
Yes, this is the problem: How do we test Drush commands provided by modules?

This PR tries to make this process easier. See the test README.md for details.

EDIT: Actually this PR does two things:

1. Provides an official/canonical way to add modules to be tested to the codebase of SUT (via `tests/composer.json`). The benefit is that you can use all the composer features, like applying patches, or symlinking from a local dir.
2. Ensures that the PHPUnit test suite discovers the tests shipped with such modules. You don't have to invoke the test by passing the full path to test file but you can use the fancy `--filter=` & Co.

Note: I had to change the contrib modules location to `modules/unish`. Otherwise, in `modules/contrib` they are deleted on each test class run. See `\Unish\UnishTestCase::setUpBeforeClass()`.


claudiu-cristea:
@weitzman, this is ready for review. I tested it locally and, probably, adding an automated test would be a little tricky. It's easy to be tested manually by following the `tests/README.md` instructions. Also, it would be nice if Unish tests could be run by the d.o infrastructure so that we can ship modules with testable Drush commands.

greg-1-anderson:
Did you look at the tests for devel? https://cgit.drupalcode.org/devel/tree/drush/develDrushTest.php

Does that technique not working well for your tests?

claudiu-cristea:
@greg-1-anderson, yes but Devel is a Drush core dependency, so the module is added to codebase everytime SUT is build. But what about other modules exposing Drush commands? Then, I guess, you'll have to manually pass the path to test file when running `./unish.phpunit.php`. This PR is resolving the problem with modules that are not, normally, dependencies of Drush core. Actually this PR does two things:
- Provides a canonical way to add tested modules to the code base of SUT (via `tests/composer.json`)
- Ensures that the PHPUnit test suite discovers the tests shipped with the module.

claudiu-cristea:
@greg-1-anderson @weitzman I've updated the issue description to make clear what this PR brings. Let me know if there's anything else I can do here.

tests/README.md
```diff
@@ -14,3 +14,35 @@ Advanced usage
 - XML results: `unish.phpunit.php --filter=testVersionString --log-junit results.xml`
 - Build the SUT and run test suite (slower) - `unish.clean.php`
 - Install the SUT in a given folder - `UNISH_TMP=/path/to/folder php unish.sut.php`
+
+Testing contrib module commands
+---------
+
+**Important:** The Unish tests should be placed under `tests/src/Unish` in your module. But if you still prefer other location, adjust the settings in `tests/phpunit.xml`.
+
+1. Create a copy of `tests/composer.json.dist` as `tests/composer.json`:
```
weitzman: Lets clarify that this is a copy of vendor/drush/drush/tests/composer.json.dist to the same dir, and not anything the module. Also, this is pretty unexpected. Its rare for module devs to be editing inside vendor dir. There is a danger of losing it on rebuild.

tests/README.md
```diff
@@ -14,3 +14,35 @@ Advanced usage
...
+    ```
+    cp tests/composer.json.dist tests/composer.json
+    ```
+1. Add the modules containing the Drush commands you want to test as composer dependencies in `tests/composer.json`. For example:
+    ```
+    "require": {
+        "drupal/migrate_plus": "*"
+    }
+    ```
+    Tip: If you're currently designing the module commands and you want to test your work, you can install the module as a symlink to your local directory where the development is taking place:
```
weitzman: Thanks for including this useful tip ... Lets add a link to Composer's `path` repo docs.

unish.sut.php
```diff
@@ -75,5 +79,11 @@ function unish_setup_sut($unish_sandbox) {
     symlink("$target_dir/vendor", __DIR__ . '/vendor');
   }
 
+  // Create a symlink to modules directory in Drush root, so that PHPUnit is able to discover module Unish tests.
```
weitzman: This symlink is a bit unfortunate IMO. For one thing, it looks like it will add a code to PHPStorm project when working on Drush. Maybe we can avoid it?

tests/README.md
```diff
@@ -14,3 +14,35 @@ Advanced usage
 - XML results: `unish.phpunit.php --filter=testVersionString --log-junit results.xml`
 - Build the SUT and run test suite (slower) - `unish.clean.php`
 - Install the SUT in a given folder - `UNISH_TMP=/path/to/folder php unish.sut.php`
+
+Testing contrib module commands
+---------
+
+**Important:** The Unish tests should be placed under `tests/src/Unish` in your module. But if you still prefer other location, adjust the settings in `tests/phpunit.xml`.
```
weitzman: IMO we can omit the word 'Important'.

tests/README.md
claudiu-cristea: Well, when I work on commands provided by modules, I'm opening Drush as project. I think this is an advantage because **I don't need a full Drupal install**. The single thing is that I need an external workcopy of the module. Then I'm testing the Drush commands from that module using this procedure.

unish.sut.php
claudiu-cristea: Well... I have no idea how can I remove it because this line in `phpunit.xml.dist`:

```xml
<directory>../modules/unish/*/tests/src/Unish</directory>
```

Another option would have been to add a symlink directly in the SUT install as `/path/to/sut/drush-sut/vendor/drush/drush` but the problem is that `/path/to/sut/drush-sut/vendor/drush/drush` is a symlink and it's the same.

However, I see a big advantage to bring the modules code in PhpStorm. As I already told in other comment, you're able to design Drush commands in contrib module from the Drush project. 

I will think more on this but I don't see an alternative for discovering the tests from contrib. I don't think that `phpunit.xml.dist` knows to use/read env vars so that we could pass the SUT dir (?)

weitzman:
I wonder if we can remove test discovery and just rely on the user to point to her test directory like `unish.phpunit.php /path/to/modules/migrate_plus/tests/src/Unish`. That way `phpunit`will find the test class. unish. Also, we should consider making the composer.json changes automatically, adding in a path repository. So, developer doesn't have to edit anything in the drush tree.

weitzman:
This is just a bit too awkward for module authors IMO.

Issue #3330, opened 2018-01-26
Prompts for password for user postgres


jcandan:
### Problem/Motivation

Using drush with postgres repeatedly prompts for database password. It seems #63 has become an issue again. 

I believe the culprit is Drush 9.0.0, because it only just became an issue for me when `drupal-composer` updated to Drush 9.0.0 a couple days ago.

**To reproduce this problem**:

```
composer create-project drupal-composer/drupal-project:8.x-dev my_app --stability dev --no-interaction
cd my_app/web
../vendor/bin/drush si standard --db-url=pgsql://postgres:root@127.0.0.1:5432/my_app -y
```

### Proposed resolution

Not sure if the fix in #695 could address this again, but basically that fix was to set the PGPASSWORD environment variable to the `settings.php` database password.

geerlingguy:
I'm suddenly getting this as well, starting last night, on my Drupal VM automated builds, e.g.: https://travis-ci.org/geerlingguy/drupal-vm/jobs/333562827

```
TASK [geerlingguy.drupal : Install Drupal with drush.] *************************
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["/var/www/drupalvm/drupal/vendor/drush/drush/drush", "site-install", "standard", "-y", "--root=/var/www/drupalvm/drupal/web", "--site-name=Drupal", "--account-name=admin", "--account-pass=admin", "--db-url=pgsql://drupal:drupal@localhost/drupal"], "delta": "0:00:00.408622", "end": "2018-01-26 04:27:38.036190", "msg": "non-zero return code", "rc": 1, "start": "2018-01-26 04:27:37.627568", "stderr": " [error]  Failed to create database: Password for user drupal:psql: fe_sendauth: no password supplied ", "stderr_lines": [" [error]  Failed to create database: Password for user drupal:psql: fe_sendauth: no password supplied "], "stdout": "\n // You are about to DROP all tables in your 'drupal' database. Do you want to  \n // continue?: yes.                                                             ", "stdout_lines": ["", " // You are about to DROP all tables in your 'drupal' database. Do you want to  ", " // continue?: yes.                                                             "]}
```

geerlingguy:
From https://www.drupal.org/node/2146131, I've also found:

> You might want to try setting the environment variable PGPASSWORD and then passing a flag into your drush command line.
> 
> Assume Linux/OSX
> export PGPASSWORD=password;
> drush --extra=-w commands
> 
[...]

PR #3480, opened 2018-03-26
[#3330] Fix logic leading to creation of PGPASSFILE when connecting t…

nbertram:
As discussed in #3330 Drush 9 regressed the use of a PGPASSFILE environment variable to connect to PostgreSQL databases that require a password. This failure is quite crucial in Drush 9 because of the use of the 'psql' tool to perform more operations directly, so even things like drush status will fall into a password prompt with the current code.

The issue here is reversed logic in the SqlPgsql::createPasswordFile() method:

`if (null !== ($this->getPasswordFile()) && isset($dbSpec['password'])) {`

This will only create the password file if it is _not_ null, in other words if it's already existing. See below for why getPasswordFile() always returns null regardless, making this code block unreachable in any scenario.

If you contrast this line to Drush 8's (working) version:

`if (!isset($password_file) && isset($this->db_spec['password'])) {`

you can see that it's meant to be the other way around, creating the password file if it's not already set.

This Drush 8 line entered Drush 9 in b23f454 with a slight refactor:

`if (!isset($this->getPasswordFile()) && isset($dbSpec['password'])) {`

which was corrected to avoid isset() on a method return by 2d98048:

`if (null !== ($this->getPasswordFile()) && isset($dbSpec['password'])) {`

which accidentally reversed the logic of the first part of the statement, leading to the current code. Looking at the Travis config, it looks like pgsql tests with passwords are not performed, so that regression slipped through.

In addition to this logic issue, it appears getPasswordFile() will never return anything except null, because nothing sets the password_file member variable after writing the file. This was also the case in Drush 8, where the private variable $this->password_file actually has no references whatsoever. All code to get the name of the password file in both Drush 8 and Drush 9 actually calls createPasswordFile(), which returns the name. This means a new password file is created for every command, rather than reusing the existing one, because it can never detect that it has actually already created one. I suspect this came about because further back in Drush's history the local variable $password_file was possibly static.

I've fixed these two issues, and a third one where Drush reports that psql is not available because drush_program_exists() can't handle the prepended environment variable that command() adds:

`[warning] The command 'PGPASSFILE=/tmp/drush_42DoXS psql -q' is required for preflight but cannot be found. Please install it and retry.`

My approach there was to use a regex to remove prepended environment variables before running 'command' on $program. Given it appears the pgsql class is the only one in Drush core actually prepending environment variables in this manner, a less generic approach might be warranted to fix this. It might even be better to extend the way the database commands are executed to support passing environment variables through in the environment rather than the shell line, but I'll leave that call to more seasoned Drush contributors.

This PR might also supersede #2073, as I inadvertently fixed the same issue for pg_dump in my commit, coming to the same solution as that PR.

### DIFF
--- a/includes/filesystem.inc
+++ b/includes/filesystem.inc
@@ -216,6 +216,10 @@ function drush_mkdir($path) {
 function drush_program_exists($program) {
   if (drush_has_bash()) {
     $bucket = drush_bit_bucket();
+
+    // Remove environment variables (eg. PGPASSFILE=) before testing program.
+    $program = preg_replace('#^([A-Z0-9]+=.+? )+#', '', $program);
+
     // Dont't use drush_op_system() since we don't want output during tests.
     system("command -v $program > $bucket 2>&1", $result_code);
     return $result_code === 0 ? TRUE : FALSE;
--- a/src/Sql/SqlPgsql.php
+++ b/src/Sql/SqlPgsql.php
@@ -15,9 +15,8 @@ class SqlPgsql extends SqlBase
 
     private function createPasswordFile()
     {
-        $password_file = null;
         $dbSpec = $this->getDbSpec();
-        if (null !== ($this->getPasswordFile()) && isset($dbSpec['password'])) {
+        if (null == ($this->getPasswordFile()) && isset($dbSpec['password'])) {
             $pgpass_parts = [
             empty($dbSpec['host']) ? 'localhost' : $dbSpec['host'],
             empty($dbSpec['port']) ? '5432' : $dbSpec['port'],
@@ -34,10 +33,10 @@ class SqlPgsql extends SqlBase
                   $part = str_replace(['\\', ':'], ['\\\\', '\:'], $part);
             });
             $pgpass_contents = implode(':', $pgpass_parts);
-            $password_file = drush_save_data_to_temp_file($pgpass_contents);
-            chmod($password_file, 0600);
+            $this->password_file = drush_save_data_to_temp_file($pgpass_contents);
+            chmod($this->password_file, 0600);
         }
-        return $password_file;
+        return $this->password_file;
     }
 
     public function command()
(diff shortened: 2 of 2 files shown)


weitzman:
Thanks everyone. Hopefully @tstoeckler gets a chance to review this soon.

weitzman:
I went ahead and merged this as its causing some pain and can be changed later. Thanks all for the help. We are short on postgres maintainers at this time.

DhruveOnMars:
I just installed drush 9.3.0 and when I run `drush status` from the root of my Drupal installation, I get the same error: `[warning] The command 'PGPASSFILE=/tmp/drush_42DoXS psql -q' is required for preflight but cannot be found. Please install it and retry.`.
After that, it still gives me the relevant information.

I'm on a Mac, using pgsql. 

What can I do to fix this? Or should I use a different version of drush?

DhruveOnMars:
Hey @nbertram yeah I ran that and it says 9.3.0

I'll create a new Drupal installation and add drush to see if it's a problem I'm having globally or specific to this project.

I guess it shouldn't make too much of a difference what version of Drupal I'm using, but in case it does, I'm using 8.5.3

DhruveOnMars:
Hey @nbertram I created a new Drupal installation and installed drush. Before running Drupal, I ran `drush status` and it was fine, but after installing Drupal with a pgsql database, I ran `drush status` and I still got the same warning as before

nbertram:
@DhruveOnMars Could you verify that you do in fact have the psql command available in the shell you're running Drush from?

Reading the code in DrupalBoot.php, it looks like it's still going to report the command with the environment variable on it when it can't find psql. My patch only made it find psql correctly by removing the variable, but the error when it's not available is likely to still be the one with the variable on it...

Something like `psql -V` should output something along the lines of `psql (PostgreSQL) 9.5.12` if the command is there. If not, you might need to install a postgresql-client package or similar.

DhruveOnMars:
@nbertram thank you! It looks like I didn't have it available. I re-installed postgresql using brew, and now I don't get that warning when I run `drush status`

nbertram:
Glad that helped!

Yeah Drush 9 started shelling out to psql instead of bootstrapping Drupal for some operations, so that's a new requirement compared to Drush 8.

PR #3482, opened 2018-03-28
Remove outdated connection between name and machine name

### LINKED ISSUE #3481: Notice: Undefined index: name in InputHandler.php
When performing `drush generate test-unit` the following notice occurs:

Undefined index: name vendor/drush/drush/src/Commands/generate/Helper/InputHandler.php on line 102

```
            $default_machine_name = function ($vars) use ($modules, $root_directory) {
                $machine_name = array_search($vars['name'], $modules);
                return $machine_name ?: Utils::human2machine(isset($vars['name']) ? $vars['name'] : $root_directory);
            };
```

Chi-teck:
This should resolve #3481.

### DIFF
--- a/src/Commands/generate/Helper/InputHandler.php
+++ b/src/Commands/generate/Helper/InputHandler.php
@@ -97,9 +97,7 @@ class InputHandler extends BaseInputHandler
 
             // Machine name question.
             $questions['machine_name']->setAutocompleterValues(array_keys($modules));
-            $default_machine_name = function () use ($root_directory) {
-                return $root_directory ? Utils::human2machine($root_directory) : '';
-            };
+            $default_machine_name = $root_directory ? Utils::human2machine($root_directory) : '';
             $this->setQuestionDefault($questions['machine_name'], $default_machine_name);
 
         // Theme related generators.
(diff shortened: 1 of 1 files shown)


src/Commands/generate/Helper/InputHandler.php
```diff
@@ -89,17 +89,16 @@ protected function preprocessQuestions(array &$questions, $destination)
             // Name question.
             if (isset($questions['name'])) {
                 $questions['name']->setAutocompleterValues($modules);
-                $default_name = function ($vars) use ($root_directory) {
-                    return Utils::machine2human(isset($vars['machine_name']) ? $vars['machine_name'] : $root_directory);
+                $default_name = function () use ($root_directory) {
+                  return $root_directory ? Utils::human2machine($root_directory) : '';
```
weitzman: Is the anonymous function needed? How about
```
$default_name = $root_directory ? Utils::human2machine($root_directory) : '';
```

Chi-teck: Surely, it is not needed now.

src/Commands/generate/Helper/InputHandler.php
```diff
@@ -89,18 +89,15 @@ protected function preprocessQuestions(array &$questions, $destination)
             // Name question.
             if (isset($questions['name'])) {
                 $questions['name']->setAutocompleterValues($modules);
-                $default_name = function ($vars) use ($root_directory) {
-                    return Utils::machine2human(isset($vars['machine_name']) ? $vars['machine_name'] : $root_directory);
+                $default_name = function () use ($root_directory) {
```
weitzman: Looks like this anon function can be removed as well.

Chi-teck: Done.

PR #3484, opened 2018-03-28
Fix the way to create PostgreSQL database

gitsignore:
This pull request should make it possible to create a PostgreSQL database with Drush sql:create command.
In PostgreSQL backticks don't exist, so they must be replaced by quotes.

### DIFF
--- a/src/Sql/SqlPgsql.php
+++ b/src/Sql/SqlPgsql.php
@@ -76,7 +76,7 @@ class SqlPgsql extends SqlBase
     public function createdbSql($dbname, $quoted = false)
     {
         if ($quoted) {
-            $dbname = '`' . $dbname . '`';
+            $dbname = '"' . $dbname . '"';
         }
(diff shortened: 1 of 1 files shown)


weitzman:
Could anyone verify this PR?

gitsignore:
To verify :

1. Init a Drupal project (with Drush >= 9.0.0)
2. Configure it with PostgreSQL in the settings :
` $databases['default']['default'] = [
        'driver' => 'pgsql',
        'database' => 'your-db-name',
        'username' => 'your-db-username',
        'password' => 'your-db-password',
        'host' => 'localhost',
        'prefix' => '',
 ]; `
3. Use Drush (>= 9.0.0) command to create the database :
`drush sql:create`
4. Check in PostgreSQL that the database is created.

weitzman:
Merged, Its low risk. Thanks.

PR #3485, opened 2018-03-29
Fix drush_batch_command() unexpected output

phcorp:
This suppresses unexpected output for `src\Drupal\Commands\core\BatchCommands.php`.

It was already done there for `src\Commands\core\UpdateDBCommands.php`: https://github.com/drush-ops/drush/commit/4050545b447da01d1d6998ebdfbdbf7acb04b137

### DIFF
--- a/src/Drupal/Commands/core/BatchCommands.php
+++ b/src/Drupal/Commands/core/BatchCommands.php
@@ -2,6 +2,7 @@
 namespace Drush\Drupal\Commands\core;
 
 use Drush\Commands\DrushCommands;
+use Symfony\Component\Console\Output\OutputInterface;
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This doesn't look like a good way to disable output for a subcommand to me. I haven't tested this, but I would expect that `$this->output()->setVerbosity()` would have a lasting effect e.g. on any log output that happened after `drush_batch_command()` returns.

 `drush_batch_command()` is the legacy implementation from Drush 8. It seems that this needs to be modernized to take its own OutputInterface `$output` object so that we will have better control over its operation.

pfrenssen:
In the meantime this has been fixed in #4090.

PR #3488, opened 2018-03-30
fix: add missing return

gitsignore:
Fix missing return from this [commit](https://github.com/drush-ops/drush/commit/500ec257857ed04cf6c76569d6593df7bcdf7571)


weitzman:
Thanks.

PR #3489, opened 2018-03-30
Surface Drupal 8 Messages in the log.

### LINKED ISSUE #3486: How to print a message to the console from a module when called with drush 9
I made a custom module that implements hook_rebuild(). There's a drupal_set_message() in it. If I clear the caches from Drupal's UI, the message is shown. But it's not shown if I execute drush cr. Is there an easy way to print my message in both cases?
I tried both following commands with Drupal 8.5 and Drush 9.2.1, but I only get "[success] Cache rebuild complete."

`\Drupal::messenger()->addMessage($message); \Drupal::logger('my_module')->info($message);`

weitzman:
Fixes #3486


greg-1-anderson:
+1 LGTM

PR #3490, opened 2018-03-31
Add Language commands

Sutharsan:
Add drush commands to add and list (configurable) languages.
```
drush language:add nl,fr
drush language:add nl --no-translations
```
```
drush language:info
```

Testing drush locale commands will benefit from having drush commands to add and list languages of a Drupal installation. Testing locale related code in Drush 8 will benefit even more as there is no easy way to add a language in Drush 8. See #3251 and #3445.


Sutharsan:
The current code works and is ready for initial review.
Tests will follow.
Drush 8 code will follow after that.

tests/LanguageAddTest.php
```diff
@@ -0,0 +1,83 @@
...
+ *  @group slow
+ *  @group pm
+ */
+class LanguageAddCase extends CommandUnishTestCase
+{
+
+    public function testLanguageInfo()
+    {
+        $sites = $this->setUpDrupal(1, true);
+        $uri = key($sites);
```
weitzman: We now default uri to 'dev' so no need to set it anymore. And I think 'root' gets autodiscovered as well. Omit them both.

Sutharsan:
>Also, I'm hesitant to add commands to Drush8 at this point.
Having a language:add is beneficial to #3445 because it makes code testable. Let me know if it is yes or no.

weitzman:
What do you think about merging this into master and 8.x with the command hidden? We do that for other “internal” commands.

Sutharsan:
That is Ok with me. Did not know that option.

weitzman:
Merged, and hidden in a follow up commit.

Issue #2261, opened 2016-07-08
make-convert composer.lock --> make assumes package versions are in the "webflo" format


balsama:
packagist.drupal-composer.org converts standard drupal tags (e.g. 8.x-1.2) to 8.1.2. But the new packager.drupal.org drops the major digit (since it has to packagers, one for 7 and one for 8) and tacks a zero on to the end. E.g. 8.x-1.2 => 1.2.0.

make-convert assumes they are in the "webflo" format:
https://github.com/drush-ops/drush/blob/master/commands/make/make.drush.inc#L674

We currently use this command to generate makefiles for the D.O tarballs, but those makefiles will point to invalid versions of modules once we switch to the nee packager.drupal.org.

I'm not sure if this is something worth fixing since we really shouldn't be spending time supporting the legacy tarball. But I wanted to document it either way.

weitzman:
Thanks for documenting it, and for preferring the Composer way :)

greg-1-anderson:
Yes, this should probably be updated. Distributions are in a bad state right now, as they require drush make; however, drush make is deprecated, as it becomes untenable as soon as a project uses a module with Composer dependencies. This command also helps people transition from make to composer, so we should make sure it supports both formats.

balsama:
Just FYI. We (the developers of lightning) are actually no longer blocked on this. We implemented our own [lock->make converter](https://github.com/acquia/lightning/blob/8.x-2.x/package) because we had issues with how [make-convert pointed to the git repos instead of D.O](https://www.drupal.org/node/2827227).

jcnventura:
We can close this now.

PR #3499, opened 2018-04-04
Fix drush make-convert --format=composer

jcnventura:
This patch fixes the composer.json make conversion.

As the maintainer of the [7.x branch of drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project/tree/7.x), I really want to convince people to leave drush make behind and move over to composer.json. Even for Drupal 7 sites.

However, the current "drush make-convert --format=composer" has several flaws:
1. It still uses webflo's packagist.drupal-composer.org, which is no longer active.
2. It assumes Drupal 8 installer-paths.
3. It uses versioning system from webflo's packagist (see #2261)
4. It doesn't support drupal-composer/preserve-paths for D7 sites.

This patch fixes all the above, moving to the packages.drupal.org packagist.

### DIFF
--- a/commands/make/make.drush.inc
+++ b/commands/make/make.drush.inc
@@ -567,21 +567,89 @@ function drush_make_convert_make_to_composer($info) {
   $core_major_version = substr($info['core'], 0, 1);
   $core_project_name = $core_major_version == 7 ? 'drupal/drupal' : 'drupal/core';
 
-  // Add default projects.
-  $projects = array(
-    'composer/installers' => '^1.0.20',
-    'cweagans/composer-patches' => '~1.0',
-    $core_project_name => str_replace('x', '*', $info['core']),
-  );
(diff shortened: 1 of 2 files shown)


jcnventura:
Suggested labels:
area-make
drush8

weitzman:
Code looks good to me. There is a test fail - https://travis-ci.org/drush-ops/drush/jobs/362191867. You can ignore other fails in this run as they are unrelated.

jcnventura:
@weitzman Tests all pass. The test was expecting webflo's packagist version number.

There's a webflo's packagist composer.json + composer.lock file still in tests/makefiles, but I think those are examples to be used to convert back into a drush make format (hard to understand why someone would want to do that). I'm not touching that at the moment.

weitzman:
Thanks.

PR #3521, opened 2018-04-14
Add command to import .po files

Sutharsan:
This PR adds a command to import translation files similar to what can be done in the interface (at /admin/config/regional/translate/import)

This is part of an initiative to provide tools for handling interface translations as part of site deployment.
Translation export: #3253


src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +157,186 @@ public function update($options = ['langcodes' => self::REQ])
...
+     * @usage drush locale-import nl drupal-8.4.2.nl.po
+     *   Import the Dutch drupal core translation.
+     * @usage drush locale-import nl custom-translations.po --type=custom --override=all
+     *   Import customized Dutch translations and override any existing translation.
+     * @aliases locale-export
+     * @throws \Exception
+     */
+    public function import($langcode, $file, $options = ['type' => self::OPT, 'override' => self::OPT])
+    {
+        if (!file_exists($file)) {
```
TessaBakker: Use drush_file_not_empty() it checks for more than only existence.

TessaBakker: Maybe move this to a validate hook?

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -128,4 +157,186 @@ public function update($options = ['langcodes' => self::REQ])
...
+
+        $language = $this->getTranslatableLanguage($langcode, true);
+
+        $this->getModuleHandler()->loadInclude('locale', 'translation.inc');
+        $this->getModuleHandler()->loadInclude('locale', 'bulk.inc');
+
+        $translationOptions = _locale_translation_default_update_options();
+        $translationOptions['langcode'] = $language->getId();
+        $translationOptions['customized'] = $this->convertCustomizedType($options['type']);
+        $override = $this->convertOverrideOption($options['override']);
```
TessaBakker: I'm missing the override settings from /admin/config/regional/translate/settings

Sutharsan: All override options are covered by `override`. If no override is provided the ones from `_locale_translation_default_update_options()` are used.

Sutharsan:
@TessaBakker Did you have a change for another review?

ideb:
This looks very similar to the current locale-check / locale-update. How does this relate to this existing functionality?

TessaBakker:
@ideb locale:import can import local PO-files like the old D8 drush_locale_sync project

TessaBakker:
@ideb yes it does, but only from project, when they are defined in *.info.yml

locale:import can do more things like:
- importing po-files from any specified location
- importing custom translations and flag them different than your Drupal setting

Also this Drush 9 command will replace the Drush 8 command in [drush_locale_sync](https://drupal.org/project/drush_locale_sync).

Issue #3414, opened 2018-02-23
--tail option for drush watdchdog-show is missing in drush 9


zuernBernhard:
drush ws --tail 

                                                          
  [Symfony\Component\Console\Exception\RuntimeException]  
  The "--tail" option does not exist.

zuernBernhard:
Drush version : 9.1.0

greg-1-anderson:
Some existing Drush commands were simplified in Drush 9. I think that `--tail` would still be a welcome addition, if someone wanted to submit a PR.

thursdaybw:
I don't agree that removing valuable features counts as 'simplified'.

thursdaybw:
To get around this, at least on linux, this command achieves a similar result
`watch -n 1 drush ws`

olofjohansson:
Agreed. Tail is very valuable when developing since you can see hidden problems as they occur.

Tried https://stackoverflow.com/a/9574123 for macOS (similar to watch), but it doesn't cut it since it prints the 10 latest messages over and over again.

I'd really need tail in order to follow the log messages properly while developing.

greg-1-anderson:
Valuable features require maintainers. The way to keep the functionality you need working in Open Source software is to contribute pull requests to improve the state of affairs.

For those who strongly value or need `ws --tail`, Drush 8 is still maintained, and the option is working there. Folks can stay on `8` until the features they need are available in master.

serundeputy:
started on this @ DrupalConNA Nashville ... hope to finish this weekend.

serundeputy:
filed a PR: https://github.com/drush-ops/drush/pull/3523

@greg-1-anderson I'd appreciate a review; the Circle CI builds are failing, but it does not seem related to this PR; if it is perhaps you can help me understand how it does.

PR #3523, opened 2018-04-15
#3414: Adding --tail option to drush ws.

serundeputy:


weitzman:
Code looks good to me. Can anyone test it?

src/Drupal/Commands/core/WatchdogCommands.php
```diff
@@ -62,9 +64,30 @@ public function show($substring = '', $options = ['format' => 'table', 'count' =
             $row = $this->formatResult($result, $options['extended']);
             $table[$row->wid] = (array)$row;
```
rbayliss: This initial set of records just hangs out in memory until you exit the tail process. I think you should check for $tail earlier to avoid selecting them initially.

src/Drupal/Commands/core/WatchdogCommands.php
```diff
@@ -62,9 +64,30 @@ public function show($substring = '', $options = ['format' => 'table', 'count' =
...
             $this->logger()->notice(dt('No log messages available.'));
         } else {
+            if ($tail) {
+                // Query for new watchdog messages every 2 seconds.
+                while (true) {
+                    $rsc = Database::getConnection()->select('watchdog', 'w')
+                        ->fields('w')
+                        ->condition('wid', $last_wid, '>')
+                        ->orderBy('wid', 'ASC')
+                        ->execute();
```
rbayliss: This query doesn't respect the $where clause.  I'm not sure if it needs to, but I figured I'd point it out.

src/Drupal/Commands/core/WatchdogCommands.php
```diff
@@ -62,9 +64,30 @@ public function show($substring = '', $options = ['format' => 'table', 'count' =
...
+                    while ($result = $rsc->fetchObject()) {
```
rbayliss: Won't we be selecting all records since the beginning of time in the table?  That could be destructive.  I think we'd want to find the max WID in the table, and set that to the initial $last_wid.  Also, $last_wid isn't defined yet when we use it on the first trip through.

serundeputy: yes; not defined; but `ws` is all ready scoping it to 10 records.

amaisano:
Bump.

koppieesq:
@amaisano if you want to help, here are some suggestions:
1. take a look at the automated tests that failed.  If you understand why they failed, then submit a fix.  If you don't understand why, then try to research it.
1. If you still don't understand why the tests failed, you can at least paste in the error message from the test log.  That will be helpful too.
1. @rbayliss left a few comments back in April.  Someone needs to review those comments and decide whether we should make changes, or ignore the comments.  (I'm not suggesting anything about the quality of the comments; I simply haven't read them myself.)
1. This pull request hasn't been merged with `master` since April.  Try re-merging it (preferably with a rebase, to put the PR commits on top).  Maybe the tests will magically start passing?  It's been known to happen!
1. If you don't know how to use git, this is a great opportunity to learn.  It's an important skill if you want to work with Drupal, and it will open doors professionally.

Open source code relies on hundreds of volunteers who work for free.  Asking them to hurry up is more likely to get a resentful response.  Especially with the latest version of Drush, which was completed by a pretty small group of core contributors.  They removed a lot of features because no one was supporting those parts of the code.  All it takes is someone who's willing to put in the effort.  That could be you!

Bonus: Badcamp is this weekend.  It's a large regional Drupal convention in the San Francisco Bay Area.  If you can make the trip, there will always be someone in the Coders Lounge who would be happy to walk you through the steps.

Thanks for being involved!!!

novia713:
yes, please. i need this badly

weitzman:
It's in latest drush.

scubafly:
You can now use `drush wt`

PR #3528, opened 2018-04-17
Add field:create command

DieterHolvoet:
This PR adds a command to create fields, similar to the field-create command that was present in Drush 8. It has already been used in multiple projects at the company I work at, and is in my opinion stable enough to be considered to be merged in Drush.

### Tests
I do believe that this command should be backed by tests, but I have no past experience with PhpUnit and I can't seem to get the existing tests passing on my local machine, even without my changes.

### Naming
I named the class `FieldCreateCommands` instead of `FieldCommands` because I think future (smaller) field-related commands should be kept seperate from the field-create command. The class is already almost 650 lines long and adding other commands would make it a bit too cluttered.

### DIFF
--- a/src/Drupal/Commands/core/FieldCreateCommands.php
+++ b/src/Drupal/Commands/core/FieldCreateCommands.php
@@ -185,8 +185,7 @@ class FieldCreateCommands extends DrushCommands
                 $this->input->getOption('cardinality') ?? $this->askCardinality()
             );
 
-            if (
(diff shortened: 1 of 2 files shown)


weitzman:
Yes, lets add tests before merging. We have a problem with tests that just arose, as you can see on the results of this PR.

weitzman:
Lets skip this command, as we now have `generate field` which is somewhat similar.

PR #3529, opened 2018-04-19
fix Drush command sql-query with option "--db-prefix" in use

deminy:
Pull request #2882 works for Drupal 8 but breaks in Drupal 7. This new pull request makes it work well with both Drupal 8 and Drupal 7. Thanks for code review from @jorrit.


weitzman:
Instead, lets add a prefixTables() method in the version specific sql drivers (drupal 8, drupal 7, etc.)

PR #3531, opened 2018-04-20
 pm:enable option to skip translations

Sutharsan:
There are situations when a developer does not want to download and import interface translation files while enabling a module. For example:
- When enabling large number of modules
- When enabling a module to try it out. (Translation import can not be reverted)
- When working on a limited bandwith connection, e.g. while traveling

Usage example: 
```
drush en devel --skip-translations
```


src/Drupal/Commands/pm/PmCommands.php
```diff
@@ -79,7 +80,7 @@ public function enable(array $modules)
         if (!$this->getModuleInstaller()->install($modules, true)) {
             throw new \Exception('Unable to install modules.');
         }
-        if (batch_get()) {
+        if (batch_get() && !$options['skip-translations']) {
```
TessaBakker: Are you sure there are no other batches, beside translation import? Maybe removing the batch that holds the translation import is a possibility?

Sutharsan: No, I am not sure. Removing the batch is not possible. I think the chance is small that disabling the batch will cause an actual problem. We have been running Drupal 8 and Drush without batches untill only two weeks ago, when processing batches was introduced. I think there is no other batch function at enabling of modules since no one reported a problem with it.

The alternative is to disable the import function. I have coded it, but it is not a simple fix.

Sutharsan: @TessaBakker What to do with this remaining point?

weitzman:
I think its surprising to set persistent config is order to suppress translations for one request.

src/Drupal/Commands/pm/PmCommands.php
TessaBakker: Adding batches, to support batches and then skipping batches to support an edge case. It doesn't sound great. What can I expect as a module developer, will the "very important" hook_modules_installed() I wrote be supported by Drush or not?

Sutharsan:
>I think its surprising to set persistent config is order to suppress translations for one request.
I agree, reverting the workaround made on April 28.

Sutharsan:
Sorry. The fork that this PR was based on got lost. PR is continued at #4389

PR #3542, opened 2018-04-27
Fixing pm:security to properly identify security updates.

grasmash:


src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -78,7 +78,19 @@ public function security()
...
                         }
-                    } else {
+                    }
+                    // Compare exact versions that are insecure.
+                    elseif (preg_match('/^[[:digit:]](?![-*><=~ ])/', $conflict_constraint)) {
+                        $exact_version = $conflict_constraint;
+                        if (Comparator::equalTo($package['version'], $exact_version)) {
+                            $this->securityUpdates[$name] = [
+                                'name' => $name,
+                                'version' => $package['version'],
```
grasmash: This does not correctly indicate the newly available, secure version.

src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -112,4 +78,128 @@ public function suggestComposerCommand($result, CommandData $commandData)
...
+        }
+        // Compare exact versions that are insecure.
+        elseif (preg_match('/^[[:digit:]](?![-*><=~ ])/',
+            $conflict_constraint)) {
+            $exact_version = $conflict_constraint;
+            if (Comparator::equalTo($package['version'],
+                $exact_version)) {
+                // $version_parts = explode('.', $package['version']);
+                $version_parser = new VersionParser();
+                $constraints = $version_parser->parseConstraints($package['version']);
```
grasmash: This is a WIP. Need to calculate min-version.

src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -112,4 +78,151 @@ public function suggestComposerCommand($result, CommandData $commandData)
...
+     *
+     * @return array
+     *   An associative array containing name, version, and min-version keys.
+     */
+    public static function determineUpdatesFromConstraint(
+        $conflict_constraint,
+        $package,
+        $name
+    ) {
+        // Only parse constraints that follow pattern like "<1.0.0".
```
grasmash: If both `<8.5.1` and `8.5.0` conflict constraints are defined in drupal-security-advisories, it may cause multiple available updates to be registered. We need to somehow sort the list of conflicts such that the most relevant one is evaluated first, then we need to return and stop evaluating subsequent constraints.

weitzman:
Fingers crossed for @grasmash here.

grasmash:
I will revisit this next week.

grasmash:
I think that what remains to be done is:

1. Verify that upstream changes in drupal-security-advisories provide the expected new metadata.
2. Resolve the outstanding code comment.

PR #3545, opened 2018-05-03
Fix PHP 7.2 warning regarding count()

pfrenssen:
I am getting some warnings when building a legacy Drupal 7 project in PHP 7.2:

```
count(): Parameter must be an array or an object that implements Countable make.drush.inc:631
```


weitzman:
thanks

PR #3553, opened 2018-05-08
Continue #3166. Fix incorrect script name when Drupal uses a domain + subdir.

weitzman:


weitzman:
The test fails were unrelated so merging.

PR #3559, opened 2018-05-15
Update user cancel test to match upstream DCG changes

Chi-teck:
Content entity generator has got "translatable" option.
https://github.com/Chi-teck/drupal-code-generator/pull/10

### DIFF
--- a/tests/UserTest.php
+++ b/tests/UserTest.php
@@ -109,6 +109,7 @@ class UserCase extends CommandUnishTestCase
             'entity_type_id' => 'unish_article',
             'entity_base_path' => 'admin/content/unish_article',
             'fieldable' => 'no',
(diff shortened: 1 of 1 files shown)


Chi-teck:
Btw, in the next DCG release content entity generator will get new "bundle" option which will likely break this test again.
https://github.com/Chi-teck/drupal-code-generator/pull/12

greg-1-anderson:
Do you bump your major version for breaking changes?

Chi-teck:
@greg-1-anderson, this is not a breaking change. I suppose `drush generate` command sill works well. The test failed because it is executing `content-entity` command programmatically simulating user interaction . This is not something that should be considered as public API.

weitzman:
I upped composer.json and merged this.

Issue #3569, opened 2018-05-29
Drush 9 aliases path is wrong when using symlink


rotemreiss:
**Background:**
Versions - Drupal 8, Drush Launcher Version: 0.5.0, Drush Commandline Tool 9.3.0 .

We have three environments, and we are using Drush aliases manually and with our CI tools.
Recently rsync start failing when pushing files from staging (local) to production (`drush rsync @mysite.stglocal@mysite.prod`), and I couldn't figure out why, since mysite.site.yml wasn't changed recently.

I also couldn't reproduce the issue on my local, so I start debugging the issue directly on our staging with `vim` and `print_r()` (fun fun fun!), and found out that the problem is in `AliasRecord` in a method named `root`. That method is calling `FsUtils::realpath($root);` which in its turn returns the "realpath", but it obviously assumes that the real path is the same on both source and destination, which is not true on our case, since we have symlink on our staging environment (/var/www/html/site -> /var/www/html/stgsite).

I have prepared a patch to only check the real path for local aliases, hopefully, I didn't miss anything. Creating a PR in a few minutes, please review :)

Thanks.
Rotem

rotemreiss:
Fix has been merged.

PR #3570, opened 2018-05-29
#3569, Drush 9 aliases path is wrong when using symlink

rotemreiss:
https://github.com/drush-ops/drush/issues/3569


weitzman:
Looks reasonable. I'll leave it open for a bit in case @greg-1-anderson wants to chime in on the need to do that realpath() call.

PR #3573, opened 2018-06-01
Mark drush_backend_invoke_concurrent as deprecated

deviantintegral:
As mentioned in Slack at https://drupal.slack.com/archives/C62H9CWQM/p1527875860000040

I also fixed a typo in another docblock I saw as I was scrolling through.

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -621,7 +621,8 @@ function _drush_backend_adjust_options($site_record, $command, $command_options,
 /**
  * Execute a new local or remote command in a new process.
  *
- * n.b. Prefer drush_invoke_process() to this function.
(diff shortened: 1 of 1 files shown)


includes/backend.inc
```diff
@@ -621,7 +621,8 @@ function _drush_backend_adjust_options($site_record, $command, $command_options,
 /**
  * Execute a new local or remote command in a new process.
  *
- * n.b. Prefer drush_invoke_process() to this function.
+ * @deprecated as of Drush 9.4.0 and will be removed in Drush 10. Instead, use
```
deviantintegral: A guess on the version number, what should this be?

weitzman:
That version number looks good to me.

There is much more in this file thats deprecated but we can start here.

PR #3578, opened 2018-06-10
Support var-dumper 4.x

Chi-teck:
Symfony 4.1 comes with new cool feature called [VarDumper server](https://symfony.com/blog/new-in-symfony-4-1-vardumper-server). This is very useful especially for debugging non-HTML responses and unit tests. Unfortunately, it is not possible to take advantage of VarDumper server on Drupal installation witth Drush  because of dependency conflict.

As a bonus, fixing this would also resolve the issue with installing Drush after Drupal console (https://github.com/hechoendrupal/drupal-console/issues/3249), which is probably caused by this Composer bug https://github.com/composer/composer/issues/7329.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
     "symfony/event-dispatcher": "~2.7|^3",
     "symfony/finder": "~2.7|^3",
     "symfony/process": "~2.7|^3",
-    "symfony/var-dumper": "~2.7|^3",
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
All the tests are green, so I guess we don't have any problems from any sort of API change. Thanks.

Issue #3495, opened 2018-04-03
drush 9 site-install with drupalvm not finding sites-subdir on single site install


dgpalmer:
Hi,

I am using DrupalVM with Acquia BLT for my local development environment. This has been working well until I updated to BLT 9 & Drush 9 last week. When I try to build my VM or install drupal via drush, drush site-install cannot find the sites subdir and gives me an error upon installation.

For example with Drupal VM I get this:

TASK [geerlingguy.drupal : Install Drupal with drush.] *************************
fatal: [umgch]: FAILED! => {"changed": true, "cmd": ["/var/www/umgch/vendor/drush/drush/drush", "site-install", "standard", "-y", "--root=/var/www/umgch/docroot", "--site-name=UMG Content Hub", "--account-name=admin", "--account-pass=admin", "--db-url=mysql://drupal:drupal@localhost/drupal"], "delta": "0:00:00.205538", "end": "2018-04-03 17:54:38.336713", "msg": "non-zero return code", "rc": 1, "start": "2018-04-03 17:54:38.131175", "stderr": " [error]  Could not determine target sites directory for site to install. Use --sites-subdir to specify. ", "stderr_lines": [" [error]  Could not determine target sites directory for site to install. Use --sites-subdir to specify. "], "stdout": "", "stdout_lines": []}

If I try to install it via drush, see my drush status to see "sites/default" site path.
$ drush @umgch status
 Drupal version   : 8.5.1
 Site URI         : http://umgch.vm
 DB driver        : mysql
 DB hostname      : localhost
 DB port          : 3306
 DB username      : drupal
 DB name          : drupal
 Database         : Connected
 Drupal bootstrap : Successful
 Default theme    : stark
 Admin theme      : seven
 PHP binary       : /usr/bin/php7.1
 PHP config       : /etc/php/7.1/cli/php.ini
 PHP OS           : Linux
 Drush script     : /usr/local/bin/drush
 Drush version    : 9.2.3
 Drush temp       : /tmp
 Drush configs    : /var/www/umgch/vendor/drush/drush/drush.yml
                    /var/www/umgch/drush/drush.yml
 Install profile  : lightning
 Drupal root      : /var/www/umgch/docroot
 Site path        : sites/default
 Files, Public    : sites/default/files
 Files, Private   : /var/www/umgch/files-private
 Files, Temp      : /tmp

$ drush @umgch site-install
 [error]  Could not determine target sites directory for site to install. Use --sites-subdir to specify.

[...]

PR #3591, opened 2018-06-25
Fix si breaking with uri and default directory

neclimdul:
Fallback to default site directory in SubdirFromUri check


neclimdul:
followup to #3495 to fix default directory resolution failure when uri is provided.

greg-1-anderson:
lgtm thx

PR #3594, opened 2018-06-28
Fixes #3592: Prevent race conditions in cache clears.

### LINKED ISSUE #3592: Customize the drush cache directory - avoid rebuild collisions
Using Drush 9, if two processes try to clear the Drush cache at the same time, it creates a race condition that sometimes results in an error:
> Failed to remove directory "/home/foo/.drush/cache/factory": rmdir(/home/foo/.drush/cache/factory): Directory not empty

This is much more likely to happen on shared file systems such as Gluster or NFS, presumably because the cache clear takes 1-2 seconds and there's more time to produce a conflict.

This is easy to reproduce on a shared filesystem. Simply open two or three terminals and run `drush cc drush` in each one simultaneously.

I think it's a reasonable request for Drush cache clears to be multi-process safe, because these commands often happen automatically as part of a CI or deployment process, and these processes often _must_ run in parallel for efficiency.

If there's no way to make the cache clears more tolerant of slow file systems or running in parallel, it would at least help to have a configuration option to disable caching altogether, either permanently or on a request-by-request basis.

Note that other folks have reported the same error with Symfony cache clears, but it seems unlikely that Symfony is going to address this at the Filesystem level: https://github.com/symfony/symfony/issues/27578

danepowell:
Fixes #3592 

This is maybe more of a proof of concept, I put some todos in about ways that it could be improved if you approve of the general concept. It does work however.

### DIFF
--- a/src/Cache/FileCache.php
+++ b/src/Cache/FileCache.php
@@ -129,6 +129,15 @@ class FileCache implements CacheInterface
         $bin_dir = $this->cacheDirectory();
         $files = [];
         if (empty($cid)) {
+            // TODO: Extract this whole bit into a generic "remove" function so it can be reused below for specific bins.
+            $semaphore = Path::join($bin_dir, 'semaphore');
+            // See if our semaphore lock exists, and if so abort this cache clear in order to prevent race conditions.
+            // TODO: Check last file modification time, and if it's more than a few seconds assume the lock is stale and clear it.
+            if ($fs->exists($semaphore)) {
+                return;
+            }
+            // Write a semaphore lock file into the cache directory we are removing in order to prevent race conditions.
+            $fs->dumpfile($semaphore, null);
             $fs->remove($bin_dir);
         } else {
             if ($wildcard) {


src/Cache/FileCache.php
```diff
@@ -129,6 +129,15 @@ public function clear($cid = null, $wildcard = false)
         $bin_dir = $this->cacheDirectory();
         $files = [];
         if (empty($cid)) {
+            // TODO: Extract this whole bit into a generic "remove" function so it can be reused below for specific bins.
+            $semaphore = Path::join($bin_dir, 'semaphore');
```
greg-1-anderson: I don't think it works to put your semaphore in `$bin_dir`. You want the semaphore to be deleted at the end of the cache clear, and you cannot control what order the contents of `$bin_dir` will be deleted. The semaphore might even be deleted first.

You might consider using [flock](http://php.net/manual/en/function.flock.php) on a file parallel to $bin_dir (e.g. `$bin_dir.lock`)

greg-1-anderson:
I think the idea here is workable if the implementation is fleshed out. It might also be nice to make a small reusable locking class. I don't know if such a thing already exists.

src/Cache/FileCache.php
danepowell: Makes sense, although for what it's worth this does seem to work fine in practice. I think Symfony Filesystem takes a depth-first approach to deleting directories, so the semaphore file in the root-level cache directory is always one of the last to get removed.

danepowell:
Closing in favor of #3597

josephdpurcell:
Note: this approach did not solve the problem for me, see https://github.com/acquia/blt/issues/2957#issuecomment-505942336.

PR #3598, opened 2018-06-29
Highlight need to run composer in D7 manual install docs

wturrell:
You still need to run composer install at the very end of a Drush 7 manual install, to populate the vendor directory.  This is a bit confusing if you're scanning the install options quickly and have chosen not to use the "composer" methods.  Also, adds exact error message displayed, to aid those searching.


weitzman:
thats already mentioned later. no need to duplicate imo.

PR #3600, opened 2018-07-02
Canonicalize directory

### LINKED ISSUE #3596: config-status is not reflecting config splits
**Versions used**
Drupal version   : 8.5.4
Drush version    : 9.3.0

**Steps to reproduce**
- Install config_split module.
- Set up a config split  with the name `local` and add the devel module to it. (in my example I also added the chosen module).
- Export the config split configs `drush csex local`
- Export the rest of the configs `drush cex` and confirm the devel related configs are not exported in the default directory.
- Check if all configs are exported well by importing again `drush cim`
- Check the status of configs with `drush config-status`

**Results**
image
image

**Expected Results**
I expect `drush config-status` to reflect the actual status of config export, like seen in `admin/config/development/configuration`, taking into account the config splits that are active, hence in this example resulting in `No differences between DB and sync directory.`

Chi-teck:
This should fix #3596.
I've added _Path::canonicalize()_ as it is done in `drush cim` command.
https://github.com/drush-ops/drush/blob/9.3.0/src/Drupal/Commands/config/ConfigImportCommands.php#L168

### DIFF
--- a/src/Drupal/Commands/config/ConfigCommands.php
+++ b/src/Drupal/Commands/config/ConfigCommands.php
@@ -340,7 +340,7 @@ class ConfigCommands extends DrushCommands
      */
     public function getStorage($directory)
     {
-        if ($directory == \config_get_config_directory(CONFIG_SYNC_DIRECTORY)) {
+        if ($directory == Path::canonicalize(\config_get_config_directory(CONFIG_SYNC_DIRECTORY))) {
             return \Drupal::service('config.storage.sync');
         } else {
             return new FileStorage($directory);


Chi-teck:
Currently at least three command files related to configuration use the same code to get config storage service.
```
if ($directory == \config_get_config_directory(CONFIG_SYNC_DIRECTORY)) {
    return \Drupal::service('config.storage.sync');
} else {
    return new FileStorage($directory);
}
```
We may need to refactor them, to use trait or base class to reduce code duplication.

greg-1-anderson:
Can we inject `CONFIG_SYNC_DIRECTORY` via Drupal DI?

Chi-teck:
Symfony 3.2 [allows this](https://symfony.com/blog/new-in-symfony-3-2-php-constants-in-yaml-files) but Drupal serializer does not pass the required `SymfonyYaml::PARSE_CONSTANT` flag to Yml parser.

greg-1-anderson:
OK, that's fine. To answer the question you asked, though, I'm +1 on a trait over a base class here (unless we feel this is likely to be of such wide use that we want to move it to an existing base class).

weitzman:
Merged. A trait would still be welcome to centralize the duplicated code.

PR #3602, opened 2018-07-03
Site alias location filters

greg-1-anderson:
Allow site aliases to be referred to by a 'location' component of the alias name that is based on the name of the folder the alias file is stored in.

See the PR at https://github.com/consolidation/site-alias/pull/1

### DIFF
--- a/examples/example.site.yml
+++ b/examples/example.site.yml
@@ -71,6 +71,25 @@
 #       - /etc/drush/sites
 # @endcode
 #
+# It is also possible to create a named location that contains a group
+# of related site aliases. For example, if you had a number of sites
+# for local schools, you could define a location named "schools":
+#
+# @code
+# drush:
+#   paths:
+#     alias-path:
+#       - '${env.home}/.drush/sites/schools'
+# @endcode
+#
+# Site aliases stored in this directory may then be referenced by its
+# full alias name, including its location, e.g.:
+#
+#   $ drush @schools.example.dev
+#
+# Such alias files may still be referenced by their shorter name, e.g.
+# `@example.dev`.
+#
 # If you run the command `drush core:init`, this configuration will
 # be done for you automatically.
 #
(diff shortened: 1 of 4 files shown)


greg-1-anderson:
Motivation:

1. The ability for an isp to put all of their aliases in a group so that users can list them via `drush sa @isp` is very important.
2. If a user is moving a site `foo` from one isp to another, they are likely to also name the site `foo` in the target location. If the isp's are creating the site aliases, it would be very inconvenient for the user if both sites could not be named `foo`, as they will probably create the new site before discovering the conflict.

Beyond these use cases, maintaining the user experience of being able to address a site as either `@isp.foo.dev` or `@foo.dev` is important. Naming all of the aliases `isp-foo` would prevent the use of the short form.

See the above-referenced PR for implementation. Unlike the previous "alias groups" implementation, this version does not add much additional complexity to the site alias package.

examples/example.site.yml
```diff
@@ -71,6 +71,25 @@
...
+#
+# Site aliases stored in this directory may then be referenced by its
+# full alias name, including its location, e.g.:
+#
+#   $ drush @schools.example.dev
+#
+# Such alias files may still be referenced by their shorter name, e.g.
+# `@example.dev`.
+#
 # If you run the command `drush core:init`, this configuration will
```
weitzman: this sentence is awkwardly placed. i'm not sure what 'this' refers to.

greg-1-anderson:
Yeah, checking aliases into the site requires a local working copy of the sites codebase to work. The use case of not having a local working copy is more of a convenience for "drush for isps". The "Download Drush Aliases" button is popular.

I'll fix up the text in example.site.yml that follows the text I added.

greg-1-anderson:
I updated this PR to use site-alias:^1, and also corrected the documentation.

Note that users do not get the `@location...` feature unless they opt-in to it by adding an alias file location to their configuration text. I have a suggestion related to that. We could reduce the `example.site.yml` file to just the canonical use-case, and then put more advanced topics in `advanced.site.yml` or something like that?

If we do that, then is adding a global `@foo.dev` for your local working copy part of the canonical use-case? Doing that requires adding a location to your drush.yml config file. Maybe enough folks want to do that that it should go in the canonical documentation -- or maybe we want to keep the canonical documentation literally as short as possible, and put everything else in the "advanced" file.

weitzman:
I think we should try to keep one example file. Lets try over time, in the same file, to fence off the Advanced info.

greg-1-anderson:
I'll merge here for now and try adding an "ADVANCED" section at the bottom later.

PR #3603, opened 2018-07-04
Extend the life of the global Drush 8

greg-1-anderson:
**THESIS**: Drush should continue to support Drupal sites that are **NOT** managed by Composer until such a time as Drupal itself requires that all sites be managed by Composer.

**GIVEN**: Drush 9 is not suitable for non-Composer-managed sites. While it is possible (although deprecated) to run Drush 9 via a global install, Drush 9 does not have access to the `pm-*` commands. It might be possible to use a global Drush 9 and install/update modules through the Drupal admin web UI; however, lack of a CLI mechanism for managing code is antithetical to Drush charter, so this option should not be considered viable.

**THEREFORE**: Drush 8 support for Drupal 8.4.x, 8.5.x, and 8.6.x at a minimum remains important.

This PR is a continuation of https://github.com/drush-ops/drush/pull/2787. It uses the same strategy that Drush 9 utilizes during the early bootstrap, and postpones using any dependencies until the target Drupal site is identified. At that time, Drush loads the autoload file for Drupal and gives it priority, preventing any class mixing / dependency hell. Drush 8 is even better at this, because it uses fewer external dependencies than Drush 9 does.

It remains possible to use Drush 8 as a site-local Drush inside a Composer-managed Drupal site.

**PROPOSAL**

- Merge this PR and tag it as Drush 8.2.0 when the next 8.x release is made.
- Mark Drupal 8.4.x and 8.5.x as supported by Drush 8.2+ in the documentation.
- Declare the aspirational goal that Drush 8 support will continue (e.g. to Drupal 8.6.x) until Drupal requires Composer to install any contrib module.
- Drush 8 remains in "maintenance fixes only".

**NON-GOALS**

- Non-Composer users remain second-class citizens. We make no provisions to allow for the use of modules that require Composer (out-of-scope for Drush).
- Drush make remains deprecated.
- It remains impossible to use a global Drush with any Composer-managed site that contains a site-local Drush. The site-local Drush must always be used if present; the user must remove the site-local Drush in order to use a global Drush.


joachim-n:
Would it not be less maintenance work to port the PM-* commands to Drush 9, then end-of-life Drush 8 and so no longer have to maintain it? Or are there issues with running Drush 9 on older versions of Drupal 8?

greg-1-anderson:
I would be +1 on porting pm-* commands to Drush 9 if we had someone to do the work and maintain pm going forward. This is definitely more work than continuing to maintain Drush 8 with critical fixes only. Old versions of Drupal are not a huge concern for Drush 9. Drush 9 is "happens to work" on Drupal 8.3.x, and supported for Drupal 8.4.0+. I don't think we need to go further back than that.

There are other issues in addition to the pm-* commands, though. There are a number of Drush 8 modules that have Drush commands that have not been ported yet. Progress is being made in this area for sure, but we're not there yet. (I should survey contrib on drupal.org and see where we're at with this.) A bigger issue is that the distribution packaging infrastructure on drupal.org is still based on drush make; I'm not sure when this will be unnecessary for Drupal 8 distributions. Some distributions are hosting on GitHub + packagist, as the drush-make-based packaging on drupal.org does not support Composer yet.

Clearly, we cannot continue Drush 8 forever and ever. At some point we're going to have to say "look, this old stuff doesn't work any more, port to Drush 9 or go home." The question is, when do we do that? If Drush 8 stopped working in the Drupal 8.6.x timeframe, then Pantheon would just be broken for a large portion of our customers (for more reasons than just pm-*). From my perspective, it simply wouldn't be acceptable to say "your site isn't supported without Composer" until Drupal itself is requiring Composer to run (at least for installing contrib modules). Until then, I think it's important, and not too much work, to support folks with legacy ties to Drush 8.

weitzman:
Lots of kudos for the engineering here, and for the clear writeup. Love your citation of the Drush charter. Your words summoned a document that always existed but we've never seen or discussed.

Sadly, I do not support the PR for a few reasons ...

1. Practically speaking ...
    1. I find the changes in this PR to be too deep for a branch in maintenance mode. I know that we aren't touching the commands at all, but I still think this is too much. 
    1. We have years of neglect with make and pm commands. And now the rest of the commands join them in neglect, as I'm more motivated to work on Drush9. I would much rather see new sites using Drush9, and I think this PR gives conflicting signals on that front. I have seen your suggestion to keep Drush8 in maint mode but that doesn't allay my concern about signaling.
    1. This also gives confusing signaling for [drush-launcher](https://github.com/drush-ops/drush-launcher) which is our recommended global launcher.
1. Philosophically speaking, I don't exactly concur with the Thesis ... 
    1. I'm disappointed by the pace of Drupal core development, and especially by its near ignorance of dependency management. A movement is afoot now to improve Composer treatment, but its late, and it doesn't go far enough (perhaps I'll say more on that topic elsewhere).
    1. Drush9's product approach is a great model IMO. It accepts the Composer project as the master, and only asks to be included as a dependency. Further, Drush leaves all codebase management to Composer and Git. 
   1. This product approach reduces our surface, which vastly reduces our maintenance burden. Thats my main gripe with Drupal core - it still wants to serve too many masters. Product management is primarily about choosing which masters to serve and which to ignore.
   1. Drush is a not just a product but also my advocacy platform. I'm trying to nudge the Drupal community toward the above world-view. This PR is antithetical to that goal, as it extends and legitimizes a model that I find problematic (see the deeply problematic [Composer Manager](https://www.drupal.org/project/composer_manager)).
    1. I have confidence that the vast majority of Drush users are capable of learning Composer. For starters, they have already mastered Drush, a more niche and less focused CLI tool. And most of these folks have also mastered Git in the past few years. Composer is a breeze compared to Git. Of course some folks use Drush and Git but struggle with Composer. But I think enough of them can get over the hump that I'm not going dedicate resources to the few that cant.

Perhaps this PR can live on as a well known branch/fork. I can see ISPs using it.

greg-1-anderson:
> Your words summoned a document that always existed but we've never seen or discussed.

Not sure what you mean by "always existed" -- it seems that every maintainer might have a slightly different notion of what the charter should be, but as far as I know this is, as you said, something we have not yet seen in print. This is why I started with a thesis rather than a statement of fact. Perhaps a paragraph or two in the Drush documentation would be helpful.

I agree with most of your bullet points above. I strongly agree that Composer-managed Drupal sites is the direction everyone should be moving towards. There is no good reason why an individual Drupal coder / site builder who is already using Drush cannot learn to use Composer for their new sites.

However, the reason that maintaining Drush 8 is useful for ISPs is that ISPs have customers. A lot of an ISPs customers have existing sites that are clipping along just fine, and don't / shouldn't need an engineer to rework them as long as they have someone who can push a button to do the security updates.

Drupal core maintainers have decided that Drupal 8.6.x and Drupal 8.7.x and perhaps one or several more releases after that will continue to support tarball downloads and the non-Composer workflow. It's unclear when the current work to support a workflow similar to `drupal-composer/drupal-project` will become the preferred mechanism for installing and managing Drupal sites on drupal.org. While it is certainly appropriate for Drush maintainers to advocate the right way to do things, and steer the future development of Drupal core to more maintainable technologies, and this advocacy can be a net positive for the community, to me, taking the stance that non-Composer users cannot apply security updates for their existing Drupal 8 sites as a supported (maintenance) workflow (for >1yr) is bad for the Drupal community as a whole and will only lead to more cli fragmentation.

It is therefore disappointing to consider forking as the solution here. If this is the strategy that we're going to move forward with, we need to decide what branch name it should live in. I would propose using `8.2.x` for this purpose. It would be best if the forked Drush repo could be identical to the official Drush repo save for the addition of release tags.

weitzman:
> always existed

I just mean that we have all carried shared notions about what drush is meant to do. I love calling that a charter, whether or not it manifests as an actual document.

I propose using a branch called 8.x-isp. That branch can live in this repo. I am not a fan of calling it 8.2.x.

PR #3606, opened 2018-07-05
Set maintenance mode to update when running updates

### LINKED ISSUE #3604: Drush should set MAINTENANCE_MODE constant when running updates
See https://www.drupal.org/project/drupal/issues/2910353

That is a critical bug that results in data loss by for example overwriting translations. Drupal will start to throw an exception instead of overwriting the data. To avoid that, it relies on the MAINTENANCE_MODE in the latest patch to never load overrides then.

Drush does not set that constant anymore for D8 and it should do so again.

Berdir:
This is the PR for 9.x/master

Unlike 8.x, post updates are unfortunately not re-using the code from core, so we have to duplicate it there.

Fixes #3604

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -164,6 +164,11 @@ class UpdateDBCommands extends DrushCommands
     {
         $function = $module . '_update_' . $number;
 
+        // Disable config entity overrides.
+        if (!defined('MAINTENANCE_MODE')) {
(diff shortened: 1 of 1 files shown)


weitzman:
Can you say a bit more? We already wrap the update process in `\Drupal::service('state')->set('system.maintenance_mode', true);`. Thats not enough?

weitzman:
Also, can you elaborate on which sort of config entity overrides you are referring to? how do they cause troubles?

PR #3608, opened 2018-07-05
Fix getenv validation for COLUMNS variable.

heilop:


greg-1-anderson:
What version of PHP are you using? `?:` has been valid since PHP 5.3, so I don't see the motivation here. What sort of a problem are you seeing?

heilop:
I'm using PHP 7.1.19 version. But I think it was my wrong.

greg-1-anderson:
OK np

PR #3609, opened 2018-07-09
Throw a more relevant exception in the bootstrap phase

claudiu-cristea:


greg-1-anderson:
LGTM

src/Boot/BootstrapManager.php
```diff
@@ -385,26 +386,22 @@ public function bootstrapValidate($phase)
      * @param \Consolidation\AnnotatedCommand\AnnotationData $annotationData
      *   Optional annotation data from the command.
      *
-     * @return bool
```
weitzman: we should keep returning TRUE upon success, for backward compat. can be marked as deprecated.

PR #3612, opened 2018-07-09
Improve site alias example documentation.

greg-1-anderson:
The primary goal of this rewrite is to pull the canonical site alias example up to the top of the example file, so that beginning users can read just the first section and get the overall idea on how to use the recommended workflow for site alias files that are local to a single site.

More advanced workflows and use cases are pushed down to the lower sections of the document.

### DIFF
--- a/examples/example.site.yml
+++ b/examples/example.site.yml
@@ -1,18 +1,87 @@
 #
 # Example of valid statements for an alias file.
+
+# Basic Alias File Usage
 #
-# To convert legacy alias (*.aliases.drushrc.php) to yml, run the
-# site:alias-convert command.
+# In its most basic form, the Drush site alias feature provides a way
+# for teams to share short names that refer to the live and staging sites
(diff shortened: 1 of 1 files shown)


examples/example.site.yml
```diff
@@ -1,66 +1,74 @@
...
-# Drush site aliases always contain one or more environments;
-# for example, a site may have "dev", "test" and "live"
-# environments. The default environment is "dev"; the dev
-# environment of a site named "example" may therefore be
-# referred to as either "@example.dev" or "@example".
+# Following these steps, a cache-rebuild on the live environment would be:
 #
-# A canonical alias named "example" that points to a local
-# Drupal site named "http://example.com" looks like this:
+#   $ drush @live cache-rebuild
```
weitzman: lets use : in command name.

greg-1-anderson: Muscle memory from the ancient past...

Issue #3610, opened 2018-07-09
custom commands are no longer getting picked up


joachim-n:
I've just updated to Drush 9.3.0.

My custom command is no longer getting picked up.

It's available on Packagist as drupal-code-builder/drupal-code-builder-drush.

Composer installs the package at: web/drush/contrib/drupal-code-builder-drush

In that, there is a class:

```
namespace Drush\Commands\CodeBuilder;

use Consolidation\AnnotatedCommand\AnnotationData;
use Consolidation\AnnotatedCommand\CommandData;
use Consolidation\AnnotatedCommand\CommandError;
use Drush\Commands\DrushCommands;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;

/**
 * Provides commands for generating code with the Drupal Code Builder library.
 */
class CodeBuilderCommands extends DrushCommands {
```

Drush list doens't pick this up.

weitzman:
See bottom of the release notes - https://github.com/drush-ops/drush/releases/tag/9.3.0. We made a change in order to have code match the docs.

greg-1-anderson:
Sorry for the breaking change, but the published implementation was a bug. 😞

joachim-n:
> Authors of Drush9 global commands (i.e. commands that don't live within a module) will want to make sure they use the namespace Drush/Commands and not the briefer Drush.

So does that mean it *must* be Drush\Commands\MyClass? It can't have an extra namespace like mine does?

I have pointed out before how this is a flaw, as it means that all custom commands have to share the same namespace for their classes, and thus have to resort to pseudonamespacing their short classnames with a package prefix to avoid potential clashes!

greg-1-anderson:
I remember that old discussion, but I don't think it ever reached any conclusion.

If your source tree looks like this:
```
src/
└── Commands
        └── Foo
            ├── MyClass.php
            └── PolicyCommands.php
```
Then in this instance your namespace would be `Drush\Commands\Foo`. This is not ideal.

As I mentioned previously, Robo has a Composer-based solution for this that allows for flexible namespaces, but Drush has not adopted this mechanism. We probably don't want lots of different ways to create global extensions, as these are less common.

I agree that the current solution is not great, but I'm not sure what the best way to allow dynamic namespaces should be. Suggestions and PRs welcome.

joachim-n:
> Then in this instance your namespace would be Drush\Commands\Foo. This is not ideal.

[...]

Issue #3566, opened 2018-05-24
My global command is no longer found with 9.3.0


jurgenhaas:
First of all, the namespace in the global command is `Drush\Commands` and it lives in `PROJECT_ROOT/drush/contrib/dev_modules/src/Commands/DevModulesCommands.php`

In the drush.yml I have defined the include path and when running with debug I can see that the path is included:

```
drush dev --debug
 [preflight] Redispatch to site-local Drush: /var/www/html/vendor/drush/drush/drush.
 [preflight] Config paths: /etc/drush/drush.yml,/var/www/html/vendor/drush/drush/drush.yml,/var/www/html/drush/drush.yml
 [preflight] Alias paths: /var/www/html/web/drush/sites,/var/www/html/drush/sites
 [preflight] Commandfile search paths: /var/www/html/vendor/drush/drush/src,./drush/contrib
 [debug] Bootstrap further to find dev
```

Still, the command isn't found. The module in question is https://www.drupal.org/project/dev_modules

Can you please help me to understand what's wrong?

weitzman:
global drush commands no longer distribute as modules. they usually are just placed in a /drush dir. if you want to distrbute one, you would do so as a composer package thats of type=drupal-drush.  if the user has https://github.com/drupal-composer/drupal-project/blob/8.x/composer.json#L68 then it should get downloaded to /drush/contrib/[name].

our generator for drush commands currently assumes your command will ship in a module. perhaps thats what tripped you up.

jurgenhaas:
Thanks @weitzman, the way this "module" is defined now has been working up until drush 9.2 just fine and we came up with the workaround with the infrastructure team to make it work although it's not really a product.

I'm happy to change this to a composer package. Can you explore the structure inside /drush/contrib/[name] then?

weitzman:
Yes, we changed this in 9.3, as the existing behavior was a bug. I tried to note this in the release notes but I was still fuzzy on details.

This is working for me:

Filename: `$PROJECT_ROOT/drush/Commands/dev_modules/DevModulesCommands.php`
Namespace: `Drush\Commands\dev_modules`

ping @greg-1-anderson in case I've missed a nuance here. we do need to clarify docs at http://docs.drush.org/en/master/commands/#global-drush-commands. For example, the above namespace is longer than the one mentioned there.

greg-1-anderson:
[...]

PR #3615, opened 2018-07-11
Fixed global commands not being discovered where Composer installs th…

joachim-n:
See #3610

### DIFF
--- a/src/Config/ConfigLocator.php
+++ b/src/Config/ConfigLocator.php
@@ -413,6 +413,9 @@ class ConfigLocator
     /**
      * Get the commandfile paths according to preflight arguments.
      *
+     * This searches for commandfiles in '$root/drush/' and
+     * 'dirname($root)/drush'.
+     *
      * @param $preflightArgs
      *
      * @return array
@@ -457,19 +460,39 @@ class ConfigLocator
     }
 
     /**
-     * Return all of the commandfile paths in any '$root/drush' or
-     * 'dirname($root)/drush' directory that contains a composer.json
-     * file or a 'Commands' or 'src/Commands' directory.
+     * Return all of the commandfile paths in the given directories.
+     *
+     * Within each given directory, a commandfile may exist in any of the
+     * following structures:
+     *  - contrib/$package/Commands/$commandfile
+     *  - contrib/$package/src/Commands/$commandfile
+     *  - Commands/$package/Commands/$commandfile
+     *  - Commands/$package/src/Commands/$commandfile
      */
     protected function getSiteCommandFilePaths($directories)
     {
         $result = [];
 
-        $directories = array_filter($directories, 'is_dir');
-
         if (empty($directories)) {
             return $result;
         }
+
+        // In each drush directory, allow a command-providing package to be in
+        // the following subdirectories patterns. This is due there being
+        // different variants of Composer installer-paths configuration for
+        // "type:drupal-drush" packages.
+        $subdirectories = [
+            '/contrib',
+            '/Commands',
+        ];
+        $search_directories = [];
+        foreach ($directories as $directory) {
+            foreach ($subdirectories as $subdirectory) {
+                $search_directories[] = $directory . $subdirectory;
+            }
+        }
+
+        $search_directories = array_filter($search_directories, 'is_dir');
 
         // Find projects
         $finder = new Finder();
@@ -477,8 +500,8 @@ class ConfigLocator
             ->ignoreUnreadableDirs()
             ->followLinks()
             ->path('#src/Commands$|Commands$#')
-            ->in($directories)
-            ->depth('<= 3');
+            ->in($search_directories)
+            ->depth('<= 2');
 
         foreach ($finder as $file) {
             $result[] = dirname($file->getRealPath());
--- a/src/Config/ConfigLocator.php
+++ b/src/Config/ConfigLocator.php
@@ -475,7 +475,8 @@ class ConfigLocator
         $finder = new Finder();
         $finder->directories()
             ->ignoreUnreadableDirs()
-            ->path('#^src/Commands$|^Commands$#')
+            ->followLinks()
+            ->path('#src/Commands$|Commands$#')
             ->in($directories)
             ->depth('<= 3');
 


greg-1-anderson:
I'm really unsure about why we have both `depth('<=3')` AND `^` in the path regex.

I'm ambivalent about all of the file searching here.  Maybe we should call the finder twice, once at `drush` and `web/drush` with a depth of one, and once at `drush/contrib` and `web/drush/contrib` with a depth of 2 or 3.  That way we match what Composer is currently doing, and we don't waste time looking for `Commands` directories inside `drush/sites` et. al.

richardbporter:
I've got this issue as well and would love to see this (or some variation) merged in. 

What is left to do here - explicitly call the finder twice and update docs?

greg-1-anderson:
The main requirements are 1) passing tests and 2) no excessive searching. If someone has created `DRUPAL_ROOT/drush/vendor` or has put npm or other build results somewhere inside of `drush`, while this is not recommended, it should not cause Drush to search through those directories for commandfiles. We can have some deep searching in well-known locations like `DRUPAL_ROOT/drush/commands`, but we should be more conservative when searching at `DRUPAL_ROOT/drush`.

richardbporter:
Actually, I'm just going to use the new namespaces recommended in the docs: http://docs.drush.org/en/master/commands/#global-drush-commands 

I think that works better anyway since autoloading classes in global Drush commands was problematic, for me at least. Thanks!

joachim-n:
> Actually, I'm just going to use the new namespaces recommended in the docs: http://docs.drush.org/en/master/commands/#global-drush-commands

Do those work though? The point of this issue is that commands aren't being discovered.

```
Nested (e.g. Commandfile is part of a Composer package)

    Filename: $PROJECT_ROOT/drush/Commands/dev_modules/ExampleCommands.php
    Namespace: Drush\Commands\dev_modules
```

Can you confirm if the above works?

richardbporter:
Yes, that works for me with Drush 9.4.0. What didn't work for me was nesting further, like under a src/ directory:
```
Filename: $PROJECT_ROOT/drush/Commands/foo/src/FooCommands.php
Namespace: Drush\Commands\Foo
```

joachim-n:
```
    Filename: $PROJECT_ROOT/drush/Commands/dev_modules/ExampleCommands.php
    Namespace: Drush\Commands\dev_modules
```

Oh wait -- does 'dev_modules' in the filename need to match 'dev_modules' in the namespace?

If that's the case, then that won't work for projects that have hyphens in the name. I've already filed an issue for that.

richardbporter:
Yes, hyphens will not work but you can get around that by specifying a custom install name for you package: https://github.com/composer/installers#custom-install-names

greg-1-anderson:
A PR in [consolidation/annotated-command](https://github.com/consolidation/annotated-command) to transliterate the dashes to underscore would be welcome.

joachim-n:
Ok so consolidation/annotated-command#155 is by design, I am told over there.

I don't understand how this problem is meant to be fixed.

greg-1-anderson:
I'm not sure either. I'm open to suggestions. Above I said this PR would be fine if it avoided excessive searching should anyone drop a large folder in their __DRUPAL__/drush directory. I'm also in favor of deprecating commandfile searching and coming up with a more clear declarative API.

@weitzman was just using some plugin mechanism that @Chi-teck extracted from one of his projects. Maybe that would be helpful here?

richardbporter:
Can you change the installer name in your package's composer.json to match the namespace? 

Ex.
```
"extra": {
        "installer-name": "DevModules"
    }
```

That's what I was referring to in https://github.com/drush-ops/drush/pull/3615#issuecomment-420289698. Sorry if I'm missing something.

joachim-n:
> Can you change the installer name in your package's composer.json to match the namespace?

Yup, I did that.

Annotated Command assumes the namespace is something different, and so the command class is not picked up.

> I'm not sure either. I'm open to suggestions. Above I said this PR would be fine if it avoided excessive searching should anyone drop a large folder in their DRUPAL/drush directory. I'm also in favor of deprecating commandfile searching and coming up with a more clear declarative API.

I'm burnt out on this. Drush 9 breaks my command package completely. I don't know what direction to take, as Drush/AC/Robo seem to have different and conflicting requirements. I just want ONE way of registering a command to work so my package works -- I don't care how.

greg-1-anderson:
I'm -1 on Drush global commands and +1 on standalone commands (see https://github.com/g1a/starter) that `exec` out to Drush to do introspection on Drupal sites. We're also working on factoring parts of Drush out as libraries to make it possible to do more without exec'ing Drush.

AC does not have a plugin mechanism. Robo has a recommended plugin mechanism, but Drush does not use it. Drush focused on a plugin mechanism for modules, and stopped short of providing a clean mechanism for global commands. If you want to keep on the global command bandwagon, you're going to have to push through and make the darn thing work. There's not a lot of demand for this, though, and as I said above and before, we're de-emphasizing global commands because it really complicates the dependency hell problem, especially if you want to allow your global plugins to have dependencies.

I'm sorry this has been rough on you; no one has had time to fix all of the hard problems yet.

joachim-n:
> AC does not have a plugin mechanism. Robo has a recommended plugin mechanism, but Drush does not use it.

But correct me if I'm wrong -- you're one of the principal authors of all three packages! Why don't they agree on how to do things?

> If you want to keep on the global command bandwagon, you're going to have to push through and make the darn thing work. 

Sorry, I am done here. I am probably the only user of the Drupal Code Builder Drush command, since now Drush comes bundled with code generator of its own, people will be using that, even if it is very limited compared to DCB. So carrying on banging my head against a wall here is just a waste of my time and sapping my morale.

webflo:
I agree with @joachim-n, drush command discovery is a mess (sorry). I would like to see a clear vision (or example) whats supported and whats not.

I packaged my drush commands as composer package (e.g. https://github.com/webflo/config_file_import). The 2.x worked for Drush 9, but is completely broken in Drush 9.3 and 9.4. The instructions on https://docs.drush.org/en/master/commands/#global-drush-commands are not composer friendly because we are mixing custom (project specific) commands with "vendored" commands.

I tried to add an additional folder to the discovery via "drush/drush.yml", but this does not as intended. 

```
drush:
  paths:
    include:
      - "${drush.vendor-dir}/../drush/contrib"
```

richardbporter:
Here is a global command working with Drush 9.4.0 if that helps anyone. 

https://github.com/richardbporter/drush-users-commands

greg-1-anderson:
Drush users-commands requires:
```
"drush/Commands/{$name}": ["type:drupal-drush"]
```
If different global command have different Composer installer requirements, then they won't work together. I agree that global commands are messed up right now. It would be good if someone could come up with a proposal and PR that worked in a consistent way & didn't involve too much deep directory searching.

greg-1-anderson:
OK, regarding `drush/Commands/{$name}`, that was changed in https://github.com/drupal-composer/drupal-project/pull/394. Unfortunately, there's a problem with this scheme that @zaporylie pointed out in https://github.com/drush-ops/drush/issues/3566#issuecomment-391952290. Unfortunately, no one took notice. This is the same issue that @webflo mentions above (https://github.com/drush-ops/drush/pull/3615#issuecomment-421304600).

It seems that per https://github.com/drush-ops/drush/issues/3566#issuecomment-391719845 (and per @richardbporter in https://github.com/drush-ops/drush/pull/3615#issuecomment-421366734), site-wide Drush commands (those that are part of a site, but not packaged with a module) do work, but **only** for commands `require`d via Composer. There is no good, documented solution for global Drush commands included directly in the repository of a site (site-specific site-wide commands).

I think there is nothing for it but to support the paths `drush/contrib` and `drush/custom`, as we did in Drush 9.2.x and earlier, and as this PR is attempting to do. 

I am continuing this work in https://github.com/drush-ops/drush/pull/3687

PR #3619, opened 2018-07-15
Clarify support policy.

greg-1-anderson:
Some folks have been confused about the Drush support policy, and some seem to feel as if their questions are rejected if they are closed. This PR attempts to clarify the documentation to point out that the Drush issue queue is not the preferred means for providing community support, and closing a support request issue does not mean that the question was invalid or resolved.


greg-1-anderson:
We addressed this with different documentation changes.

PR #3623, opened 2018-07-18
Adding drush cache directory to core:status

rjgwiz:
Adding drush cache directory as an output field in drush status (`@command core:status`) as a sanity check for the new `cache-directory` config/`DRUSH_PATHS_CACHE_DIRECTORY` env variable added in #3597.

Example usage: `drush status --field=drush-cache-directory`

### DIFF
--- a/src/Commands/core/StatusCommands.php
+++ b/src/Commands/core/StatusCommands.php
@@ -69,7 +69,7 @@ class StatusCommands extends DrushCommands implements SiteAliasManagerAwareInter
      *   files-path: Files, Public
      *   temp-path: Files, Temp
      *   %paths: Other paths
-     * @default-fields drupal-version,uri,db-driver,db-hostname,db-port,db-username,db-name,db-status,bootstrap,theme,admin-theme,php-bin,php-conf,php-os,drush-script,drush-version,drush-temp,drush-cache-directory,drush-conf,install-profile,root,site,files,private,temp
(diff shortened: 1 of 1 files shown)


weitzman:
I think this should be omitted from the default fields.

weitzman:
Or perhaps it should be in %paths such as `drush st %paths`. At least that makes sense from a [config hierarchy perspective](https://github.com/drush-ops/drush/blob/master/examples/example.drush.yml#L75) - (DRUSH.PATHS.CACHE_DIRECTORY)

rjgwiz:
> I think this should be omitted from the default fields.

Makes sense

> Or perhaps it should be in %paths such as drush st %paths. At least that makes sense from a config hierarchy perspective - (DRUSH.PATHS.CACHE_DIRECTORY)

That makes sense from a drush runtime config perspective, though the other `paths` all seem to be arrays and there can is only one cache dir (similar to temp dir) . Also when `%paths` gets printed, it's "Other Paths" with a list of folders. I think cache path having its own label would be very helpful

rjgwiz:
I'm stumped on the [ci error](https://circleci.com/gh/drush-ops/drush/2841?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). I can't see where this change would be impacting `UpdateDBTest::testFailedUpdate()`

greg-1-anderson:
Those same tests passed in ci/circleci: build and ci/circleci: build_highest. Re-running build_56 to see if perhaps it was a transient failure

PR #3625, opened 2018-07-20
fix #3624: Change usage of '--config-dir' site-install option.

### LINKED ISSUE #3624: Add support for core config installer
# Description
Thanks to https://www.drupal.org/project/drupal/issues/2788777 and https://www.drupal.org/project/drupal/issues/2980670 core now supports installing directly from config without the need of a profile. This doesn't just happen out of the box, there is a new `config_install_path` parameter in `install_state` that toggles a lot of behaviors in the installer.

To support installing from config in Drush, SiteInstallCommands will need to be updated to support providing this.

# Goal
To supporting installing from config directly in Drush.

vpeltot:
Fix #3624

Regarding https://www.drupal.org/project/drupal/issues/2788777 that allow a site to be installed from an existing configuration from 8.6.x version, here an update to change the usage of `--config-dir` option.

Up to now, this option allowed to import a full set of configuration after installation.
From now, configuration does not import anymore after installation, but during installation.

### DIFF
--- a/src/Commands/core/SiteInstallCommands.php
+++ b/src/Commands/core/SiteInstallCommands.php
@@ -71,7 +71,7 @@ class SiteInstallCommands extends DrushCommands implements SiteAliasManagerAware
      * @aliases si,sin,site-install
      *
      */
-    public function install(array $profile, $options = ['db-url' => self::REQ, 'db-prefix' => self::REQ, 'db-su' => self::REQ, 'db-su-pw' => self::REQ, 'account-name' => 'admin', 'account-mail' => 'admin@example.com', 'site-mail' => 'admin@example.com', 'account-pass' => self::REQ, 'locale' => 'en', 'site-name' => 'Drush Site-Install', 'site-pass' => self::REQ, 'sites-subdir' => self::REQ, 'config-dir' => self::REQ, 'existing-config' => FALSE])
+    public function install(array $profile, $options = ['db-url' => self::REQ, 'db-prefix' => self::REQ, 'db-su' => self::REQ, 'db-su-pw' => self::REQ, 'account-name' => 'admin', 'account-mail' => 'admin@example.com', 'site-mail' => 'admin@example.com', 'account-pass' => self::REQ, 'locale' => 'en', 'site-name' => 'Drush Site-Install', 'site-pass' => self::REQ, 'sites-subdir' => self::REQ, 'config-dir' => self::REQ, 'existing-config' => false])
     {
         $additional = $profile;
         $profile = array_shift($additional) ?: '';
--- a/src/Commands/core/SiteInstallCommands.php
+++ b/src/Commands/core/SiteInstallCommands.php
@@ -20,6 +20,25 @@ class SiteInstallCommands extends DrushCommands implements SiteAliasManagerAware
     use SiteAliasManagerAwareTrait;
 
     /**
+     * Drupal version exploded.
+     *
+     * eg. [8,6,0-alpha1]
+     *
+     * @var array
+     */
+    protected $version_parts = [];
+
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks. We do need to remain compatible with 8.5 sites for a good while. if needed, you can tell what version of Drupal we are dealing with via 
```
$drupal_root = Drush::bootstrapManager()->getRoot()
$version = Drush::bootstrap()->getVersion($drupal_root);
```

vpeltot:
@weitzman I forgot that there is a lot of Drupal not up to date 😄 

So, here a new commit to keep the backward compatibility + add new option to allow the installation from the config sync directory (`--existing-config`).

bircher:
I tested the PR branch in one of my projects with Drupal 8.6.0-alpha1 and it worked. :tada: 

We are using https://github.com/openeuropa/task-runner on that project, and the drush option would have to be added there too.

For others finding this, for now I just added the following to runner.yml for the normal `run drupal:site-install` to work:
```
drupal:
  pre_install:
    - { task: "symlink", from: "../../../../../config/sync", to: "${drupal.root}/core/profiles/minimal/config/sync" }
  post_install:
    - { task: "remove", file: "${drupal.root}/core/profiles/minimal/config/sync" }
```

weitzman:
The code looks fine, but it is confusing to see two options which do nearly the same thing. Would it be too thorny to reuse same option and have its behavior be different with 8.6+?

neclimdul:
I don't know, they seem different to me but we're probably thinking about it differently.
1) --existing-config
  - No value.
  - Reads from `$config_directories[CONFIG_SYNC_DIRECTORY]` like cex and cim
2) --config-dir='foo'
  - Has a value
  - Installs from the exact path given which can be separate from other config commands.

weitzman:
I think those are too similar and will confuse

mike-potter:
Tested the --existing-config option.  If I don't specify the "minimal" profile Drush assumes "standard" even though the existing config has "minimal" set in core.extensions.yml.  The same is true when using --config-dir option.  Does Drush need to know the profile when installing via config?  If so, then it should probably read from config. But this is minor enough that I'd RTBC the current version if it will expedite getting this released before 8.6.0.  People are going to need this asap to start using 8.6 since config-installer no longer exists.

Regarding the --existing-config vs --config-dir: they do different things. I agree they are somewhat similar but I also think they are properly named.  --config-dir is the drush-only compatibility option for previous users. I don't expect people to use this much in 8.6+ workflow as the config sync should be set in settings.php already.  --existing-config is the name of the new option in core so you don't want Drush changing this.

weitzman:
Thanks for testing. I'm warming up to the two option solution. Maybe we explicitly say that `--existing-config ` is recommended for Drupal 8.6+

Where is `--existing-config` declared in core?

weitzman:
I bet that Drush does not need to pass a profile to install_drupal() when using the new --existing-config code path. Thats a bug we should fix in this PR. Its probably wise to leave the --config-dir code path alone as changes can mess up existing sites.

mike-potter:
existing_config was added in the patch committed to 8.6 here: https://www.drupal.org/project/drupal/issues/2980670

neclimdul:
Yes, both methods where explicitly added as part of that patch.

weitzman:
Yes, it looks like Drush is passing Standard. Maybe we can just remove the fallback to Standard and let Drupal handle it? See https://github.com/drush-ops/drush/blob/2c9a38d1db1f78505ca28cfe44e3db565f02e9e8/src/Commands/core/SiteInstallCommands.php#L137-L173.

weitzman:
Lets keep working on option descriptions and default profile in new PRs.

PR #3629, opened 2018-07-24
Add support in config import single command

rotemreiss:


weitzman:
I dont see why we would use a new command for this. Could't we re-use --partial and let it point to a single file? Would that make the code flow too complex?

rotemreiss:
I'll look into it once again, although I think that it crossed my mind.

navneet0693:
This will really a lot helpful!

mxr576:
Is there a new PR that would add this functionality to `drush cim --partial` as suggested in https://github.com/drush-ops/drush/pull/3629#issuecomment-409328264?

PR #3633, opened 2018-07-26
Removing unused D8 Database reference. Adding copy of Random class.

### LINKED ISSUE #3632: sql-sanitize doesn't work for drupal 7 or 6 in drush 8.x
**Describe the bug**
In Drush 8 `Drush\Commands\core\SanitizeCommands` attempts to make use of two classes that are only available in Drupal 8. 

```
use Drupal\Component\Utility\Random;
use Drupal\Core\Database\Database;
```

In the case of `Database`, the class is imported but never used int `SanitizeCommands`. 

**To Reproduce**
Attempt to run `drush sql-sanitize` using Drush 8 on a Drupal 7 site.

**Expected behavior**
The sanitization hooks would be executed and the database would be altered.

**Actual behavior**
Received this error message:

```
WD php: Error: Class 'Drupal\Core\Database\Database' not found in Drush\Sql\SqlBase->query_prefix() (line 205 of                                       [error]
/home/.composer/vendor/drush/drush/lib/Drush/Sql/SqlBase.php).
Error: Class &#039;Drupal\Core\Database\Database&#039; not found in Drush\Sql\SqlBase->query_prefix() (line 205 of /home/.composer/vendor/drush/drush/lib/Drush/Sql/SqlBase.php).
Drush command terminated abnormally due to an unrecoverable error.                                                                                     [error]
```

**Workaround**
No simple workaround possible.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.1.15
| Drupal version? | 7.59
| PHP version     | 7.1
| OS?             | Linux

**Additional information**
This issue was introduced in 3a10fa05180c05912800c253ae758781695fb38d.

daggerhart:
This PR fixes #3632.

It resolves the issues with the import of Drupal 8 classes in 2 ways. 

1. It removes the unused import of `Drupal\Core\Database\Database`.
2. It creates a copy of the `Drupal\Component\Utility\Random` component as an internal Drush class, `Drush\Utility\Random`.

This allows `sql-sanitize` to run without issues.


weitzman:
Thanks for the fine issue report and PR. Looks good at first glance.

We used to test on older drupal versions but apparently don't do that anymore. Pity.

weitzman:
I will say thats its a bit surprising that we have not received more complaints about this. I would think that folks are sanitizing on older Drupal versions.

daggerhart:
Yeah I'm not sure why this hasn't been reported yet. Maybe people aren't making new D7 sites, and they're not updating drush for their older sites? ¯\_(ツ)_/¯

Also, I'm not sure that what is failing in that travis build relates to this PR at all. You see anything I need to clean up or fix to get this merged?

weitzman:
The random class is only instanciated for drush 8 sites. See https://github.com/drush-ops/drush/blob/0dfa90eccbb2c457579d00528a99203ee148db09/lib/Drush/Commands/core/SanitizeCommands.php#L106-L108. I dont think there is a problem here.

How does removing a unused `use` statement fix anything? I'm fine with removing it, but I'm curious.

daggerhart:
Thanks @weitzman, you're right. 

I've tested again using the 8.x branch, and it works fine. Looks like this commit fixes the error I was running into -- https://github.com/drush-ops/drush/commit/6ba88ce6b9738da683f68ff0a68d40e7817e20e9

So I just need to wait for the next 8.x release, or use a checkout of drush on my server.

No need to replace the Random class for this specific issue. I'm going to close this commit and the related issue.

PR #3642, opened 2018-08-07
Try consolidation/config useMergeStrategyForKeys 

greg-1-anderson:
This will allow drush.paths.include and drush.paths.alias-path to merge values defined in different drush.yml files.

Without this, if a user defines commandfile paths or site alias paths in more than one location, only the values from the most-recently read config file will be used, and the others will be ignored.


greg-1-anderson:
I forgot that the `isolation` directory has its own composer.json file. Doh. composer-test-scenarios generates derivative composer.json files on the fly. Maybe I should use composer-test-scenarios to do Drush isolation testing.

src/Preflight/Preflight.php
```diff
@@ -284,8 +284,8 @@ public function preflight($argv)
         // report on it from Drush status or wherever else it may be needed.
         $configFilePaths = $this->configLocator->configFilePaths();
         $config->set('runtime.config.paths', $configFilePaths);
-        $this->logger()->log(dt('Config paths: ' . implode(',', $configFilePaths)));
-        $this->logger()->log(dt('Alias paths: ' . implode(',', $paths)));
+        $this->logger()->log(dt('Config paths: ' . implode(',', (array)$configFilePaths)));
```
weitzman: This cast has a slight smell. Why would configFilePaths() sometimes return an array and sometimes not?

greg-1-anderson: There was a bug. I can take that out.

PR #3644, opened 2018-08-08
Isolation via test scenarios

greg-1-anderson:


composer.json
```diff
@@ -76,10 +82,21 @@
...
+    ],
     "api": "PATH=$HOME/bin:$PATH sami.phar --ansi update sami-config.php",
     "sami-install": "mkdir -p $HOME/bin && curl --output $HOME/bin/sami.phar http://get.sensiolabs.org/sami.phar && chmod +x $HOME/bin/sami.phar",
-    "isolation": "cd isolation && vendor/bin/phpunit --colors=always",
-    "post-update-cmd": "cd isolation && composer update"
+    "scenario": "scenarios/install",
+    "unit": "phpunit --colors=always",
+    "functional": "./unish.phpunit.php",
+    "post-update-cmd": [
+        "create-scenario isolation --autoload-dir isolation --autoload-dir internal-copy --keep '\\(psr/log\\|consolidation/config\\|site-alias\\|var-dumper\\|symfony/finder\\|drupal-finder\\|path-util\\|sebastian/version\\|xhprof\\)' 'phpunit/phpunit:^5.5.4'",
```
weitzman: these two calls happen after every drush update for all users (not just require-dev)? but test-scenarios package wont be present?

greg-1-anderson: This won't run for site-local Drush users, but you are correct that this will cause problems for anyone who installs a global drush via Composer and then uses `composer update --no-dev`.

This typically isn't a problem for Robo applications, as folks can update their phar via the `self:update` command, and those who are not using the phar typically don't use `--no-dev`. Drush doesn't use this, though.

This won't cause a problem at the moment because [Robo itself is including composer-test-scenarios in its 'require' section](https://github.com/consolidation/Robo/blob/master/composer.json#L30). I did this as a workaround for a similar problem with dependencies.io when I was trying to run `composer update --no-dev` in an attempt to only generate a PR when there were updates in the non-dev components. This didn't work out, though, so I'm probably going to move composer-test-scenarios back to require-dev in Robo eventually (even though it's very small).

Maybe what I need to do long-term is make composer-test-scenarios a Composer installer, and have it do nothing other than define a `create-scenarios` hook that runs from `post-update-cmd` when it is present. That would probably work.

Since the problem is kind of niche and is currently masked by Robo, are you okay with merging this as-is for now? You did approve the PR, but I wanted to double-check post explanation.

weitzman: Yes, we lets proceed. 

It would be helpful if you do the long term fix before moving composer-test-scenarios package to the require-dev section in Robo.

greg-1-anderson: Agreed

Issue #3645, opened 2018-08-11
Replace "config-export" alias with different string


damienmckenna:
The alias of the "config-export" command, "cex", is inappropriate given systemic sexisim in most contemporary societies, another alias should be provided as an alternative.

greg-1-anderson:
`cxp`?
`ce`?

damienmckenna:
`ce` would work.

TessaBakker:
Don't forget `cim`, it would be really strange to have two different types of aliases.

weitzman:
I disagree with the premise of this issue. I find am not seeing the logical proof that starts at cex and ends at "promotes sexism". If anyone cares to connect those dots, please write a blog post and link to it from here.

damienmckenna:
Try this: next time you're in a crowded room call out some drush commands and aliases, see which one of them makes people feel awkward.

damienmckenna:
I guess it comes down to wanting to remove problematic word / phrase usage, in this case removing a sexualized word from somewhere that it has no reason to exist.

Would you be comfortable saying the phrase "drush cex" in a keynote speech at DrupalCon, a large group of folks at your client's office, or a classroom of teenagers learning web development for the first time, or a small group learning to take care of a college department's website? What message would it send? How would the male members of the audience feel hearing it? How would the non-male members of the audience feel hearing it? Would it make people feel awkward? Would some folk start joking around that they're "sexing" their Drupal site? Would it make some people wonder why working with Drupal has to involve running "sex" commands every day? How would that feel to a woman just learning Drupal for the first time, maybe questioning whether she should even in IT given how much there is to learn and how much she has to fight and/or ignore structural sexism?

greg-1-anderson:
I am adding more comments because I feel that we could have been more constructive here, and that's bothering me.

First, I want to address the premise of the issue, which I skipped over last time.

Yes, I would be entirely comfortable talking about "drush cex" at the DrupalCon keynote speech, or any other venue. I am comfortable with this because I pronounce this shortcut as "drush sea ee ex". I talk about Drush a lot, and so far I have never had anyone think I was talking about "drush sex" or "drush es ee ex" when I said "drush sea ee ex".  I haven't ever had any confusion or difficulty in understanding with "drush cex", because the Drush config-export command is very commonly used, and most people have existing experience with it. When I hear other people talking about the config-export command, I predominantly hear "config-export" or "sea ee ex".

[...]

PR #3646, opened 2018-08-12
Rename config-export alias.

damienmckenna:
For #3645.


greg-1-anderson:
I don't think we should completely remove `cex`, as that would break scripts. Do we need to have hidden shortcuts, or is it sufficient to have an alternative for folks to use?

greg-1-anderson:
To be clear, I am strong -1 on breaking scripts here. We could potentially hide `cex` and `cim` (although I see no reason to hide `cim`), but we should not outright remove them prior to Drush 10.

It would take extra work to implement hidden command shortcuts, of course, but breaking automation is a non-starter.

PR #3647, opened 2018-08-12
New aliases for config-import/export.

damienmckenna:
This is the changes from #3645 for the 8.x branch.

### DIFF
--- a/commands/core/config.drush.inc
+++ b/commands/core/config.drush.inc
@@ -91,7 +91,7 @@ function config_drush_command() {
   $items['config-export'] = array(
     'description' => 'Export configuration to a directory.',
     'core' => array('8+'),
-    'aliases' => array('cex'),
+    'aliases' => array('ce'),
     'arguments' => array(
       'label' => "A config directory label (i.e. a key in \$config_directories array in settings.php). Defaults to 'sync'",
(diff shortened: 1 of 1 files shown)


TessaBakker:
This will introduce a conflict with ultimate cron: https://cgit.drupalcode.org/ultimate_cron/tree/ultimate_cron.drush.inc?h=8.x-2.x#n74 
If a possible nasty word is the issue, could we hide aliases with a simple setting? Other option, remove all aliases in 9.x and allow users to create their own and leave 8.x as it is, so implementations won't break.

greg-1-anderson:
Drush 9 and Drush 8 both have stable releases, so breaking existing scripts by removing existing command shortcuts cannot be done. We could consider removing `cex` completely in Drush 10, but in Drush 9 and earlier the best we could do is hide it.

Drush 8 is in maintenance fixes only. I suggest:

- Drush 8: Add the new command shortcut only.
- Drush 9: Add the new command shortcut, and optionally hide `cex` (if someone steps up to make the PR)
- Drush 10: Remove `cex`

damienmckenna:
I've updated both PRs to put back the existing aliases, so that now there are two new aliases for each one.

PR #3648, opened 2018-08-13
Add an 'ssh.tty' option to site aliased.

greg-1-anderson:
Usually Drush decides whether or not to create a tty for remote ssh commands based on whether the current Drush command is running in interactive mode or not. When in interactive mode, a tty is created; otherwise, no tty is created.

Sometimes, a remote server will not grant the ssh user the permission to create a tty. In this instance, it is possible to force Drush to never create a tty via this new option.


weitzman:
The code looks like it does not affect non-interactive requests, so the doc is only partially correct.

greg-1-anderson:
You're right -- I should fix the code to match the docs.

PR #3656, opened 2018-08-21
Drush 8.x: Adding cwd option to core-cli

DimeZilla:
so that we can change the default working directory.

Hi I love drush and I love the addition of psysh through core-cli. I'd love an option for changing the working directory while I'm working on a custom module or something so I added a cwd option to the core-cli command. This way I can do something like this:

```
drush core-cli --cwd="$(pwd -P)"
```

Now, when the interactive psysh shell is lifted, the current working directory is wherever I am working.

I hope this is useful.

### DIFF
--- a/commands/core/cli.drush.inc
+++ b/commands/core/cli.drush.inc
@@ -19,6 +19,7 @@ function cli_drush_command() {
     'topics' => array('docs-repl'),
     'options' => array(
       'version-history' => 'Use command history based on Drupal version (Default is per site).',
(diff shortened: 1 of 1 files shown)


damiankloip:
Yes, I like this. It actually mirrors what Psysh does for its own bin file too. Thanks!

greg-1-anderson:
We need this on the 9.x branch. New features should be added on `master` first to avoid inconsistencies.

damiankloip:
Yes, it's a very good point.

PR #3664, opened 2018-08-25
Make config:set --value_format=yaml work.

fago:
I tried to use config:set --format=yaml and noted it's not working as the --format option is already used for the output format. Attached PR makes it usable by renaming it to value_format and by setting the values correctly.


weitzman:
#3666

PR #3673, opened 2018-09-07
Fix 3376. When an update fails, updatedb is still exiting with 0.

weitzman:
Both exception branches have same 2 lines.


dpagini:
I know it's a little late here. I pulled this patch and it's solving the issue for me. Thank you for the fix!

PR #3680, opened 2018-09-12
Fixes #3672: Call Boot::terminate() in a shutdown handler.

### LINKED ISSUE #3672: Services with needs_destruction tags are not destroyed
**Describe the bug**
Services tagged with needs_destruction that are instantiated as part of a script run ('drush scr') are not destroyed at the end. 

**To Reproduce**
Configure a search API index to "index immediately".
Change an entity in a script executed with drush scr scriptname.php

**Expected behavior**
The change should be reflected in the index even before the next cron run, i.e. immediately (read: "after the end of the script run").

**Actual behavior**
The change is only reflected in the index after the next cron run which means the search API feature "index immediately" is broken. This is because the destruct() method of the service is not called. In GUI mode this is done via HTTPKernel->terminate() in KernelDestructionSubscriber->onKernelTerminate() which receives a PostRequestEvent.

**Workaround**
Dunno. Calling this manually does not seem like the correct solution.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.3.0
| Drupal version? | 8.5.x
| PHP version     | 7.0
| OS?             | Mac

**Additional information**
I opened a core issue, too, as I am not sure if this is a drush or a drupal core bug:
https://www.drupal.org/project/drupal/issues/2997740

greg-1-anderson:
Not sure if this is effective, but it at least does not crash.

### DIFF
--- a/src/Boot/BaseBoot.php
+++ b/src/Boot/BaseBoot.php
@@ -21,6 +21,7 @@ abstract class BaseBoot implements Boot, LoggerAwareInterface, ContainerAwareInt
 
     public function __construct()
     {
+        register_shutdown_function([$this, 'terminate']);
     }
 
     public function findUri($root, $uri)


weitzman:
Is there an issue where we call emptyBoot->terminate(), and also a version specific one? I seem to recall that we change boot objects during the bootstrap. Maybe this doesn't matter in practice.

greg-1-anderson:
The empty boot terminate does not do anything, and in general, a terminate method does not do anything unless it was constructed. I therefore think it's safe to terminate all of the boot objects.

If this were a concern, I could move the shutdown handler to the bootstrap manager, and only terminate the one active bootstrap object.

PR #3681, opened 2018-09-12
Fix #667. Allow enabling of test modules.

### LINKED ISSUE #667: D8 and test module enable
Since sun broke it test modules can't be enabled any more. There's some obscure setting that fixes it, please add it to drush. https://drupal.org/node/2198713

weitzman:


weitzman:
Unfortunately, Drush cant provide this. I added the Settings code to drush en, pmu and pml. I was able to add/uninstall and list test modules. However, the services provided by those modules was unavailable because core does a discovery which I cant easily fake. So, I dont think we should enable modules that kinda sorta work. See \Drupal\Core\DrupalKernel::moduleData.

PR #3682, opened 2018-09-16
Move sandbox and SUT alongside Drush, instead of in a tmp dir

weitzman:
Here is a significant restructuring for Unish, Drush's functional test suite. For a while now, Drush has run tests in "sut" and "sandbox" directories, which were located within the system's temp dirs.

This PR moves these into top level directories within the Drush package itself. I've added drupal/core as a require-dev dependency and it installs into the /sut directory. Regular Drush users see no difference since require-dev will be ignored.

I've removed files and logic related to the distant SUT. Gone are:
- unish.sut.php
- unish.phpunit.php
- DRUSH_AUTOLOAD_PHP
- UNISH_TMP
- The Path repository in the SUT

To tun tests now, you run `phpunit --configuration tests` (just like a few years ago). This same code is provided as `composer functional`.

Another happy byproduct of this PR is that when you develop on Drush, you have a full Drupal installation in your IDE. You benefit from completion, help etc.

This PR changes a couple scenario files, since they were auto-changed after a `composer update`. I assume thats OK.


greg-1-anderson:
We could maybe use [g1a/composer-test-scenarios](https://github.com/g1a/composer-test-scenarios) for this. Might need to fix that outstanding issue with dev dependencies first. In any event, I'm 👍 on this with or without test-scenarios. I can try to add that in later in a follow-on PR.

PR #3686, opened 2018-09-17
Update cron.md - invert env / docroot order

Agence-IDIX:
cd to the docroot shall be made prior to calling env / drush


weitzman:
I think you dropped the && by accident. Its needed

PR #3687, opened 2018-09-17
Retain 'contrib' / 'custom' paths for site-wide Drush extensions

greg-1-anderson:
...as in Drush 9.2.x and earlier.

Also retains the 'Commands' path for the benefit of those who switched their global commands for Drush 9.3.0 / Drush 9.4.0.


docs/commands.md
```diff
@@ -65,13 +65,15 @@ Commandfiles that don't ship inside Drupal modules are called 'global' commandfi
      - Filename: $PROJECT_ROOT/drush/Commands/ExampleCommands.php
      - Namespace: Drush\Commands
 1. Nested (e.g. Commandfile is part of a Composer package)
-    - Filename: $PROJECT_ROOT/drush/Commands/dev_modules/ExampleCommands.php
+    - Filename: $PROJECT_ROOT/drush/contrib/dev_modules/ExampleCommands.php
```
weitzman: Its a bit awkward to have `sites` alongside contrib/custom. Its more natural to me to have Commands alongside sites.

greg-1-anderson:
This PR still has the deep-searching problem. I'm working on fixing that. I'll also reverse the order of `Commands` and `contrib` in the documentation.

I think that it might also be useful to support `contrib/Commands` as an alternate form, though. That would allow a Composer-managed Drush extension to package both Drush 8 and Drush 9 commands in the same project.

src/Config/ConfigLocator.php
```diff
@@ -462,29 +462,11 @@ protected function getIncludedCommandFilePaths($commandPaths)
...
-        if (empty($directories)) {
-            return $result;
-        }
+        $directories = ["$root/drush", dirname($root) . '/drush'];
 
-        // Find projects
-        $finder = new Finder();
-        $finder->directories()
-            ->ignoreUnreadableDirs()
-            ->path('#^src/Commands$|^Commands$#')
```
greg-1-anderson: This PR removes the undocumented location `src/Commands`.

greg-1-anderson:
See also the supporting feature in https://github.com/consolidation/annotated-command/pull/156

docs/commands.md
```diff
@@ -47,31 +47,53 @@ Altering Drush Command Info
...
 
 For an example, see the alterer class provided by the testing 'woot' module: `tests/resources/modules/d8/woot/src/WootCommandInfoAlterer.php`.
 
-Global Drush Commands
+Site-Wide Drush Commands
 ==============================
 
-Commandfiles that don't ship inside Drupal modules are called 'global' commandfiles. See the [examples/Commands](/examples/Commands) folder for examples. In general, it's better to use modules to carry your Drush commands. If you still prefer using a global commandfiles, here are two examples of valid commandfile names and namespaces:
+Commandfiles that are added directly to a Drupal site that do not ship inside Drupal modules are called 'site-wide' commandfiles. See the [examples/Commands](/examples/Commands) folder for examples. In general, it's better to use modules to carry your Drush commands. If you still prefer using site-wide commandfiles, here are some examples of valid commandfile names and namespaces:
```
weitzman: "... added to a Composer project .."?

weitzman: Can we say a sentence about why its better to put commandfiles in modules? How about "can inject Drupal services"

docs/commands.md
```diff
@@ -47,31 +47,53 @@ Altering Drush Command Info
...
+Commandfiles that are not part of any Drupal site are called 'global' commandfiles. Global commandfiles are not supported by default; in order to enable them, you must configure your `drush.yml` configuration file to add an `include` search location.
+
+For example:
+
+drush:
+  paths:
+    include:
+      - '${env.home}/.drush/commands'
+      
+With this configuration in place, global commands may be placed as described in the Site-Wide Drush Commands section above. Global commandfiles may not declare any dependencies of their own; they may only use those dependencies already made available by Drush.
```
weitzman: s/already made available by Drush/already available via the autoloader.

tests/AnnotatedCommandTest.php
```diff
@@ -48,10 +48,28 @@ public function testGlobal()
         unlink($target);
     }
 
+    public function siteWideCommands()
+    {
+        $options = [
+            'yes' => null,
```
weitzman: Do you really need $options?

tests/UnishTestCase.php
```diff
@@ -107,6 +107,9 @@ public static function cleanDirs()
                     self::recursiveDelete($target);
                 }
             }
+            // Bring back any committed assets removed above.
+            $webroot = self::webroot();
+            exec("git -C $webroot checkout -- .");
```
weitzman: Clever! But can we guarantee that we have a git clone? Can't someone start with a tarball from composer? Maybe thats something we don't care about?

greg-1-anderson: Hm good point. The code as written deletes the Drush commandfiles I added. I was also vaguely concerned that there might be other necessary items that were getting deleted. It might be better to just be careful about what we delete in the cleanup step.

richardbporter:
Thanks for doing this. The differentiation between site-wide and global commands is very useful. Turns out I've been using site-wide commands.

I tested this with my site-wide commands structured for Drush 9.3.0/9.4.0 and they work fine. It picked up `PROJECT_ROOT/drush/Commands/foo/FooCommands.php` as well as the hooks (`PROJECT_ROOT/drush/Commands/foo/FooHooks.php`).

My previous structure pre-9.3.0 was `PROJECT_ROOT/drush/Commands/foo/src/Commands/FooCommands.php` and ``PROJECT_ROOT/drush/Commands/foo/src/Hooks/FooHooks.php`. I tested that but it didn't seem to work which looks like is by design based on https://github.com/drush-ops/drush/pull/3687#discussion_r218624102. I mentioned in the other PR that I didn't think that structure was super useful anyway so probably not a big deal.

Thanks!

greg-1-anderson:
I allowed the searching to go one level deeper, which allows site-wide commandfiles to now be stored in `PROJECT_ROOT/drush/Commands/foo/src/FooCommands.php`. However, I didn't go any further, so `PROJECT_ROOT/drush/Commands/foo/src/Commands/FooCommands.php` is still not allowed. If folks really wanted that, we could bump the search depth all the way up to `4`, but I'm already concerned with how much things could bog down if a large file were placed in `PROJECT_ROOT/drush/Commands`. Fortunately, large directories inside `PROJECT_ROOT/drush` other than `Commands` / `Hooks` are not searched.

Overall I'm still unhappy with command discovery, but I think that I've put enough time into this for now. The current state at least is clear and documented, and should work well enough to serve the needs of the small number of site-wide commands that exist.

Issue #3572, opened 2018-06-01
`drush cex` doesn't remove `language` directory when it should


boobaa:
Steps to reproduce:
- Install a D8 site with two languages.
- Export its config via `drush cex`. A `language` directory will be created under `sync`.
- Delete the non-default language.
- Re-export the config via `drush cex`.

Actual results:
- Drush says `The .yml files in your export directory (sites/default/files/config_whatever_hash/sync) will be deleted and replaced with the active config.`
- The `language` directory along with all its .yml files are still present.
- Subsequent `drush cex` commands will also report their deletion but won't actually delete them either.

Expected results:
- The `language` directory should be gone altogether with all the subdirectories and .yml files in it.

This is not a D8 core bug: when the config is first exported from the web UI, the tarball does contain the `language` directory, but when it's exported with all the non-default languages deleted, the `language` directory is gone from the tarball.

yobottehg:
I can confirm this from multiple previous experiences. This is also true for single files inside these language directories (For example a view translation that got deleted).

DerH4NNES:
I dont't think that this is a drush problem. I can reproduce that problem with the drupal console too.

cydgit:
drush cex doesn't delete collections. The above pull request fixes it.

Sylry:
It seems it breaks the cex command with that exception on a site where I have several languages:
`In FileStorage.php line 254:
FilesystemIterator::__construct(conf/drupal/config/sync/language/ar): failed to open dir: No such file or directory`

If I comment the lines of what was merged, it works like before.

Sylry:
A quick fix is to try/catch the `$target_collection->deleteAll();` which was added 
Another one is to try/catch the `$this->storage->deleteAll($prefix)` in config_filter FilteredStorage

But it might be a better solution to check in drupal core (`FileStorage::deleteAll()`) if the directory exists before trying to instanciate the FilesystemIterator

vrwired:
I ran into same issue and did not have config_split installed. It did export full via UI though seemed to take a while. I also temp commented out latest merge on vendor/drush/drush/src/Drupal/Commands/config/ConfigExportCommands.php 

I got back
 [success] Configuration successfully exported

dpacassi:
I have the same issue with `Drush 9.5.2`, rolling back to `Drush 9.3.0` solves the issue for me.

joergLin:
[...]

PR #3690, opened 2018-09-19
Fix config:export not deleting collections

cydgit:
Fix for #3572.


alexpott:
This looks like a good solution and inline with deleting all the configuration from the default collection that happens above.

PR #3694, opened 2018-09-22
Be compatible with PHPUnit 6.

weitzman:


.circleci/config.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       - run: $HOME/drush/.circleci/setup.sh
       - run: composer remove --dev webflo/drupal-core-strict --no-update
       - run: composer require --dev drupal/core:8.7.x-dev --no-update
+      - run: composer config platform.php 7.2
```
weitzman: This brings in PHPUnit 6 during build_highest

tests/CommandUnishTestCase.php
```diff
@@ -377,49 +377,6 @@ public function drush($command, array $args = [], array $options = [], $site_spe
...
-   * has been invoked.
-   *
-   * Runs the test case and collects the results in a TestResult object.
-   * If no TestResult object is passed a new one will be created.
-   *
-   * @param  \PHPUnit_Framework_TestResult $result
-   * @return \PHPUnit_Framework_TestResult
-   * @throws \PHPUnit_Framework_Exception
-   */
-    public function run(\PHPUnit_Framework_TestResult $result = null)
```
weitzman: This chunk was an OO error on PHPUnit 5. Not sure how to make both 5 and 6 happy with same code. See https://circleci.com/gh/drush-ops/drush/3338

I'm sad to lose this code coverage collecting during $this->drush() calls but we are not using it anyway. If anyone can fix that error we will bring this code back

PR #3696, opened 2018-09-23
Dont fork new processes in Unish's drush() method

weitzman:
This needs a lot more work and I likely wont pursue it for a while.


weitzman:
The plan now is to reorganize tests at #3759 and then make a new class which will run these faster tests. Then tests can move over to the faster system.

greg-1-anderson:
Continuing in https://github.com/drush-ops/drush/pull/3807

PR #3706, opened 2018-09-27
More conservative variation on PR #2002

dsnopek:
See https://github.com/drush-ops/drush/pull/2002#issuecomment-424412519 where I attempted to describe this :-)

This is branched from @eXistenZNL's original PR. I had to rebase it on the 8.x branch for GitHub to say it was mergeable, which I think is why it gives us both credit for his commit. That commit is all his! Not trying to steal credit or anything.

Anyway, this variation should mean that everyone who Drush was working for in the past should keep working (because they have the 'mysql' extension available). And, for everyone that was getting broken (ie. those without the 'mysql' extension) it should start working! So, hopefully this will make the change acceptable.

I tried to run the test suite with Drupal 6 and PHP 7, but it pulls in Drupal 6.38, which is not compatible with PHP 7. The next release of the D6LTS fork of Drupal 6 (probably 6.45) will be compatible! But I don't know how to get the Drush test suite to use it.

But this does allow drush to work with Drupal 6 on PHP 7 in my limited manual testing!

### DIFF
--- a/includes/sitealias.inc
+++ b/includes/sitealias.inc
@@ -1646,7 +1646,7 @@ function drush_convert_db_from_db_url($db_url) {
       );
       $url = (object)array_map('urldecode', $url);
       $db_spec = array(
-        'driver'   => $url->scheme,
+        'driver'   => $url->scheme == 'mysqli' && extension_loaded('mysql') ? 'mysql' : $url->scheme,
         'username' => $url->user,
         'password' => $url->pass,
         'host' => $url->host,
--- a/includes/sitealias.inc
+++ b/includes/sitealias.inc
@@ -1646,7 +1646,7 @@ function drush_convert_db_from_db_url($db_url) {
       );
       $url = (object)array_map('urldecode', $url);
       $db_spec = array(
-        'driver'   => $url->scheme == 'mysqli' ? 'mysql' : $url->scheme,
+        'driver'   => $url->scheme,
         'username' => $url->user,
         'password' => $url->pass,
         'host' => $url->host,
--- a/lib/Drush/Sql/Sqlmysqli.php
+++ b/lib/Drush/Sql/Sqlmysqli.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Drush\Sql;
+
+class Sqlmysqli extends Sqlmysql {
+
+}


weitzman:
Looks reasonable to me.

greg-1-anderson:
Thanks, @dsnopek. Could you please add the failing test with Drush 6.38, and let me know what is the best way to download and update Drush 6 lts (via PM, via git, etc.), and I'll fix up the test so that it uses the right version and passes.

dsnopek:
So, I don't know for sure that all the test failures are because it's using Drupal 6.38, but I do know that 6.38 will fail bad with PHP 7.2, so it must be responsible for some. :-)

When the tests are running, I can see it say:

```
Testing on Drupal version   :  6.38
```

This will get a Drupal 6 core that's compatible with PHP 7.2:

```
git clone git@github.com:d6lts/drupal.git --branch 6.x
```

dsnopek:
@greg-1-anderson If you have a Drupal 6 site with the [mydropwizard](https://www.drupal.org/project/mydropwizard) module enabled, then Drupal will use the update status data that we publish, so `drush up` will actually work! But that needs a working Drupal 6 site already

greg-1-anderson:
That's good to know. For testing, I think I'll just pull HEAD of 6.x, or maybe just pull a hardcoded tag like 6.44.

dsnopek:
Well, 6.44 won't work with PHP 7 - those changes are only in HEAD, currently. A release is coming soon :-)

Thanks so much for working on this!

greg-1-anderson:
OK, HEAD it is!

dsnopek:
I just made a D6LTS Drupal 6.45 release which supports PHP 7.2:

https://github.com/d6lts/drupal/releases/tag/6.45

Is there anything we can do to move this PR forward?

greg-1-anderson:
Thanks for making the stable release. I haven't had a chance to get to this yet; I'll try to fit it in. It just needs tests. It appears that the current tests have a comment that claim to be testing Drupal 6, but they actually just tests the same variants of Drupal 7 as the tests above.

greg-1-anderson:
Testing with 6.45 here:
https://travis-ci.org/drush-ops/drush/builds/438914617

greg-1-anderson:
Hm I tried to check out the tag 8.45, but it still reports Drupal 8.44. Don't have time to investigate why right now.

dsnopek:
Ah, crap... I think I messed up the release process! I'll see what I can do to correct that.

dsnopek:
Ok. I rewrote history a little bit, but the release was only just made so I don't think anyone was depending on it yet. Pulling the 6.45 tag should now know that it's 6.45 :-)

greg-1-anderson:
The tests now claim to be running on Drupal 6.45, but when they get to `site-install` I get the same symptom as before:
```
PHP Fatal error:  Uncaught Error: Call to undefined function db_result() in /home/travis/build/drush-ops/drush/commands/core/drupal/site_install_6.inc:145
```

dsnopek:
What are the arguments being passed to `drush site-install`? The `--db-url` needs to use 'mysqli://' rather than 'mysql://'. I just tested locally doing `drush site-install` manually and it seemed to work

greg-1-anderson:
OK that makes sense. I haven't had time to chase down the cause, but it sounds like what you said is it. I'm guessing that the unish test harness needs to be updated in addition to the changes in the code.

eXistenZNL:
@dsnopek thanks for giving credit and thanks for the work you put into it! Let's hope that in the end this gets resolved once and for all :+1:

greg-1-anderson:
Merged in future pull requests

PR #3707, opened 2018-09-27
Compatibility with new Tideways XHProf extension.

danepowell:
Drush includes built-in functionality to profile commands using XHProf or Tideways.

Tideways recently went through a [pretty significant refactoring](https://tideways.com/profiler/blog/releasing-new-tideways-xhprof-extension), which changed the extension name and apparently some of the internals as well, although the data format is still compatible with the XHProf UI. This requires Drush to update its integration as well.

I could think of ways to improve this approach, but I wanted to get at least a working PoC for maintainers to review before putting more work into it. Primarily, it would be nice to support flags, which are now also named differently in Tideways. Also, maybe use an abstract "profiler" interface to support either XHProf or Tideways (or even different versions of Tideways), without having to use so many conditionals... maybe that's overkill.

### DIFF
--- a/src/Commands/core/XhprofCommands.php
+++ b/src/Commands/core/XhprofCommands.php
@@ -37,7 +37,7 @@ class XhprofCommands extends DrushCommands
     {
         if (self::xhprofIsEnabled()) {
             $namespace = 'Drush';
-            $run_id = self::xhprof_finish_run($namespace);
+            $run_id = self::xhprofFinishRun($namespace);
             $url = Drush::config()->get('xh.link') . '/index.php?run=' . urlencode($run_id) . '&source=' . urlencode($namespace);
             $this->logger()->notice(dt('XHProf run saved. View report at !url', ['!url' => $url]));
         }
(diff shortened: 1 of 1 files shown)


src/Commands/core/XhprofCommands.php
```diff
@@ -87,4 +86,34 @@ public static function xhprofFlags(array $config)
...
+    }
+
+    /**
+     * Disable profiling and save results.
+     */
+    public static function xhprofFinishRun($namespace)
+    {
+        if (extension_loaded('tideways_xhprof')) {
+            $data = \tideways_xhprof_disable();
+            $dir = sys_get_temp_dir();
```
weitzman: Use Drush::config()->tmp() to get temp dir.

src/Commands/core/XhprofCommands.php
```diff
@@ -53,17 +50,17 @@ public function xhprofPost($result, CommandData $commandData)
...
-            \xhprof_enable($flags);
+            self::xhprofEnable($flags);
         }
     }
 
     public static function xhprofIsEnabled()
     {
         if (Drush::config()->get('xh.link')) {
-            if (!extension_loaded('xhprof') && !extension_loaded('tideways')) {
+            if (!extension_loaded('xhprof') && !extension_loaded('tideways_xhprof')) {
```
weitzman: Perhaps add docs at top of file about tideways. Also expand Exception message to mention new Tideways.

Issue #3711, opened 2018-09-30
drush_error_handler requires an optional and deprecated argument


bangpound:
**Describe the bug**

The function `drush_error_handler` in `includes/environment.inc` requires the fifth argument `$context` to be passed even though this argument has always been optional in PHP and is deprecated in PHP 7.2.

Being optional and deprecated, [some other projects](https://github.com/symfony/dependency-injection/blob/v3.4.15/Loader/YamlFileLoader.php#L663) do not pass this argument when they call the error handler as part of their own.

The fifth argument isn't even used in the `drush_error_handler` function, so it doesn't even need to be expected in this function.

**To Reproduce**

1. After bootstrapping Drush, override the error handler with a function that also calls the previous error handler.
2. Trigger an error.
3. Get a fatal error in the error handler.

**Expected behavior**

I'd expect to see the error I triggered.

**Actual behavior**

The drush error handler causes an error that obscures the error I triggered.

**Workaround**

None!

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.x
| Drupal version? | 7.x
| PHP version     | 7.1
| OS?             | Mac and Linux

**Additional information**

This problem exists in both Drush 9 and 8.

PR #3712, opened 2018-09-30
Make drush_error_handler's fifth argument optional.

bangpound:
Since set_error_handler was introduced, the fifth argument has been optional. In
PHP 7.2 it has been deprecated. The argument is not even used in the function.

### DIFF
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -17,7 +17,7 @@ use Webmozart\PathUtil\Path;
  * Log PHP errors to the Drush log. This is in effect until Drupal's error
  * handler takes over.
  */
-function drush_error_handler($errno, $message, $filename, $line, $context) {
+function drush_error_handler($errno, $message, $filename, $line) {
   // E_DEPRECATED was added in PHP 5.3. Drupal 6 will not fix all the
   // deprecated errors, but suppresses them. So we suppress them as well.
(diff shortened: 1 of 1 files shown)


bangpound:
This would fix https://github.com/drush-ops/drush/issues/3711 for 8.x.

greg-1-anderson:
Can we just remove this parameter here as well? We want the 8.x branch to work on php 7.2 without deprecation warnings. I run 7.2 and don't recall seeing a notice here, though, so perhaps this won't be an issue until a future version of PHP.

bangpound:
I encountered this error because I instantiate a Symfony Dependency Injection container in Drupal 7. I had some YAML service files that multiple service definitions with the same name, and this caused the YamlFileLoader in the Symfony component to trigger an error. The YamlFileLoader has its own error handler but also calls whichever was already defined. When that error handler called the drush error handler with only 4 parameters, I got a fatal error.

greg-1-anderson:
Yes,  in my comment above I was intending to contrast only the difference between removing the parameter entirely vs. giving it a default value. Since I have never seen a warning here in absence of a failing situation (with a secondary error handler), my hypothesis  was that it is not necessary to remove the parameter.

The Drush error handler does not attempt to call any other pre-installed error handlers. Perhaps it should.

PR #3713, opened 2018-09-30
Make drush_error_handler's fifth argument optional.

bangpound:
Since set_error_handler was introduced, the fifth argument has been optional. In
PHP 7.2 it has been deprecated. The argument is not even used in the function.

### DIFF
--- a/includes/environment.inc
+++ b/includes/environment.inc
@@ -18,7 +18,7 @@ use Webmozart\PathUtil\Path;
  * Log PHP errors to the Drush log. This is in effect until Drupal's error
  * handler takes over.
  */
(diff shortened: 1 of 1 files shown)


bangpound:
This would fix #3711 for 9.x. I wasn't sure given the version support in 9.x if it would be better to just remove the argument since it is deprecated. Let me know if I should revise this. Thank you!

greg-1-anderson:
I think there is no ill effect in removing the parameter here. Even older versions of PHP should still behave correctly, even if the extra parameter is provided. Please LMK if you can think of any situation where there would be a problem.

bangpound:
I've removed the parameter on both pull requests.

PR #3718, opened 2018-10-04
Add --filter to many commands

greg-1-anderson:
```
$ drush9 pml --filter='id=bartik' 
 --------- ----------------- --------- --------- 
  Package   Name              Status    Version  
 --------- ----------------- --------- --------- 
  Core      Bartik (bartik)   Enabled   8.5.3    
 --------- ----------------- --------- --------- 
```
Just one field:
```
$ drush9 pml --filter='id=bartik' --field=status
Enabled
```


greg-1-anderson:
It might be possible to make the `@filter-output` implicit based on the return type of the command; however, if we did this, then there should be some other annotation to turn off the filter. If a command is already filtering the output rows, then it would be inefficient to pass it through the generic filter as well.

I don't think it would be too bad to have to add the annotation to commands that need it.

greg-1-anderson:
I updated the documentation for [filter-via-dot-access-data](https://github.com/consolidation/filter-via-dot-access-data) to illustrate how "contains" and "regex" searches are done.

greg-1-anderson:
The `role:list` command already took a `--filter` parameter that had different conventions than the filter added by this PR. To compensate, I added a feature to consolidation/filter-via-dot-access-data that allows `role:list` to specify that the default field to filter by is "perms".  If someone uses `--filter=value` without providing a filter operator, then "perms contains value" is the operation that is performed. This allows both:
- `drush role:list --filter='edit terms in tags'`
- `drush role:list --filter='perms*=edit terms in tags'`
- `drush role:list --filter='id=editor'`

docs/output-filters.md
```diff
@@ -0,0 +1,36 @@
+Output Filters
```
weitzman: Thanks for writing this ... I have two thoughts. I don't know how to handle them though.

1. Output filters sounds a lot like output formats which is a related but different concept. Everyone is going to mix them up.
1. Might be worth saying how this feature differs from grep and when to use each.

src/Commands/core/UpdateDBCommands.php
```diff
@@ -112,6 +112,8 @@ public function entityUpdates($options = ['cache-clear' => true])
      *   description: Description
      *   type: Type
      * @default-fields module,update_id,type,description
+     * @filter-output
+     * @filter-default-field module
```
weitzman: How about a rule that defining @filter-default-field implies @filter-output? No need to do both, typically.

greg-1-anderson: Should be possible.

greg-1-anderson:
> Output filters sounds a lot like output formats which is a related but different concept. Everyone is going to mix them up.

I resolved this by making one topic on adjusting command output. It covers output formats, output fields and output filters together. Since all of these options are related to the same concept, it should be clear if folks can read about them together.

> Might be worth saying how this feature differs from grep and when to use each.

Haven't worked this in yet.

- Output filters are like grep, but they match only the data within specific named fields. This makes it easier to avoid false matches.
- Filters are also applied before data is formatted, so you don't need to worry about missing out on rows that should have matched, but that were missed because the matching data was word-wrapped.
- Use `grep` directly if you need to match based on data that spans columns.

Something akin to that could go in the introduction, or maybe at the top of the "output filters" section.

src/Commands/generate/Generators/Drush/default-methods.twig
```diff
@@ -31,6 +31,7 @@
    * @command {{  machine_name }}:token
    * @aliases token
    *
+   * @filter-output
```
weitzman: should declare a default field instead

docs/output-formats-filters.md
```diff
@@ -0,0 +1,125 @@
...
+taxonomy_term
+user_admin_people
+watchdog
+who_s_new
+who_s_online
+archive
+glossary
+```
+
+Output Filters
```
weitzman: I'm thinking that this topic should be offerred automatically in help for any command that shows the --fields option.

src/Commands/generate/Generators/Drush/default-methods.twig
greg-1-anderson: How do we pick which field is the default field? Assume `id` and let the user change it?

docs/output-formats-filters.md
greg-1-anderson: Done

src/Commands/core/UpdateDBCommands.php
greg-1-anderson: Done

greg-1-anderson:
I think I've done about as much as I intend to do here for now. I recommend that we first release Drush 9.5.0, and then merge this PR to the master branch so that folks can try it out in the dev release for a while before calling it stable.

src/Commands/core/UpdateDBCommands.php
```diff
@@ -112,6 +112,7 @@ public function entityUpdates($options = ['cache-clear' => true])
      *   description: Description
      *   type: Type
      * @default-fields module,update_id,type,description
+     * @filter-default-field module
```
weitzman: I think 'type' would be a better default field here.

PR #3720, opened 2018-10-09
Dsnopek mysqli extension 2

greg-1-anderson:
Continuation of #2002


greg-1-anderson:
I'm getting `Call to undefined function db_result() in drush/commands/core/drupal/site_install_6.inc` on php 5.6 and greater.
```
drush --nocolor site-install default --db-url=mysqli://root:@127.0.0.1/unish_dev --yes
```
It doesn't matter if I use mysql or mysqli.

It does work with php 5.4, so it has something to do with the lack of the mysql driver.

dsnopek:
Hm. That command (with a modified `--db-url` or my database) works in my testing using Drupal 6.45, PHP 7.2.10 and Drush modified with my PR.

The result you're getting is exactly what I get when I use Drush on the 8.x branch without my PR!

@greg-1-anderson Are you sure that you're running the right copy of Drush with the modifications, and not another one?

greg-1-anderson:
@dsnopek I think I tried both this branch, and your PR unmodified without success. I will double-check and make sure that my Drush has the right diffs.

Assuming I am running the right Drush, could you give the exact git clone + drush site-install line you are using to successfully install Drupal 6 with PHP 7+ (I tried 7.2/5.6 (not working) and 5.4 (working))? I think that in https://github.com/drush-ops/drush/pull/3720/commits/1d90c778c3dc5d3cc4ff0e4e9e603262461cb1a5 I fixed a bug related to `--sites-subdir`, but I was still unable to install even avoiding this code path.

dsnopek:
Sure!

I'm using Lando, so here's a [.lando.yml file](https://gist.github.com/dsnopek/d92ef788e518a2373e535d78156c5082) you can use to get an environment the same as mine. It's got some special configuration to use a custom Drush from the 'drush' directory in the site code base.

Here's the exact commands:

```
git clone https://github.com/d6lts/drupal.git --branch 6.x d6p7
cd d6p7
git clone https://github.com/dsnopek/drush.git --branch mysqli-extension-2
(cd drush && composer install)

# These next bits add the lando config and run the drush command that way
wget https://gist.githubusercontent.com/dsnopek/d92ef788e518a2373e535d78156c5082/raw/5c5e9351fd08cef12b3b9ba04f854769692512e6/.lando.yml -O .lando.yml
lando start
# If you're not using Lando, just drop the 'lando' from the front of the command
lando drush --nocolor site-install default --db-url=mysqli://drupal6:drupal6@database/drupal6 --yes
```

I based the site-install command on the one you said you were running - normally I run that with fewer arguments.

Anyway, doing exactly the above worked in my testing just now! If I go into the 'drush' directory, and instead checkout the 8.x branch, then I get the error you're having.

greg-1-anderson:
Did all of that and still got:
```
./drush/drush site-install default --db-url=mysqli://root:@127.0.0.1/d6ltsdb --yes
PHP Fatal error:  Uncaught Error: Call to undefined function db_result() in /Users/ganderson/tmp/dsnopek-d6/d6p7/drush/commands/core/drupal/site_install_6.inc:145
```
The db url written into settings.php is the same as the one I specified:
```
$db_url = 'mysqli://root:@127.0.0.1/d6ltsdb';
```
That's expected, right?

Seems like I have the right commits from d6lts:
```
9be250ac44 2018-10-09 David Snopek Increase version (so I don't forget again) and update CHANGELOG.txt
0ee2e6ebf6 2018-10-09 David Snopek Revert "Added packaging info for the 6.45 release"
bbc3f0d558 2018-10-09 David Snopek Added packaging info for the 6.45 release
```
I'm not using Lando, but I wouldn't expect that to change the behavior.

dsnopek:
@greg-1-anderson Hm. Yeah, that all seems right, as far as the code for Drupal and Drush.

Is it possible that the 'mysqli' extension isn't installed for your PHP 5.6 and 7.2? Because, actually, it's super weird that you get that error with PHP 5.6 - we use vanilla Drush with Drupal 6 and PHP 5.6 all the time, that should just work.

Maybe you just have the 'pdo_mysql' extension installed but not 'mysqli'? That would work with Drupal 7 & 8, I think, because they can use PDO, but not Drupal 6 - it actually needs the 'mysql' or 'mysqli' extensions.

greg-1-anderson:
@dsnopek I was going off of this:
```
php > var_export(extension_loaded('mysqli'));
true
```
Is that right, or might that be showing availability through the PDO mechanism?

Note that the tests are failing on Travis the same way they are failing on my system. Maybe you can merge this branch into yours and see if you can install what is needed or do what is needed to make the tests pass.

dsnopek:
Hm, no, I think if `extension_loaded("mysqli")` is returning `TRUE` you should be good.

That was my last good idea... :-/

greg-1-anderson:
Discovered that `site-install` works on my system with the following php.ini settings:
```
mbstring.internal_encoding = pass
mbstring.http_input = pass
mbstring.http_output = pass
```

per: https://www.drupal.org/project/drupal/issues/2353901#comment-9235025

I'll inject these in the Travis tests tomorrow and see if that allows the tests to pass.

dsnopek:
Oh, interesting! I'm willing to bet Lando and the other environments I tried had that baked into the php.ini already. I can't wait to see if this passes on Travis. :-)

greg-1-anderson:
Well drat; the .travis.yml already has `mbstring.http_input` and `mbstring.http_output` set to `pass`. I removed `mbstring.internal_encoding = pass` from my local system, and the install still worked.

There must still be something else wrong on Travis. Maybe I'll be able to find it if I change it to run the install in debug mode.

greg-1-anderson:
There were a bunch of failures, but I think I got them all. If 7a836be comes back green, then I'll rebase with 8.x and bring back the tests I disabled temporarily to speed things up.

greg-1-anderson:
Rebased with the 8.x branch. Squashed my commits down to 1, but left commits from @dsnopek and @eXistenZNL for credit.

greg-1-anderson:
Looks like some of the convenience changes I put in break with Drupal 7 and 8.

tests/Unish/UnishTestCase.php
```diff
@@ -372,13 +377,56 @@ function fetchInstallDrupal($env = 'dev', $install = FALSE, $version_string = NU
...
+      chmod("$site/settings.php", 0777);
     }
     else {
       @mkdir($site);
       touch("$site/settings.php");
     }
   }
 
+  // pm-download cannot download from d6lts, so we will rough in our own download function
+  function downloadDrupal6($drupal_6_test_version, $root) {
```
weitzman: Have you tried [drush_download_file()](http://api.drush.org/api/drush/includes%21drush.inc/function/drush_download_file/8.0.x). That has optional caching. I'm fine with keeping this if you prefer.

tests/pmDownloadTest.php
```diff
@@ -7,6 +7,9 @@
   */
 class pmDownloadCase extends CommandUnishTestCase {
   public function testPmDownload() {
+    if (UNISH_DRUPAL_MAJOR_VERSION == '6') {
+      $this->markTestSkipped("pm-* no longer supported with Drupal 6.");
```
weitzman: What motivated this? I dont have a problem with it per se but its a little odd to keep working on D6 support and neutering it at same time.

tests/Unish/UnishTestCase.php
greg-1-anderson: Is `drush_download_file()` available to Unish? Usually we avoid using Drush methods from the test framework. I was reluctant to break this pattern.

tests/pmDownloadTest.php
greg-1-anderson: There are no supported releases for Drupal 6 any longer, which makes the `pm-*` commands less useful. However, I think I went too far & may have introduced a problem that didn't formerly exist. Probably the only problem with `pm-*` is `pm-updatecode` and `pm-updatecodestatus`. I don't plan to fix these, but I shouldn't break things that are still working. I'll investigate further.

weitzman: Are there no SAs coming out for Drupal 6 contrib modules anymore? If so I guess it makes sense to stop support for those commands.

I'd like for pm-download to continue to work. Thats quite popular.

tests/Unish/UnishTestCase.php
weitzman: Ah right. Makes sense.

IMO we can start breaking that pattern in Drush 9 but lets leave 8 alone.

tests/pmDownloadTest.php
greg-1-anderson: I'm stepping back through the tests I marked as skipped and ensuring that they are skipped for a good reason. For example, the `generate make` test relies on the bootstrap theme, which isn't available for some Drupal versions. This could be fixed, but I'm just skipping it because I don't want to maintain pm or make.

I'm going to avoid breaking them, though.

dsnopek: > Are there no SAs coming out for Drupal 6 contrib modules anymore?

The D6LTS vendors publish security releases, but they are not in the update status data from Drupal.org, and there isn't a "generic" source of update status information.

myDropWizard (the company I work for) publishes alternative update status information which includes the D6LTS releases we create and it can be used by Drush via [a module](https://www.drupal.org/project/mydropwizard). But I'm not sure if that can really be used in the Drush tests, or if it's appropriate to rely on vendor-specific update status info...

greg-1-anderson: We could rely on that module in the tests, and include a pointer to it in the documentation. Let's leave that for a follow-on issue that we can do if desired.

It turned out to be unnecessary to skip most of these tests. I clarified why I was skipping the ones that remain skipped.

dsnopek:
Huzzah!

greg-1-anderson:
For huzzah part II, see #3791.

greg-1-anderson:
And huzzah part III: https://github.com/pantheon-systems/drops-6/pull/26

PR #3724, opened 2018-10-12
Unstructured data fields

greg-1-anderson:
Branched from #3718; merge that PR first. For an explanation of how to use this, see https://github.com/consolidation/output-formatters/pull/69

Depends on a dev version of output-formatters; should make a stable release there before merging here. Help text for `--fields` option could use improvement.
```
$ ./drush help site:alias
Show site alias details, or a list of available site aliases.

Examples:
  drush site:alias      List all alias records known to drush.     
  drush site:alias @dev Print an alias record for the alias 'dev'. 

Arguments:
  [site] Site alias or site specification. 

Options:
  --format[=FORMAT] Format the result data. Available formats:                                          
                    csv,json,list,php,print-r,string,tsv,var_dump,var_export,xml,yaml [default: "yaml"] 
  --fields=FIELDS   Dot notation of fields to include in output.                                        
  --field=FIELD     Select just one field, and force format to 'string'.                                
  --filter[=FILTER] Filter output based on provided expression                                          

Topics:
  drush topic docs:aliases                Creating site aliases for running Drush on remote sites.                   
  drush topic docs:output-formats-filters Output formatters and filters: how to control the output produced by Drush 
                                          commands        
```
The output-formats-filters documentation should also be updated to include a description of the new features here.

Finally, note that the operation of `--fields` varies subtly between commands that return RowsOfFields and those that return UnstructuredData. The mode of the former cannot be used by the later, but commands that return RowsOfFields could in theory work with the features currently only available to commands that return UnstructuredData.  e.g. `drush status --fields='Drupal version as vers'` does not currently work, but potentially could if there were a heuristic for casting the RowsOfFields data into an UnstructuredData object. This could potentially happen if `--fields` contained a `.` or an ` as `. Not completely sure about this, but might try it tomorrow.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
     "consolidation/annotated-command": "^2.9.1",
     "consolidation/config": "^1.1.0",
     "consolidation/filter-via-dot-access-data": "^0.4",
-    "consolidation/output-formatters": "^3.1.12",
+    "consolidation/output-formatters": "dev-unstructured-data-field-selection as 3.2.1",
     "consolidation/robo": "^1.1.5",
     "consolidation/site-alias": "^1.1.5",
     "grasmash/yaml-expander": "^1.1.1",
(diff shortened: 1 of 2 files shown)


weitzman:
Looks great. The easiest improvement to help text is to manually add a couple example Usage items to sa command.

greg-1-anderson:
Data returned as RowsOfFields or PropertyList will now be automatically converted to UnstructuredListData or UnstructuredData (respectively) if the user specifies any unstructured data access features in the `--fields` option, such as `--fields=a.b.c` or `--fields='foo as bar'`. Field restructuring does not work with `--format=table`, and automatic format adjustment is not done, so you'll usually need to add `--format=yaml` when this feature is used.

PR #3729, opened 2018-10-15
Use Symfony process instead of custom exec functions

weitzman:
A modernized version of #673.


src/Sql/SqlBase.php
```diff
@@ -277,7 +312,12 @@ public function alwaysQuery($query, $input_file = null, $result_file = '')
         // We show the query when --debug is used and this function created the temp file.
         $this->logQueryInDebugMode($query, $input_file_original);
 
-        $success = drush_always_exec($exec);
+        // This Process deliberately ignores simulated mode.
```
greg-1-anderson: Maybe we could just create a Symfony Process object directly if we don't want to respect simulated mode? That would also bypass logging, which may or may not be an issue.

weitzman: Yeah, I think the logging is useful, so I'm inclined to use our Process object.

weitzman:
Next I'd like to take a pass over sqlite and pgsql.

PR #3733, opened 2018-10-15
Set a non-zero exit status for "sql-create" on failure

byrond:
This PR ensures `drush sql-create` returns a non-zero status when it fails.  See https://github.com/drush-ops/drush/pull/3732 and https://github.com/drush-ops/drush/pull/2925.

### DIFF
--- a/commands/sql/sql.drush.inc
+++ b/commands/sql/sql.drush.inc
@@ -240,7 +240,11 @@ function drush_sql_create() {
     }
   }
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Thanks for this improvement.

byrond:
Thanks, @greg-1-anderson!  I appreciate your feedback and guidance on #3732 on the most appropriate place for this fix.  It will help us with some of the automation we're working on in our build tool.

PR #3735, opened 2018-10-15
Fix #3730. Remove patch to drupal/core 

### LINKED ISSUE #3730: drush updb fails when using services in Drush 9.5.0
**Describe the bug**
As of Drush 9.5.0, I'm no longer able to revert features using the features manager service in post update hooks. This may be related to https://www.drupal.org/project/drupal/issues/2863986, which was applied in Drush 9.5.0 as the following patch: https://www.drupal.org/files/issues/2863986-2-49.patch

An example of the error returned:

```
[notice] Update started: MODULE_post_update_1
[error]  You have requested a non-existent service "features.config.installer".
[error]  Update failed: MODULE_post_update_1
[error]  Update aborted by: MODULE_post_update_1
[error]  Finished performing updates.
```

**To Reproduce**
Include the following in a `MODULE.post_update.php` or `MODULE.install` file. You must have the features module installed for this example, though other modules that include services may have similar behaviors.

```php
$feature_list = [
  'some_feature_module',
];

\Drupal::service('features.manager')->import($feature_list, TRUE);

return t('Reverted @features', [
  '@features' => implode(', ', $feature_list),
]);
```

**Expected behavior**
`drush updb` should proceed without failing.

**Actual behavior**
`drush updb` failed with previously working code.

**Workaround**
Run `drush fr some_feature_module` to manually revert the features in question, or revert drush to 9.4.x.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.5.0
| Drupal version? | 8.5.7 and 8.6.1
| PHP version     | 5.6 and 7.2
| OS?             | Linux

weitzman:
The offending patch used to only be applied to our test system but I inadvertently made it so that Drupal installs that have extra config enable_patching: true in their composer.json extra config get the patch too. 

For now removing the patch and the test that depends on it. We'll find a better way. I made a [quick feature request in cweagans/composer-patches](https://github.com/cweagans/composer-patches/issues/237)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -125,11 +125,6 @@
       "sut/themes/contrib/{$name}": ["type:drupal-theme"],
       "sut/drush/contrib/{$name}": ["type:drupal-drush"]
     },
-    "patches": {
-      "drupal/core": {
-        "Allow updating modules with new service dependencies https://www.drupal.org/project/drupal/issues/2863986": "https://www.drupal.org/files/issues/2863986-2-49.patch"
-      }
-    },
     "branch-alias": {
         "dev-master": "9.x-dev"
     }
--- a/scenarios/isolation-phpunit4/composer.json
+++ b/scenarios/isolation-phpunit4/composer.json
@@ -105,11 +105,6 @@
       "sut/themes/contrib/{$name}": ["type:drupal-theme"],
       "sut/drush/contrib/{$name}": ["type:drupal-drush"]
     },
-    "patches": {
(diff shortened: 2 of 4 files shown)


pfrenssen:
So the `enable_patching: true` option will recursively apply all the patches from dependencies. This is indeed not the intention to have this patch applied for any other reason that running our tests.

I'll make a followup issue where we can discuss possible approaches.

PR #3738, opened 2018-10-16
Restore skipped test testUpdateModuleWithServiceDependency()

### LINKED ISSUE #3737: Restore skipped test testUpdateModuleWithServiceDependency()
One of our tests requires a patch to be applied to Drupal core in order to fix a core bug that interferes with the test. We were initially applying the patch using [cweagans/composer-patches](https://github.com/cweagans/composer-patches) but this had the unintended side effect that the patch would be applied on our user's projects if a certain flag was set. The patch was removed in #3735, and also the test was disabled to we can maintain a green build for now.

The scope of this issue is to allow this test to run again.

Possible solutions:

1. The best possible solution would be to fix the bug in core and get it committed and released. The test would then be able to run on the versions of core that include the fix. Ref. [Issue #2863986: Allow updating modules with new service dependencies](https://www.drupal.org/project/drupal/issues/2863986).
2. A second possibility is to implement a new `patches-dev` feature in `composer-patches` that allows to only apply certain patches in case a development build is being created, and these patches would not propagate to depending projects. Ref. https://github.com/cweagans/composer-patches/issues/237
3. As a quick fix we could include a couple of commands in our CI pipeline to download the patch and apply it.

pfrenssen:
Fixes #3737

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,7 @@ jobs:
       - checkout
       - run: $HOME/drush/.circleci/setup.sh
       - run: composer install
+      - run: $HOME/drush/.circleci/patch.sh
       - run: composer functional
 
   build_highest:
@@ -37,6 +38,7 @@ jobs:
       - run: composer require --dev drupal/core:8.7.x-dev --no-update
       - run: composer config platform.php 7.2
       - run: composer install
+      - run: $HOME/drush/.circleci/patch.sh
       - run: composer functional
 
   build_56:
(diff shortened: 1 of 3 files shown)


weitzman:
Any chance you can enhance the test so it gets skipped if a non-patched Drupal is detected? That way tests can pass locally as well.

pfrenssen:
I think a good way to detect it is to check whether the test module named `new_dependency_test` exists.

pfrenssen:
I added a simple check that looks if the test module that is added by the patch exists in the filesystem.

claudiu-cristea:
The trick looks good and fixes the problem.

PR #3742, opened 2018-10-16
Improve tests README

fancyweb:
- The docker php service name was changed in https://github.com/drush-ops/drush/pull/3728 but not the doc.
- There is no test named `testVersionString`, so when you execute the examples you end up with a `No tests executed!` message. IMO, this is bad for the DX. When I copy paste an example from a doc, I expect it to work. I updated the `--filter` argument with an actually existing test (I chose it randomly).


weitzman:
Thanks

PR #3744, opened 2018-10-16
Allow symfony/finder 4

fancyweb:
The `symfony/finder` component is not a core Drupal dependency. So it not technically locked to `^3.4`.

I have reviewed the usage that is made of this component in drush. The only thing that is used is the `Symfony\Component\Finder\Finder` class.

I have checked every method of this class that is used by drush. Then I have read the CHANGELOG of the component and did not see any obvious impact for drush. Then I made a diff between the `3.4` branch and the `master` branch of this class file and manually checked changes that could break drush usage of this component. I did not find any.

It would be great to accept v4 of this component, so projects that require drush and this component directly can use the latest version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
     "symfony/config": "^3.4",
     "symfony/console": "^3.4",
     "symfony/event-dispatcher": "^3.4",
-    "symfony/finder": "^3.4",
+    "symfony/finder": "^3.4 || ^4.0",
     "symfony/process": "^3.4",
     "symfony/var-dumper": "^3.4",
     "symfony/yaml": "^3.4",


greg-1-anderson:
Tests are green so I'm +1 on this. 

@weitzman: The greater question is should we ensure that we actually test with multiple versions of the Symfony components? As it is, this change means that we'll be testing with ^4 all the time. This is probably fine; we could consider adding a `lowest` test configuration to the functional tests. That wouldn't be hard, but I don't know if it would be worth the extra time. Maybe convert the php 5.6 test to a `lowest` test?

deviantintegral:
> Maybe convert the php 5.6 test to a lowest test?

👍 to this.

This comes up if you run core's command to update PHPUnit for testing on PHP 7, and then try to install drush afterwards. You have to force a downgrade with `composer require symfony/finder:^3.4` and then install drush, which works but isn't intuitive.

PR #3746, opened 2018-10-17
Fix #3741 Handle global Drush9 edge case

### LINKED ISSUE #3741: Drush tries to bootstrap the SUT site when installed standalone (out of a Drupal install) via composer
**To Reproduce**

```
$ cd /tmp
$ composer require drush/drush
$ vendor/bin/drush

PHP Warning:  require(/tmp/vendor/drush/drush/sut/../vendor/autoload.php): failed to open stream: No such file or directory in /tmp/vendor/drush/drush/sut/autoload.php on line 14
PHP Fatal error:  require(): Failed opening required '/tmp/vendor/drush/drush/sut/../vendor/autoload.php' (include_path='.:/usr/share/php') in /tmp/vendor/drush/drush/sut/autoload.php on line 14
```

**Workaround**

An installation based on `git clone` + `composer install` works as expected.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.x
| Drupal version? | -
| PHP version     | 7.1
| OS?             | GNU/Linux

weitzman:
See  - https://github.com/drush-ops/drush/issues/3741.

### DIFF
--- a/src/Config/Environment.php
+++ b/src/Config/Environment.php
@@ -59,6 +59,10 @@ class Environment
         }
 
         $this->siteLoader = require $autloadFilePath;
+        if ($this->siteLoader === false) {
+            // Nothing more to do. See https://github.com/drush-ops/drush/issues/3741.
+            return $this->loader;
+        }
         if ($this->siteLoader === true) {
             // The autoloader was already required. Assume that Drush and Drupal share an autoloader per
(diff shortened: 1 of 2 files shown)


Kingdutch:
The comment in the sut/autoload.php should probably say `use include instead of require` instead of `use include instead of return`. Especially since return is still used and this wasn't changed :)

Issue #3721, opened 2018-10-11
Array to string conversion in TsvFormatter.php on line 35


MPParsley:
**Describe the bug**
`drush locale:check` gives this notice: `Array to string conversion in TsvFormatter.php on line 35`

**To Reproduce**
`drush locale:check`

**Expected behavior**
No notice should appear.

**Actual behavior**
A notice: `Array to string conversion in TsvFormatter.php on line 35`

**Workaround**
-

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.4
| Drupal version? | 8.6
| PHP version     | 7.2.10
| OS?             | Mac

**Additional information**
At the end of the script this message appears:
`Array`

greg-1-anderson:
What output do you get with `drush locale:check --format=yaml`?  Whether or not you get this warning is likely dependent on the result returned by the command, which may vary based on the state of your Drupal installation.

MPParsley:
I get this error:
`  The "--format" option does not exist.  `

Possibly not related but note that I also encountered #3722 and https://www.drupal.org/project/drupal/issues/3005958.

greg-1-anderson:
`locale:check` does not return any data, so it should not be running the format manager / tsv formatter.

When I run this command, I get no output and no warning. Could you provide steps to reproduce from a fresh install?

MPParsley:
It might be related to the PHP version.
I'm on `7.2.10` and a colleague has the same project setup on `7.2.3` and doesn't have this issue.

The array that is sent to the format manager is a list of all modules and the manager is unsuccessfully trying to format it as a table.

I'll document the steps.

greg-1-anderson:
You're using [locale:check](https://github.com/drush-ops/drush/blob/9.4.0/src/Drupal/Commands/core/LocaleCommands.php#L45)? I'm also using php 7.2.10. local:check does not use the formatter manager, so I'm a bit perplexed as to why this is going through the tsv formatter.

MPParsley:
Yes, I'm using `drush locale:check`. See [this screencast](https://www.dropbox.com/s/m4gjk3kmdkfshpd/recording.mov?dl=0).

greg-1-anderson:
writeUsingFormatter should not be called unless this function returns true:
```
    protected function dataCanBeFormatted($structuredOutput)
    {
        if (!isset($this->formatterManager)) {
            return false;
        }
        return
            is_object($structuredOutput) ||
            is_array($structuredOutput);
[...]

PR #3750, opened 2018-10-19
batch:process experiment

greg-1-anderson:
Make batch:process return unstructured list data, and produce no output by default.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
     "consolidation/annotated-command": "^2.9.1",
     "consolidation/config": "^1.1.0",
     "consolidation/filter-via-dot-access-data": "^0.4",
-    "consolidation/output-formatters": "^3.3.1",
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
@MPParsley: Could you please test this and see if it resolves your issues with #3721?

Although UnstructuredListData behaves very similarly to an array, it is not exactly the same. I am uncertain how the return value from `batch:process` is used, and I don't know if our test coverage is complete enough to reveal any problems. If you discover anything that is broken by this PR, we could do a slightly less robust but more b/c thing and continue to just return the `array` here.

MPParsley:
@greg-1-anderson, this looks good. I tested it and no issues came up.

greg-1-anderson:
@weitzman Any thoughts on this? Do you think it's safe to change the output type of batch:process here?

weitzman:
This is going in as part of #3758

PR #3752, opened 2018-10-19
Removed link to VersionEye

bjxrn:
VersionEye was shut down at the end of 2017. See https://blog.versioneye.com/2017/10/19/versioneye-sunset-process/

The link now goes to a page which gives a certificate security warning, and if you continue redirects you to something completely different (Médica Panamericana).


greg-1-anderson:
Thanks for cleaning that up.

PR #3754, opened 2018-10-22
fix - symfony/filesystem wasn't required but was used

Laiah:
Symfony/FileSystem component is used but isn't required in composer.json file. 

Let's allow 4.0 too because this component isn't a Drupal Core dependency and the usage inside Drush is compatible with it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,6 +47,7 @@
     "symfony/config": "^3.4",
     "symfony/console": "^3.4",
     "symfony/event-dispatcher": "^3.4",
+    "symfony/filesystem": "^3.4 || ^4.0",
     "symfony/finder": "^3.4",
     "symfony/process": "^3.4",
     "symfony/var-dumper": "^3.4",


greg-1-anderson:
I don't think that Drush needs to explicitly declare direct dependencies on things brought in by our dependencies. We currently have:
```
$ composer why symfony/filesystem
chi-teck/drupal-code-generator  1.27.0   requires  symfony/filesystem (~2.7|^3)         
consolidation/robo              1.3.1    requires  symfony/filesystem (^2.5|^3|^4)      
consolidation/self-update       1.1.3    requires  symfony/filesystem (^2.5|^3|^4)      
symfony/config                  v3.4.17  requires  symfony/filesystem (~2.8|~3.0|~4.0)
```
For some components, it makes sense to me to have an explicit requirement in Drush, in case our dependency stops requiring the component we are using. However, Drush is using `consolidation/Robo` as a general Symfony framework. Robo is not going to stop using core Symfony components. I also don't think that Robo would start using symfony/filesystem:^5 (at some point in the future) unless a) doing so would be generally safe for its clients.

So, I'm generally -1 on listing each component we need unless we have specific need e.g. to declare a minimum supported version requirement. However, I care 1 out of 10. If we do move forward here, then we should probably analyze all of the dependencies of all of our dependencies (output of `composer info`) and add everything that we call directly to our composer.json file.

fancyweb:
> it makes sense to me to have an explicit requirement in Drush, in case our dependency stops requiring the component we are using

This is exactly why any project needs to explicitly require its dependencies.

Drush is not Robo. We do not know what Robo developers are going to do. They could techincally stop using this Symfony components and stay on the same major version.

> Robo would start using symfony/filesystem:^5
That's another thing that could technically happen. 

>  If we do move forward here, then we should probably analyze all of the dependencies of all of our dependencies (output of composer info) and add everything that we call directly to our composer.json file

That's exactly what we sould do to improve the quality of Drush. There are 2 other Symfony components that are used and that are not required.

weitzman:
FYI, Greg is the Robo maintainer so yes, he does know what Robo is going to do. Lets go with his preference.

PR #3755, opened 2018-10-22
fix - remove the unused "use" statements

Laiah:
I removed all the "use" statements that are not used in the files.

It would be great to use PHP code sniffer in the CI :) 'cause it pollutes the code


greg-1-anderson:
Thanks for this -- very helpful. 👍 to merge after tests pass.

We do use phpcs during the CI. If there are extra options or other tools we could use to catch things like this, we'd be open to adding more validation.

PR #3758, opened 2018-10-24
Deprecate backend.inc in favor of a new site-process library

weitzman:
See [site-process](https://github.com/consolidation/site-process).

See questions in the diff. Can be tested with `composer -v sut @example.live ssh -- -v -s`

Output shows new and then old (proc_open) shell command:
```
~/drush (siteProcess *=) $ composer -v sut @example.live ssh -- -v -s
> sut: ./drush --uri=dev '@example.live' 'ssh' '-v' '-s'
 [notice] Simulating: ssh -o PasswordAuthentication=example -t www-admin@service-provider.com cd /path/on/service-provider
Calling proc_open(ssh -o PasswordAuthentication=example -t www-admin@service-provider.com 'cd /path/on/service-provider && bash -l');
```

This is nowhere near ready - we'll be hacking more at Badcamp.


src/Commands/core/UpdateDBCommands.php
```diff
@@ -203,25 +199,25 @@ public function updateDoOne($module, $number, $dependency_map, &$context)
                     Database::startLog($function);
                 }
 
-                $this->logger()->notice("Update started: $function");
+                Drush::logger()->notice("Update started: $function");
```
greg-1-anderson: This seems bad, but if it gets the tests to green, I guess we can circle back to figuring out why the logger reference is gone or corrupt in the subprocess call later.

This is better than making a separate .inc file anyway.

src/Preflight/PreflightArgs.php
```diff
@@ -419,7 +419,14 @@ public function isBackend()
      */
     public function setBackend($backend)
     {
-        return $this->set(self::BACKEND, $backend);
+        if ($backend == 'json') {
```
greg-1-anderson: `!is_numeric($backend)` and `$args[] = "--format=$backend";`?

src/Commands/sql/SqlSyncCommands.php
```diff
@@ -127,29 +125,40 @@ public function databaseName(AliasRecord $record)
...
             'gzip' => true,
             'result-file' => $options['source-dump'] ?: 'auto',
         ];
         if (!$options['no-dump']) {
             $this->logger()->notice(dt('Starting to dump database on source.'));
-            $return = drush_invoke_process($sourceRecord, 'sql-dump', [], $dump_options, $backend_options);
-            if ($return['error_status']) {
-                throw new \Exception(dt('sql-dump failed.'));
-            } elseif (Drush::simulate()) {
+            // Set --backend=json. Drush 9.6+ changes that to --format=json. See .
```
greg-1-anderson: Missing a reference here

src/Commands/core/NotifyCommands.php
```diff
@@ -88,9 +88,10 @@ public static function shutdownSendText($msg, CommandData $commandData)
             }
         }
 
-        if (!drush_shell_exec($cmd, $msg)) {
-            throw new \Exception($error_message . ' ' . dt('Or you may specify an alternate command to run by setting notify:cmd in a drush.yml file'));
-        }
+        // Keep backward compat and prepare a string here.
+        $cmd = sprintf($cmd, Escape::forSite(Drush::aliasManager()->getSelf(), $msg));
+        $process = Drush::process($cmd);
```
greg-1-anderson: Why not omit Escape::forSite and just:
`$process = Drush::process($cmd, [$msg])`?

weitzman: the 2nd argument to process() is cwd, not a replacement array. And if there were two items in the $cmd array, it would quote the first long item which is invalid I think.

greg-1-anderson: My mistake. I meant:
```
$process = Drush::process([$cmd, $msg]);
```
That will escape the commands for you. Hm, that uses Symfony escaping instead of our improved escaping. Maybe we should have Drush::process use the Escape utilities to escape all of the parameters for the local machine. `Escape::shellArg()` is the API to use to just escape for the local system.

PR #3759, opened 2018-10-26
Test reorg

greg-1-anderson:


sut/drush/sites/synctest.site.yml
```diff
@@ -0,0 +1,14 @@
+remote:
```
weitzman: I suspect that this file, and the similar one above, were committed in error. The current sut does not have these. I think they are generated by the tests and cleaned up at the end.

Issue #3408, opened 2018-02-21
Support remote-to-remote aliases on sql-sync/rsync


mikeryan776:
#3046 (and at least one other issue I've lost track of) references the removed functionality in Drush 9 of being able to sync directly between two remote aliases, and the desire to restore it at some point. I can't find an open issue for implementing this, so (unless someone can point me to one) let this stand as that feature request.

Context: the workflow [described by geerlingguy](https://www.jeffgeerling.com/blog/2017/soup-nuts-using-drupal-vm-build-local-and-prod) for syncing between a remote production server and local DrupalVM instance.

weitzman:
Try the workaround mentioned in that issue `drush @synctest.remote2 rsync @synctest.remote1 @self`. That, run the command at a remote so that your sql:sync is now using 1 remote and a @self.

joelpittet:
There are are no keys in the car... I mean VM. The work around seems to not work because from within the VM there are no SSH keys to the remote servers. This used to work in 8.x (still does).

I can explicitly work around the `rsync` because the files are mounted, but not sure how to work around the `sql-sync`?

`rsync` workaround:
```
drush rsync @self.prod:%files ./web/sites/default/files```

Tawreh:
Up until very recently we've been able to sync databases between two environments on Acquia using the following command: 

    $ drush sql:sync @site.prod @site.test --structure-tables-key=common 

This isn't working anymore with the error: 

    [notice] Starting to dump database on source.
    [success] Database dump saved to /home/site/drush-backups/site_backup_20181019_072149.sql.gz
    [notice] Starting to discover temporary files directory on target.
    [error]  Cannot specify two remote aliases. Instead, use one of the following alternate options:

       `drush  rsync @self `
       `drush  rsync @self site.test@sitestg.ssh.prod.acquia-sites.com:/location/site_backup_20181019_072149.sql.gz

    Use the second form if the site alias definitions are not available at . 
    [notice] Copying dump file from source to target.
    [error]  core-rsync failed. 

This has only become an issue within the last month or so and may be related to Acquia updating Drush (??) but it's making it difficult to copy databases around. I'd like to avoid copying certain tables (CRM data) so that we can have different test data per environment. 

I'm also not running `rsync` so I don't think I can use the suggestions provided but I'm a little stuck as to where to go from here.

greg-1-anderson:
[...]

PR #3760, opened 2018-10-26
Generate full path for rsync only if alias is remote.

mirsoftacquia:
This should fix drush sql:sync if the local alias is different from @self (needed for most of drupal multisite project where each site uses different alias but they are all in fact in local). It should be also starting point to support sync between two remotes (possible, with tunnelling rsync transfer through local).

This is related to the discussion in https://github.com/drush-ops/drush/issues/3408 , where I will write the closer comment.

Please forgive my ignorance but it was not in my time possibilities to set up the test suite and it is very likely the SqlSyncTest.php will need to be updated as well. I would appreciate if someone from drush authors could take over that.

### DIFF
--- a/src/Commands/sql/SqlSyncCommands.php
+++ b/src/Commands/sql/SqlSyncCommands.php
@@ -204,9 +204,14 @@ class SqlSyncCommands extends DrushCommands implements SiteAliasManagerAwareInte
             if (($runner == 'target') || ($runner == 'destination')) {
                 $runner = $targetRecord;
             }
+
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Drush used to consider `localhost` and `127.0.0.1` to be the same as a local alias, but folks who develop using a docker container need to address localhost on some alternate port as a remote system (i.e. accessed via ssh).

Does it work for your use case to simply remove `host` and `user` from your local alias records? That is the expected convention.

PR #3764, opened 2018-10-29
Fixes #3761. Taking into account unreadable output of 'stty size' when...

### LINKED ISSUE #3761: Drush outputs malformed tables when run in Docker container, when allocating pseudo-tty and piping/redirecting output of docker exec.
**Describe the bug**
When run in docker container, by using `docker exec`, assuming `docker exec` has allocated pseudo-tty with `-t` flag, and output of the command is being piped or redirected away, `Drush` runs with COLUMNS=0, which breaks output of tables returned from `docker exec`.

**To Reproduce**
Create docker container with drush, and run any drush command outputting tables with `docker exec -t`, then redirect/pipe output. The output will be messed up.

**Expected behavior**
Properly formatted tables.

**Actual behavior**
Tables are malformed because drush assumes that COLUMNS=0.

**Workaround**
You can avoid `-t` flag of `docker exec`. Without allocating of pseudo-tty drush output is correct. In the case of docker-compose, there is `-T` flag disabling pseudo-tty allocation that also fixes the problem. There may be some other undesirable effects to that though.

You can also pass the "COLUMNS" environmental variable to docker, to set this value manually.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.x
| Drupal version? | 7.x
| PHP version     | 7.1
| OS?             | Linux

**Additional information**
Originally described here: https://stackoverflow.com/questions/53027036/why-piping-output-of-docker-compose-exec-to-grep-breaks-it , I'm creating this issue after diagnosing the exact cause existing in Drush.

lukasz-zaroda:
…determining number of columns, to avoid unreadable tables.

It just makes additional check for already existing assumption. Without it, we are getting unexpected state which in some situations causes COLUMNS to be assumed to be 0, which breaks display of tables.

### DIFF
--- a/includes/preflight.inc
+++ b/includes/preflight.inc
@@ -441,7 +441,7 @@ function _drush_preflight_columns() {
   if (!($columns = getenv('COLUMNS') ?: 0)) {
     // Trying to export the columns using stty.
     exec('stty size 2>&1', $columns_output, $columns_status);
-    if (!$columns_status) $columns = preg_replace('/\d+\s(\d+)/', '$1', $columns_output[0], -1, $columns_count);
+    if (!$columns_status && is_numeric($columns_output[0])) $columns = preg_replace('/\d+\s(\d+)/', '$1', $columns_output[0], -1, $columns_count);
 
     // If stty fails and Drush us running on Windows are we trying with mode con.
     if (($columns_status || !$columns_count) && drush_is_windows()) {


greg-1-anderson:
What is the actual output of `stty size` in your use case?

lukasz-zaroda:
In my docker container output of `stty size` is `stty: standard input`, but only in specific circumstances, like when I'm using allocation of pseudo-tty + pipe/redirection.

includes/preflight.inc
```diff
@@ -441,7 +441,7 @@ function _drush_preflight_columns() {
   if (!($columns = getenv('COLUMNS') ?: 0)) {
     // Trying to export the columns using stty.
     exec('stty size 2>&1', $columns_output, $columns_status);
-    if (!$columns_status) $columns = preg_replace('/\d+\s(\d+)/', '$1', $columns_output[0], -1, $columns_count);
+    if (!$columns_status && is_numeric($columns_output[0])) $columns = preg_replace('/\d+\s(\d+)/', '$1', $columns_output[0], -1, $columns_count);
 
     // If stty fails and Drush us running on Windows are we trying with mode con.
     if (($columns_status || !$columns_count) && drush_is_windows()) {
```
greg-1-anderson: I think it would be better to fix the problem here, as I can imagine instances where is_numeric is true, but the preg_replace still does not work correctly. Just change this so that default behavior happens when $columns comes out as zero (or less).

lukasz-zaroda: What about lines 459-461? We could maybe change it into:
```
     if (empty($columns) || !is_numeric($columns)) {
       $columns = 80;
     }
```
That would cover all wrong values of $columns I think.

lukasz-zaroda:
Hmm, am I the only one that has an issue with the "Start conversation" button? I couldn't start conversation on line 459, some JS error in console. :/

includes/preflight.inc
```diff
@@ -441,7 +441,7 @@ function _drush_preflight_columns() {
   if (!($columns = getenv('COLUMNS') ?: 0)) {
     // Trying to export the columns using stty.
     exec('stty size 2>&1', $columns_output, $columns_status);
-    if (!$columns_status) $columns = preg_replace('/\d+\s(\d+)/', '$1', $columns_output[0], -1, $columns_count);
+    if (!$columns_status && is_numeric($columns_output[0])) $columns = preg_replace('/\d+\s(\d+)/', '$1', $columns_output[0], -1, $columns_count);
```
lukasz-zaroda:
```suggestion
    $matched = FALSE;
    if (!$columns_status) {
        preg_match('/\d+\s(\d+)/', $columns_output[0], $matches, 0);
        if (isset($matches[1])) {
            $columns = $matches[1];
            $matched = TRUE;
        }
    }
```

includes/preflight.inc
lukasz-zaroda:
```suggestion
    if (($columns_status || !$matched) && drush_is_windows()) {
```

includes/preflight.inc
greg-1-anderson: Looks good in concept. Maybe `$matched = preg_match...` would be sufficient if you made the regex bounded with `^` and `$` for an exact match.

lukasz-zaroda:
```suggestion
    $matched = FALSE;
    if (!$columns_status && $matched = preg_match('/^\d+\s(\d+)$/', $columns_output[0], $matches, 0)) {
        $columns = $matches[1];
    }
```
Good idea, it could be improved like this.

greg-1-anderson:
The new github feature for making code suggestions is kind of neat, but not really necessary for commenting on your own code. What you're doing is fine; could you just roll your suggestions into the actual PR?

lukasz-zaroda:
Sorry, had to play with a new toy for a second :) , it's included now.

greg-1-anderson:
LGTM - thanks for fixing this.

PR #3765, opened 2018-10-30
Pin consolidation/site-alias to 1.1.5 until Drush supports 1.1.6.

pingers:
Drush 9.5.2 can install consolidation/site-alias 1.1.6, which is incompatible. We can pin the version until those changes go in to a Drush release.

See https://github.com/consolidation/site-alias/pull/11#issuecomment-433678408

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
     "consolidation/config": "^1.1.0",
     "consolidation/output-formatters": "^3.1.12",
     "consolidation/robo": "^1.1.5",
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
This only helps the master branch, and I hope to have this fixed before the next stable release. Folks can use this branch as a guide if their Drush is broken.

greg-1-anderson:
Fixed in https://github.com/consolidation/site-alias/pull/15

PR #3766, opened 2018-10-30
Fix bug where pm:security misses required security updates.

anemirovsky:
The bug here is that 'packages-dev' and 'packages' are both arrays with
numeric keys. Using the '+' operator on them, according to PHP's
documentation (https://secure.php.net/manual/en/language.operators.array.php),
means that if a key exists in both, whatever array is on the right-hand side
will have that key's value ignored and the left-hand side's value will be
used.

To ensure that the 'packages' array is appended to the 'packages-dev' array,
we have to use array_merge, which always appends numeric keys
(https://secure.php.net/manual/en/function.array-merge.php).


weitzman:
Thanks for tracking this down.

Issue #3749, opened 2018-10-18
drush pm:security -n wrong status-code


asodastreamer:
**Describe the bug**
drush returns a statuscode indicating that there is an error, although there is none.
This **wil** terminate good scripts tasked with updating a drupal-site in the moment when they need to run flawlessly.

**To Reproduce**
Install a Drupal-Site expecting a core-update e.g. 8.6.1.
Run `drush pm:security -n ; echo $?` .

**Expected behavior**
I expect to see a non-zero status-code only when there is an error causing drush not to perform the task requested, because of varying conditions.
Running `drush pm:security -n` should not return non-zero when there is an update as getting an update is expected behaviour and not an error.

**Actual behavior**
A script written to stop when an error occures as to not do any damage will not be able to perform an update.
e.g.

`
function handleErrors() {
$statusCode=$?;
......
exit $statusCode;
}
trap handleErrors ERR

.....
......drush pm:security -n --pipe......
..... update-action never executed.........
`

**Workaround**
Disable Trap with `trap - ERR`.
This has the drawback that you will not be able to respond to an actual error happening.

### System Configuration
Linux
Drupal 8.6.1
Drush Commandline Tool 9.2.3
PHP 71

**Additional information**
None.

weitzman:
I think current behavior is also fine, but lets keep this open for a while to collect opinions/feedback.

asodastreamer:
> I don't quite recall the history, but this is perhaps by design.
> 
> We could consider introducing an option to allow the client to override the default error reporting strategy. We shouldn't change the default until a major release, as that would be a breaking change.

That would be awesome, because working with it in a bash-like-manner is not really reliable that way, because all of the sudden there is a corner-case popping up and handling it completely separate from the rest just because of the status-code makes it all the more complicated.

greg-1-anderson:
A pull request to kick things off would be welcome.

weitzman:
I'm not so sure that this is something I'd like to make configurable. I'm hesitant to add switches for core plumbing.

Also, the exit code of this command was deliberately chosen - #3087. Its convenient for CI to work this way.

asodastreamer:
> I'm not so sure that this is something I'd like to make configurable. I'm hesitant to add switches for core plumbing.
> 
> Also, the exit code of this command was deliberately chosen - #3087. Its convenient for CI to work this way.

Hi, not sure about this but apparently I cannot find anything in regards to  the exit-code.
[...]

PR #3768, opened 2018-11-01
Add a command to check if Drupal has been bootstrapped.

floretan:
There are many cases where we need to check if Drupal has been installed from scripts, for example:

- In a deployment script, run updates if Drupal is already running, make a clean install otherwise
- In a Kubernetes readiness check, start serving traffic to a pod only when Drupal has been installed

At the moment, there is no dedicated command to check for the bootstrap phase, one needs to run `drush status` and parse the output looking for the "Successful" string, which is not ideal.

This PR adds a set of commands that execute successfully if Drupal has been bootstrapped to a certain level. If not specified the default level is "full", as that's probably the most common use case.

### DIFF
--- a/src/Commands/core/CheckBootstrapCommands.php
+++ b/src/Commands/core/CheckBootstrapCommands.php
@@ -1,0 +1,88 @@
+<?php
+
+namespace Drush\Commands\core;
+
+use Drush\Drush;
+use Consolidation\SiteAlias\SiteAliasManagerAwareInterface;
+use Consolidation\SiteAlias\SiteAliasManagerAwareTrait;
+
+class CheckBootstrapCommands extends DrushCommands implements SiteAliasManagerAwareInterface
+{
+    use SiteAliasManagerAwareTrait;
+
+    /**
+     * This command executes successfully if Drupal has been fully bootstrapped.
+     *
+     * @see DRUSH_BOOTSTRAP_DRUPAL_FULL
+     *
+     * @command check-bootstrap:full
+     * @aliases check-bootstrap, cb
+     * @bootstrap max
+     */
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks. Have you tried `./drush st --format=string bootstrap`. That returns only the word 'Successful' as appropriate. I don't think Drush core needs a new command for this. I would accept a patch that adds this as a Usage example.

To see all the keys that `status` offers, run `drush status --format=yaml --fields=*`

floretan:
Yes, we used something quite similar:
```
bootstrapped=$(drush status --field=bootstrap);
if [[ $bootstrapped = "Successful" ]];
then
  // do something.
fi
```
Having to parse the output is quite verbose and also error prone. The proposed command would make the same script much more clear and understandable:
```
if drush check-bootstrap;
then
  // do something.
fi
```

Another option would be to update `drush status` so that it would return a non-zero exit code when a specific field is selected and that field is empty. Changing the behavior of a command that has been around for a long time is generally not a good idea though.

floretan:
One more variation that came up in a discussion at DrupalCamp Baltics: we could add a flag to `drush status` that would make it return an execution code according that reflects what is being output. Would you be open to including this kind of change?

greg-1-anderson:
Related status code discussion in https://github.com/drush-ops/drush/issues/3749

weitzman:
If you just want a proper exit code, you can run any full bootstrap command like `drush pml`. If you get a 0 back, you are installed. 

I see that you would rather this be fixed in a "right" way. I'm inclined to say that the current methods are good enough, and I'm not really looking to improve it with a new command or configurable exit code. As I said, a Usage example would be welcome.

Issue #3725, opened 2018-10-12
Getting "A site alias is required. The way you call ssh command has changed" with drush 8.1.17


lmakarov:
**Describe the bug**
`drush @alias ssh` is broken in Drush 8.1.17, when Drush is installed as a project level Composer dependency.

**To Reproduce**
- `composer require drush/drush:8.1.17`
- `drush @alias ssh`

**Expected behavior**
SSH session is successfully established with the remote host.

**Actual behavior**

```
$ drush @alias ssh
A site alias is required. The way you call ssh command has changed to `drush @alias ssh`.
```

**Workaround**

There are two options:

1. Downgrading the project level drush dependency to `8.1.16`

```
composer require drush/drush:8.1.16
```

2. Removing the project level drush and installing a global one (works even with `8.1.17`)

```
composer remove drush/drush
composer global require drush/drush:8.1.17
```

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.1.17
| Drupal version? | 8.5.6
| PHP version     | 7.1
| OS?             | Linux

**Additional information**
Other reports of this issue:
- https://github.com/drush-ops/drush/pull/3428#issuecomment-424654547
- https://github.com/amazeeio/lagoon/issues/542

Presumably caused by: https://github.com/drush-ops/drush/pull/3428

nickveenhof:
I'll try to reproduce this. But we haven't seen this happen yet. Are you using any specific alias file? Can you also provide the full debug output of the command? Eg drush with the  -vvv flags. This way we can follow along what happens.

lmakarov:
Here are more details:

```
docker@cli:/var/www/source/drupal$ drush --version
Drush Launcher Version: 0.6.0
 Drush Version   :  8.1.17 
```

```
docker@cli:/var/www/source/drupal$ drush sa @bhgeacsf.01dev
$aliases["bhgeacsf.01dev"] = array (
  'root' => '/var/www/html/bhgeacsf.01dev/docroot',
  'ac-site' => 'bhgeacsf',
  'ac-env' => '01dev',
  'ac-realm' => 'enterprise-g1',
  'uri' => 'bhgeacsf01dev.enterprise-g1.acquia-sites.com',
  'remote-host' => 'bhgeacsf01dev.ssh.enterprise-g1.acquia-sites.com',
  'remote-user' => 'bhgeacsf.01dev',
  'path-aliases' => 
  array (
    '%drush-script' => 'drush8',
  ),
);
[...]

PR #3777, opened 2018-11-09
Fix ssh to site alias

rocketeerbkw:
There have been many reports of ssh issues with 8.1.17. See #3725. I believe the changes in #3428 are the source of the breakage.

I'm able to connect successfully if I use `--local`: `drush --local @alias ssh`. I'm also able to connect if I use only a global drush. The combination of a global drush calling a local drush breaks.

I find myself very confused with the variable names. The command setting `handle-remote-commands` sounds to me like it should run the command remotely, but the docs say ```set to TRUE if `drush @remote mycommand` should be executed locally rather than remotely dispatched```.

The `ssh` command has `handle-remote-commands` set to `TRUE` so it should always be executed locally.

In the changes for #3428, a new variable `$must_use_site_local` is defined. To me it sounds like that means the command must be executed locally, but the code checks that drush `--local` is **not** set.

Based on the "global+local drush" case, the previous code and the variable names, I believe the bug is a missing `!` check. The `$must_use_site_local` should check that `--local` is `!empty()` instead of `empty()`.

I will provide test setup details and logs of 8.1.16 working and 8.1.17 broken in the first comment.

### DIFF
--- a/includes/preflight.inc
+++ b/includes/preflight.inc
@@ -613,7 +613,7 @@ function drush_preflight_command_dispatch() {
     $local_drush = find_wrapper_or_launcher($root);
   }
   $is_local = drush_get_option('local');
-  $must_use_site_local = !empty($root) && !empty($local_drush) && empty($is_local);
+  $must_use_site_local = !empty($root) && !empty($local_drush) && !empty($is_local);
   // If the command sets the 'handle-remote-commands' flag, then we will short-circuit
   // remote command dispatching and site-list command dispatching, and always let
   // the command handler run on the local machine.


rocketeerbkw:
I've revised my PR based on a better understanding. It now passes tests related to the ssh and site alias functions.

The current test failures reported in Travis are ones I cannot reproduce locally.

greg-1-anderson:
Thanks for tracking this down. I re-ran the tests and they came back green.

ayalon:
@greg-1-anderson : Can you please create a new version with that fix? Drush 8.1.18 ist still broken and we have to constantly downgrade drush on all installations because drush @alias ssh is not working. Thanks a lot.

rocketeerbkw:
@ayalon you can use `drush --local @alias` instead of downgrading everything until this gets released

ayalon:
Thanks. But we us drush @alias ssh in all our deployment scripts. So its easier to downgrade everywhere instead of changing all the deployment scripts.

greg-1-anderson:
Yeah I have been on the verge of a Drush 8 release for a couple of weeks now. Sorry for the delay, will try to wrap it up and get it out soon.

PR #3780, opened 2018-11-10
Remove DRUSH_EXECUTION_COMPLETED and DRUSH_EXIT_CODE handling

### LINKED ISSUE #3749: drush pm:security -n wrong status-code

weitzman:
I think its OK to remove these during a minor release since they are long ancient and non-recommended APIs. Happy to discuss.

One notable change is that pm:security now exits with 0 when there an update is found. This fixes #3749. I had to do this since I can no longer set a 1 exit code AND print my structured output, even when the command returns a ResultData object.


greg-1-anderson:
Are we just giving up on the "Drush command terminated abnormally due to an unrecoverable error"  feature? While that caused a lot of confusion, not having any indication that the website called `exit()` might be even worse.

I'm not sure about changing the exit-status behavior of pm:security in a minor release, though. This could interfere with folks' security-update checking scripts. It might be better to enhance annotated command / output formatters to allow for returning structured data with an exit code.

greg-1-anderson:
Here's a PR in annotated command that would allow both an exit code and result data to be returned by the same annotated command:

https://github.com/consolidation/annotated-command/pull/167

If this works well for our purpose, then I could make a stable release with this addition.

weitzman:
That looks great. Thanks Greg.

greg-1-anderson:
I'm okay with losing "Drush command terminated abnormally due to an unrecoverable error" if you really want to, but I sort of think we should keep it. We could try without for a while and see how it goes.

weitzman:
I brought it back in recent commits. https://github.com/drush-ops/drush/pull/3780/files#diff-00e1d309f15af5b596b61de17b456130R36

PR #3792, opened 2018-11-21
Forward-port #3720: Do not force scheme to 'mysql'.

greg-1-anderson:
While this was only necessary for Drupal 6, it seems that leaving this in place is unnecessary at best, and worst case might be a time-bomb that bites us later.


greg-1-anderson:
mysqli is build-in on all versions of php from 5.0 onward, according to https://en.wikipedia.org/wiki/MySQLi, so I think this is safe to break. The version on the 8.x branch is more conservative, but in the other direction (does not allow Drush to switch to `mysql` unless that driver is available)

greg-1-anderson:
@weitzman Should we merge this, or forget about it? I don't feel strongly in either direction.

weitzman:
Lets skip this.

PR #3794, opened 2018-11-22
Remove unused code from bootstrap

weitzman:
Primarily the d6 and d7 boot candidates are removed. I thought we used these to provide a friendly error message when you run against d6 or d7. We dont even do that - I'm guessing we lost that in Drush 9. Its not really needed with drush9 not being a supported launcher anymore. So, lets just toss all this code.


weitzman:
After this PR we have a rather odd class hierarchy where there is only ever one of Boot < DrupalBoot < DrupalBoot8. I'd love input into how this should work in reparation for moving to https://github.com/consolidation/bootstrap

PR #3795, opened 2018-11-22
Remove context.inc usage in Boot classes

weitzman:
Main change is that the bootstrap phase is now tracked with a property on the Boot class.

This builds upon the other PR from today #3794

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -57,6 +57,7 @@
  * drush_backend_get_result().
  */
 
+use Drush\Drush;
 use Drush\Log\LogLevel;
 use Drush\Preflight\PreflightArgs;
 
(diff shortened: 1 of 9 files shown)


weitzman:
I found one more call to context. 

https://github.com/drush-ops/drush/blob/01cce0838c68efde84b811965d085bc3c9ec387d/src/Boot/DrupalBoot8.php#L203

I personally have never used this 'ignored-modules' feature and am not sure it still works. We long ago commented out the tests for it. If anyone wants to keep it, please suggest how to reimplement it.

https://github.com/drush-ops/drush/blob/049d2a4b36338178c0c7cf89b8e7caf9769524c9/tests/functional/AnnotatedCommandTest.php#L216-L220

greg-1-anderson:
I don't use `ignored-modules`, and can't remember the use-case for it. Seems dubious in the era of the service container.

PR #3796, opened 2018-11-23
fix locale:import alias

ms-slasher13:


weitzman:
The command name is the same as the alias. That also looks incorrect. I guess the alias should be removed.

Sutharsan:
I think this is a remainder from from copying the locale-import docblock.
Agree to make this change.

TessaBakker:
Thanks for the fix @ms-slasher13

PR #3800, opened 2018-11-24
Use Drupal DB during bootstrap instead of CLI

weitzman:
Removes a dependency on the SQL classes and Drush users may no skip installation of `mysql` CLI client if they don't want to run sql commands. This is nice for mnimalist Docker pwoered environments.

### DIFF
--- a/src/Boot/DrupalBoot.php
+++ b/src/Boot/DrupalBoot.php
@@ -169,34 +169,6 @@ abstract class DrupalBoot extends BaseBoot
      */
     public function bootstrapDrupalDatabaseValidate()
     {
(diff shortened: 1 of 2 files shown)


weitzman:
Unfortunately, site:install uses DB API so we can't quite install a site without the `mysql`. This is doable with a bit more work, unless the database needs to be created. I'll see what I can do.

weitzman:
I dont think we need to tackle site:install right now. We've done the job of further compartmentalizing the Boot classes.

greg-1-anderson:
Looks like we maybe need to catch an exception so that bootstrap max can work cleanly when the requisite table does not exist.

I agree that this is a net benefit even if we don't get site:install converted right away.

PR #3811, opened 2018-12-01
Collect log messages in Logger instead of in context system.

weitzman:
And instruct backend to use the new methods.


greg-1-anderson:
We could probably also use a way for the integration tests to clear all of the cached log messages.

Bonus points: have caching of log messages disabled by default, and only enable it in backend mode. The integration tests currently do not test backend mode, so clearing becomes unimportant. Clearing would still be nice to have, though, in case we enhance the integration tests to handle backend later.

PR #3821, opened 2018-12-10
Use Circle 2.1 - Build Processing also enabled in CircleCI prefs

weitzman:


greg-1-anderson:
Fine by me, but what affect does this have?

greg-1-anderson:
This is neat, but the downside is that the Mysql test gets a Postgres container it doesn't need, the Postgres test gets a Mysql container it doesn't need, and the sqlite test gets Mysql and Postgres containers it doesn't need.

Is there a way to easily conditionalize the inclusion of additional containers? If not, I'd prefer extra duplication in the CircleCI config.yml file to using resources we don't need or writing code to spin up docker containers ourselves.

Overall -1 on this.

PR #3822, opened 2018-12-10
Add sut alias record to repository rather than building it dynamically

greg-1-anderson:
We don't need to dynamically generate these any more.

### DIFF
--- a/sut/drush/sites/sut.site.yml
+++ b/sut/drush/sites/sut.site.yml
@@ -1,0 +1,5 @@
+common:
+  root: /Users/ganderson/local/drupal/drush9/drush/sut
+'*':
+  uri: ${env-name}
+  dbUrl: 'mysql://root:@127.0.0.1/unish_${env-name}'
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
I should point out that this is using [wildcard aliases](https://github.com/consolidation/site-alias#wildcard-environments-experimental), which means that we have a new feature. `drush @sut.foo st` is the same as `drush --root=/path/to/sut --uri=foo st`. Drupal won't find a `foo` directory, so it will select `default`. The database creds will be read from settings.php rather than taken from the alias. You can create `foo` via site:install.

If this is undesirable, we can explicitly list all of the environments rather than use the wildcard.

sut/drush/sites/sut.site.yml
```diff
@@ -0,0 +1,5 @@
+common:
+  root: /Users/ganderson/local/drupal/drush9/drush/sut
```
weitzman: Is this a mistake? If not please add a comment. My path is different. 

Also add link to wildcard aliases docs.

greg-1-anderson: That's a mistake.

PR #3823, opened 2018-12-11
Allow integration tests to test commands that have confirmations

greg-1-anderson:
Pass --yes to confirm.

### DIFF
--- a/tests/integration/WatchdogTest.php
+++ b/tests/integration/WatchdogTest.php
@@ -5,13 +5,15 @@ namespace Unish;
 /**
  * @group commands
  */
-class WatchdogCase extends CommandUnishTestCase
+class WatchdogTest extends UnishIntegrationTestCase
 {
 
     public function testWatchdog()
     {
-        $this->setUpDrupal(1, true);
-        $this->drush('pm-enable', ['dblog']);
+        $this->drush('pm-enable', ['dblog'], [], self::IGNORE_EXIT_CODE);
+        $this->drush('watchdog-delete', ['all'], ['yes' => true]);
+        $output = $this->getErrorOutput();
+        $this->assertContains('All watchdog messages have been deleted', $output);
 
         $eval1 = "\\Drupal::logger('drush')->notice('Unish rocks.');";
         $this->drush('php-eval', [$eval1]);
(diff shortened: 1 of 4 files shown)


tests/unish/UnishIntegrationTestCase.php
```diff
@@ -133,7 +133,7 @@ protected function buildCommandLine($command, $args, $options)
         }
         // insert drush command options
         foreach ($options as $key => $value) {
-            if (!isset($value)) {
+            if (!isset($value) || $value === true) {
```
greg-1-anderson: Usually, `'yes' => true` gives `--yes`, and `'yes' => '1'` gives `--yes=1`. I've always found the Unish-style `'yes' => null` to be confusing. Would it be okay to switch to `true` everywhere?  We should also support `null` for external unish clients.

greg-1-anderson:
Just one trivial question before merging. We can stick with `'yes' => null` if you feel strongly about it.

tests/unish/UnishIntegrationTestCase.php
weitzman: Yes feel free to change that.

tests/unish/UnishIntegrationTestCase.php
```diff
@@ -102,7 +102,7 @@ public function drush($command, array $args = [], array $options = [], $expected
     protected function buildCommandLine($command, $args, $options)
     {
         $global_option_list = ['simulate', 'root', 'uri', 'include', 'config', 'alias-path', 'ssh-options', 'backend', 'cd'];
-        $options += ['root' => $this->webroot(), 'uri' => self::INTEGRATION_TEST_ENV]; // Default value.
+        $options += ['root' => $this->webroot(), 'uri' => self::INTEGRATION_TEST_ENV, 'strict' => 0]; // Default value.
```
weitzman: Why disable strict mode?

tests/integration/WatchdogTest.php
```diff
@@ -5,13 +5,15 @@
...
  */
-class WatchdogCase extends CommandUnishTestCase
+class WatchdogTest extends UnishIntegrationTestCase
 {
 
     public function testWatchdog()
     {
-        $this->setUpDrupal(1, true);
-        $this->drush('pm-enable', ['dblog']);
+        $this->drush('pm-enable', ['dblog'], [], self::IGNORE_EXIT_CODE);
```
weitzman: Why ignore exit code here?

tests/unish/UnishIntegrationTestCase.php
greg-1-anderson: Disabling strict mode bypasses argument validation. We can try without this.

tests/integration/WatchdogTest.php
greg-1-anderson: We might enable modules again and again. I guess that enabling an already-enabled module is not an error, so we can remove this.

greg-1-anderson:
It worked to remove strict mode; it was also fine to remove the ignore errors flag.

PR #3827, opened 2018-12-14
Allow integration tests to feed stdin to Drush commands.

greg-1-anderson:
One more thing that's now easy to do in the tests.


tests/unish/UnishIntegrationTestCase.php
```diff
@@ -99,6 +104,13 @@ public function drush($command, array $args = [], array $options = [], $expected
         return $return;
     }
 
+    protected function setStdin($contents)
+    {
+        $path = $this->writeToTmpFile($contents);
```
weitzman: If you write to `getSandbox()`, then cleanDirs() will clean up automatically. Would remove all changes in UnishTestCase ... Just a thought.

greg-1-anderson: Yeah I like to run locally with UNISH_DIRTY, and I still want these to go away. Maybe segregate the sandbox into invariant-deletion and only-if-not-dirty deletion would work

weitzman: Sure, an alwaysDelete subdir of the sandbox sounds good

greg-1-anderson:
Better?

PR #3838, opened 2019-01-02
Fixed config location prompt has an unnecessary full stop

joachim-n:
There should not be a full stop here, as the framework adds a colon:

```
 Choose a source.:
  [0] Cancel
  [1] sync
  [2] vcs
```


weitzman:
Thanks.

PR #3842, opened 2019-01-04
Use process manager

greg-1-anderson:
c.f. https://github.com/consolidation/site-process/pull/21

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "37071355a8551900672d4ca3017c0cc4",
+    "content-hash": "ce2c6da1a024e7dcb692af29fcc24b79",
     "packages": [
         {
             "name": "chi-teck/drupal-code-generator",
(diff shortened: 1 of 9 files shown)


src/SiteAlias/ProcessManager.php
```diff
@@ -0,0 +1,112 @@
...
+        }
+        array_unshift($args, $siteAlias->get('paths.drush-script', $defaultDrushScript));
+
+        return $this->siteProcess($siteAlias, $args, $options, $options_double_dash);
+    }
+
+    /**
+     * @inheritdoc
+     *
+     * Use ProcessManager::drush() instead of this method when calling Drush.
```
weitzman: This should be Drush::drush() still.

composer.lock
```diff
@@ -696,12 +696,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/consolidation/site-process.git",
-                "reference": "cf749637dd63e161768fc707c7456fc138df269c"
+                "reference": "3d198ec83c4bbe11851bd934ed839d03944c3eec"
```
weitzman: I guess you will get us back on a stable release of site-process soon.

src/SiteAlias/ProcessManager.php
greg-1-anderson: If a class can inject a reference to the process manager, then it should prefer this method over Drush::drush(). Both will be available, though, so I'll add Drush::drush() back to the comment.

PR #3848, opened 2019-01-07
Issue #1523: Add schema support for PostgreSQL - test.

RoSk0:
DO NOT MERGE!!! TEST ONLY BRANCH!!!


RoSk0:
Finally those tests fail as expected...

But now, the way I wrote them, this depends on #3849

weitzman:
We should not need a testing PR thats different from the code change PR.

PR #3849, opened 2019-01-07
DB prefix ignored when installing site.

RoSk0:


RoSk0:
Test only, will add fix when test fail.

RoSk0:
Tests failed.

Pushed fix.

weitzman:
The fix looks good. The test only runs on postgres. Is that needed? I would think we would want a universal test ... I'd be OK with no test here if thats easier.

RoSk0:
Hi Moshe,

I removed special test behaviour for PostgreSQL. I was too influenced by #1523 when I wrote that test.

weitzman:
This was reverted as it caused sqlite tests to fail.

RoSk0:
OK, traced it down to src/Sql/SqlSqlite.php::creds(). SQLite driver ignore prefix completely when it should treat as core driver, adding as suffix to the DB file name (see Drupal\Core\Database\Driver\sqlite\Connection::__construct() ).

So we cannot rely on actual table name being same across different DBs...

Question: should we remove SQLite from this test? Or should we modify test conditions for it?

RoSk0:
Excluded SQLite from test and opened follow up #3912.

weitzman:
Did you mean to push something here?

RoSk0:
Hm, I did, but haven't checked that nothing changed for this PR. Created new PR #3920.

PR #3854, opened 2019-01-08
Move some methods from Drush:: to DrushConfig::

greg-1-anderson:
Depend on injected config object rather than static `Drush::` methods wherever possible.

### DIFF
--- a/src/Drush.php
+++ b/src/Drush.php
@@ -233,6 +233,8 @@ class Drush
      * Return the Drush logger object.
      *
      * @return LoggerInterface
+     *
+     * @deprecated Use injected logger instead. @see Drush::drush()
      */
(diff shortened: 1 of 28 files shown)


src/Drush.php
```diff
@@ -339,6 +345,8 @@ public static function process($commandline, $cwd = null, array $env = null, $in
      * @param mixed|null $input   The input as stream resource, scalar or \Traversable, or null for no input
      * @param int|float|null $timeout The timeout in seconds or null to disable
      * @return Process
+     *
+     * @deprecated Use injected process manager instead.
```
weitzman: We should give an example here IMO

greg-1-anderson: 👍

PR #3855, opened 2019-01-08
Propagate option negations to updatedb:status

zaporylie:
## Current behavior

When running `drush updb --no-post-updates` one can notice post-update hooks listed among regular update hooks.  Answering `yes` to `Do you wish to run the specified pending updates? (yes/no)` **doesn't run these updates**. 

## Expected behavior

`drush updb --no-post-updates` doesn't show post-update hooks in the update the summary.

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -53,6 +53,8 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
         $updatedb_options = [
             'entity-updates' => $options['entity-updates'],
             'post-updates' => $options['post-updates'],
+            'no-entity-updates' => $options['no-entity-updates'],
+            'no-post-updates' => $options['no-post-updates'],
         ];
         $process = Drush::drush($this->siteAliasManager()->getSelf(), 'updatedb:status', [], $updatedb_options);
         $process->mustRun();


greg-1-anderson:
Could you give steps to reproduce? (Mention some Drupal / contrib module version that has post update hooks.)

The fix here doesn't look appropriate to me. If `--no-post-updates` are specified on the commandline, then the `post-updates` option should be passed through with a value of `false`. `$options['no-post-updates']` will also be true as a side-effect, but `$options['post-updates']` is the value that should be used in code. I don't see the code using `$options['no-post-updates']`, so I'm not sure why this PR would work, or why the bug is happening.

greg-1-anderson:
Oh, I get it.

If `--post-updates` is false, then `Drush::drush()` omits the `--post-updates` option entirely -- that is, it behaves as if the default value for `--post-updates` is false rather than true. By explicitly placing `no-post-updates` in the `$options` array, it spurs `Drush::drush()` into placing `--no-post-updates` on the commandline, which in turn is converted into `$options['post-updates'] = false`, just as if a user had used the `--no-post-updates` flag.

If we were going to stick with this hack, then I would prefer:
```
        $updatedb_options = [
            'no-entity-updates' => !$options['entity-updates'],
            'no-post-updates' => !$options['post-updates'],
        ];

```
It would be niftier, though, if `Drush::drush()` would convert `$options['post-updates'] = false` into `--no-post-updates`. However, in order for this to work, it would need to know the default value for boolean options. I'm not sure if it does or does not -- would need to look into it. If it does not, then perhaps sticking with the hack, and formalizing it as a convention might be the way to go after all.

PR #3856, opened 2019-01-08
Adjust conflict string temporarily in pm:security

weitzman:
This is a bandaid until https://github.com/drupal-composer/drupal-security-advisories/pull/11 is merged.


webflo:
@weitzman Please try the two new branches.

- https://github.com/drupal-composer/drupal-security-advisories/tree/7.x-v2 
- https://github.com/drupal-composer/drupal-security-advisories/tree/8.x-v2

PR #3861, opened 2019-01-12
Clear the cache prior running the config importer

### LINKED ISSUE #3860: Config cache should be cleared before running config import
**Describe the bug**

There are cases when the cache is out-of-sync just before running `drush config:import`. This happens most likely after a DB restore, when the cache backend is elsewhere (Redis, Memcache). The database is restored but has a different content than the previous DB, the one that was the base of the actual cache stored in memory. When running `drush config:import` in such circumstances the config 
change list might not be computed correctly.

**To Reproduce**

# Use a Drupal install with Redis or Memcache.
# Change a value in some config YAML file from the config sync dir.
# Clear the cache and perform some operations to warm-up the config cache.
# Restore a different version of the database: `$ drush sqlc < /tmp/dump.sql`.
# Run the import: `$ drush config:import`.

Repeat this sequence 3-5 several times.

**Expected behavior**

Expecting the same change list every time.

**Actual behavior**

Different change lists are displayed each time.

**Workaround**

A cache rebuild `$ drush cache:rebuild` just before step 5. fixes the issue.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.x
| Drupal version? | 8.x
| PHP version     | 5.6/7.1
| OS?             | Mac

**Additional information**

N/A

claudiu-cristea:
Fixes #3860

### DIFF
--- a/src/Drupal/Commands/config/ConfigImportCommands.php
+++ b/src/Drupal/Commands/config/ConfigImportCommands.php
@@ -4,6 +4,7 @@ namespace Drush\Drupal\Commands\config;
 use Consolidation\AnnotatedCommand\CommandError;
 use Consolidation\AnnotatedCommand\CommandData;
 use Drupal\config\StorageReplaceDataWrapper;
+use Drupal\Core\Cache\CacheBackendInterface;
 use Drupal\Core\Config\ConfigManagerInterface;
 use Drupal\Core\Config\StorageComparer;
 use Drupal\Core\Config\ConfigImporter;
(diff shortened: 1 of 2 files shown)


weitzman:
Does Drupal itself clear cache right before the importer? If not, I'd rather keep cache clearing separate from config import

claudiu-cristea:
Well, I think Drupal core should do that. Haven't check

weitzman:
FYI, we now have `drush cc bin config`

claudiu-cristea:
That would help to be ran before import. However, I still think the importer should work on clean table. The config cache is not meant for this kind of operations, which are rare and low level.

PR #3875, opened 2019-01-19
FIXES #3874: Wrap 'pipefail' in "bash -c".

### LINKED ISSUE #3874: Illegal option 'pipefail' in sql:dump with gzip on Ubuntu
**Describe the bug**
When running `drush sql:dump --gzip` with drush 9.6.0-beta4 on Ubuntu 16.04 LTS the task fails with a "set: illegal option -o pipefail".  It appears that the system execs on Ubuntu may be using `sh`/`dash` instead of `bash`, as `sh` does not support `pipefail`.

**To Reproduce**
Execute `/path/to/drush sql:dump --gzip > /path/to/dump.sql.gz`

**Expected behavior**
A gzipped dump file would appear at `/path/to/dump.sql.gz`

**Actual behavior**
The command encounters a fatal error:
```
# /var/www/vendor/bin/drush sql-dump --gzip --debug > /backups/db/test.sql.gz
 [preflight] Config paths: /var/www/vendor/drush/drush/drush.yml,/var/www/drush/drush.yml
 [preflight] Alias paths: /var/www/web/drush/sites,/var/www/drush/sites
 [preflight] Commandfile search paths: /var/www/vendor/drush/drush/src,/var/www/drush
 [bootstrap] Starting bootstrap to max [0.23 sec, 8.43 MB]
 [bootstrap] Drush bootstrap phase: bootstrapDrupalRoot() [0.23 sec, 8.43 MB]
 [bootstrap] Change working directory to /var/www/web [0.23 sec, 8.43 MB]
 [bootstrap] Initialized Drupal 8.6.7 root directory at /var/www/web [0.23 sec, 8.62 MB]
 [bootstrap] Drush bootstrap phase: bootstrapDrupalSite() [0.23 sec, 8.91 MB]
 [bootstrap] Initialized Drupal site xxx at sites/default [0.24 sec, 9.08 MB]
 [bootstrap] Drush bootstrap phase: bootstrapDrupalConfiguration() [0.24 sec, 9.08 MB]
 [debug] Add service modifier [0.24 sec, 9.3 MB]
 [info] Executing: command -v mysql [0.25 sec, 9.74 MB]
 [info] sql:query: SHOW TABLES; [0.25 sec, 9.8 MB]
 [info] Executing: mysql --defaults-file=/tmp/drush_0XiJU0 --database=xxx --host=xxx --port=3306 --silent -A < /tmp/drush_mjVvMo [0.25 sec, 9.8 MB]
 [info] Executing: set -o pipefail; mysqldump --defaults-file=/tmp/drush_MHYv8C  xxx --host=xxx --port=3306 --no-autocommit --single-transaction --opt -Q  | gzip -f [0.28 sec, 9.81 MB]
> sh: 1: set: Illegal option -o pipefail

In SqlCommands.php line 231:

  [Exception]
  Unable to dump database. Rerun with --debug to see any error message.


Exception trace:
 () at /var/www/vendor/drush/drush/src/Commands/sql/SqlCommands.php:231
 Drush\Commands\sql\SqlCommands->dump() at n/a:n/a
 call_user_func_array() at /var/www/vendor/consolidation/annotated-command/src/CommandProcessor.php:257
 Consolidation\AnnotatedCommand\CommandProcessor->runCommandCallback() at /var/www/vendor/consolidation/annotated-command/src/CommandProcessor.php:212
 Consolidation\AnnotatedCommand\CommandProcessor->validateRunAndAlter() at /var/www/vendor/consolidation/annotated-command/src/CommandProcessor.php:178
 Consolidation\AnnotatedCommand\CommandProcessor->process() at /var/www/vendor/consolidation/annotated-command/src/AnnotatedCommand.php:302
 Consolidation\AnnotatedCommand\AnnotatedCommand->execute() at /var/www/pds/releases/master-2019-01-17-221207/vendor/symfony/console/Command/Command.php:255
 Symfony\Component\Console\Command\Command->run() at /var/www/vendor/symfony/console/Application.php:971
 Symfony\Component\Console\Application->doRunCommand() at /var/www/vendor/symfony/console/Application.php:248
 Symfony\Component\Console\Application->doRun() at /var/www/vendor/symfony/console/Application.php:148
 Symfony\Component\Console\Application->run() at /var/www/vendor/drush/drush/src/Runtime/Runtime.php:118
 Drush\Runtime\Runtime->doRun() at /var/www/vendor/drush/drush/src/Runtime/Runtime.php:49
 Drush\Runtime\Runtime->run() at /var/www/vendor/drush/drush/drush.php:69
 require() at /var/www/vendor/drush/drush/drush:4

sql:dump [--result-file RESULT-FILE] [--create-db] [--data-only] [--ordered-dump] [--gzip] [--extra EXTRA] [--extra-dump EXTRA-DUMP] [--format [FORMAT]] [--fields FIELDS] [--field FIELD] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-d|--debug] [-y|--yes] [--no] [--remote-host REMOTE-HOST] [--remote-user REMOTE-USER] [-r|--root ROOT] [-l|--uri URI] [--simulate] [--pipe] [-D|--define DEFINE] [--database [DATABASE]] [--target [TARGET]] [--db-url DB-URL] [--skip-tables-key SKIP-TABLES-KEY] [--structure-tables-key STRUCTURE-TABLES-KEY] [--tables-key TABLES-KEY] [--skip-tables-list SKIP-TABLES-LIST] [--structure-tables-list STRUCTURE-TABLES-LIST] [--tables-list TABLES-LIST] [--notify] [--xh-link XH-LINK] [--druplicon] [--] <command>
```

**Workaround**
Removing `--gzip` resolves the problem

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.6.0-beta4
| Drupal version? | 8.6.7
| PHP version     | 7.2.8-1+ubuntu16.04.1+deb.sury.org+
| OS?             | Ubuntu 16.04.5 LTS "xenial"

**Additional information**
This problem is not present in 9.5.x versions of drush and was likely introduced in #3818

greg-1-anderson:
This is almost right.

What we are trying to do here is wrap the 'pipefail' command in "bash -c" so that we can ensure that the shell is bash, since pipefail is a bash feature. We use interpolation on the pipefail string so that it is easy for the caller to alter the template as needed to support other shells.

The default pipefail string is: `bash -c "set -o pipefail; {{cmd}}"`

Also, 'pipefail' doesn't have anything to do with ssh, so I renamed the configuration name from 'ssh.pipefail' to 'sh.pipefail'.

Shortcomings:

1. This works if pipefail is empty, or if it is the default value. However, it is also possible that someone might use something like `set -o pipefail; {{cmd}}`, which for example should work if the user know their default shell is bash. However, at the moment this code always calls `addslashes()` on the contents of `$cmd`, even in instances like this where doing so would be wrong (although you wouldn't notice unless cmd contains quote characters). Perhaps what we need to do is make other interpolation shortcuts that indicate whether the substitution value should be escaped in any way. Maybe `//cmd//` or `{/cmd/}` or something like that for backslash-escaped config values? Do we want to consider shell-escaping at the same time?

2. It might be helpful to pull the 'pipefail' code out into a utility class.

### DIFF
--- a/src/Sql/SqlBase.php
+++ b/src/Sql/SqlBase.php
@@ -184,13 +184,7 @@ class SqlBase implements ConfigAwareInterface
             $file .= $file_suffix;
             $cmd .= ' > ' . Escape::shellArg($file);
         }
-        if ($pipefail) {
-            $interpolator = new Interpolator();
-            $replacements = [
-                'cmd' => addslashes($cmd),
-            ];
-            $cmd = $interpolator->interpolate($replacements, $pipefail);
-        }
+        $cmd = $this->addPipeFail($cmd, $pipefail);
 
         $process = Drush::shell($cmd, null, $this->getEnv());
         // Avoid the php memory of saving stdout.
@@ -198,6 +192,33 @@ class SqlBase implements ConfigAwareInterface
         // Show dump in real-time on stdout, for backward compat.
         $process->run($process->showRealtime());
         return $process->isSuccessful() ? $file : false;
+    }
+
+    /**
+     * We have three possibilities for $pipefail:
+     *
+     *   - empty: Return $cmd unmodified
+     *   - simple string: Return $cmd appended to $pipefail
+     *   - interpolated: Add slashes to $cmd and insert in $pipefail
+     *
+     * The last is particularly for environments such as Ubuntu
+     * that use something other than bash as the default shell. To
+     * make pipefail work right in this instance, we must wrap it
+     * in 'bash -c', since pipefail is a bash feature.
+     */
+    protected function addPipeFail($cmd, $pipefail)
+    {
+        if (empty($pipefail)) {
+            return $cmd;
+        }
+        if (strpos($pipefail, '{{cmd}}') === false) {
+            return $pipefail . ' ' . $cmd;
+        }
+        $interpolator = new Interpolator();
+        $replacements = [
+            'cmd' => addslashes($cmd),
+        ];
+        return $interpolator->interpolate($replacements, $pipefail);
     }
 
     /*
--- a/src/Sql/SqlBase.php
+++ b/src/Sql/SqlBase.php
@@ -195,13 +195,16 @@ class SqlBase implements ConfigAwareInterface
     }
 
     /**
-     * We have three possibilities for $pipefail:
+     * Handle 'pipefail' option for the specified command.
      *
+     * @param string $cmd Script command to execute; should contain a pipe command
+     * @param string $pipefail Script statements to insert into / wrap around $cmd
+     * @return string Result varies based on value of $pipefail
      *   - empty: Return $cmd unmodified
      *   - simple string: Return $cmd appended to $pipefail
      *   - interpolated: Add slashes to $cmd and insert in $pipefail
      *
-     * The last is particularly for environments such as Ubuntu
+     * Interpolation is particularly for environments such as Ubuntu
      * that use something other than bash as the default shell. To
      * make pipefail work right in this instance, we must wrap it
      * in 'bash -c', since pipefail is a bash feature.
--- a/src/Sql/SqlBase.php
+++ b/src/Sql/SqlBase.php
@@ -11,6 +11,7 @@ use Robo\Contract\ConfigAwareInterface;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Process\Process;
 use Webmozart\PathUtil\Path;
+use Consolidation\Config\Util\Interpolator;
 
 class SqlBase implements ConfigAwareInterface
 {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
The reason I'm ambivalent about this feature is that I feel that it *is* a primary purpose of Drush (and Robo) to make it easy to convolve the shell commands you need to run to perform your devops/maintenance tasks in a standard and straightforward way. However, the operations required for customization should be simple and obvious, not baroque and undiscoverable.

Maybe function-call syntax would be appropriate here:
```
bash -c "set -o pipefail; {{addslashes(cmd)}}"
```
That, at least, would be readable (somewhat?). The available functions would be very limited, just `addslashes` and `escapeshell`, perhaps.

greg-1-anderson:
So, thinking about this a little more, it is not great to start to turn the interpolation system into a mini-interpreter. Something like "wrap in bash" is better implemented in code. This could perhaps be a feature of site process. SqlBase would then have to use site process.

🤔

PR #3876, opened 2019-01-19
Make Log message status transparent

karate:
I've noticed that colored log statuses have a black background, which looks ugly on semi-transparent terminals (I guess on all terminals with a different background color as well). Is there a reason to force the bg color?

### DIFF
--- a/src/Log/Logger.php
+++ b/src/Log/Logger.php
@@ -103,9 +103,9 @@ class Logger extends RoboLogger
         }
 
         if ($this->output->isDecorated()) {
-            $red = "\033[31;40m\033[1m[%s]\033[0m";
-            $yellow = "\033[1;33;40m\033[1m[%s]\033[0m";
-            $green = "\033[1;32;40m\033[1m[%s]\033[0m";
+            $red = "\033[31m\033[1m[%s]\033[0m";
+            $yellow = "\033[1;33m\033[1m[%s]\033[0m";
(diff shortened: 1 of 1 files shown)


weitzman:
I suspect the rationale is to assure that the text is visible even when a white background is used (we use white text). How do other CLI's handle this. I dont think we need to get fancy here.

A couple places where Style is defined:
1.  https://github.com/drush-ops/drush/blob/323519c376faad7c4b44d86c156b38b2cf3f05e0/src/Style/DrushStyle.php#L46-L59
1. https://github.com/drush-ops/drush/blob/4a34a28f4f1d610775c6bb990ce6f0150f9ddf39/src/Log/Logger.php#L197-L198

karate:
I don't think we can safely assume that everyone uses black opaque backgrounds in their terminals.

The terminal emulator's theme should make sure that all available font colors are readable on each background. The CLI should respect that.

weitzman:
Perhaps someone can share before/after screenshots so we know whats changing. Use whatever terminal theme you usually use. Thanks.

karate:
I'm using `gnome-terminal` in `i3wm`, with `compton` for compositing. Here's a demo:
transparency demo

greg-1-anderson:
That screenshot looks like Drush 8, but this PR is set to merge into master (Drush 9).

karate:
@greg-1-anderson you are correct. My patch was for Drush 8. I apologize.

PR #3880, opened 2019-01-22
make drush run with php-fpm

matevoux:
Make drush run with php-FPM
proof of concept

### DIFF
--- a/cgdrush
+++ b/cgdrush
@@ -1,0 +1,14 @@
+#/bin/bash
+
+export SCRIPT_NAME=drush 
+CGDRUSH=$( readlink -f "$0" )
+SCRIPT_PATH=$( dirname "$CGDRUSH" )
+export SCRIPT_FILENAME=$SCRIPT_PATH/drush.php 
+export REQUEST_METHOD=GET 
+export args="drush" 
+for i in $* 
+  do
+  arg=$(urlencode $i)
+  args="$args&$arg"
+done
+cgi-fcgi -bind -connect /var/run/php/php7.2-fpm.sock | tail +3
--- a/drush.php
+++ b/drush.php
@@ -55,10 +55,12 @@ if (file_exists($autoloadFile = __DIR__ . '/vendor/autoload.php')
     throw new \Exception("Could not locate autoload.php. cwd is $cwd; __DIR__ is " . __DIR__);
 }
 
-$argv = explode(' ',$_SERVER['args']);
-$argc = count($argv);
-$_SERVER['argc'] = $argc;
-$_SERVER['argv'] = $argv;
+if(array_key_exists('args', $_SERVER)) {
+    $argv = explode(' ',$_SERVER['args']);
+    $argc = count($argv);
+    $_SERVER['argc'] = $argc;
+    $_SERVER['argv'] = $argv;
+}
 
 
 // Set up environment
--- a/drush.php
+++ b/drush.php
@@ -56,8 +56,9 @@ if (file_exists($autoloadFile = __DIR__ . '/vendor/autoload.php')
 }
 
 if(array_key_exists('args', $_SERVER)) {
-    $argv = explode(' ',$_SERVER['args']);
-    $argc = count($argv);
+    preg_replace_callback("/(\S+)|(\".+\")|('.+')/" ,
+             function($m) use (&$argv, &$argc) { $argv[$argc++] =  preg_replace('/(^[\'"])|([\'"]$)/','',$m[0]); },
+            $_SERVER['args']);
     $_SERVER['argc'] = $argc;
     $_SERVER['argv'] = $argv;
 }
--- a/drush.php
+++ b/drush.php
@@ -56,12 +56,13 @@ if (file_exists($autoloadFile = __DIR__ . '/vendor/autoload.php')
 }
 
 if(array_key_exists('args', $_SERVER)) {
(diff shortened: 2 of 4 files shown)


drush.php
```diff
@@ -54,6 +55,14 @@
     throw new \Exception("Could not locate autoload.php. cwd is $cwd; __DIR__ is " . __DIR__);
 }
 
+if(array_key_exists('args', $_SERVER)) {
+    $argv = explode(' ',$_SERVER['args']);
```
greg-1-anderson: I think we'd need to do a better job here if we're going to support this, as a simple `explode` on spaces provides no way for quoted / escaped arguments.  I wouldn't expect that all of the tests would pass until this is right.

matevoux: I've tried something with regex.
But not sure, must be tested with real quoted param

matevoux: another way should be to encode each base 64 param in a shell script calling drush with php-fpm and keep space as separator for arg varaiable and add a the base 64 decoding after the explode

What do you think the best way ?

greg-1-anderson: I think we need to take a step back and figure out what the use-case is, and how that would relate to the standard Drush that uses php-cli.

If you are just making your own fork of Drush, then you can base64 or json encode your arguments and do whatever you want, and it really won't matter. I don't have a strong opinion on what you do in this case; it's whatever is most convenient for you.

This is looking a little too edge-case for inclusion in Drush core at the moment, though. The regex parser is fragile and doesn't give good error messages for unbalanced quotes, et. al.; the base64 method would be inconvenient for most clients. There is also the question of selecting between ConsoleOutput and an OutputBuffer. It is unclear on how this would be installed an used.

So I'd try to define those items first. Start off assuming this will be your own thing. Make your own copy of drush.php or whatever to get you going. Then figure out what your cli client is going to look like, and how it's going to identify and call your php-fpm-embedded Drush. Ultimately, this could probably become something that one composer-require's inside of a Drupal project along with drush/drush:^9.

matevoux: ok got something with encoding (url one, will need to install the urlencode shell command).
Use case is case (for me) running drush migrate import by cron every 15 minutes for  importing remote content (one from youtube, other via rss feed of a few websites).
But everyone  using drush on production server may be interested by taking profit of opcache as drupal class are in cache.
For consoleOutput i have an  idea, will send a patch soon that define a child class of consoleOutput and overriding the constructor to open php://stderr and php://output insteand of  php://stderr and php://stdout

greg-1-anderson:
Is it possible to just open `php://stdout` in php-fpm? Is it possible to just assign to STDOUT?

I'm wondering if we can adapt instead of convert.

matevoux:
I think i got it work !

matevoux:
stdout could be opened but go to log rather than output console and need to change php.ini

matevoux:
in last version stdout use for CLI and output for FPM.
so NO CHANGE for cli use (and all non fpm)

drush.php
```diff
@@ -63,6 +73,6 @@
 // Preflight and run
 $preflight = new Preflight($environment);
 $runtime = new Runtime($preflight);
-$status_code = $runtime->run($_SERVER['argv']);
+$status_code = $runtime->run($argv);
```
greg-1-anderson: I think $argv should be assigned `$_SERVER['argv']` in the non-php-fpm code path.

cgdrush
```diff
@@ -0,0 +1,14 @@
...
+export SCRIPT_NAME=drush 
+CGDRUSH=$( readlink -f "$0" )
+SCRIPT_PATH=$( dirname "$CGDRUSH" )
+export SCRIPT_FILENAME=$SCRIPT_PATH/drush.php 
+export REQUEST_METHOD=GET 
+export args="drush" 
+for i in $* 
+  do
+  arg=$(urlencode $i)
+  args="$args&$arg"
```
greg-1-anderson: How does `args` get passed to php-fpm?

drush.php
matevoux: both are setted to the same, by decoding args (line 58 to 65).
you'd rather using $_SERVER['argv'] or $argv  line 76 ?

cgdrush
matevoux: shell script encode them from command line and store them to args env var then php get them in $_SERVER['args'] and decode line 58 to 65.

matevoux: in case you're wondering the problem is send array to php-fpm for argv

matevoux:
Args is not set in php cli.It my variable setted and exported in shell.So cli version skip the code in the if bracket while cGi version decode parameters and assign both variable in both versions (server array and variable ).Same in symfony overriden class.If called from php-fpm return false, else fall back to parent method.Le 28 janv. 2019 22:27, Greg Anderson <notifications@github.com> a écrit :@greg-1-anderson commented on this pull request.



In drush.php:
> @@ -63,6 +73,6 @@
 // Preflight and run
 $preflight = new Preflight($environment);
 $runtime = new Runtime($preflight);
-$status_code = $runtime->run($_SERVER['argv']);
+$status_code = $runtime->run($argv);

Maybe line 58 is false in php-cli?

matevoux:
Argv is set by php !
see http://php.net/manual/en/reserved.variables.argv.php

greg-1-anderson:
Derp. I'm used to just using `$_SERVER` for this.

weitzman:
I'm uninclined to support this use case. Or at least I dont understand when its broadly useful.

PR #3882, opened 2019-01-23
fix: process whitelisting of fields correctly

e0ipso:
The current logic is unsetting the cardinal index instead.


e0ipso:
I think this was a copy&paste bug from another foreach in the same function.

weitzman:
Thanks

Issue #3883, opened 2019-01-24
--ssh-options argument doesn't override ssh.options when using site:ssh


minorOffense:
**Describe the bug**
When using a site alias file which has an entry for ssh options using the site:ssh command the alias file options aren't overridden by the command line argument value. But every other drush command favours the CLI argument over the alias file settings.

**To Reproduce**
Create a site alias file with some ssh options defined.

E.g.
```yaml
ssh:
  options: -l username
```

Then call a drush command that uses ssh 

e.g.
```
drush --ssh-options="-A" rsync @myalias @self
```
The alias ssh settings are overridden when rsync opens the SSH connection (which is good!)

However if you call site:ssh as the command
```
drush --ssh-options="-A" @myalias ssh echo "Hello"
```
Drush uses the settings from the alias file and ignores the ``--ssh-options`` parameter.

**Expected behavior**
Drush should always prefer the CLI arguments over the alias file defaults.

**Actual behavior**
It uses the ssh options from the alias file.

**Workaround**
Not really...

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.5
| Drupal version? | 8.x
| PHP version     | 7.1
| OS?             | Only tested on Linux but presume it would be for all.

greg-1-anderson:
Aside: this is related to the current bug in 9.6.0-rc1 where --ssh-options are ignored. When we fix that, we should take care of this use-case as well.

greg-1-anderson:
In progress at https://github.com/drush-ops/drush/pull/3885

weitzman:
lets talk there.

greg-1-anderson:
#3885 did not fix this issue. Need a follow-up PR to continue this.

greg-1-anderson:
Another interesting thing about this is that in Drush 9, --ssh-options are remapped to `'-Dssh.options'`. The definition of `-D` is that it provides **default** configuration values, e.g. that is overridden by configuration values that may appear in a drush.yml file or an alias.

There are definitely use-cases for providing default config values, though (e.g. bash aliases), so the definition of `-D` should not be changed. Perhaps there needs to be a different but similar construct to `-D` for config overrides.

greg-1-anderson:
I'm starting to question whether we should fix this, or at least whether we should postpone the decision until after the Drush 9.6.0 release.

In Drush 9, there are three objects that need to interact in order for this to work as desired:
- `InputInterface $input`: The commandline arguments
[...]

PR #3885, opened 2019-01-25
Allow configuration to provide default values for site alias properties.

greg-1-anderson:
More work will be required to allow configuration to provide overrides for site alias properties.


greg-1-anderson:
This does not fix #3883 yet

PR #3886, opened 2019-01-25
Detect os

ms-slasher13:


ms-slasher13:
updates consolidation/site-process

greg-1-anderson:
Looks like the unescaping changes to getOutputAsJson may have broken some use cases.

greg-1-anderson:
Temporarily reverted the fix in site-process until the regressions can be corrected.

greg-1-anderson:
In another PR it looks like `command` works on Windows. If this is the case, then my advice was undesirable.

I think we're still waiting for updates in site process before continuing here.

ms-slasher13:
Changes here are for testing https://github.com/consolidation/site-process/pull/24 with drush.

greg-1-anderson:
Travis is failing due to code style - easy fix. Appveyor is failing due to being unable to find the php binary for linting. Not sure why it's broken, since php is found in the step before. Php should not lint differently on Windows, so you may just take this check out of the Appveyor job if you don't want to fix it.

.scenarios.lock/php5/composer.json
```diff
@@ -64,7 +68,7 @@
         "consolidation/output-formatters": "^3.3.1",
         "consolidation/robo": "^1.4.4",
         "consolidation/site-alias": "^2.1.1",
-        "consolidation/site-process": "^1.1.1",
+        "consolidation/site-process": "dev-master",
```
greg-1-anderson: I merged your PR and made a release, so this can become `^1.1.2` now.

composer.json
```diff
@@ -40,7 +44,7 @@
     "consolidation/output-formatters": "^3.3.1",
     "consolidation/robo": "^1.4.4",
     "consolidation/site-alias": "^2.1.1",
-    "consolidation/site-process": "^1.1.1",
+    "consolidation/site-process": "dev-master",
```
greg-1-anderson: I merged your PR and made a release, so this can become `^1.1.2` now.

composer.json
```diff
@@ -86,6 +90,9 @@
     "process-timeout": 9600,
     "platform": {
       "php": "7.0.11"
+    },
+    "github-oauth": {
+      "github.com": "154fbc0cb25b34e837a43f018d63b2e6803a568e"
```
greg-1-anderson: Not sure why this is here; you should remove it, and revoke the credentials.

src/Exec/ExecTrait.php
```diff
@@ -88,7 +89,11 @@ public function startBrowser($uri = null, $sleep = 0, $port = false, $browser =
      */
     public static function programExists($program)
     {
-        $process = Drush::shell("command -v $program");
+        $cmd = "command -v $program";
+        if (Escape::isWindows()) {
+            $cmd = "where $program";
```
greg-1-anderson: I think I was in error recommending that we use `where` here. If `command` works, then it would be preferable to remove this.

ms-slasher13: `command -v $program` doesn't work for mysql on windows
`mysql -V` works

What do you prefer?

greg-1-anderson: OK, I guess I misunderstood the comment in the other issue. If `command -v` does not work, then I prefer to leave this as-is, and continue to use `where`

greg-1-anderson:
Please run `composer update` and commit all the files (especially in the scenarios directory); that should make the tests green, and this will be ready to merge.

greg-1-anderson:
Thanks for contributing this fix. It's very helpful to have assistance with Windows issues.

PR #3890, opened 2019-01-28
Fix logging and status checking in Integration tests

weitzman:


tests/unish/UnishIntegrationTestCase.php
```diff
@@ -97,9 +99,9 @@ public function drush($command, array $args = [], array $options = [], $expected
         $cache = &drush_get_context();
         $cache = [];
 
-        if ($expected_return != self::IGNORE_EXIT_CODE) {
+        // if ($expected_return != self::IGNORE_EXIT_CODE) {
```
weitzman: I'm not clear what `self::IGNORE_EXIT_CODE` is doing. I've commented out its only use since it was blocking failure/success checking always.

greg-1-anderson: Since the tests are passing, this must not be used any longer. The goal of this change was to allow for tests to use the `drush` command to test the output of some command, and yet be indifferent to the exit code. I don't remember what my use-case for that was, but apparently it was a transitory need.

PR #3891, opened 2019-01-28
Add array type hint to QueueCommands::run

eiriksm:
Ok, so I realize this is a bit of an edge case.

I have added a command file that extends a drush core command (QueueCommands). It lives in a custom module. I override the ::run method for some very specific reason.

As with most projects, I am doing code style checking of all custom code. Against Drupal coding standards. So including this new file. And it would require me to add a type hint to the $options parameter to pass. Which would then give a warning, because the overriding class now changes the function signature of run.

So here is a quick PR just for my very narrow use case :)

### DIFF
--- a/src/Drupal/Commands/core/QueueCommands.php
+++ b/src/Drupal/Commands/core/QueueCommands.php
@@ -58,7 +58,7 @@ class QueueCommands extends DrushCommands
      * @validate-queue name
      * @option time-limit The maximum number of seconds allowed to run the queue
      */
-    public function run($name, $options = ['time-limit' => self::REQ])
+    public function run($name, array $options = ['time-limit' => self::REQ])
     {
         $time_limit = (int) $options['time-limit'];
         $start = microtime(true);


greg-1-anderson:
In general, I think it's better to have a Drush command rather than extend a Drush command.

If we include it here, we should include it in all command methods. I would not mind doing this, except I think perhaps this might cause an error, not a warning, if someone else was overriding this command and did not include the `array` typehint. If so, that would make this a breaking change. Is this fix important enough to ignore the b/c break, or can we get by until Drush 10?

eiriksm:
I agree. And for sure about the BC break.

And I totally understand if you want it included for all command methods. For me, all it takes is a patch line in my composer.json, which i now get by referring to this PR :)

Do you want me to do a sweep of all methods while I am at it, or how would this be easiest to approach?

greg-1-anderson:
I'd like @weitzman to weigh in, but I think my preference would be to hold off until Drush 10. In that scenario, it might be premature to correct all commandfiles now.

weitzman:
I'd welcome to PR to add that type hint for all commands. I also think that Drush10 would be a good time to merge it.

greg-1-anderson:
This usually isn't done for Robo annotated commands. As mentioned above, this is a breaking change, so we'd have to do it for all commands for Drush 11 if we are going to do it. I don't think it's worth it, though, for every Drupal module to have to decide if their command files are going to work only on Drush 10 or only on Drush 11.

Recommend closing.

PR #3899, opened 2019-02-01
git:// is discontinued on drupal.org - use https://

AronNovak:
https://twitter.com/drupal_infra/status/1088571024918994945

### DIFF
--- a/commands/pm/package_handler/git_drupalorg.inc
+++ b/commands/pm/package_handler/git_drupalorg.inc
@@ -52,7 +52,7 @@ function package_handler_download_project(&$request, $release) {
     $repository = $username . '@git.drupal.org:project/' . $request['name'] . '.git';
   }
   else {
-    $repository = 'git://git.drupal.org/project/' . $request['name'] . '.git';
+    $repository = 'https://git.drupal.org/project/' . $request['name'] . '.git';
   }
   $request['repository'] = $repository;
   $tag = $release['tag'];


weitzman:
code looks good.

@greg-1-anderson Any idea whats up with the failure?

greg-1-anderson:
Syntax error in the Circle config file -- maybe Circle got more strict in parsing, as this used to work. I'll take a look.

greg-1-anderson:
Ah, Drush 8 is still on Circle CI 1.x.  The claim is that [CircleCI will discontinue 1.x on 15 March](https://twitter.com/circleci/status/1091113311221895169), but my experience is that 1.x builds are already failing.  I'm going to be AFK for most of the day, but I'll see if I can fix this, either before I leave or after I get back.

greg-1-anderson:
This PR seems to be non-overlapping with #3902.

drumm:
Yes, this pull request is independent of the others, and is higher priority than the others. We do not plan to turn the git:// protocol support back on, since this is not at all supported by GitLab.

It would be good to get this and https://github.com/drush-ops/drush/pull/3901 into a release before we do the final cutover to GitLab.

greg-1-anderson:
Turns out that the Circle tests are only the lint check. Fixed that on the 8.x branch and merged here.

drumm:
This pull request is still the highest priority one. The git:// protocol is not supported by Drupal.org and clones made with this drush functionality will currently fail. Does GitHub support re-opening pull requests?

greg-1-anderson:
The tests started failing when I merged this, so I reverted it.  I chose this one to merge first originally because it had green tests, whereas the other PRs did not.

Rather than re-opening here, it would be better to make just one PR for the drupal.org tests and get it to green. Since the 8.x branch tests are still broken anyway, I can re-apply this change directly to that branch if it would help.

greg-1-anderson:
See #3915

PR #3902, opened 2019-02-01
Replace http with https for git.drupal.org URLs

drumm:
We should no longer be using http whenever possible.


weitzman:
looks ok to me. leaving for @greg to merge.

greg-1-anderson:
Let's continue in #3915

PR #3906, opened 2019-02-04
Fix #3894. Assume pre-escaped for single arg calls to ssh command.

### LINKED ISSUE #3894: Over-escaping in drush ssh when using compound statement
from @mishac in #drush

```
$ drush -v @dev ssh "cd .. && pwd"
[info] Executing: ssh -o PasswordAuthentication=no USER@HOST 'cd /var/www/vhosts/HOST/dev/web && cd .. '\''&&'\'' pwd' 
```

Note odd escaping of `&&`

We may need to assume already-escaped in this case.

weitzman:
Gets rid of --legacy which was not even declared as an option.

### DIFF
--- a/src/Commands/core/SshCommands.php
+++ b/src/Commands/core/SshCommands.php
@@ -22,7 +22,7 @@ class SshCommands extends DrushCommands implements SiteAliasManagerAwareInterfac
      * @usage drush @mysite ssh
      *   Open an interactive shell on @mysite's server.
      * @usage drush @prod ssh ls /tmp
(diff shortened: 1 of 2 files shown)


weitzman:
Those people would be getting 'option not recognized' today. Its not declared.

greg-1-anderson:
Robo picks up options declared in the parameter list even if they are not declared as annotations. Using a version of Drush that does not contain this change behaves as you say, though. I'm not sure what causes the change in behavior; our custom help should not affect option parsing.

Issue #3117, opened 2017-10-29
batch process fails to start new thread after reaching 50% of available memory


giorgio79:
When executing a drush batch-process, it does not restart after the available memory fills up. Is this a bug?

Using migrate-import, the new thread start works fine.
```
Prepared 153480 out of 206941 [3034.35 sec, 512.28 MB]                                                                                                                                                                            [ok]
Batch process has consumed in excess of 50% of available memory. Starting new thread [3034.35 sec, 512.27 MB]                                                                                                                         [batch]
Returned from hook drush_core_batch_process [3034.35 sec, 512.26 MB]                                                                                                                                                                  [debug]
Command dispatch complete [3034.35 sec, 512.2 MB]                                                                                                                                                                                    [notice]
 Timer  Cum (sec)  Count  Avg (msec)
 page   3034.013   1      3034013.05

```

weitzman:
sounds like a bug ... what version of drush? please provide steps to
reproduce, if possible.

giorgio79:
Drush Version   :  8.1.13

This issue happens with all of my drush batch-process -es. I just figured out today with debug and verbose calls where it fails. For now I am running drush batch process in a for loop.
Have not yet tried on other systems.

weitzman:
Perhaps @pfrenssen has an idea about this as he recently touched the batch code..

pfrenssen:
This doesn't ring any bells for me. I have been going through parts of the batch code recently but it was always with a specific goal in mind. I still have to gain a more complete picture of it. I wasn't even aware that it has functionality built in to preserve memory.

giorgio79:
Yeah, this issue is strange. It does not happen in the browser fortunately...

giorgio79:
Here is the code
https://github.com/drush-ops/drush/search?utf8=%E2%9C%93&q=Batch+process+has+consumed+in+excess+of+50%25&type=

Looks like the code is meant to stop the batch process if the memory limit is reached. I don't get though how it can have a memory limit. The whole concept of batch is meant to handle low memory and resource environments, and break a task to smaller pieces.

weitzman:
I think this feature should be removed from drush.

ryross:
I've also run into this issue. It would be nice if drush could start a new process once it hits 50% memory usage.

ryross:
I might have a solution for some of you.

[...]

PR #3909, opened 2019-02-05
Fix #2762, #3117. Remove respawn after 50% memory exhaustion during batch processing.

### LINKED ISSUE #2762: Drush updatedb / updb keeps running the same operations.
The sandbox variable is unavailable in the new thread which is created once 50% memory usage has been reached.

In case of large batches which are set based on an empty sandbox the batch will loop forever.

weitzman:
This code is broken. Since [migrate already handles low memory](https://github.com/drupal/drupal/blob/8.6.x/core/modules/migrate/src/MigrateExecutable.php#L256), this code no longer serves its original purpose.

### DIFF
--- a/includes/batch.inc
+++ b/includes/batch.inc
@@ -52,13 +52,12 @@ class DrushBatchContext extends ArrayObject {
  * Process a Drupal batch by spawning multiple Drush processes.
  *
  * This function will include the correct batch engine for the current
- * major version of Drupal, and will make use of the drush_backend_invoke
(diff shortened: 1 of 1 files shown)


markcarver:
Regardless of what it was "originally" intended for, it was part of Drush's public batch API.

Removing this is not a BC "solution" for code that actually **relies** on this functionality (which is hard to get around in Drush due to the way batches are handled).

Instead of removing this, the affected code (referenced in the issues above), could simply not use this functionality of the API.

PR #3917, opened 2019-02-07
Fix Array to string conversion in TsvFormatter.php on line 35 #3721

MPParsley:
See 3721


weitzman:
At minimum this will need to update composer.lock,

Waiting for @greg-1-anderson for a code review.

greg-1-anderson:
I updated composer.lock, but the tests are failing. See #3918.

To update composer.lock here, just run `composer update`. Be sure to also commit the files in the scenarios directory that are also updated.

src/Drupal/Commands/core/BatchCommands.php
```diff
@@ -17,7 +17,7 @@ class BatchCommands extends DrushCommands
      *
      * @return \Consolidation\OutputFormatters\StructuredData\UnstructuredListData
      */
-    public function process($batch_id, $options = ['format' => 'json'])
+    public function process($batch_id, $options = ['format' => 'null'])
```
weitzman: The caller expects JSON to be returned. See https://github.com/drush-ops/drush/blob/92b2359f8d448adba5fcf1e0d2ecd1584c8c7d73/includes/batch.inc#L152-L153

composer.json
```diff
@@ -37,7 +37,7 @@
     "consolidation/annotated-command": "^2.10.2",
     "consolidation/config": "^1.1.0",
     "consolidation/filter-via-dot-access-data": "^1",
-    "consolidation/output-formatters": "^3.3.1",
+    "consolidation/output-formatters": "^3.4",
```
weitzman: This does look useful. After you run composer update, please commit the changes that happen in .scenario directory.

greg-1-anderson:
Drush now depends on output-formatters ^3.4.

PR #3924, opened 2019-02-08
Follow symlinks when looking for Drush extensions.

greg-1-anderson:
Any objections to invariantly following symlinks when searching for Drush extensions?  We already limit the search depth to 3, so I think that this probably won't have any impact.

If you think there might be odd symlinks in folk's search paths, then perhaps we could define a configuration value to control whether symlinks are followed or not.

My use-case is I am making an example sitewide Drush extension project with tests. I have a `sut` set up similar to what Drush does, and I want to symlink from the sut's drush/Commands/example-drush-extension directory to the root of the project. With this change, that setup allows `vendor/bin/drush example:param foo` to run the Drush command defined at the project root.


weitzman:
I'm not too keen on this at first blush. I feel a bit burned by symlinks like this in earlier versions of drush. Could you not add config or --include so that the commandfile gets loaded?

greg-1-anderson:
Hm yeah `--include` will probably work. I'll try it that way.

greg-1-anderson:
Hm, thinking about my options. `--include` does not work because the namespace is wrong.

greg-1-anderson:
I can't think of a good alternative involving `--include`, so I'm going to try again with this PR coupled with configuration, so that this can be enabled only for test suites.

greg-1-anderson:
Is this good enough? The only alternative I can think of here would be to use Robo plugins, but that would be a larger change.

greg-1-anderson:
The other option might be to add some syntax where both a namespace and a path can be specified.
```
drush --include=foo:/path foo:cmd
```
and:
```
drush:
  include:
   foo: /path
```
Maybe that's better?

greg-1-anderson:
The thing we have trouble with is setting the default value of a boolean option. The time we do that is when applying a default value for an option from config. In this instance, the config value is read directly, so there is no problem.

I'm not sure if you read my previous comment before posting your review. If you don't mind the config-enabled symlink, then I'll just merge this and move forward. If you think the alternative (namespaced include paths) is substantially better, I'll do the extra work.

weitzman:
Let’s do the pr as is, instead of the namespace idea.

PR #3932, opened 2019-02-14
Namespaced includes

greg-1-anderson:
#3924 (allow following symlinks when searching for commandfiles) worked okay on Drush 9, but it was poison to Drush 8. Running any Drush 8 command in the presence of a Symlink from sites/drush/Commands/foo to the project root (where `vendor` and the `sut` are located) caused Drush 8 to spend two minutes just to find command files. Encouraging this sort of symlink to be created is bad in any environment where Drush 8 may be used.

This PR backs out the support for followig symlinks, and replaces it with a better mechanism that allows the namespace parts for an include file to be specified with the include path. It also implements a couple of features that were documented but unimplemented in Drush 9.


greg-1-anderson:
With this PR, the config to set up testing looks like this:
```
 drush:
  # Instructs Drush to include the current working directory (env.PWD) as a
  # search path for commands, and to use \Drush\Commands\example_drush_extension
  # as the base namespace for commands found there.
  include:
    'Commands/example_drush_extension': '${env.PWD}'
```
I've also considered an alternate design that would just have clients exactly specify the class name and path of the code they want to test, e.g. like this:
```
drush:
  commands:
    '\Drush\Commands\example_drush_extension\ExampleCommands': '${env.cwd}/ExampleCommands.php'
```
The later is less flexible, but a lot more clear. Since this feature is primarily intended for testing, perhaps we do not need commandfile discovery.  I already implemented the later strategy, and it works fine. I thought code complexity would be lower with the second implementation, but it's similar. I can push the implementation if you'd like to consider it.

examples/example.drush.yml
```diff
@@ -56,6 +56,12 @@ drush:
     # or multiple --include options may be provided. Drush 8 and earlier did
     # a deep search in ~/.drush and /usr/share/drush/commands when loading
     # command files.
+    #
+    # For testing, specify the namespace component in the key. e.g.:
```
weitzman: For testing?

src/Config/ConfigLocator.php
```diff
@@ -447,12 +447,28 @@ protected function getBuiltinCommandFilePaths()
...
-            if (is_dir($commandPath)) {
-                $searchpath[] = $commandPath;
+        foreach ($commandPaths as $key => $commandPath) {
+            // Check to see if there is a `#` in the include path.
+            // This indicates an include path that has a namespace,
+            // e.g. `namespace#/path`.
+            if (is_numeric($key) && strpos($commandPath, '#') !== false) {
+                list($key, $commandPath) = explode('#', $commandPath, 2);
+            }
+            $sep = (DIRECTORY_SEPARATOR == '\\') ? ';' : ':';
```
weitzman: Is that an isWindows check? I think we have that method somewhere.

examples/example.drush.yml
greg-1-anderson: The point of the comment is to indicate to users that this feature is intended for use in tests, not for configuring Drush extensions in a site. There's no technical reason to make that limitation; the word "testing" could be removed if desired.

greg-1-anderson:
A Drush extension looks something like this:
```
<?php

namespace Drush\Commands\example_drush_extension;

/**
 * Edit this file to reflect your organization's needs.
 */
class ExampleCommands extends DrushCommands
{
```

This file must be at a specific path, e.g. `__DRUPAL_ROOT__/drush/Commands/example_drush_extension/ExampleCommands.php`.

In order to use the extension, Drush needs to know both the namespace and the path. In the current code, this requirement is satisfied by the commandfile discovery component. The discovery class starts at a given path, e.g. __DRUPAL_ROOT__/drush, and with a given base namespace, e.g. \Drush. If a commandfile is found at Commands/example_drush_extension/ExampleCommands.php, then this is converted into a namespace fragment (Commands\example_drush_extension\ExampleCommands) which is appended to the namespace base.

This PR alters the initial base namespace using features already present in the discovery class. This allows us to load an extension that is located at __PROJECT_ROOT__/ExampleCommands.php and make Drush act as if it were inside a Commands/example-drush-extension directory.

This is a bit hard to follow, so I thought that instead it might be better to just have the configuration file fully specify the path and the namespace directly.

greg-1-anderson:
I pushed a commit that implements support for `drush.commands`.  The namespace addition for `drush.includes` is also still supported, although we probably do not need both.

PR #3942, opened 2019-02-20
Fixes #3940: Do not get confused by the SUT

### LINKED ISSUE #3940: Drush cannot bootstrap Drupal when started outside of Drupal root
This happens on Drush ^9.6-beta4.
When I run Drush outside Drupal root the following error shows up.
> PHP Fatal error:  Uncaught Error: Call to a member function findUri() on null in /var/www/d8/docroot/vendor/drush/drush/src/Boot/BootstrapManager.php:132

I think it caused by [this change](https://github.com/drush-ops/drush/commit/f7fd4b7f41ce868e8c476eed4b729e1d728c8886#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R114).

Drupal Finder reads `installer-paths` section when searching for Drupal root.

greg-1-anderson:
If Drush is installed as a dependency of a Drupal site (as we stipulate it always should be) and the cwd / alias does not find a site, then search for the site that Drush is installed in, and fall back to the sut only as a last resort.

Note that this PR does not fix the case where your cwd is inside the Drush project root for a Drush site installed inside a Drupal site. That is an edge case that not too many folks should encounter.

### DIFF
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -343,8 +343,17 @@ class Preflight
     {
         if ($selectedRoot || $fallbackPath) {
             $foundRoot = $this->drupalFinder->locateRoot($selectedRoot);
+            // If we did not find a site at the selected root, check the
+            // PARENT directory of the fallback path. This will find a site
+            // that Drush is installed in while avoiding the SUT.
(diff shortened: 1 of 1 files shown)


Chi-teck:
It works for me.

PR #3947, opened 2019-02-27
Make pm:security test more flexible

greg-1-anderson:
The current pm:security test fails if there is also a pending security release for some other module, or Drupal core. Make the check a little more flexible to avoid this problem.


greg-1-anderson:
Came back green here: https://circleci.com/workflow-run/b40f32d7-069c-4ece-9f71-e0d5fc96e1f0

PR #3949, opened 2019-02-28
Appveyor

ms-slasher13:


weitzman:
Thanks for working on this.

I brought in latest changes into appveyor branch resolved a conflict so this PR now shows just the one changed line.

So I guess the idea here is that the --answers value is not actually bring sent to Bash so it doesn't need to be escaped via Windows? The change here is a little unclear to me but I dont think I have to understand it fully. I'll merge this soon, even if nobody replies.

ms-slasher13:
> --answers value is not actually bring sent to Bash so it doesn't need to be escaped via Window

Yes, that's the idea

PR #3951, opened 2019-02-28
Add --database-target and --database-key options to the sql:sync command

BramDriesen:
In previous versions of Drush it was possible to select the target/source database if you had multiple, this seems to be omitted from the latest version brining a big limitation in the options to sync multiple databases.

By simply adding the annotation the option is back to do this.


weitzman:
OK, but do the new options work as expected?

BramDriesen:
I only tested --target and --database those worked fine for me!

```
drush sql:sync @silver @self --database=external -y
You will destroy data in drupal8 and replace with data from xxxxxxxxx.be/beolregapi.

 // Do you really want to continue?: yes.

 [notice] Starting to dump database on source.
 [success] Database dump saved to /home/cicd-customer/drush-backups/xxxxxxx.sql.gz
 [notice] Starting to discover temporary files directory on target.
You will delete files in /tmp/xxxxxxxx.sql.gz and replace with data from xxxxx@xxxxxxx:/home/xxxxx.sql.gz

 // Do you want to continue?: yes.

 [notice] Copying dump file from source to target.
 [notice] Starting to import dump file onto target database.
```
After which data appeared fine in the database. 

My approach works because underlaying in the `sql:sync` the standard `drop` and `rsync` command is being used which both support the options from `optionset_sql`

BramDriesen:
I only tested --target and --database those worked fine for me!

```
drush sql:sync @silver @self --database=external -y
You will destroy data in drupal8 and replace with data from xxxxxxxxx.be/xxxx.

 // Do you really want to continue?: yes.

 [notice] Starting to dump database on source.
 [success] Database dump saved to /home/xxxxx/drush-backups/xxxxxxx.sql.gz
 [notice] Starting to discover temporary files directory on target.
You will delete files in /tmp/xxxxxxxx.sql.gz and replace with data from xxxxx@xxxxxxx:/home/xxxxx.sql.gz

 // Do you want to continue?: yes.

 [notice] Copying dump file from source to target.
 [notice] Starting to import dump file onto target database.
```
After which data appeared fine in the database. 

My approach works because underlaying in the `sql:sync` the standard `drop` and `rsync` command is being used which both support the options from `optionset_sql`

BramDriesen:
This is also reflected in the `--help` command.

```
~> drush sql:sync --help
Copy DB data from a source site to a target site. Transfers data via rsync.

Examples:
  drush sql:sync @source @self Copy the database from the site with the alias 'source' to the local site.
  drush sql:sync @self @target Copy the database from the local site to the site with the alias 'target'.
  drush sql:sync #prod #dev    Copy the database from the site in /sites/prod to the site in /sites/dev (multisite installation).

Arguments:
  source A site-alias or the name of a subdirectory within /sites whose database you want to copy from.
  target A site-alias or the name of a subdirectory within /sites whose database you want to replace.

Options:
  --no-dump                                     Do not dump the sql database; always use an existing dump file.
  --no-sync                                     Do not rsync the database dump file from source to target.
  --runner=RUNNER                               Where to run the rsync command; defaults to the local site. Can also be 'source' or 'target'.
  --create-db                                   Create a new database before importing the database dump on the target machine.
  --db-su=DB-SU                                 Account to use when creating a new database (e.g. root).
  --db-su-pw=DB-SU-PW                           Password for the db-su account.
  --target-dump=TARGET-DUMP                     The path for storing the sql-dump on target machine.
  --source-dump[=SOURCE-DUMP]                   The path for retrieving the sql-dump on source machine.
  --skip-tables-key=SKIP-TABLES-KEY             A key in the $skip_tables array. @see example.drush.yml
  --structure-tables-key=STRUCTURE-TABLES-KEY   A key in the $structure_tables array. @see example.drush.yml
  --tables-key=TABLES-KEY                       A key in the $tables array.
  --skip-tables-list=SKIP-TABLES-LIST           A comma-separated list of tables to exclude completely.
  --structure-tables-list=STRUCTURE-TABLES-LIST A comma-separated list of tables to include for structure, but not data.
  --tables-list=TABLES-LIST                     A comma-separated list of tables to transfer.
  --database[=DATABASE]                         The DB connection key if using multiple connections in settings.php. [default: "default"]
  --target[=TARGET]                             The name of a target within the specified database connection. Defaults to default [default: "default"]
  --db-url=DB-URL                               A Drupal 6 style database URL.

Topics:
  drush topic docs:aliases               Creating site aliases for running Drush on remote sites.
  drush topic docs:policy                Example policy file.
  drush topic docs:configuration         Configuration overview with examples from example.drush.yml.
  drush topic docs:example-sync-via-http Extend sql-sync to allow transfer of the sql dump file via http.

Aliases: sql-sync
```

weitzman:
I'm not so inclined to show these new options. They are confusing - do they refer to the source or the target? If you really want to use them, you can setup the options using drush config or an alias. If using command specific, set them up on sql:dump, not sql:sync.

BramDriesen:
If I'm not mistaken you're confusing 2 things here. The drush aliases can be used for the _arguments_ like I'm doing in my example to target a specific site/environment (`drush sql:sync @source @target`). But this will **only** take the `[default][default]` database configured in your `settings.php`. To my knowledge (and the drush aliases documentation) you can't specify which specific database to target.

Let's say I have a website with 3 databases (not multisite), there is no way with `drush sql:sync` to sync them. This is exactly what the `--database` and `--target` options are for and how those are being utilised in the `rsync` and `sql:drop` command.

Why are you inclined to add an option that was (mistakingly?) removed from drush 9? It's clearly a missing feature that has been existing in the last 3 versions of drush. They where called a bit different in previous versions though: `--source-database` and `--target-database` with this as description: _A key in the $db_url (D6) or $databases (D7+) array which provides the data._ (I feel the description is better now!).

To my view this change is still fine, and with the information added in the `--help` command, it should be clear to everyone what it does.

BramDriesen:
> They are confusing - do they refer to the source or the target?

After reading this again I know understand your question (I think). They refer to "database target" (e.g ['default']['default'] in both the @source and in the @target.

So meaning I can do `drush sql:sync @production @staging --database=foo --target=bar` which would translate to: _sync the foo-bar database of production into staging_. Of course this would need to correspond to a configured database as `['foo']['bar']` in the `settings.php`.

greg-1-anderson:
Compare to Drush 8:
```
  --source-database                         A key in the $db_url (D6) or $databases (D7+) array which provides  
                                           the data.                                                           
 --target-database                         A key in the $db_url (D6) or $databases (D7+) array which shall     
                                           receive the data.                                                   
```

I think earlier versions of Drush also had `--source-target` and `--target-target`; these   may have been removed to avoid confusion.

greg-1-anderson:
I had to go back to Drush 4 before I found `--source-target` and `--target-target`. These apparently were removed in Drush 5, but we never received a complaint. Maybe `--source-database` / `--target-database` are sufficient. I think that these should be controlled by separate options, rather than having a single option that controls both source and target. This implies some option-remapping.

BramDriesen:
Okay fair point, the option names might indeed be a bit confusing since they share the name of an argument. What would be the approach for this? I'm in no means an expert in the drush codebase 😄 but glad to help fix this if you give me some pointers on what exactly needs to be done.

greg-1-anderson:
In sql:sync:

- Remove @optionset_sql
- Define `--source-database` and `--target-database`
- Pass `--source-database` through to sql:dump as `--database`
- Use `--target-database` as `--database` to import database received

Did not walk through the code, so there may be wrinkles. That's the general idea though.

BramDriesen:
Sorry for the delay, but I finally got around to make the changes. When working on it I was getting confused myself about the options, so I renamed them to be more clear and I added a description & example usage command.

Not sure though about the options mapping that I did, since I'm not that familiar with the Drush source code. Maybe there is another way of doing this? 

I confirm however that the changes I did work as expected.

BramDriesen:
Not really sure what caused the failed test, so some assistance is welcome here :)

weitzman:
Thanks for working on this. 

The test fail is happenning because there are new --database and --target options on the call to sql:dump. The expectation needs to have these.

However, I'm still unconvinced. The option descriptions are not clear that they refer to the database on the source. Also, I think this will add those 2 new options to the rsync and sql-query calls later. Not a big deal I suppose.

Could you try an alternate approach? Without this PR, pass --database and --target and also pass --strict=0. This should get rid of the "option not recognized" error and sql:sync will hopefully pass along the options to sql:dump. If this works we can document it in an Example and be done.

BramDriesen:
> However, I'm still unconvinced. The option descriptions are not clear that they refer to the database on the source.

They point to both the database on the source and on the target. Meaning e.g. in the example: `drush sql:sync @source @self --database-key=foo --database-target=bar` means that you need a database configured as `$databases['foo']['bar']` both on your source server (@source) as on your target server (@self).

Let me test your second approach.

BramDriesen:
Passing `--strict=0` indeed suppresses the warning that the `--database` and `--target` options don't exist on the sql:sync command, but is that really the way to go?? It feels a bit,... wrong to do it like that. In that case I would still prefer the original change I did to just add the annotation which fixed the issue as well. Instead of documenting an unintended option with a suppression of strict checking.

weitzman:
Just tested my suggestion. It worked fine. This is what I ran, inside the drupal container that ships with drush. `./drush sql:sync @self --uri=dev @self --strict=0 --target=foo --database=bar -s -v`.

Maybe its not as discoverable as having named options on sql:sync, but I think keeping the option list small and simple-ish is valuable too.

I'll add docs now.

weitzman:
Doc added at f5af9f63a

BramDriesen:
Okay seams fair to me

BramDriesen:
Reopening this because there is I think no way to achieve this now.

BramDriesen:
https://github.com/drush-ops/drush/commit/63f284846270b5cd9d781ed94d570688c1b2c688#diff-fb2c277772bf5979ffd2cd9676de3c88

PR #3953, opened 2019-02-28
Drush make is no longer maintained (fixes #3946).

### LINKED ISSUE #3946: --cache-duration-releasexml also to drush make
The problem has risen due to Drupal security updates where a module is released but not yet propagated to the list where it says it exists. When you 'drush make' the project too soon, drush will cache that particular version of missing module and will save that information locally. Now deleting all the caches manually or using 'drush cc drush' works but that deletes the downloaded files also. Without clearing the caches in all instances all next deploys will fail.

There's a option currently for 'drush dl':
--cache-duration-releasexml             Expire duration (in seconds) for release XML. Defaults to 86400 (24 hours).
That would be nice to have also for 'drush make'. I understand now that if I would do 'drush make --cache-duration-releasexml=0...' I would get always the latest release info from d.org and also keep the caches of already downloaded files.

Alternative would be always deleting the caches but that would make builds slower and would give extra load to d.org.

tormi:
As per #3946, Drush make is no longer maintained. Let's inform people about this and propose alternatives.

### DIFF
--- a/docs/make.md
+++ b/docs/make.md
@@ -1,6 +1,8 @@
 
 Drush make
 ----------
+**Drush make is no longer maintained ([#3946](https://github.com/drush-ops/drush/issues/3946#issuecomment-467861007)), use [drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project) template for Drupal projects!**
+
 Drush make is an extension to drush that can create a ready-to-use drupal site,
 pulling sources from various locations. It does this by parsing a flat text file
 (similar to a drupal `.info` file) and downloading the sources it describes. In


weitzman:
LGTM. Leaving for @greg to merge.

greg-1-anderson:
Yeah I am 👍 on this. We do not have a maintainer, and have not for years.

It would also be cool if we got a maintainer for Make who was willing to curate the make issues. Seems like a lot of people are still depending on Drush make, despite the large number of bugs and years of neglect.

I'm going to go ahead and merge here since @weitzman gave his +1 while I was typing. We can revisit if anything changes.

PR #3956, opened 2019-03-01
Make Log message status transparent

karate:
I've noticed that colored log statuses have a black background, which looks ugly on some semi-transparent terminals. Is there a reason to force the bg color?

I'm using `gnome-terminal` in `i3wm`, with `compton` for compositing. Here's a screenshot with my fix:
transparency demo

### DIFF
--- a/lib/Drush/Log/Logger.php
+++ b/lib/Drush/Log/Logger.php
@@ -55,9 +55,9 @@ class Logger extends AbstractLogger {
         $green = "[%s]";
       }
       else {
-        $red = "\033[31;40m\033[1m[%s]\033[0m";
-        $yellow = "\033[1;33;40m\033[1m[%s]\033[0m";
-        $green = "\033[1;32;40m\033[1m[%s]\033[0m";
+        $red = "\033[31m\033[1m[%s]\033[0m";
+        $yellow = "\033[1;33m\033[1m[%s]\033[0m";
+        $green = "\033[1;32m\033[1m[%s]\033[0m";
       }
 
       $verbose = drush_get_context('DRUSH_VERBOSE');


weitzman:
dupe of #3876

weitzman:
Lets get master branch fixed first

karate:
I don't think that this applies to master branch.

Drush 9 doesn't have transparent background on status messages, that's why I closed #3876 

Here's the output of Drush 9:
drush 9

weitzman:
OK, I did not realize that this changed.

karate:
So, can we re-open re PR?

weitzman:
github wont let me.

greg-1-anderson:
Reopen is disabled for me as well. Odd.

greg-1-anderson:
I'm not sure about this PR, though. I wonder if the message might be hard to read on some terminals if we did not force the background color.

In general I am in favor of enhancing the appearance of Drush's output, but let's make sure that any improvements we make are widely usable. I don't think that the black background looks too bad, and Drush 8 is in maintenance-only mode. Maybe we should leave the logs as they are in Drush 8, and focus efforts on making the Drush 9 logs look their best.

karate:
As I said in #3876, the terminal emulator's theme should make sure that all available font colors are readable on each background. Overriding the theme's colors sounds like bad practice.

greg-1-anderson:
You made two assertions:

1. The terminal emulator's theme should make sure that all available font colors are readable on each background
2.  Overriding the theme's colors sounds like bad practice.

There's a discrepancy here that I don't understand, and perhaps you can explain. How does one use just "available" font colors without "overriding the theme's colors"?  I only know how to set text and background colors to specific values; for example, I could set the text color to black. That wouldn't look good on a black background. What sort of terminal commands does one have to use in order to be sure that one is only using "available" colors?

To move forward here, we need two things:

1. Start with the master branch (Drush 9). Note that Drush 9 inherits log colors from [Robo](https://github.com/consolidation/robo). It's possible to override the Robo colors in Drush, but it might be even better to put the color enhancements in the upstream project, which could also benefit from improved appearance. (I am also a maintainer of Robo, so switching projects won't be any slower than working here.)
2. Demonstrate that the colors selected work well in more than one Terminal theme, ideally one light and one dark.

karate:
I'm not an expert, but here's what I understand:

Drush uses 3/4 bit [ANSI Escape Codes](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors) to print colored text.

That's 8 basic colors and 8 secondary (bright). These codes tell the terminal emulator to print a color (eg: Yellow), but the exact RGB value is decided by the terminal emulator. Most modern terminal emulators (eg: gnome-terminal, konsole) even support different color schemes.

So, when a terminal emulator uses a color scheme with a white backgroung for example, it usually makes sure that at least all 8 basic colors are visible (by making the yellow more orange-ish for example).

What I'm saying is that we shouldn't force a black backgoung, and should respect the terminal's color scheme. Maybe "overriding the theme's colors" was not very accurate.

I'm not using Drush 9, and I'm not familiar with Robo.

I can get some more screenshots with different color schemes, but I don't think that would be of much use, since I'll be setting the colors myself.

karate:
To add to the previous comment, a terminal may not even have a black background.

For example, my terminal emulator may have a color scheme with a blue background. It's up to the terminal emulator to make sure that that ANSI color `34` (blue) is readable on that background.

Here's another example with the current code (forced black background)
Forced black BG

And the PR (no background set)
No BG set

PR #3959, opened 2019-03-02
Try running all tests together

greg-1-anderson:
This has failed in the past; let's see if it has cleared up.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ jobs:
       - run: composer install
       - run: $HOME/drush/.circleci/patch.sh
       - run: composer test
-      - run: composer exec php-coveralls -v
+      - run: composer exec coveralls -v
 
   # See scheduled workflow
   test_71_sqlite:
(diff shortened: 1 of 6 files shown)


greg-1-anderson:
I got the tests to pass in 449328c, but moving on to check our coverage on Coveralls I ran into some trouble. We're running out of memory in various instances, e.g. doing `composer update` for highest / lowest testing, and just running tests.

It is probably possible to reduce the memory that Composer needs by moving phpunit, coveralls and the code sniffer into the docker container. If we run out of memory testing with code coverage on, though, I'm not sure if there's anything to be done about that.

greg-1-anderson:
If we ran all of the tests in CircleCI, we could stop testing on Travis and Shippable. The main motivation for doing this, though, is to get uniform code coverage repots across all three test suites. If we're blocked from doing that by memory limitations, I'm not sure there's a good way to continue here.

greg-1-anderson:
May return to this someday - but not today.

Issue #3961, opened 2019-03-03
Allow other packages to use the integration test helper classes


Sweetchuck:
As a Drush command developer I also want to write _integration_ tests for commands.

There are helper class under `tests/unish/` directory,
and I can reach them if I add the following lines to my `composer.json`
```json
{
    "require-dev": {
        "drush/drush": "dev-master#9.6.0-rc3"
    },
    "autoload-dev": {
        "psr-4": {
            "Unish\\": "vendor/drush/drush/tests/unish/"
        }
    }
}
```
But the problem is that a lot of thing is hard-coded and impossible to override them, because the `self::` keyword is used instead of `static::`, or the _private_ access modifier is used.

The hard-coded things are the following:
* Path to composer root
* Path to web root
* Path to Drush executable
* path to Git executable
* List of global options (--simulate, --root, --uri etc...)
* Multi value options can appear only once. (--include --config)
* etc ...

greg-1-anderson:
I am -1 on Drush extensions reaching inside Drush and using its test code. I know devel does this.

I'd prefer to see DrushTestTraits that extensions could include and use on their own. See [example-drush-extension](https://github.com/drush-ops/example-drush-extension), in progress. The test traits there need a little work before they can be extracted and re-used, but I think they are close. Perhaps you could help out there?

Ideally, Drush could use the same traits. I don't remember if I have come across any reason why that  would be inconvenient.

weitzman:
Dupe of #3530. I'd like to see this come to fruition as greg describes.

Sweetchuck:
You don't have to encourage others to use the testing environment this way, but at least do not close the door until there is no another option.

greg-1-anderson:
Your PR is still open. We tend to not keep all the duplicate issues open in this queue while work is in progress.

PR #3962, opened 2019-03-03
Issue #3961 - Allow other packages to use the integration test helper…

Sweetchuck:
Issue #3961 - Allow other packages to use the integration test helper classes.

### DIFF
--- a/tests/phpunit.xml.dist
+++ b/tests/phpunit.xml.dist
@@ -8,17 +8,16 @@
       tests without defining these variables if you use the default values.
 -->
 <phpunit
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="../vendor/phpunit/phpunit/phpunit.xsd"
     bootstrap="../vendor/autoload.php"
-    backupGlobals="false"
-    backupStaticAttributes="false"
-    syntaxCheck="false"
     colors="true">
   <testsuites>
     <testsuite name="unit">
-      <directory suffix="Test.php">unit</directory>
+      <directory>unit</directory>
     </testsuite>
     <testsuite name="integration">
-      <directory suffix="Test.php">integration</directory>
+      <directory>integration</directory>
     </testsuite>
     <testsuite name="functional">
       <directory>functional</directory>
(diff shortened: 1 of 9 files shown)


greg-1-anderson:
These changes look beneficial. I haven't reviewed closely enough to say that we definitely want everything here, but I gave it a quick look over and I think that we want most if not all. The "lowest" test is currently failing.

tests/phpunit.xml.dist
```diff
@@ -8,17 +8,16 @@
...
+    xsi:noNamespaceSchemaLocation="../vendor/phpunit/phpunit/phpunit.xsd"
     bootstrap="../vendor/autoload.php"
-    backupGlobals="false"
-    backupStaticAttributes="false"
-    syntaxCheck="false"
     colors="true">
   <testsuites>
     <testsuite name="unit">
-      <directory suffix="Test.php">unit</directory>
+      <directory>unit</directory>
```
greg-1-anderson: I'm not sure we want this change. Other projects won't be using our phpunit.xml.dist directly, will they? If someone copies this template to their own project, they may remove the suffix at that time if they do not want it.

For core Drush, I think we want to constrain that tests end in `Test.php`.

greg-1-anderson:
Except as noted, I am 👍 on the changes here as general improvements (code correctness) of Unish.

If you put the `Test.php` suffix back, rebase or merge with master, and get the `lowest` tests passing, then I'm in favor of merging.

tests/phpunit.xml.dist
Sweetchuck: OK, I can put it the suffix back. The only reason I removed it because it has a default value.
https://github.com/sebastianbergmann/phpunit/blob/6.5.14/phpunit.xsd#L164

greg-1-anderson: Oh, ok. In that case I am 👍 on taking it out.

greg-1-anderson:
So just rebase / merge with master and get the "lowest" test passing and this is ready to go in.

greg-1-anderson:
Note that the test utils went through a big refactor; external code is now expected to use Drush Test Traits. See the docs and the devel module.

Some of the changes here might still be beneficial; please open a new PR based on the new code if you want to push this forward.

PR #3966, opened 2019-03-04
Use --uri from commandline over uri in an alias

greg-1-anderson:
This is a start, but it looks like there are a few failures in the functional tests.

### DIFF
--- a/src/Preflight/PreflightArgs.php
+++ b/src/Preflight/PreflightArgs.php
@@ -276,6 +276,11 @@ class PreflightArgs extends Config implements PreflightArgsInterface
         return $this->get(self::URI, $default);
     }
 
+    public function hasUri()
+    {
+        return $this->has(self::URI);
+    }
+
     /**
      * Set the uri option
      */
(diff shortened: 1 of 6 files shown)


greg-1-anderson:
The tests like to insert `--uri=dev` by default, even if no one asked for that, for historical reasons. With this PR, that addition becomes significant, which broke some tests.

I think that the most recent commit gets us down to about one remaining failure.

greg-1-anderson:
Somewhat related to https://github.com/drush-ops/drush/issues/3968

weitzman:
The way to do this today is:

Old: `drush --uri=http://example.com`
New: `DRUSH_OPTIONS_URI=http://example.com drush ...`

greg-1-anderson:
`DRUSH_OPTIONS_URI` is a decent workaround (it does work fine today), but `--uri` on the commandline has better DX and is familiar. There's no downside to making the enhancement to take `--uri` from the commandline.

weitzman:
LGTM. Please close any other related issues that are still open,

PR #3971, opened 2019-03-06
WIP: Try testing sql:sync with a local Drupal site that has a site-lo…

greg-1-anderson:
…cal Drush 9.5.

Not high priority; it's the remote case that is broken. Local works fine.


greg-1-anderson:
Might come back to this later. Maybe not.

PR #3974, opened 2019-03-07
Add a test for UserAbortException

greg-1-anderson:
\Throwable is documented as being first available in PHP 7:

http://php.net/manual/en/class.throwable.php

### DIFF
--- a/src/Exceptions/UserAbortException.php
+++ b/src/Exceptions/UserAbortException.php
@@ -7,7 +7,7 @@ namespace Drush\Exceptions;
  */
 class UserAbortException extends \Exception
 {
-    public function __construct($message = "Cancelled.", $code = 0, \Throwable $previous = null)
+    public function __construct($message = "Cancelled.", $code = 0, \Exception $previous = null)
     {
         parent::__construct($message, $code, $previous);
     }
--- a/tests/unit/UserAbortExceptionTest.php
+++ b/tests/unit/UserAbortExceptionTest.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Unish;
+
+use PHPUnit\Framework\TestCase;
+
+/**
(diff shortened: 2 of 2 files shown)


greg-1-anderson:
This PR documents what works and what does not vis-a-vis `Throwable` and PHP 5.x.

src/Commands/core/UpdateDBCommands.php
```diff
@@ -205,14 +205,14 @@ public static function updateDoOne($module, $number, $dependency_map, &$context)
...
                 $ret['results']['success'] = true;
-            } catch (\Throwable $e) {
-                // PHP 7 introduces Throwable, which covers both Error and Exception throwables.
-                $ret['#abort'] = ['success' => false, 'query' => $e->getMessage()];
-                Drush::logger()->error($e->getMessage());
             } catch (\Exception $e) {
                 // In order to be compatible with PHP 5 we also catch regular Exceptions.
                 $ret['#abort'] = ['success' => false, 'query' => $e->getMessage()];
                 Drush::logger()->error($e->getMessage());
+            } catch (\Throwable $e) {
```
weitzman: I got the current ordering from https://github.com/drupal/drupal/blob/8.6.x/core/includes/bootstrap.inc#L597-L613. Happy to go with your ordering if thats better.

greg-1-anderson: Interesting. Maybe I need another test.  😛

greg-1-anderson: Yeah, sure enough, the Drupal-inspired order works. Would be pretty surprising / impossible for it to be otherwise.

greg-1-anderson:
Kept the same order as Drupal for catching both Exception and Throwable.

PR #3978, opened 2019-03-09
Don't call bootstrapValidate commands repeatedly, and add --uri tests

rmuit:
These are two followups to #3847 which are not super related, but I hope putting them in one PR turns out to be less work.

1) I found out why bootstrapDrupalSiteValidate() is called a lot more lately (since #3820):
* BootstrapManager::bootstrapMax() has a loop for each phase, in which it calls bootstrapValidate()
* bootstrapValidate() only checks its $result_cache for _the current target phase_ is filled, and if it is not, validates _all phases from 0 up to this target phase_ again.

I'm going to just assume that was not intentional and that it can be fixed this way.

--

2) It was suggested to add a test, to add against regressions with the --uri flag.

So here it is. I'm guessing a fairly simple addition to CoreTest::testOptionsUri() will do the trick, even though that means it's now used to test two slightly different things.

Note - I just assumed that the 'expected' URLs are the ones that are returned by 'drush status' at this moment. I'm not 100% sure if that's true; you're in a better position to judge:
* test.uri/ leads to outputting "http://test.uri/"  for the Site URI. That wasn't the case before #3820 - i.e. in 9.6.0-beta3 (and all the way down to drush6), "test.uri/" was output instead. 
* I could imagine that http://test.uri and http://test.uri/ should actually both output the same result (i.e. without trailing slash) - but this was never the case before, so...

--

If this PR (containing _both_ commits) is applied to the commit from before #3847, the test for "--uri=test.uri/subpath" fails as expected.

### DIFF
--- a/tests/functional/CoreTest.php
+++ b/tests/functional/CoreTest.php
@@ -47,13 +47,6 @@ class CoreCase extends CommandUnishTestCase
         'format' => 'json',
         'uri' => 'OMIT', // A special value which causes --uri to not be specified.
         ];
-        // Modify settings.php to check whether site paths correctly resolve.
-        $root = $this->webroot();
-        $fp = fopen($root . '/sites/sites.php', 'a+');
-        fwrite($fp, '
-$sites["test.uri"] = "dev";
-$sites["test.uri.subpath"] = "stage";
-');
         foreach ([
                    'test.uri' => ['http://test.uri', 'sites/dev'],
                    'test.uri/' => ['http://test.uri/', 'sites/dev'],
--- a/tests/functional/CoreTest.php
+++ b/tests/functional/CoreTest.php
@@ -40,23 +40,47 @@ class CoreCase extends CommandUnishTestCase
 
     public function testOptionsUri()
     {
-        // Put a yml file in the drush folder.
+        // Test whether a URI in a config file resolves correctly, and test
+        // various URI values for their expected Site URI and path.
         $drush_config_file = Path::join($this->webrootSlashDrush(), 'drush.yml');
-        $test_uri = 'http://test.uri';
-        $options_with_uri = [
-        'options' => [
-        'uri' => $test_uri,
-        ],
-        ];
-        $options = [
+        $command_options = [
         'format' => 'json',
         'uri' => 'OMIT', // A special value which causes --uri to not be specified.
         ];
-        file_put_contents($drush_config_file, Yaml::dump($options_with_uri, PHP_INT_MAX, 2));
-        $this->drush('core-status', [], $options);
-        unlink($drush_config_file);
-        $output = $this->getOutputFromJSON();
-        $this->assertEquals($test_uri, $output->uri);
+        // Modify settings.php to check whether site paths correctly resolve.
+        $root = $this->webroot();
+        $fp = fopen($root . '/sites/sites.php', 'a+');
+        fwrite($fp, '
+$sites["test.uri"] = "dev";
+$sites["test.uri.subpath"] = "stage";
+');
+        foreach ([
+                   'test.uri' => ['http://test.uri', 'sites/dev'],
+                   'test.uri/' => ['http://test.uri/', 'sites/dev'],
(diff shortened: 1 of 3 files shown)


tests/functional/CoreTest.php
```diff
@@ -40,23 +40,47 @@ public function testSiteSelectionViaCwd()
...
-        unlink($drush_config_file);
-        $output = $this->getOutputFromJSON();
-        $this->assertEquals($test_uri, $output->uri);
+        // Modify settings.php to check whether site paths correctly resolve.
+        $root = $this->webroot();
+        $fp = fopen($root . '/sites/sites.php', 'a+');
+        fwrite($fp, '
+$sites["test.uri"] = "dev";
+$sites["test.uri.subpath"] = "stage";
+');
```
rmuit: I'm not sure if I should have reverted sites.php after this test finishes... You're not using "test.uri" anywhere else apparently, because tests still succeed.

weitzman:
I'm not sure about the test (someone else should review it), but the bootstrapValidate() change looks right to me. I'll merge just that part, unless @greg-1-anderson chimes in otherwise.

tests/functional/CoreTest.php
```diff
@@ -40,23 +40,47 @@ public function testSiteSelectionViaCwd()
 
     public function testOptionsUri()
     {
-        // Put a yml file in the drush folder.
+        // Test whether a URI in a config file resolves correctly, and test
+        // various URI values for their expected Site URI and path.
         $drush_config_file = Path::join($this->webrootSlashDrush(), 'drush.yml');
-        $test_uri = 'http://test.uri';
```
greg-1-anderson: We still want to test with `--uri=http://test.uri`. Looks like you removed this test.

rmuit: Two answers, to make surer I'm not misinterpreting:
* This test _is_ still being done; it's just done as part of the loop.
* This (the current/old test) isn't testing with --uri=http://test.uri; it's only writing "http://test.uri" to the config file and testing with --uri=OMIT on the commandline. It's slightly hard to see because of naming, that $options_with_uri is not the --uri option on the commandline. Which is also why I renamed the variables.

rmuit: I could _add a new_ drush invocation that tests --uri on the commandline (not writing config fille), though. Either with only one flag, or with all of them.

greg-1-anderson: I'll have to re-read it later in context, but I think that `$options_with_uri` *does* have the `--uri` option on the commandline. At least that was my impression.

greg-1-anderson: Ah, I see now. Tests look good.

tests/functional/CoreTest.php
greg-1-anderson: Shouldn't hurt to leave sites.php in place; you could even commit this to the git repository if you wanted. Optional.

rmuit:
Pushed a 3rd commit moving the $sites configuration into sut/sites/example.sites.php as suggested... since that is the file which is copied.
Tests OK locally.

Commit can be deleted if it turns out that keeping example.sites.php pristine is a better idea.

greg-1-anderson:
I forgot that we overwrite sites.php with example.sites.php repeatedly. Modifying example.sites.php would be okay, but Drupal Scaffold is going to overwrite this on `composer update`. We could either revert the last commit as suggested, or stop Drupal Scaffold from updating example.sites.php. I don't think we need an up-to-date version of this file for the sut.

greg-1-anderson:
If we set `extras.drupal-scaffold.excludes` to `sites/example.sites.php`, then Drupal Scaffold would not overwrite the example.sites.php file.

@weitzman Do you have a preference about committing example.sites.php vs dynamically writing sites.php?

weitzman:
I dont have a preference.

greg-1-anderson:
I have a slight preference for committing static fixtures; I'll add the Drupal Scaffold exclude and merge in a little bit.

greg-1-anderson:
We are not using Drupal Scaffold at the moment, so this is good to go.

PR #3979, opened 2019-03-10
LanguageAdd test fails on Highest test

Sutharsan:
`LanguageAddCase::testLanguageAddWithTranslations()` fails when running the 'test_72_highest' test scenario.

The test fails because the dev-master version of the devel module is used in the test. The translation import process requires a module to have a 'project' definition in its info.yml file. But this is only added by the drupal.org packaging system and is not included in a dev release.

There are several possible solutions:
1. Use a stable release of the devel module in the test.
2. Add `interface translation project: devel` to the devel.info.yml. 
3. Use the locale_test module instead of the devel module.
4. Use a custom test module instead of devel.
5. Use `hook_system_info_alter()` to set 'project' for the devel module.
6. Modify the content of the devel.info.yml file inside the SuT.

Re 1. I failed to get this done. I can't explain why a dev version is downloaded by composer.
Re 2. 'interface translation project' serves as a placement of the 'project' definition. But this requires a change in devel module.
Re 3. locale_test is a hidden core module. This requires `$settings['extension_discovery_scan_tests'] = TRUE;` to be set in settings.php of the SuT. I failed to get this working from within the `LanguageAddCase` class.
Re 4. This module must have 'interface translation project' set and (optionally) have it own translation files.
Re 5. As far as I understand this requires a module to be added to the SuT that utilizes this hook.
Re 6. This is rather hacky, but can be done from within the test.

This PR is an initial start of option 4, the custom module. The custom module can be found at https://github.com/Sutharsan/language_test, but should be relocated to either drupal.org or the https://github.com/drush-ops project.


greg-1-anderson:
Perhaps as an interim solution we could simply mark this test as skipped when the project definition is missing.

Sutharsan:
Yes, that is definitively an solution. Did not think of that.
Either check the content of the devel.info.yml for the absence of 'project:' or use `drush pm:list` and check for '-dev' in the version string.

weitzman:
How about we add that line to devel.info.yml? At least the test would run and verify functionality?

Any chance we can emit a notice or debug message in this case so that folks can understand why the feature isn't working?

weitzman:
I started adding this key to devel but its apparently discouraged in https://drupal.stackexchange.com/questions/233946/what-does-the-project-key-in-module-info-yml-do. is there any way Drush can stop requiring this key?

Sutharsan:
The tests requires this key because it identifies the module as being (interface) translatable. The same key is used by Update module to identify modules sourced at d.o. The interface translation can not do without. The only alternatives I know:
- interface translation project (a locale-only alternative for the 'project' key)
- hook_system_info_alter() (to add the (interface translation) project key programatically)

greg-1-anderson:
I don't think it's worth solving this problem just to make the tests run in `highest`. If the proposed solutions would also benefit folks who are actually using dev modules and would like to translate them, then I'd be +1 on pushing forward on one of them. I'm uncertain about the use-case and how the parts fit together, but let's frame solutions in terms of how they benefit end users, and just mark the test skipped until then.

My opinion, anyway.

weitzman:
I implemented the skip in https://github.com/drush-ops/drush/pull/3979

PR #3987, opened 2019-03-11
Do not setInput on the process object unless STDIN actually was redirected.

greg-1-anderson:
If we can't tell, we'll assume no input redirection was done.

Without this, the remote side may hang accessing the passed STDIN. This seems to only be a problem if the remote side cannot create a tty.


src/Runtime/RedispatchHook.php
```diff
@@ -93,11 +94,17 @@ public function redispatch(InputInterface $input)
         // The options the user provided on the commandline will be included
         // in $redispatchArgs.
         $redispatchOptions = [];
+        if (!TerminalUtils::stdinIsTerminal()) {
+            // Force non-interactive mode for the remote command when redirecting stdin
+            $redispatchOptions['no-interaction'] = true;
```
greg-1-anderson: I don't think this is the right way to solve this part of the problem. I'll push another commit in a bit.

greg-1-anderson:
This seems to work in ad-hoc testing.

src/Utils/TerminalUtils.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Drush\Utils;
+
+class TerminalUtils
+{
+    public static function stdinIsTerminal($default = true)
```
weitzman: Seems like isTty() would be more consistent name. 

Could so with doxygen here if you have time.

src/Runtime/RedispatchHook.php
```diff
@@ -96,8 +97,11 @@ public function redispatch(InputInterface $input)
 
         $aliasRecord = $this->siteAliasManager()->getSelf();
         $process = $this->processManager->drushSiteProcess($aliasRecord, $redispatchArgs, $redispatchOptions);
-        $process->setTty($this->getConfig()->get('ssh.tty', $input->isInteractive()));
-        $process->setInput(STDIN);
+        if (!TerminalUtils::stdinIsTerminal()) {
```
greg-1-anderson: I am tempted to simplify this entire block to:
```
$process->setDefaultRedirectBehavior(STDIN, $input, $config, 'ssh.tty');
```
That's too many parameters, though.

Maybe:
```
$process->setDefaultRedirectBehavior(STDIN, $input, $this->getConfig()->get('ssh.tty'));
```

src/Commands/sql/SqlCommands.php
```diff
@@ -142,8 +143,11 @@ public function cli(InputInterface $input, $options = ['extra' => self::REQ])
     {
         $sql = SqlBase::create($options);
         $process = $this->processManager()->shell($sql->connect(), null, $sql->getEnv());
-        $process->setTty($this->getConfig()->get('ssh.tty', $input->isInteractive()));
-        $process->setInput(STDIN);
+        if (!TerminalUtils::stdinIsTerminal(false)) {
```
greg-1-anderson: @weitzman Do you think it would be worthwhile to simplify this with a $process helper?
```
$process->setDefaultRedirectBehavior(STDIN, $input, $this->getConfig()->get('ssh.tty'));
```
Then the TerminalUtils equivalent moves into the site-process library.

I tend to think this is better. Thinking about it.

greg-1-anderson: Decided against the wrapper function for now. Maybe later.

PR #3989, opened 2019-03-12
Fix #3979. Skip testLanguageAddWithTranslations() when using a dev snapshot of devel

weitzman:


weitzman:
Highest test runs each night so the full story will be reveled tonight. I'll adjust if this continues to fail.

PR #3994, opened 2019-03-13
Fix $this being used in a static context in Drush.php.

echernyavskiy:
Discovered via the _PHPCompatibility.Variables.ForbiddenThisUseContexts.OutsideObjectContext_ PHPCompatibility PHPCS rule:

> "$this" can no longer be used in a plain function or method since PHP 7.1.


greg-1-anderson:
Thanks.

Issue #3930, opened 2019-02-13
_drush_backend_get_stdin() return FALSE on Windows 10


xandeadx:
I work on issue #3315 and found a bug. If execute command `drush make ...` drush creates new process using [proc_open()](https://github.com/drush-ops/drush/blob/8.x/includes/backend.inc#L379).
Step 2 - drush send options in process stdin using [fwrite()](https://github.com/drush-ops/drush/blob/8.x/includes/backend.inc#L382).
Step 3 - in the new process drush read options from stdin using [_drush_backend_get_stdin()](https://github.com/drush-ops/drush/blob/8.x/includes/bootstrap.inc#L566).

**Problem**
In Windows 7 it's work good, `_drush_backend_get_stdin()` return array, but Windows 10 function return FALSE. This is because [stream_select($read, $write, $except, 1)](https://github.com/drush-ops/drush/blob/8.x/includes/backend.inc#L1276) always return `0`.

If comment (or delete) next code, function as expected will return array:

~~~
  if (drush_is_windows()) {
    $read = array($fp);
    $write = NULL;
    $except = NULL;
    $changed_streams = stream_select($read, $write, $except, 0);
    if ($changed_streams < 1) {
      return FALSE;
    }
  }
~~~

I don't know why `stream_select()` return `0`.

Drush version: 8.x-dev
PHP version: 7.2, 7.1, 7.0, 5.6 (tried at al)
Windows version: 10 (1809)
Shell: cmd, cygwin, git bash (tried at al)

greg-1-anderson:
`stream_select()` is checking to see whether there is anything ready to be read on the stdin stream. The way this code is written presents a possible race condition. You could try adding a very short sleep before selecting, to see if that gives Win10 time enough to present stdin to php.

I don't know if there's a better way. In Drush 9 we no longer use backend invoke; we will remove it in Drush 10.

xandeadx:
I try add `sleep(1)` before `stream_select()` and set stream_select `tv_set` param to `1` but it's does not help.

Without next code all works good:

~~~
    if ($changed_streams < 1) {
      return FALSE;
    }
~~~

greg-1-anderson:
Maybe conditionalize the whole branch so that the stream_select is skipped on Windows 10 and later, but runs for Windows 7 and non-Windows machines. (I do not know what to advise for Windows 8.)

[...]

PR #2231, opened 2016-06-20
Prevent drush make failures

acrollet:
Resolves #1887

PR #4002, opened 2019-03-19
Update backend.inc

majithiyakavan:
I worked on issue #3930.

I changed condition 

`if ($changed_streams > 0) {
      return FALSE;
 }`

Now `drush status --backend` and `drush make` both commands working fine in windows 10 and windows 7 also.

### DIFF
--- a/includes/backend.inc
+++ b/includes/backend.inc
@@ -1279,7 +1279,7 @@ function _drush_backend_get_stdin() {
     // stream_select will return FALSE for streams returned by proc_open.
     // That is not applicable to us, is it? Our stream is connected to a stream
     // created by proc_open, but is not a stream returned by proc_open.
-    if ($changed_streams < 1) {
+    if ($changed_streams > 0) {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
I think this solution has cropped up before. It's not impossible to do this, but I think it causes blocking in some circumstances.

greg-1-anderson:
How does `drush status --backend` behave with this PR?

See also #2231

majithiyakavan:
`drush status --backend` gives this as output in Windows 10.
 
image

and I expect, it behaves same in windows 7 and Linux.

Also I  will refer #2231 and try to get that, in which circumstances this causes blocking.

greg-1-anderson:
And `drush8 ev 'print(json_encode(array("test" => "XYZZY")));' | drush status --backend` works also?  If so, perhaps this fix did the trick.

Could you update the comments to explain what is happening?

weitzman:
No activity

PR #4006, opened 2019-03-20
Add a note about code generators depending on composer

rodrigoaguilera:


weitzman:
Drush's install docs only allow for site local installation via Composer. So we don't officially support a site that isn't built by Composer. Thats why I'm hesitating on this PR.

rodrigoaguilera:
I think is important to note that before drush 9 adding commands to a module made drush optional but since they are a service now the class that service extends from need to exists. I think this can be useful for people coming from drush < 9.

I'm in favor for only supporting composer installation but this is an issue I faced and I wish it was stated somehow in the docs.

Thank you for considering it. Great work on drush :)

weitzman:
Is this specific to generators or also true of commandfiles? They are services too.

rodrigoaguilera:
I did not read the docs properly and I was adding my generator to mymodule.services.yml instead of drush.services.yml. I will fix my module. the docs are alright.

Thank you for the support.

Issue #2909, opened 2017-08-28
Relative root paths no longer seem to work


jhedstrom:
When switching from Drush 8 to Drush 9, I noticed that relative `--root` paths don't seem to work for providing the core location.

Here's output for a build that exists in `build/html`:

**With relative `--root`**
```
$> vendor/bin/drush -r build/html status
 PHP binary    : /usr/local/Cellar/php70/7.0.12_5/bin/php
 PHP config    : /usr/local/etc/php/7.0/php.ini
 PHP OS        : Darwin
 Drush script  : /.../vendor/bin/drush
 Drush version : 9.0-dev-g65c2383
 Drush temp    : /tmp
 Drush configs :
 Drush aliases : /.../.drush/aliases.drushrc.php
                 /.../.drush/drupalEight.aliases.drushrc.php
```

**With no `--root`** (behaves the same as with absolute path for root arg)
```
$> vendor/bin/drush status

 Drupal version : 8.4.0-beta1
 Site URI       : http://default
 PHP binary     : /usr/local/Cellar/php70/7.0.12_5/bin/php
 PHP config     : /usr/local/etc/php/7.0/php.ini
 PHP OS         : Darwin
 Drush script   : /.../vendor/bin/drush
 Drush version  : 9.0-dev-g65c2383
 Drush temp     : /tmp
 Drush configs  :
 Drush aliases  : /.../.drush/aliases.drushrc.php
                  /.../.drush/drupalEight.aliases.drushrc.php
 Drupal root    : /.../build/html
 Site path      : sites/default
```

greg-1-anderson:
We're rewriting the bootstrap / preflight at the moment. See https://github.com/drush-ops/drush/pull/2843. Let's make sure it works there.

weitzman:
Looks fixed to me.

pingers:
Relative paths working here too - thanks!

pjcdawkins:
I've found a relative `--root` path does not work in Drush 9.5.2. I suppose there was a regression?

greg-1-anderson:
How about in 9.6.0-rc3?

greg-1-anderson:
Hm, yes, broken in HEAD of master too.

pjcdawkins:
Hi @greg-1-anderson I did some digging and opened https://github.com/webflo/drupal-finder/pull/40

greg-1-anderson:
[...]

PR #4019, opened 2019-03-27
#2909: Relative root paths no longer seem to work

JTubex:
**Describe the bug**
#2909 
When executing any drush command in a relative root path, the command times out and gives a segmentation fault.

**To Reproduce**
execute `drush cr`.

**Expected behavior**
A cache rebuild.

**Actual behavior**
The command timed out and gave a segmentation fault.

When executing drush cr -vvv the result was the following:

```
[preflight] Redispatch to site-local Drush: /var/www/..../htdocs/vendor/drush/drush/drush.
[preflight] Redispatch to site-local Drush: /var/www/..../htdocs/vendor/drush/drush/drush.
[preflight] Redispatch to site-local Drush: /var/www/..../htdocs/vendor/drush/drush/drush.
....
```

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.6.x. 9.5 still works
| Drupal version? | 8.6
| PHP version     | 7.2
| OS?             | Linux

### DIFF
--- a/src/Config/Environment.php
+++ b/src/Config/Environment.php
@@ -34,7 +34,7 @@ class Environment
     public function __construct($homeDir, $cwd, $autoloadFile)
     {
         $this->homeDir = $homeDir;
-        $this->originalCwd = Path::canonicalize($cwd);
+        $this->originalCwd = FsUtils::realpath($cwd);
         $this->etcPrefix = '';
         $this->sharePrefix = '';
         $this->drushBasePath = Path::canonicalize(dirname(dirname(__DIR__)));
--- a/src/Config/Environment.php
+++ b/src/Config/Environment.php
@@ -34,7 +34,7 @@ class Environment
     public function __construct($homeDir, $cwd, $autoloadFile)
     {
         $this->homeDir = $homeDir;
-        $this->originalCwd = FsUtils::realpath($cwd);
+        $this->originalCwd = Path::canonicalize(FsUtils::realpath($cwd));
         $this->etcPrefix = '';
         $this->sharePrefix = '';
         $this->drushBasePath = Path::canonicalize(dirname(dirname(__DIR__)));
--- a/src/Commands/LegacyCommands.php
+++ b/src/Commands/LegacyCommands.php
@@ -54,7 +54,9 @@ class LegacyCommands extends DrushCommands
     }
 
     /**
-     * The pm-updatestatus command was deprecated. Please see `composer show` and `composer outdated`. For security release notification, see `drush pm:security`.
+     * The pm-updatestatus command was deprecated. Please see `composer show`
+     * and `composer update --dry-run`. For security release notification,
+     * see `drush pm:security`.
      *
      * @command pm:updatestatus
      * @aliases ups,pm-updatestatus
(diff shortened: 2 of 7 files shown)


weitzman:
Greg suggest Path::canonicalize() instead of realpath().

JTubex:
@weitzman, Path::canonicalize() did not resolve the issue.

The problem is that in `Drush\Preflight\RedispatchToSiteLocal::redispatchIfSiteLocalDrush()` $vendor is the absolute path and $root is the symlinked absolute path. So the check on rule 38 in `Drush\Preflight\RedispatchToSiteLocal`  will fail because the symlinked path to drush is not the same as the real path to drush.

weitzman:
I just approved https://github.com/webflo/drupal-finder/pull/40. Hopefully @webflo will merge it soon. Does that fix it for you?

JTubex:
@weitzman, The problem still occurs with the fix in webflo/drupal-finder#40.

JTubex:
@weitzman, I was able to find where the bug is located.

1. e.g. Site is located in /var/www/release/1.0.0/htdocs
2. Create a symlink /var/www/current -> /var/www/release/1.0.0
3. The website is now accessible in /var/www/current/htdocs.
4. Execute drush in /var/www/current/htdocs.

The problem occurs in repo Webflo/Drupal-finder.

In file `DrupalFinder\DrupalFinder::locateRoot` line 43 the check is_link will return `FALSE` because the webroot /var/www/current/htdocs is not a symlink, but /var/www/current is.

JTubex:
BTW: I created webflo/Drupal-finder#41

src/Config/Environment.php
```diff
@@ -34,7 +34,7 @@ class Environment
     public function __construct($homeDir, $cwd, $autoloadFile)
     {
         $this->homeDir = $homeDir;
-        $this->originalCwd = Path::canonicalize($cwd);
+        $this->originalCwd = FsUtils::realpath($cwd);
```
greg-1-anderson: `FsUtils::realpath` is a lot safer than a raw `realpath` call, so perhaps this is okay. If we add this, it means that we'll have to work harder in the AppVeyor tests, because this will convert Windows paths from using `/` to `\`.

This only introduces a handful of failures in the current tests, so I suppose it is tractable.

JTubex: @greg-1-anderson, I'm not sure what the right way is to fix the tests. Can you point me in the right direction? Thanks!

greg-1-anderson: Our typical strategy here is to normalize the output before asserting. For example, if you str_replace each `\` with a `/`, then you can compare the path invariantly again.

JTubex: Makes sense. Let's see if this fixes the tests.

greg-1-anderson:
This PR uses `realpath` rather than `canonicalize`. That is necessary, because this use-case involves symlinks. Symlinks and `realpath` have been problematic to support in the past, but perhaps might work out here.

weitzman:
That's fine. I'm arguing against the replace calls in tests. Instead, whatever is writing cwd to config should canonicalize first.

JTubex:
What about this change?
```suggestion
$this->originalCwd = Path::canonicalize(FsUtils::realpath($cwd));
```

weitzman:
Thanks for your persistence, @JTubex.

JTubex:
Great!

Issue #4020, opened 2019-03-28
Drush user:login cumbersome to use


BramDriesen:
**Describe the bug**
While I understand you can use a drush command inside a drush command likse so: `drush user:login --browser=firefox --name=$(drush user:information --mail="drush@example.org" --fields=name --format=string)`

This is very cumbersome to use and requires a lot of typing. Therefore I'm asking why the `--uid` and `--mail` options are removed in drush 9.

**Expected behavior**
`drush uli --uid=123`
`drush uli --mail=foo@bar.com`

**Workaround**
Using commands in commands.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.3.0
| Drupal version? | 8.6.5
| PHP version     | 7.1
| OS?             | Mac/Linux

BramDriesen:
It's quite easy to implement this back into the command from what I can see looking at the code.

weitzman:
Yeah, seems like those could two be added

BramDriesen:
Okay, I'll port them to Drush 9 :)

BramDriesen:
Awesome 🎉

obriat:
This also should be done for all user commands. At least for `user:block` & `user:unblock`, if you want to script disabling uid 1, specially  if you already change is name. => #4541

PR #4021, opened 2019-03-29
#4020 Added extra options for the drush user:login command

BramDriesen:
See #4020 for more information.


weitzman:
Code looks good. Will merge when it comes back green

BramDriesen:
After #c4ff6f9 it should be ok 😉

weitzman:
Thank you.

Issue #4015, opened 2019-03-26
Drush 9.6 runserver TTY mode requires /dev/tty to be read/writable


bkildow:
**Describe the bug**
After upgrading from Drush 9.5 to 9.6, my CI environment fails to run `drush runserver` with the following error:

```
  [Symfony\Component\Process\Exception\RuntimeException]  
  TTY mode requires /dev/tty to be read/writable.
```

**To Reproduce**
Ran `vendor/bin/drush runserver --no-interaction --debug` in CI

**Expected behavior**
It would run successfully and start a server

**Actual behavior**
```
[notice] HTTP server listening on 127.0.0.1, port 8888 (see http://127.0.0.1:8888/), serving site, sites/default [0.72 sec, 59.05 MB]
In Process.php line 1042:
                                                          
  [Symfony\Component\Process\Exception\RuntimeException]  
  TTY mode requires /dev/tty to be read/writable. 

Exception trace:
 () at /opt/atlassian/pipelines/agent/build/vendor/symfony/process/Process.php:1042
 Symfony\Component\Process\Process->setTty() at /opt/atlassian/pipelines/agent/build/vendor/drush/drush/src/Commands/core/RunserverCommands.php:80
 Drush\Commands\core\RunserverCommands->runserver() at n/a:n/a
 call_user_func_array() at /opt/atlassian/pipelines/agent/build/vendor/consolidation/annotated-command/src/CommandProcessor.php:257
 Consolidation\AnnotatedCommand\CommandProcessor->runCommandCallback() at /opt/atlassian/pipelines/agent/build/vendor/consolidation/annotated-command/src/CommandProcessor.php:212
 Consolidation\AnnotatedCommand\CommandProcessor->validateRunAndAlter() at /opt/atlassian/pipelines/agent/build/vendor/consolidation/annotated-command/src/CommandProcessor.php:178
 Consolidation\AnnotatedCommand\CommandProcessor->process() at /opt/atlassian/pipelines/agent/build/vendor/consolidation/annotated-command/src/AnnotatedCommand.php:302
 Consolidation\AnnotatedCommand\AnnotatedCommand->execute() at /opt/atlassian/pipelines/agent/build/vendor/symfony/console/Command/Command.php:255
 Symfony\Component\Console\Command\Command->run() at /opt/atlassian/pipelines/agent/build/vendor/symfony/console/Application.php:978
[...]

PR #4033, opened 2019-04-04
Fix TTY detection for runserver command.

danepowell:
In #4015 we attempted to fix the problem of setting TTY in environments that don't support it, which would throw an exception.

The problem is that the method by which Symfony detects whether an environment supports TTY is more complex than just checking whether /dev/tty is writable, so this could still fail in some environments (e.g. Acquia Pipelines.)

This PR updates the detection to use the exact same check as Symfony, so it's guaranteed to avoid this error.

Note that Symfony Process implemented an "isTtySupported" static method to encapsulate this check in v4, which we could convert to once Drush upgrades to Symfony Process 4.x:
https://github.com/symfony/process/commit/7bc3a902d88f2a65091773b36b640b264285c29f


greg-1-anderson:
Thanks for finding this check in Symfony.  I think what would be best would be:

1. Move this check to a utility class in consolidation/process.
2. Have the check call through to the Symfony static method if it is available.
3. Use this check anywhere Drush tries to use the posix methods.

Lettucez:
This would explain why all of our pipeline builds have been failing.

image

Thank you for your work in getting this addressed.

danepowell:
Okay I added a util class to site-process: https://github.com/consolidation/site-process/pull/41

I updated this PR to depend on that. So you'll need to review/merge that PR first, release a new version of consolidation/site-process, then we'll need to bump the site-process version number in Drush's composer.json.

> Use this check anywhere Drush tries to use the posix methods.

Do you mean that you want me to change every call of `setTty()` in Drush to be `setTty(Tty:isTtySupported())`?

It's a little weird to me that we have to manually check whether TTY is supported for every process, and then manually set it, when Symfony clearly knows if TTY is supported or not and could just enable it automatically somehow. But maybe I'm overthinking this.

danepowell:
I updated the site-process dependency to 2.0.2, so that just leaves the issue of fixing all of the other Posix / TTY calls in Drush.

danepowell:
> Do you mean that you want me to change every call of setTty() in Drush to be setTty(Tty:isTtySupported())?

@greg-1-anderson can you respond to my question above, so we can get this finished? Thanks!

greg-1-anderson:
I don't know the answer to that question. If you go ahead and do that anyway, and the answer is "no", then I'd spot it in the diff. 😄 The answer is "probably yes".

greg-1-anderson:
Maybe `TerminalUtils::isTty` is no longer needed and can be removed?

danepowell:
I already removed the whole `TerminalUtils` class.

greg-1-anderson:
I think I was looking at just one commit instead of the whole patch.

LGTM

RoSk0:
Huge thanks to @danepowell and @greg-1-anderson for fixing this!

Maybe it's time for a new release? We have some good stuff committed since last one https://github.com/drush-ops/drush/compare/9.6.2...master

Ping @weitzman

PR #4035, opened 2019-04-05
In queue:run, set the lease time for a claimed item to the remaining execution time

lems3:
Set the lease time for a claimed item to the remaining execution time of the queue instead of using the default of 60 minutes, to have a similar behavior than a queue processed by cron.

### DIFF
--- a/src/Drupal/Commands/core/QueueCommands.php
+++ b/src/Drupal/Commands/core/QueueCommands.php
@@ -66,8 +66,9 @@ class QueueCommands extends DrushCommands
         $end = time() + $time_limit;
         $queue = $this->getQueue($name);
         $count = 0;
+        $remaining = $time_limit;
 
-        while ((!$time_limit || time() < $end) && ($item = $queue->claimItem())) {
+        while ((!$time_limit || $remaining > 0) && ($item = $queue->claimItem($remaining))) {
             try {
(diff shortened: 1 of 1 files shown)


weitzman:
Seems fine to me. Input welcome @juampynr @davereid @damiankloip

weitzman:
If you dont pass a time limit, then `remaining` check fails and nothing gets run?

lems3:
Unless mistaken (I could be), when you're not passing a time limit, the first while condition resolve to true, with the !$time_limit clause.
Otherwise, shouldn't the old condition, time() < $end, cause the same issue ?

damiankloip:
Hmm, Not sure. As `self::REQ` is the default value. Which is an alias of `\Symfony\Component\Console\Input\InputOption::VALUE_REQUIRED` which is `2`.

juampynr:
Looking at the diff I don't see major changes in the resulting logic. @lems3: can you add tests?

m4olivei:
Here is an alternative that builds on this: https://github.com/drush-ops/drush/pull/4083.

I've added the generic exception handling that core has and added a test.

weitzman:
@juampynr @damiankloip - see new PR.

greg-1-anderson:
Closing in favor of #4083.

PR #4037, opened 2019-04-08
Fixes #4034: Ensure that status code is nonzero on invalid shutdown

### LINKED ISSUE #4034: drush site-install returns improper exit code
If site installation fails, Drush exists with `0`.

```
$ ./vendor/bin/drush site-install -y minimal --sites-subdir default --config-dir ../config/sync

 // You are about to DROP all tables in your 'sites/default/files/.sqlite' database. Do you want to continue?: yes.

 [notice] Starting Drupal installation. This takes a while.
TypeError: Argument 2 passed to ...
 [warning] Drush command terminated abnormally. Check for an exit() in your Drupal site.

$ echo $?
0
```

| Q               | A
| --------------- | ---
| Drush version  | 9.6.1
| Drupal version | 8.6.13
| PHP version     | 7.3.3
| OS             | Mac

Probably the same issue as in #2223

greg-1-anderson:
If we get to the shutdown handler in an error state, but the status code is still zero, force it to 1.

### DIFF
--- a/src/Runtime/ShutdownHandler.php
+++ b/src/Runtime/ShutdownHandler.php
@@ -8,6 +8,7 @@ namespace Drush\Runtime;
 
 use Drush\Drush;
 use Drush\Log\LogLevel;
+use Drush\Commands\DrushCommands;
 use Webmozart\PathUtil\Path;
 
 use Psr\Log\LoggerAwareInterface;
@@ -45,7 +46,7 @@ class ShutdownHandler implements LoggerAwareInterface, HandlerInterface
             // Make sure that we will return an error code when we exit,
             // even if the code that got us here did not.
             if (!Runtime::exitCode()) {
-                Runtime::setExitCode(1);
+                Runtime::setExitCode(DrushCommands::EXIT_FAILURE);
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Runtime/ShutdownHandler.php
```diff
@@ -42,6 +42,11 @@ public function shutdownHandler()
 
         if (!Drush::config()->get(Runtime::DRUSH_RUNTIME_COMPLETED_NAMESPACE)) {
             Drush::logger()->warning('Drush command terminated abnormally. Check for an exit() in your Drupal site.');
+            // Make sure that we will return an error code when we exit,
+            // even if the code that got us here did not.
+            if (!Runtime::exitCode()) {
+                Runtime::setExitCode(1);
```
weitzman: Lets use \Drush\Commands\DrushCommands::EXIT_FAILURE

src/Runtime/ShutdownHandler.php
```diff
@@ -42,6 +42,11 @@ public function shutdownHandler()
 
         if (!Drush::config()->get(Runtime::DRUSH_RUNTIME_COMPLETED_NAMESPACE)) {
             Drush::logger()->warning('Drush command terminated abnormally. Check for an exit() in your Drupal site.');
+            // Make sure that we will return an error code when we exit,
+            // even if the code that got us here did not.
+            if (!Runtime::exitCode()) {
```
weitzman: Should we unmark this method as deprecated?

greg-1-anderson: I don't think so. I'd like to eventually rework exit code handling. Really, the `Runtime` class should be considered internal to Drush. We should put APIs that we want folks to use in the Drush class.

Issue #3914, opened 2019-02-07
Wrongly identifies Drupal 8.5.10 needing a security update


if-kenn:
**Describe the bug**
It looks like Drush misidentifies core security updates for Drupal 8.5.x. We have quite a few sites that have not been updated to Drupal 8.6.x and because of this bug it plays havoc with our automation systems, sending out false alerts.

**To Reproduce**
On a Drupal 8.5 site issue the following command:
```
drush ups --security-only --fields=name,existing_version,candidate_version"
```

**Expected behavior**
That core would not be identified as needing a security update

**Actual behavior**
This output is given:
```
Short Name  Installed Version  Proposed version
drupal      8.5.10             8.6.8
```

**Workaround**
None that I know.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.1.15
| Drupal version? | 8.5.10
| PHP version     | 7.2.14
| OS?             | Mac

**Additional information**
We have also seen this on previous core updates that are not the most recent Drupal 8 version, e.g. Drupal 8.4 exhibiting the same behavior with 8.5. We also are able to reproduce this on Linux.

greg-1-anderson:
The Drush pm-* commands are in maintenance only mode, and there is no active maintainer. A PR to fix this would be appreciated.

weitzman:
Your best bet is to upgrade to latest Drush 9.6 beta.

if-kenn:
@weitzman since we have to support Drupal 7, that is not an option.

if-kenn:
I was not aware that Drush 8 is no longer fully supported. Especially surprised since Drupal 7 will be supported until Nov 2021.

I guess we will have to alter our automation to use a specific version of Drush based on the version of Drupal. I was hoping to avoid that due to the complication it creates.

greg-1-anderson:
@if-kenn Maintaining the Drush 8 codebase is much more difficult than the Drush 9 codebase. We are still making a best-effort attempt to keep Drush 8 compatible with each new minor release of Drupal 8, but at some point that will not be possible. In the meantime, we will fix bugs and consider low-impact pull requests to enhance Drush 8. We will not be making any serious effort to add new features to this branch, though. This is really not very different than what is happening with Drupal 7.

[...]

PR #4038, opened 2019-04-08
Fix --security-only for make-update; improve non-bootstrapped version detection

bradjones1:
Note: I'm not sure if this is appropriate to target against `8.x` directly, but since `master` is basically 9.x, and this is an 8.x only change, I'm not sure what other option I have?

Return more useful version information when not bootstrapped; ensure security update is actually logical to suggest.

Refs/fixes (?) https://github.com/drush-ops/drush/issues/3914 - though it was closed for lack of active maintenance on 8.x.

### DIFF
--- a/commands/pm/pm.drush.inc
+++ b/commands/pm/pm.drush.inc
@@ -1642,12 +1642,14 @@ function pm_parse_version($version, $is_core = FALSE) {
       else {
         $version_parts = array(
           'version' => '',
+          // This could be incorrect (e.g., returns 2.1 and not a valid Drupal
+          // version), but keeping for BC. All other lines were historically ''.
           'drupal_version'  => $core_parts['major'] . '.x',
-          'project_version' => '',
-          'version_major'   => '',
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
I'd still take PRs such as this one for the 8.x branch with passing tests. The blocking issue is that tests are failing on the 8.x branch due to changes of behavior on Drupal.org due to the GitLab migration. I think these are false negatives; someone will need to fix these tests so that 8.x fixes can move forward.

bradjones1:
@greg-1-anderson Thanks. I'll see if I can get anyone on our team here to spend some time on it.

Note to self: failing test deeplink https://travis-ci.org/drush-ops/drush/jobs/517383694#L733

greg-1-anderson:
I've fixed the tests on the 8.x branch; if you rebase here, perhaps your tests will pass.

bradjones1:
Interesting - actually looks like the test may need to be updated to account for the fact we now return data, rather than empty strings when not bootstrapped: https://travis-ci.org/drush-ops/drush/jobs/551599919#L355

PR #4045, opened 2019-04-16
Add jsonapi command

weitzman:
This command takes a URL and returns the same JSON that Dru[al usually emits. Its built with JSONAPI in mind but any URL that emits JSON could be used with it.

Even if you have easy access to a browser, the command is handy because it bypasses authentication.

I was unable to find a PHP package that would syntax highlight the JSON for use in the shell. So the Examples recommend pairing this with `jq`

The PR also updates our testing system to use Drupal 8.7 (and updates codesniffer).


weitzman:
OK, this is green now. Travis refuses to test latest commits so its failure can be ignored.

I got this to pass by enhancing the PHP5 scenario to use Drupal 8.6. We can remove PHP5 support at our leisure (soon).

PR #4048, opened 2019-04-22
Use the version of the service dependencies patch for Drupal 8.7.x.

pfrenssen:
We are applying a patch to fix a core bug that is causing our tests to fail. Now that we are on Drupal 8.7.x we no longer need to support the 8.6.x version of the patch.

We are seeing test failures because the patch can no longer be applied.

### DIFF
--- a/.circleci/patch.sh
+++ b/.circleci/patch.sh
@@ -8,11 +8,6 @@
 
 cd $HOME/drush/sut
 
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks. We are keeping php5+Drupal8.6 test around for a little bit longer (see composer.json's extra section). I'll come back to this PR once we drop 8.6 on master branch.

PR #4061, opened 2019-05-05
Change import and export commands to use the new core API

bircher:
Once this lands in core we could use the new services in drush too.
https://www.drupal.org/project/drupal/issues/3047812

### DIFF
--- a/src/Drupal/Commands/config/ConfigExportCommands.php
+++ b/src/Drupal/Commands/config/ConfigExportCommands.php
@@ -64,6 +64,15 @@ class ConfigExportCommands extends DrushCommands
     /**
      * @return StorageInterface
      */
+    public function getConfigStorage()
+    {
+        // @todo: deprecate this method.
+        return $this->getConfigStorageExport();
+    }
+
+    /**
+     * @return StorageInterface
+     */
     public function getConfigStorageSync()
     {
         return $this->configStorageSync;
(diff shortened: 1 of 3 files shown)


bircher:
This can be tested with config_split config_ignore etc when using the patch from config_filter: https://www.drupal.org/project/drupal/issues/3047812#comment-13094680

bircher:
The core issues have been committed:
See https://www.drupal.org/node/3037022
and https://www.drupal.org/node/3066005

andypost:
Guess drush 10 will arrive before 8.8

weitzman:
Those test failures look unrelated (and odd). Ping @greg-1-anderson.

We need to stay compatible with Drupal 8.7 as well. Folks dont upgrade that quickly.

bircher:
The PR proposed here still works with Drupal 8.6 as the dependency injection is optional `@?config.storage.export` see also https://symfony.com/doc/current/service_container/optional_dependencies.html#ignoring-missing-dependencies

So the backwards compatibility concern should be addressed with this.

It is maybe worth mentioning that Drupal 8.8 will for sure have the export storage but may not have the import transformer because the new api is not used in core itself, in which case modules wanting to take advantage of it will rely on a workaround which adds the service again through contrib.
But if drush doesn't use the new api then the contrib solution would also have to add drush commands to address it.

greg-1-anderson:
Those are odd failures all right. The failure itself is a false failure. The test is checking to see that a php deprecation message is not emitted; however, it is failing because a system-level debug warning is emitted at the same time.

Ideally, this debug warning would not be emitted either; however, I think that the failure is more related to the environment of the docker container used by the Circle test than Drush. I don't know why this is happening here and not elsewhere, though. Maybe if I rerun the test, the failure will just go away.

greg-1-anderson:
As suspected, a re-run cleared up the problem.

PR #4068, opened 2019-05-10
Preserve environment variables in $_SERVER when creating request.

mfb:
Drush creates a request when bootstrapping, but does not preserve the `$_SERVER` environment variables.

With this PR, environment variables found in `$_SERVER` can also be found in `\Drupal::requestStack()->getCurrentRequest()->server` as would be expected.

(There might be some reason this isn't already the case, and a clean environment is created, but I'm not sure what that would be... :)

### DIFF
--- a/src/Boot/DrupalBoot8.php
+++ b/src/Boot/DrupalBoot8.php
@@ -145,10 +145,9 @@ class DrupalBoot8 extends DrupalBoot implements AutoloaderAwareInterface
             $parsed_url = parse_url('http://' . $uri);
         }
 
-        $server = [
-            'SCRIPT_FILENAME' => getcwd() . '/index.php',
-            'SCRIPT_NAME' => isset($parsed_url['path']) ? $parsed_url['path'] . 'index.php' : '/index.php',
-        ];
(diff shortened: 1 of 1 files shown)


weitzman:
Drush only prepares what is necessary, and nothing more.

mfb:
Is there a recommended way for scripts to put environment variables in their expected place? Or is it better to not expect environment variables to be in the request and use `$_SERVER`?

weitzman:
Not sure I follow. 

Use `getenv()` to read env variables. One of many ways to set an env variable is a .env file like https://github.com/drupal-composer/drupal-project/blob/8.x/.env.example

mfb:
Yes `getenv('VAR')` or `$_SERVER['VAR']` works fine, I was talking about code that accesses environment variables via something like `$this->requestStack->getCurrentRequest()->server->get('VAR')` - this latter syntax shouldn't be used w/ Drush code since it doesn't fill its fake request w/ environment variables, unlike a real request.

weitzman:
There are not Get variables in drush because there is not http. I agree that code like that is not cli friendly.

PR #4082, opened 2019-05-24
Convert Drush Make to Composer more accurately

bartlangelaan:
This PR introduces a few improvements for converting Drush Makefiles to Composer.

- Every library is added as a `package` type to the `repository` array of Composer. This makes them ready to be installed right after running the convert command.
- The caret (`^`) is removed in front of packages. The Drush Make command always installs the exact version that is defined in the makefile, so Composer should also install this exact version.
- Instead of using `Enter {$project['name']} patch #$key description here` as a patch name, it uses the original patch name. 
- The `drupal/core` package is marked as fully incompatible, instead of only marking `8.*` incompatible. This makes sure higher versions also won't be installed in the future.

### DIFF
--- a/commands/make/make.drush.inc
+++ b/commands/make/make.drush.inc
@@ -590,7 +590,7 @@ function drush_make_convert_make_to_composer($info) {
     );
 
     $conflict = array(
-      'drupal/core' => '8.*',
+      'drupal/core' => '*',
     );
 
     $extra = array(
@@ -649,7 +649,7 @@ function drush_make_convert_make_to_composer($info) {
     switch ($project['type']) {
       case 'core':
         $project['name'] = $core_project_name;
-        $projects[$project['name']] = '^' . str_replace('x', '*', $project['version']);
+        $projects[$project['name']] = str_replace('x', '*', $project['version']);
         break;
 
       default:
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
@bartlangelaan This PR introduced build failures, so I reverted it in #4110. If you would like to re-submit, I'd be happy to review and re-merge your improvements. At the time this build was submitted, the 8.x branch was already failing, which masked the new test failures. Drush 8 is green again, so it will be easier to determine the correctness of the re-roll.

commands/make/make.drush.inc
```diff
@@ -649,7 +649,7 @@ function drush_make_convert_make_to_composer($info) {
     switch ($project['type']) {
       case 'core':
         $project['name'] = $core_project_name;
-        $projects[$project['name']] = '^' . str_replace('x', '*', $project['version']);
+        $projects[$project['name']] = str_replace('x', '*', $project['version']);
```
greg-1-anderson: Versions will not always contain `.x`, so this change makes some versions that were previously converted to `^1.2.3` to be converted to `1.2.3` (pinned to a specific version). The old line wasn't quite right either, as it might produce `^1.*` instead of `1.*`. A correct transformation should handle both of these cases.

commands/make/make.drush.inc
```diff
@@ -760,8 +795,8 @@ function drush_make_convert_version_to_composer($version) {
       $cver = $version;
     }
     else {
-      // Replace '1.x' with '^1.*'.
-      $cver = '^' . str_replace('x', '*', $version);
+      // Replace '1.x' with '1.*'.
+      $cver = str_replace('x', '*', $version);
```
greg-1-anderson: I believe this is like https://github.com/drush-ops/drush/pull/4082/files#diff-cca65dcdb439384668dcd83746546302R652

PR #4083, opened 2019-05-24
Issue #4032 queue run lease time

### LINKED ISSUE #4032: Queue:run has a different behavior than the processQueues() section of Drupal cron.
When inspecting the code, I saw a couple of differences between the command `queue:run` and the function `processQueues()` from Drupal Cron.

Here are the sources for each functions : 
https://github.com/drush-ops/drush/blob/master/src/Drupal/Commands/core/QueueCommands.php#L61

https://git.drupalcode.org/project/drupal/blob/8.6.13/core/lib/Drupal/Core/Cron.php

In Queue:run :

- When an item is claimed, there is no time limit set, so the claim will use the default time, which sits at one hour now.
- The Queue:run command doesn't handle a generic \Exception.

The second point may not be relevant. The first one though allows a cron run to release items from the queue when it finishes. Implementing the same approach in the drush queue:run command would give it a similar behavior than a cron run.

An alternative to this would be to handle requeuing items in my QueueWorker::processItem() function, and return a SuspendQueueException when I hit an item I have just processed so the queue stop running. It seems more logic though to expect the queue:run command to processeach items once, and let the devs requeue items if they want it processed again in the same function call.

So default behavior would make queue:run process each item once and free them when it finishes, and dev can implement requeuing for instant reprocess if needed.

For the exact implementation, I would use the difference between time() and $end, so the items are freed as soon as the function calls end. We also can directly use the $time-limit parameter if we really want to stick to the current cron behavior.

m4olivei:
Resolves #4032. Builds on #4035.

Sorry if this it the wrong way about this. @lems3, I tried to make the base repository your repo, but Github wouldn't let me. I can adjust if necessary and shown the way.


lems3:
I wouldn't know how to handle the fork/PR honestly, I'm not used to that workflow on github.

m4olivei:
@lems3 Ok. Well I've included your commit here and added the generic exception handing and a test.

weitzman:
Fixed in #4322

Issue #3372, opened 2018-02-08
Very verbose output if a post-update has a batch process


pfrenssen:
When executing a batch process in a post-update the output included unnecessary lines about starting and completing each individual batch:

```
 [notice] Update started: joinup_subscription_post_update_subscribe_to_community_content
 [notice] 1000 memberships updated.
 [ok] Update completed: joinup_subscription_post_update_subscribe_to_community_content
 [notice] Update started: joinup_subscription_post_update_subscribe_to_community_content
 [notice] 2000 memberships updated.
 [ok] Update completed: joinup_subscription_post_update_subscribe_to_community_content
 [notice] Update started: joinup_subscription_post_update_subscribe_to_community_content
 [notice] 3000 memberships updated.
 [ok] Update completed: joinup_subscription_post_update_subscribe_to_community_content
```

It would be easier to read if the redundant lines are removed:

```
 [notice] Update started: joinup_subscription_post_update_subscribe_to_community_content
 [notice] 1000 memberships updated.
 [notice] 2000 memberships updated.
 [notice] 3000 memberships updated.
 [ok] Update completed: joinup_subscription_post_update_subscribe_to_community_content
```

claudiu-cristea:
PR #4089

pfrenssen:
Awesome, thanks for working on this :+1:

PR #4089, opened 2019-06-01
Reduce useless verbosity of batch update output 

claudiu-cristea:
Fixes https://github.com/drush-ops/drush/issues/3372


claudiu-cristea:
Running only the test (2cd1e3e) proves the bug.

PR #4090, opened 2019-06-01
Reduce the useless verbosity of batch update output (#3372)

### LINKED ISSUE #3372: Very verbose output if a post-update has a batch process

claudiu-cristea:
Fixes #3372

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -2,6 +2,7 @@
 namespace Drush\Commands\core;
 
 use Consolidation\Log\ConsoleLogLevel;
+use DrushBatchContext;
 use Consolidation\OutputFormatters\StructuredData\RowsOfFields;
 use Consolidation\OutputFormatters\StructuredData\UnstructuredListData;
 use Consolidation\SiteAlias\SiteAliasManagerAwareInterface;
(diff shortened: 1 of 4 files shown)


src/Commands/core/UpdateDBCommands.php
```diff
@@ -2,6 +2,7 @@
 namespace Drush\Commands\core;
 
 use Consolidation\Log\ConsoleLogLevel;
+use DrushBatchContext;
```
pfrenssen: I see that `DrushBatchContext` is a class in the global namespace, defined in `includes/batch.inc`. I'm not sure what the best practice is in this case, to define the namespace, or to directly refer to it in the code like `\DrushBatchContext`. There are no other references to this class in the codebase yet, except in `batch.inc` itself (which doesn't use the backslash since it is not needed there), and some code comments in `UpdateDBCommands` that refer to it (using the backslash prefix).

claudiu-cristea: Is this a Coding Standards issue? If, yes, we should check what PSR-2 ha sto say.

claudiu-cristea: @pfrenssen and probably we need a Drush issue/PR to move that class under a Drush namespace.

pfrenssen: OK I agree that it is in the end a coding standards issue. In PSR-2 this is not defined. I guess the real solution to this would be to move this class out of `batch.inc` and into its own namespace. That's out of scope for this PR though, so we're fine for now.

weitzman:
Thank you.

PR #4100, opened 2019-06-20
Fixes #4092: Adds test for db imports via sql:cli

### LINKED ISSUE #4092: New TTY detection in 9.7.0 breaks sql-cli input via STDIN (fixed, needs test)
**Describe the bug**
In versions of Drush prior to 9.7.0, you could provide arguments to sql-cli via STDIN. This is no longer possible in 9.7.0. Most notably, this breaks the ability to import db dumps via Drush (I'm not sure of any easier way than using sql-cli).

**To Reproduce**
Create a db dump, attempt to import it like so:
`drush sql-cli < mysqlDump.sql`

**Expected behavior**
SQL tables to be imported

**Actual behavior**
Silent failure, nothing imported.

**Workaround**
Retrieve the SQL connection information via Drush, then feed it to mysql directly in order to import data.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.7.0
| Drupal version? | 8.x
| PHP version     | 7.1

**Additional information**
Regression due to https://github.com/drush-ops/drush/pull/4033

danepowell:
Fixes #4092 

I'm going to add a few commits here. You'll see that it fails when I first add the test case because Drush's composer.lock is still using site-process 2.0.2, which exhibits the bug. The next commit (where I update site-process to 2.0.3) will pass tests.


danepowell:
Well now I'm a little stumped. The tests fail in AppVeyor, but pass locally: `docker-compose exec drupal composer functional -- --filter testSqlCli`

It's actually exhibiting the bug as if it's still using the old version of site-process, but it _is_ using 2.0.3 as expected. Not sure what's going on here.

danepowell:
Ah, didn't notice AppVeyor was windows, all sorted now :)

weitzman:
Thanks.

- We added Integration test type not that long ago. Thats preferable over functional. "Just" change your class to extend UnishIntegrationTestCase and relocate under tests/integration
- $this->drush() has a param for stdin at end. Thats better than passing it as part of the command. I would assume that redirection is the same?
- SHOW TABLES isnt valid for sqliteor postgres, I think. Drush's sqlBase has a method that does this for each DB platform. I think you can call that directly in an integration test.

danepowell:
Thanks for the feedback, I'll fix this up on Tuesday.

danepowell:
@weitzman I've been beating my head against the wall for a fair bit trying to make this work as an integration test. It seems inappropriate given that the test modifies the db, and the [testing docs](https://github.com/drush-ops/drush/blob/master/tests/README.md#about-the-test-suites) explicitly say that integration tests shouldn't do that.

Practically, the problem is that I can't seem to import the sql dump using the drush stdin parameter. I don't know whether this is a problem with the stdin param, or if Drush is actually resetting / preventing the db changes somehow. If you do somehow manage to modify the db (I did it outside of the test case), the changes persist between test runs which is obviously bad news.

danepowell:
I pushed the broken version of the integration test here if you want to see what I'm doing wrong.

tests/integration/SqlCliTest.php
```diff
@@ -0,0 +1,36 @@
...
+        if ($this->isWindows()) {
+            $this->markTestSkipped('On Windows, STDIN redirection is not supported.');
+        }
+        $this->setUpDrupal(1, true);
+        $path = Path::join(__DIR__, 'resources/sqlcli.sql');
+        // temporary: just prove that the file exists.
+        // why doesn't it get imported below?
+        $this->assertFileExists($path);
+        // Ensure SQL dumps can be imported via sql:cli.
+        $this->drush('sql:cli', [], [], null, $path);
```
weitzman: You want file_get_contents($path), instead of $path. Follow the drush() method's handling of stdin to see that a temp file will get created for you.

danepowell:
Thanks, I made those changes but it didn't affect the result. The file still just doesn't seem to be imported, and there's no errors indicating why. My guess is that MySQL is throwing an error that Drush is just swallowing. Even if I feed it gibberish for `stdin` I get no error output (using `$this->getErrorOutput()`).

weitzman:
After a significant struggle, I learned than in an integration contest, $this->drush() doesn't actually pass stdin to a subprocess because there is no subprocess. It sets up an environment such that any command thaqt implements StdinAwareInterface will see the desired stdin. sqlCommands did not implement that interface so no stdin was seen. I got the hint from https://github.com/consolidation/annotated-command/blob/master/src/Input/StdinHandler.php#L67-L69.

I updated the Doxygen for $this->drush() to make this clear and squash-merged+pushed. Thanks @danepowell for your work here.

danepowell:
Awesome, I'm glad you were able to figure that out. Sorry I wasn't able to get it across the finish line, but yeah, it was a struggle for me too :smile:  Thanks again.

PR #2620, opened 2017-02-16
Fixes #2671: Perform cache rebuild during updb in same process

tstoeckler:

PR #4103, opened 2019-06-24
Entity type provided by a module installed via update hook not available

claudiu-cristea:
If an entity type is provided by a module being installed via a (post)update hook. The next Drush command will fail to recognize the new installed definition, unless a cache rebuild is run just after `updatedb`. For example, with this hook:

```php
function woot_post_update_install_taxonomy()
{
    \Drupal::service('module_installer')->install(['taxonomy']);
    return \Drupal::entityTypeManager()->getDefinition('taxonomy_term')->id();
}
```
When running:
```bash
$ drush updb --yes
```
This will end successfully and will return `taxonomy_term`, but if I run immediately:
```bash
$ drush eval 'print \Drupal::entityTypeManager()->getDefinition("taxonomy_term")->id();'
```
...will output:
```
In EntityTypeManager.php line 150:

  The "taxonomy_term" entity type does not exist.
```

The bug is reproduced in the provided regression test. Rebuilding the cache in between is a workaround. This was not a bug before. We see this happening after moving to Drupal 8.7.x (is it probably related to https://www.drupal.org/node/3034742?)

It's interesting that the post-update recognizes the new installed entity type, as it returns its ID. This is not a Drupal issue. I tested the same scenario by running the UI update and works without error. I also run the next sequence with no errors:

```bash
$ drush en taxonomy
$ drush eval 'print \Drupal::entityTypeManager()->getDefinition("taxonomy_term")->id();'
```

The issue pops-up only when enabling the module via a (post)update.

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -23,12 +23,6 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
     protected $maintenanceModeOriginalState;
 
     /**
-     * The list of extensions that are installed before updates are applied.
-     * @var array
-     */
-    protected $originalExtensionList;
-
-    /**
      * Apply any database updates required (as with running update.php).
      *
      * @command updatedb
@@ -368,16 +362,13 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
      * @param array $results
      * @param array $operations
      */
-    public function updateFinished($success, $results, $operations)
+    public static function updateFinished($success, $results, $operations)
     {
-        // Check if updates or post-updates changed the installed extensions.
-        \Drupal::configFactory()->reset('core.extension');
-        $final_extension_list = \Drupal::config('core.extension')->getRawData();
-        if ($final_extension_list !== $this->originalExtensionList) {
-            // Invalidate the container if the installed extension list has been changed.
-            \Drupal::service('kernel')->invalidateContainer();
-        }
+        // Some (post)updates might install/uninstall extensions that require
+        // clearing the static and persistent caches and container invalidation.
+        drupal_flush_all_caches();
     }
+
 
     /**
      * Start the database update batch process.
@@ -386,10 +377,6 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
      */
     public function updateBatch($options)
     {
-        // Track the currently installed extensions so that we can detect if any
-        // new extensions are added in the update.
-        $this->originalExtensionList = \Drupal::config('core.extension')->getRawData();
-
         $start = $this->getUpdateList();
         // Resolve any update dependencies to determine the actual updates that will
         // be run and the order they will be run in.
@@ -456,7 +443,7 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
             'title' => 'Updating',
             'init_message' => 'Starting updates',
             'error_message' => 'An unrecoverable error has occurred. You can find the error message below. It is advised to copy it to the clipboard for reference.',
-            'finished' => [$this, 'updateFinished'],
+            'finished' => '\Drush\Commands\core\UpdateDBCommands::updateFinished',
             'file' => 'core/includes/update.inc',
         ];
         batch_set($batch);
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -23,6 +23,7 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
     protected $maintenanceModeOriginalState;
 
     /**
+     * The list of extensions that are installed before updates are applied.
      * @var array
      */
     protected $originalExtensionList;
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -386,6 +386,8 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
      */
     public function updateBatch($options)
     {
+        // Track the currently installed extensions so that we can detect if any
+        // new extensions are added in the update.
         $this->originalExtensionList = \Drupal::config('core.extension')->getRawData();
 
         $start = $this->getUpdateList();
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -373,7 +373,7 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
         \Drupal::configFactory()->reset('core.extension');
         $final_extension_list = \Drupal::config('core.extension')->getRawData();
         if ($final_extension_list !== $this->originalExtensionList) {
-            // If the installed extension list has been changed, clear caches and invalidate the container.
+            // Invalidate the container if the installed extension list has been changed.
             \Drupal::service('kernel')->invalidateContainer();
         }
     }
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -362,15 +362,16 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
      * @param array $results
      * @param array $operations
      */
-    public static function updateFinished($success, $results, $operations)
+    public function updateFinished($success, $results, $operations)
     {
-        // Flush all caches at the end of the batch operation. When Drupal core
-        // performs database updates it also clears the cache at the end. This
-        // ensures that we are compatible with updates that rely on this
-        // behavior.
-        drupal_flush_all_caches();
+        if ($this->cache_clear) {
+            // Flush all caches at the end of the batch operation. When Drupal
+            // core performs database updates it also clears the cache at the
+            // end. This ensures that we are compatible with updates that rely
+            // on this behavior.
+            drupal_flush_all_caches();
+        }
     }
-
 
     /**
      * Start the database update batch process.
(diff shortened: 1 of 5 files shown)


src/Commands/core/UpdateDBCommands.php
```diff
@@ -357,20 +362,26 @@ public static function updateDoOnePostUpdate($function, DrushBatchContext $conte
...
 
-
     /**
      * Start the database update batch process.
      * @param $options
      * @return bool
      */
     public function updateBatch($options)
     {
+        $this->originalExtensionList = \Drupal::config('core.extension')->getRawData();
```
pfrenssen:
```suggestion
        // Track the currently installed extensions so that we can detect if any
        // new extensions are added in the update.
        $this->originalExtensionList = \Drupal::config('core.extension')->getRawData();
```

src/Commands/core/UpdateDBCommands.php
```diff
@@ -22,6 +22,11 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
 
     protected $maintenanceModeOriginalState;
 
+    /**
+     * @var array
```
pfrenssen:
```suggestion
     * The list of extensions that are installed before updates are applied.
     * @var array
```

tests/functional/resources/modules/d8/woot/woot.module
```diff
@@ -31,3 +31,14 @@ function woot_theme() {
 
   return $theme;
 }
+
+/**
+ * Provides a helper for tests.
```
pfrenssen:
```suggestion
 * Prints the entity type ID of the taxonomy term entity type.
```

pfrenssen:
Looks good to me, thanks for the contribution!

weitzman:
Lets hold off on merging this. I want to get an old Pr linked here which touched on this issue, or a similar one.

claudiu-cristea:
> Lets hold off on merging this. I want to get an old Pr linked here which touched on this issue, or a similar one.

I can't find the source of this bug, it might not be a regression. It's very possible that this didn't surfaced. But the proof is there, in the test.

weitzman:
The issue I was thinking of is https://github.com/drush-ops/drush/issues/2617 and its PR is https://github.com/drush-ops/drush/pull/2620. I just mention them in case anyone finds them relevant.

> If an entity type is provided by a module being installed via a (post)update hook

Is it really acceptable to install modules in a post_update hook? I would think that you enable the modules you want to enable locally, and deploy a new core.extension.yml to the higher environments. Perhaps there are other use cases.

Also, the example module install does't run batches like pm:enable does. See https://github.com/drush-ops/drush/blob/40092d5dcf6f850fdf15d817c33602fe92c67c7c/src/Drupal/Commands/pm/PmCommands.php#L78-L83. Maybe that would fix the missing definition issue?

Finally, why does the cache-rebuild not fix this - https://github.com/drush-ops/drush/blob/d17e0670642524df67226585782a480289d386bd/src/Commands/core/UpdateDBCommands.php#L425-L426?

pfrenssen:
> Is it really acceptable to install modules in a post_update hook? I would think that you enable the modules you want to enable locally, and deploy a new core.extension.yml to the higher environments. Perhaps there are other use cases.

We are doing this when we are adding a new module dependency to an existing custom module that is optional in our distribution. It is possible users are running a different set of modules than we originally ship the distribution with we need to ensure somehow that they will have the module enabled after updating to our latest version. Maybe it is possible to do this in pure config, but I don't know how.

pfrenssen:
> Also, woot_post_update_install_taxonomy() does't run batches like pm:enable does

This is a very interesting suggestion. I investigated this. This section to run the batch processes was added in https://github.com/drush-ops/drush/issues/3444 to ensure that the translation imports are completed when the Locale module is enabled.

Indeed, the Locale module sets a batch in `locale_system_update()`, but nowhere in the ModuleInstaller or related code is a batch process launched. I tracked it down to `FormSubmitter::doSubmitForm()`. Basically any form is allowed to set a batch in the submit handler, and this will be executed on form submit.

So this means that somehow the Locale module is fine with translations only being imported when the module is enabled as part of a form submit. I could not find any other modules that attempt this, and it doesn't seem likely that the rebuilding of the container is handled in this way. Most probably in the "normal" install procedure (which is `ModulesListForm::submitForm()`) this just happens to work because the request ends after the modules are installed, and there is no more code running that needs the container to do entity operations.

I was also interested to understand how come this works when database updates are run in core but somehow not in Drush. It turns out that a container rebuild is triggered automatically as part of the update procedure. This is initiated through the call to `drupal_flush_all_caches()` in `DbUpdateController::batchFinished()`. As part of flushing the caches a call is done to `\Drupal::service('kernel')->invalidateContainer()`.

So I am thinking that it is probably a good idea to copy this behavior to `UpdateDbCommands::updateFinished()`. Let's just do an unconditional call to `drupal_flush_all_caches()`, without checking if any extensions have changed. This will most closely mimic the actual behavior from core, and will probably solve some other edge cases that people might be experiencing.

Still it could be a good idea to add th

src/Commands/core/UpdateDBCommands.php
```diff
@@ -22,6 +22,12 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
 
     protected $maintenanceModeOriginalState;
 
+    /**
+     * The list of extensions that are installed before updates are applied.
+     * @var array
+     */
+    protected $originalExtensionList;
```
pfrenssen: Let's remove this section, we no longer need to track if extensions are updated.

src/Commands/core/UpdateDBCommands.php
```diff
@@ -357,20 +363,28 @@ public static function updateDoOnePostUpdate($function, DrushBatchContext $conte
...
+    public function updateFinished($success, $results, $operations)
     {
-        // No code needed but the batch result bookkeeping fails without a finished callback.
-        $noop = 1;
+        // Check if updates or post-updates changed the installed extensions.
+        \Drupal::configFactory()->reset('core.extension');
+        $final_extension_list = \Drupal::config('core.extension')->getRawData();
+        if ($final_extension_list !== $this->originalExtensionList) {
+            // Invalidate the container if the installed extension list has been changed.
+            \Drupal::service('kernel')->invalidateContainer();
```
pfrenssen: Let's replace this with a simple cache flush:

```
public static function updateFinished($success, $results, $operations)
{
    // Flush all caches. Drupal core does the same after finishing the updates.
    // @see \Drupal\system\Controller\DbUpdateController::batchFinished()
    drupal_flush_all_caches();
}
```

src/Commands/core/UpdateDBCommands.php
```diff
@@ -357,20 +363,28 @@ public static function updateDoOnePostUpdate($function, DrushBatchContext $conte
...
     /**
      * Start the database update batch process.
      * @param $options
      * @return bool
      */
     public function updateBatch($options)
     {
+        // Track the currently installed extensions so that we can detect if any
+        // new extensions are added in the update.
+        $this->originalExtensionList = \Drupal::config('core.extension')->getRawData();
```
pfrenssen: We do not need this any more.

weitzman:
> So I am thinking that it is probably a good idea to copy this behavior to UpdateDbCommands::updateFinished(). Let's just do an unconditional call to drupal_flush_all_caches(), without checking if any extensions have changed.

That makes sense to me. We've had that clear and removed it a couple times over the long life of udpatedb. Please add code comments so we dont remove it again. Also, lets review the cache clears already in the command to see if any can now be removed.

claudiu-cristea:
@weitzman @pfrenssen I cannot explain the failure with Windows, it seems unrelated for me and I don't know how to trigger a rebuild.

src/Commands/core/UpdateDBCommands.php
```diff
@@ -364,8 +364,9 @@ public static function updateDoOnePostUpdate($function, DrushBatchContext $conte
      */
     public static function updateFinished($success, $results, $operations)
     {
-        // No code needed but the batch result bookkeeping fails without a finished callback.
-        $noop = 1;
+        // Some (post)updates might install/uninstall extensions that require
+        // clearing the static and persistent caches and container invalidation.
```
pfrenssen: This comment is now outdated, it still refers to the original approach.

I would make sure this mentions that this is done to mirror the approach taken in core, and link to the code responsible (`DbUpdateController::batchFinished()`).

Example:

```
// Flush all caches at the end of the batch operation. When Drupal core performs database
// updates it also clears the cache at the end. This ensures that we are compatible with updates
// that rely on this behavior.
// @see \Drupal\system\Controller\DbUpdateController::batchFinished()
```

pfrenssen:
The AppVeyor failure seems to be unrelated to this issue, I see that the master is affected by the same problem:https://github.com/drush-ops/drush/commits/master

pfrenssen:
Great, looking good to me, thanks!

pfrenssen:
> Also, lets review the cache clears already in the command to see if any can now be removed.

I looked into this. There is one other instance where the cache is cleared, in `::cacheRebuild()` but this is still needed. This performs a cache clear in between the regular updates and the post-updates, and is also matching the behavior of core. The relevant core code can be found in `\Drupal\system\Controller\DbUpdateController::triggerBatch()`:

```
    if ($post_updates) {
      // Now we rebuild all caches and after that execute the hook_post_update()
      // functions.
      $operations[] = ['drupal_flush_all_caches', []];
      foreach ($post_updates as $function) {
        $operations[] = ['update_invoke_post_update', [$function]];
      }
    }
```

src/Commands/core/UpdateDBCommands.php
```diff
@@ -364,8 +364,11 @@ public static function updateDoOnePostUpdate($function, DrushBatchContext $conte
      */
     public static function updateFinished($success, $results, $operations)
     {
-        // No code needed but the batch result bookkeeping fails without a finished callback.
-        $noop = 1;
+        // Flush all caches at the end of the batch operation. When Drupal core
+        // performs database updates it also clears the cache at the end. This
+        // ensures that we are compatible with updates that rely on this
+        // behavior.
+        drupal_flush_all_caches();
```
weitzman: Should only run if the option --cache-clear is true. PR can be merged after thats done.

claudiu-cristea:
@weitzman I've added the check. I had to make `updateFinished()` non-static otherwise it would make to complicated to pass the option.

Issue #4070, opened 2019-05-12
Drush should warn when running entity-updates on Drupal 8.7.x and higher that it's not going to work


joachim-n:
**Describe the bug**

Drupal 8.7.x removes the entity update functionality. On this version of Drupal, the 'drush entup' command appears to work, as it gives no error output, but has no effect. 

**To Reproduce**

Upgraded my site to Drupal core 8.7.2. Ran 'drush entup -y'. Ran it again. Ran it again ...

**Expected behavior**

Drush should complain or crash -- basically, tell the user that this command isn't going to work as expect and direct the user to https://www.drupal.org/project/devel_entity_updates

**Actual behavior**

Nothing.

**Workaround**

Download https://www.drupal.org/project/devel_entity_updates

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.6.2
| Drupal version? | 8.7.2
| PHP version     | 7.2
| OS?             | Mac/Linux/Windows

weitzman:
Makes sense.

Also would be nice if someone ported the Drush command in https://git.drupalcode.org/project/devel_entity_updates/blob/8.x-1.x/drush/devel_entity_updates.drush8.inc to Drush9.

joachim-n:
I think it has been. At least, once I installed that module, I was able to run entity updates. 

Looks like it's in https://git.drupalcode.org/project/devel_entity_updates/blob/8.x-1.x/src/Commands/DevelEntityUpdatesCommands.php

bradjones1:
I'm not sure the command from devel belongs "back" in Drush, but it should prompt a failure.

weitzman:
Implemented a check for entup command in #4101. updatedb is still outstanding.

mikemccaffrey:
This doesn't seem like a problem with drush, but rather with drupal core. When the code in the applyUpdates() function was removed, the error that was added should probably have been fatal, rather than E_USER_DEPRECATED, which implies that the function will be removed in the future and so does not trigger a fatal error when it is called.

Even if future versions of drush account for this feature being removed, that doesn't help anyone still using drush 8.x wondering why entity updates are saying that they succeeded, but not actually doing anything.

I created this issue in the core queue: https://www.drupal.org/project/drupal/issues/3078790

PR #4108, opened 2019-06-27
Fix updatedb --entity-updates flag in Drupal 8.7.0

alexpott:
Coming from https://github.com/drush-ops/drush/issues/4070 - it was noted that updatedb --entity-updates has the same problem.

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -27,7 +27,7 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
      *
      * @command updatedb
      * @option cache-clear Clear caches upon completion.
-     * @option entity-updates Run automatic entity schema updates at the end of any update hooks.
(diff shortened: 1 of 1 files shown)


alexpott:
Tested this locally on 8.7.0 and `drush updatedb --entity-updates` emits a warning now but soldiers on without trying to run entity updates. So I think we're in the best of all worlds until Drush no longer supports 8.6.0

src/Commands/core/UpdateDBCommands.php
```diff
@@ -40,6 +40,11 @@ public function updatedb($options = ['cache-clear' => true, 'entity-updates' =>
         require_once DRUPAL_ROOT . '/core/includes/update.inc';
         drupal_load_updates();
 
+        if ($options['entity-updates'] && version_compare(drush_drupal_version(), '8.7.0', '>=')) {
+            trigger_error(dt('Drupal removed its automatic entity-updates API in 8.7. See https://www.drupal.org/node/3034742.'), E_USER_WARNING);
```
weitzman: I think we should do `$this->logger()->warning()`

PR #4111, opened 2019-06-30
Remove backend API

weitzman:
This PR kicks off Drush 10 development. For now, Drush 10 is about removing legacy APIs (context, old site alias, context.inc, ...) and updating dependencies like Symfony, PHP, etc.

I left behind 2 functions in backend.inc that might be useful still. I'll probably get rid of them though,

See https://github.com/drush-ops/drush/milestone/5 for more about Drush 10.


weitzman:
I merged this into a new `ten` branch. Lets marinade this next version stuff there a short while before merging into master.

PR #4118, opened 2019-07-05
Fixes #3793 Can not create new config via cset

### LINKED ISSUE #3793: Drush 9 config:set : Can not create new config via cset, but it can in 8.x version
Hi Weitzman,

Good day!
I used to import a config yaml file into a remote site via cset command in drush 8.x version : 
`cat $config_file | drush @alias cset $config_name --format=yaml  --value=- -y`
It will update the config_name and create it if not exists.
It's really cool and useful.
But after I upgrade to Drush 9.5.2 version, I can't run this command that way any more, I found the config_name was been validated if exists first and also the config key was required.
I think if user uses yaml format option, he might want to update whole config object but not only one key.

So is it possible to make the config key optional when using yaml format?
And do not validate config_name just like Drush 8.x did?

Thanks in advance.
Ken

geek-merlin:


weitzman:
Sorry for 2 year delay!

This makes sense, but on the other hand its handy for most users to validate this. So, I'm thinking we keep validation on but prompt to ask user if we should proceed. That prompting should be done in cset itself, and not the validation (i.e. handle the validation exception).

geek-merlin:
Thanks for looking into this! Unfortunately i won't be able to work on this any soon.

weitzman:
> That prompting should be done in cset itself, and not the validation (i.e. handle the validation exception).

That wont work. Needs more thought.

weitzman:
Could you create a yaml file outside of drush and then do `config:import --partial`?

geek-merlin:
> (config:import --partial)
Thanks! But: That would be no win compared to "edit config-export and cim".

Your plan to use interaction sounded reasonable and i did not get why it's not feasible.

Otoh why not simply create? Typos? We may add a message "Config created - or was it a typo?" so users who created config due to a typo can `cdel` then. No kittens harmed.

alexharriesredactive:
May I add a respectful +1 for @geek-merlin's suggestion just to create the config object please @weitzman?

As far as I can tell, a mis-typed configuration name won't do any harm to the site.

Best wishes,

Alex

weitzman:
this got done in drush 12 (releasing this week)

PR #4122, opened 2019-07-11
Fixes #3845: Speculative Drush 10 compatibility

### LINKED ISSUE #3845: PHP version support for Drush 10
Placeholder issue for topics related to PHP support in Drush 10. It would be good to be able to drop 7.0, but Drush will probably follow Drupal's requirements.
 
Simplified supported for php / dependency versions in [Robo 2.x](https://github.com/consolidation/Robo/issues/548).

Drupal is planning on [dropping support for PHP 5.5 & 5.6 in March 2019](https://www.drupal.org/node/2938726).

Symfony 4 only supports PHP 7.1.3+.

Drupal might drop 7.0 on March 2019, or this might happen later. The decision is going to be tied to support for [PHP 7.1+ on various hosting providers](https://www.drupal.org/project/drupal/issues/2938725).

greg-1-anderson:
We still need to make some decisions about compatibility for Drush 10. Here is a visual representation of what that might look like. Subject to change.

For example, consolidation/site-process still needs some minor changes to work with symfony/process 4+. We will probably want to have different major versions of consolidation/site-process for symfony/process -3 vs 4+. It is still speculative whether this will pose any issues for Drush.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -28,37 +28,43 @@ Drupal Compatibility
     <th rowspan="2"> PHP </th>
     <th colspan="4"> Drupal versions </th>
   </tr>
-    <th>6</th> <th>7</th> <th>-8.3</th> <th>8.4+</th>
+    <th>6</th> <th>7</th> <th>-8.3</th> <th>8.4-8.7</th> <th>8.8+</th>
+  </tr>
+  <tr>
+    <td> Drush 10 </td>
+    <td> <a href="https://travis-ci.org/drush-ops/drush">master</a> </td>
+    <td> 7.1+ </td>
+    <td></td> <td></td> <td></td> <td></td> <td><b>✅</b></td>
   </tr>
   <tr>
     <td> Drush 9 </td>
-    <td> <a href="https://travis-ci.org/drush-ops/drush">master</a> </td>
+    <td> <a href="https://travis-ci.org/drush-ops/drush">9.x</a> </td>
     <td> 5.6+ </td>
-    <td></td> <td></td> <td></td> <td>✅</td>
+    <td></td> <td></td> <td></td> <td>✅</td> <td><b>⚠️</b></td>
   </tr>
   <tr>
     <td> Drush 8 </td>
     <td> <a href="https://travis-ci.org/drush-ops/drush">8.x</a> </td>
     <td> 5.4.5+ </td>
-    <td>✅</td> <td>✅</td> <td>✅</td> <td><b>⚠️</b></td>
+    <td>✅</td> <td>✅</td> <td>✅</td> <td><b>⚠️</b></td> <td><b>⚠️</b></td>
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
It is too early to commit here, just speculating right now. Updates near the end of the Drush 10 lifecycle , sans references to Drupal 9, no doubt.

weitzman:
In general LGTM. I personally don't see much point in updating Drush 8 so it lives on with Drupal 9.

greg-1-anderson:
Drush 8 continues to be necessary for non-Composer-managed sites. It will be good when the number of these goes to zero, but I don't think that we'll get there with Drupal 9.

weitzman:
The audience we are talking about are folks who take the significant effort to update to Drupal 9 and many of their contrib and custom modules. And then they refuse to update their Drush. I would rather spend limited resources on a different audience.

greg-1-anderson:
That's fine. Maybe by the time Drupal 9 comes out I won't have to support the non-Composer crowd either; if it came out today, though, then I would have to.

weitzman:
In addition, I'd like to remove the checkbox for the cell Drupal9+Drush9. Thoughts?

greg-1-anderson:
The impediment for upgrading from Drush 9 to Drush 10 will be contrib Drush commands for Drush 9+ that still use deprecated APIs removed in Drush 10. Other than that, it should be easy for folks to go from 9 to 10.

I don't really want to maintain Drush 9 myself, though, so I suppose the answer to that question depends on whether another maintainer steps up. If we don't have such a volunteer, then we could make the Drush 9 / Drupal 9 box empty, and, if we do that, then we should probably also make the Drush 9 / Drupal 8.4+ box a grey check ("works, but no longer supported").

greg-1-anderson:
If you wanted to support Drush 9 / Drupal 8.4+ and not Drush 9 / Drupal 9, though, that's fine with me too.

Chi-teck:
This conversation does not take into account the fact that Drupal 9 will update its dependencies (including Symfony). At this time there is no certainty  which Symfony version will be used in Drupal 9. Here is my speculation on this.
- Symfony 5 will be released In [November 2019](https://symfony.com/roadmap/5.0)
- Drupal 9 will be released in [June 2020](https://dri.es/plan-for-drupal-9#1)

That means that Drupal 9 will probably require Symfony 5. Which leads to the conclusion that to support
Drupal 8.4+/9, Drush should be compatible with Symfony `3 || 4 || 5`.

And as of Drush 8, it has to support 4 major Symfony versions (`2 || 3 || 4 || 5`).

Besides Symfony there might be some other shared dependencies that will be updated to new major releases in Drupal 9.

weitzman:
Its definately possible that Drush 11 will be needed for Drupal 9. I didn't really want to put that in the table yet.

PR #4123, opened 2019-07-11
Up version to 10 and improve unsatisfied constraint warning

weitzman:
Also warn when a commandfile doesn’t meet API constraint (was a debug log)

### DIFF
--- a/drush.info
+++ b/drush.info
@@ -1,1 +1,1 @@
-drush_version=9.8.0-dev
+drush_version=10.0.0-dev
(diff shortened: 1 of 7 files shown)


greg-1-anderson:
LGTM. Time to make the 9.x branch once this goes in.

weitzman:
When a package does does not meet its stated constraint in composer.json, we now emit the new message below Suggestions on better wording are welcome. 

> woot commands not loaded because its constraint (^9) is incompatible with Drush 10.0.0. Broaden the constraint in modules/unish/woot/composer.json (see 'extra\drush\services' section) to make the commands loadable.

greg-1-anderson:
Wording seems fine to me.

PR #4124, opened 2019-07-12
Fix #4052. Unnecessary prompt in config:export

### LINKED ISSUE #4052: drush cex --destination=/valid/destination will prompt for a destination if there is more than one item in the $config_directories array
**Describe the bug**

Running drush cex --destination=/some/valid/destination worked in Drush 8, but fails on Drush 9:

    root@a9ba710517f7:/var/www/html# ls -lah /tmp/localconfig/
    total 8.0K
    drwxr-xr-x 2 root root 4.0K Apr 24 16:13 .
    drwxrwxrwt 1 root root 4.0K Apr 24 16:13 ..
    root@a9ba710517f7:/var/www/html# drush cex --destination=/tmp/localconfig

     Choose a destination:
      [0] Cancel
      [1] deploy
      [2] sync
      [3] temp 

**To Reproduce**

drush help cex says: "--destination[=DESTINATION] An arbitrary directory that should receive the exported files. A backup directory is used when no value is provided."

This leads me to believe that, as with Drush 8, I can run:

    drush cex --destination=/tmp/localconfig
         
**Expected behavior**

I expect the configuration to be exported to /tmp/localconfig.

**Actual behavior**

I get a prompt:

    Choose a destination:
     [0] Cancel
     [1] deploy
     [2] sync
     [3] temp

**Workaround**

Use drush 8.x; I cannot find a workaround with drush 9; because this is done in a non-interactive CI environment so I cannot select a destination.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.6.2 
| Drupal version? | 8.6.15
| PHP version     | 7.2
| OS?             | Linux

**Additional information**

weitzman:


weitzman:
I merged and then immediately reverted this PR. It's wrong.

PR #4125, opened 2019-07-16
Updated subscription options in README.md

BWPanda:
Added info about GitHub's 'Releases only' watch option and removed old Version Eye link.


weitzman:
Thanks.

PR #4136, opened 2019-07-24
Issue #4135 Fix Request to have correct base URL

### LINKED ISSUE #4135: --uri still behaves unexpectedly under certain circumstances
## Background
We are using ACSF for our Drupal Hosting and using the `simple_sitemap` module to generate our sitemaps. ACSF invokes `drush cron -l https://example.com/path` when using path-based domains and leaves off the trailing slash. Site maps for the path-based domain sites are not using the correct base url and have urls that are just https://example.com.

There were a number of fixes that attempted to fix this, which I think they did in most part. More recently #3911 was closed saying that everything works as expected.

## The issue
At the end of the day simple_sitemap ends up interacting with Drupal\Core\Routing\UrlGenerator, which uses the Request to figure out the base path. Unfortunately, when Drush boots Drupal (DrupalBoot8.php) it does not use the URI with the trailing slash when generating the request. It looks like there may have just been a mistake to use $this->uri when creating the request instead of the local $url, which has had a trailing slash added.

This can be seen below snippet from 9.7.0.
```
        // Normalize URI.
        $uri = rtrim($this->uri, '/') . '/';
        $parsed_url = parse_url($uri);
        // Account for users who omit the http:// prefix.
        if (empty($parsed_url['scheme'])) {
            $this->uri = 'http://' . $this->uri;
            $parsed_url = parse_url('http://' . $uri);
        }
        $server = [
            'SCRIPT_FILENAME' => getcwd() . '/index.php',
            'SCRIPT_NAME' => isset($parsed_url['path']) ? $parsed_url['path'] . 'index.php' : '/index.php',
        ];
        $request = Request::create($this->uri, 'GET', [], [], [], $server);
        $this->setRequest($request);
```

## Example Code
I created an drush command to debug the URL generation for a node, mimicking what simple_sitemap was doing.
```
<?php

namespace Drush\Commands\test;

use Drupal\node\Entity\Node;

/**
 * A Drush commandfile for Debugging.
 */
class TestDebugCommands extends DrushCommands {

  /**
   * Test node url generation.
   *
   * @command test:debug:node-url
   * @usage drush test:debug:node-url 52
   *   Generates a node url as sitemap would.
   * @bootstrap full
   */
  public function testUrlGeneration($entity_id) {
    $entity = Node::load($entity_id);
    $url = $entity->toUrl();
    $path = $url->getInternalPath();
    $url->setOption('absolute', TRUE);
    $this->logger->warning('Generated URL: ' . $url->toString());
  }
}
```

## Actual behavior
Pretend for the following examples node 7 has the alias of /foo/bar.
* `drush -l https://example.com cgov:debug:node-url 7`
   * Produced `https://example.com/foo/bar`
* `drush -l https://example.com/test cgov:debug:node-url 7`
   * Produced `https://example.com/foo/bar`
* `drush -l https://example.com/test/ cgov:debug:node-url 7`
   * Produced `https://example.com/test/foo/bar`

## Expected Behavior
* `drush -l https://example.com cgov:debug:node-url 7`
   * Would produce `https://example.com/foo/bar`
* `drush -l https://example.com/test cgov:debug:node-url 7`
   * Would produce `https://example.com/test/foo/bar`
* `drush -l https://example.com/test/ cgov:debug:node-url 7`
   * Would produce `https://example.com/test/foo/bar`

**Workaround**
Not that I can see.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.7.0
| Drupal version? | 8.7.1
| PHP version     | 7.1 & 7.2
| OS?             | Mac & Linux

## Possible solution
I changed `$request = Request::create($this->uri, 'GET', [], [], [], $server);` to be `$request = Request::create($uri, 'GET', [], [], [], $server);` and the drush command worked. 

However that did not test all the other things that may break. Additionally, I am unsure if there are any odd request creations for the installers that would bypass this check.

bryanpizzillo:
- The Symfony request object had an incorrect base URL when using the --uri option
   with a subdirectory.
- Added tests for the url check.

Closes #4135

### DIFF
--- a/src/Boot/DrupalBoot8.php
+++ b/src/Boot/DrupalBoot8.php
@@ -137,19 +137,21 @@ class DrupalBoot8 extends DrupalBoot implements AutoloaderAwareInterface
 
         // Normalize URI.
         $uri = rtrim($this->uri, '/') . '/';
+
         $parsed_url = parse_url($uri);
 
         // Account for users who omit the http:// prefix.
         if (empty($parsed_url['scheme'])) {
             $this->uri = 'http://' . $this->uri;
-            $parsed_url = parse_url('http://' . $uri);
+            $uri = 'http://' . $uri;
+            $parsed_url = parse_url($uri);
         }
 
         $server = [
             'SCRIPT_FILENAME' => getcwd() . '/index.php',
             'SCRIPT_NAME' => isset($parsed_url['path']) ? $parsed_url['path'] . 'index.php' : '/index.php',
         ];
-        $request = Request::create($this->uri, 'GET', [], [], [], $server);
+        $request = Request::create($uri, 'GET', [], [], [], $server);
         $this->setRequest($request);
         return true;
     }
--- a/tests/functional/CoreTest.php
+++ b/tests/functional/CoreTest.php
@@ -76,6 +76,45 @@ class CoreCase extends CommandUnishTestCase
         }
     }
 
+    public function testOptionsUriRequestUrl()
+    {
+        // Test whether a URI in a config file resolves correctly, and test
+        // various URI values for their expected Site URI and path.
+        $drush_config_file = Path::join($this->webrootSlashDrush(), 'drush.yml');
+        $command_options = [
+        'uri' => 'OMIT', // A special value which causes --uri to not be specified.
+        ];
+        foreach ([
+                   'test.uri' => 'http://test.uri',
+                   'test.uri/' => 'http://test.uri',
+                   'test.uri/subpath' => 'http://test.uri/subpath',
+                   'test.uri/subpath/' => 'http://test.uri/subpath',
+                   'http://test.uri' => 'http://test.uri',
+                   'http://test.uri/' => 'http://test.uri',
+                   'http://test.uri/subpath' => 'http://test.uri/subpath',
+                   'http://test.uri/subpath/' => 'http://test.uri/subpath',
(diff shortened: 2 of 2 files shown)


bryanpizzillo:
@weitzman Any chance someone can take a look at this? Thanks!

greg-1-anderson:
Sorry for the delay. I was ambivalent about this because the improvement is awesome functionally, but I balked at a code style issue that I wasn't really sure was something to ask you to fix, since it was pre-existing.

I'd prefer it if the function just did the assignment to $this->uri once, and was consistent about whether it was using $uri or $this->uri. As it is right now, it's hard to tell be skimming the function if the code is correct for all code paths.

If you would like to make a follow-up PR to fix the code style issue, it would be appreciated. I'm just going to merge for now since this has been hanging around for a while, and it will be good to get the tests in & the functional improvement.

Thanks for the PR.

bryanpizzillo:
Thanks! @greg-1-anderson - I had the same challenge to be honest. At first I had set $this->url once, but a number of "unrelated" tests failed due to that change. (I say unrelated since they were not the URI tests specifically) I will create an issue and take a deeper look when I get a chance.

greg-1-anderson:
Yeah, the existing poor logic needs to be unravelled before that simplification can be made; if you only assigned to $this->url under the existing logic, you'd get the wrong result.

PR #4141, opened 2019-08-01
Test drupal/finder PR: https://github.com/webflo/drupal-finder/pull/44

greg-1-anderson:
DO NOT MERGE.


greg-1-anderson:
The Drupal 8.7.x tests failed because a patch would not apply, but the Drupal 8.8.x dev tests passed, which is what we're interested in here.

PR #4150, opened 2019-08-14
Fixes #4138. Escape rsync params.

### LINKED ISSUE #4138: Escape rsync params
If you place your local site in a local Google Drive-folder on Mac OSX the @self path isn't working. I guess it's due to the space in the folder name (Google Drive). I got these errors when doing an rsync. Notice how the path gets duplicated in the second error for some reason.

> rsync: link_stat "/Users/Username/Google" failed: No such file or directory (2)
> rsync: change_dir "/Users/Username/Google Drive/Path/Site//Drive/Path/Site" failed: No such file or directory (2)

I tried moving my site to my Documents-folder and rsync'ing from there works completely fine with the exact same settings.

weitzman:


weitzman:
Backported to 9.x as well.

PR #4153, opened 2019-08-15
fixes #4152

### LINKED ISSUE #4152: Fix version requirements for symfony/event-dispatcher
Drush causes the same issue as drupal core: https://www.drupal.org/project/drupal/issues/2876675

I repeat the description here:

With the release of Symfony 4.3 the event dispatching changed: https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching
This already causes issues for drupal regarding the upgrade to Symfony 4.x or 5, for example in
[#3055194] and [#3055180].

But Drupal isn't the only system that has to deal with that. Third party libraries drupal leverages directly or via contributed modules need to adapt to this change. Therefore the symfony project itself provides a migration path by decorating the event dispatcher with `LegacyEventDispatcherProxy` as described in blog post linked above.

This `LegacyEventDispatcherProxy` has been released as part of symfony/event-dispatcher 4.3.0.
The event-dispatcher has no dependency to any 4.3 versions of other symfony components but is backward compatible to symfony 3.4 which we use in Drupal 8:
```
"require-dev": {
        "symfony/dependency-injection": "~3.4|~4.0",
        "symfony/expression-language": "~3.4|~4.0",
        "symfony/config": "~3.4|~4.0",
        "symfony/http-foundation": "^3.4|^4.0",
        "symfony/service-contracts": "^1.1",
        "symfony/stopwatch": "~3.4|~4.0",
        "psr/log": "~1.0"
    },
"conflict": {
        "symfony/dependency-injection": "<3.4"
    },
```
See https://github.com/symfony/event-dispatcher/blob/4.3/composer.json

But no matter how we deal with the migration within the Drupal project, we already have a critical issue.
As an example, the upcoming 5.1.0 release of the solarium library implements the migration path as recommended by the symfony project:
https://github.com/solariumphp/solarium/pull/690

This library is required by the <a href="https://www.drupal.org/project/search_api_solr">search_api_solr</a> module that has more than 14,000 active 8.x installations already. search_api_solr injects the Drupal event dispatcher into the solarium Solr Client.

From my point that works perfectly fine. But Drupal abd Drush prohibit the update because of this version constraint:
`"symfony/event-dispatcher": "~3.4.0",`

This version constraint is simply wrong as Drupal doesn't use the event dispatcher provided by this module but only it's <code>EventDispatcherInterface</code> which has been kept compatible in symfony 4.x.

So the current situation is that Drupal and Drush block feature and security updates of third party libraries which use symfony event dispatching and adopting to symfony event dispatching changes in the proposed and compatible way. That has to be considered as critical.

Proposed resolution:
Adjust the version requirement like this:
`"symfony/event-dispatcher": "~3.4.0|~4.0"`
Afterwards composer will resolve the correct combination of module versions without upgrading the other symfony components at all.
Drupal will still use 3.4.

mkalkbrenner:
symfony/event_dispatcher version requirements

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
     "psr/log": "~1.0",
     "psy/psysh": "~0.6",
     "symfony/console": "^3.4",
-    "symfony/event-dispatcher": "^3.4",
+    "symfony/event-dispatcher": "^3.4 || ^4.0",
     "symfony/finder": "^3.4 || ^4.0",
     "symfony/process": "^3.4",
     "symfony/var-dumper": "^3.4 || ^4.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "c42d7848de97be00cf9dbbf937312242",
+    "content-hash": "3965a6d02c30d956c2718e8da8dcc53d",
     "packages": [
         {
             "name": "chi-teck/drupal-code-generator",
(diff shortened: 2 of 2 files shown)


greg-1-anderson:
This seems reasonable. Needs a `composer update --lock` to avoid the test failure.

greg-1-anderson:
I went ahead and merged this, and backported to the 9.x branch as well. I don't know that this will be of much use until https://github.com/consolidation/annotated-command/pull/190 is available, though.

PR #4155, opened 2019-08-16
Update testing for 9.x branch

greg-1-anderson:
Backporting a commit to the 9.x branch, I noticed a couple of minor things that needed changing here.

Do you think we need to split this up so that we have the same travis.yml commit on the master branch? Maybe it would be better to calculate `origin/master` / `origin/9.x` so that we can put the .circleci/config.yml change on master as well, and not have to worry about variable test configuration.


greg-1-anderson:
The changes here should be portable to the master branch.

PR #3601, opened 2018-07-02
Selective locale-check

basarkar:
Able to check and update selective projects. Will save lots of time and debugging.

PR #4156, opened 2019-08-16
Selective local check

basarkar:
Here is the drush 8.x PR https://github.com/drush-ops/drush/pull/3601

### DIFF
--- a/src/Drupal/Commands/core/LocaleCommands.php
+++ b/src/Drupal/Commands/core/LocaleCommands.php
@@ -9,6 +9,7 @@ use Drupal\Core\Language\LanguageManagerInterface;
 use Drupal\Core\State\StateInterface;
 use Drupal\language\Entity\ConfigurableLanguage;
 use Drush\Commands\DrushCommands;
+use Drush\Utils\StringUtils;
 
 class LocaleCommands extends DrushCommands
 {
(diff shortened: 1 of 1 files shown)


src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -66,17 +67,26 @@ public function __construct(LanguageManagerInterface $languageManager, ConfigFac
...
+    public function check($options = ['projects' => ''])
     {
         $this->getModuleHandler()->loadInclude('locale', 'inc', 'locale.compare');
 
+        $projects = [];
+        if ($passed_projects = $options['projects']) {
+            $projects = StringUtils::csvToArray($passed_projects);
+        }
+
+        // Remove duplicate projects.
```
Sutharsan, line 82: I'm not fond of documentation that only describes what the code does. If it does not add value, it is an additional maintenance burden.

Sutharsan: basarkar, thanks for contributing this code. I agree that it is an added value to the existing command.

It also needs a test.

Sutharsan: @basarkar, Please explain the purpose of an selective `locale:check` command if the `local:update` command is not selective. A todo in the code of the latter, even suggests it is broken.

PR #4162, opened 2019-08-23
[cron] Do not throw exception if cron is already running

driskell:
Currently, when you run cron and it is already running, it throws an exception saying Cron run failed.

This PR removes the exception throw when the cron call returns false. Checking Drupal - it will only ever return false if the cron is already running, and a warning will have already been logged. This ensures that cron installations of `drush cron --uri=<uri> --quiet` do not randomly fail

### DIFF
--- a/src/Drupal/Commands/core/DrupalCommands.php
+++ b/src/Drupal/Commands/core/DrupalCommands.php
@@ -56,10 +56,7 @@ class DrupalCommands extends DrushCommands
      */
     public function cron()
     {
-        $result = $this->getCron()->run();
-        if (!$result) {
-            throw new \Exception(dt('Cron run failed.'));
-        }
+        $this->getCron()->run();
     }
 
     /**


weitzman:
Are there other cases when this Exception would be thrown? I dont want to under-report valid failures.

driskell:
If a cron itself throws an exception, that will report as normal, as there are no `try{}catch{}` anywhere.

The cron run only ever returns false if the cron wasn't run at all, and only ever when the lock acquire failed, and only ever after reporting the message, "Another cron is already running". You can see it here: https://github.com/drupal/drupal/blob/8.8.x/core/lib/Drupal/Core/Cron.php#L129

I would imagine if there was any failure ever added inside Drupal - it would throw an Exception. Returning of errors is not something I have ever seen inside Drupal, so this conversion of return into Exception is pretty bizarre to me, as if it was an actual error case it wouldn't be returning, Drupal would've thrown.

weitzman:
Thanks for following up. 

This looks good to me. Can anyone think of a reason why its bad to report a success exit code when cron is already running?

weitzman:
Backported to 9.x as well.

PR #4168, opened 2019-09-03
Allow uri overrides on remote servers in 9.5.x.

Mramir-bounteous:
We are locked into 9.5.x with Acquia currently. We are also having an issue where any remote Drush commands we issue with the --uri flag do not pass the uri to the server. Since we are using ACSF with a large number of sites we need to be able to specify the uri in our commands.

### DIFF
--- a/src/Preflight/PreflightArgs.php
+++ b/src/Preflight/PreflightArgs.php
@@ -121,6 +121,7 @@ class PreflightArgs extends Config implements PreflightArgsInterface
             self::SIMULATE =>       \Robo\Config\Config::SIMULATE,
             self::BACKEND =>        self::BACKEND,
             self::LOCAL =>          self::DRUSH_RUNTIME_CONTEXT_NAMESPACE . '.' . self::LOCAL,
+            self::URI =>          self::DRUSH_RUNTIME_CONTEXT_NAMESPACE . '.' . self::URI,
         ];
     }
 


weitzman:
This is likely a problem with ACSF. Huge breakage would happen if Drush did not pass along URI on remote calls.

Mramir-bounteous:
@weitzman The issue appears to be client side though, 
Take for example:
```
drush @acsf01.01test -l http://sitename.test-acsf.acsitefactory.com/ cr -v
Calling proc_open(ssh -o PasswordAuthentication=no -t acsf01.01test@acsf01test.ssh.trex-prod.acquia-sites.com 'drush9  --verbose --root=/var/www/html/acsf01.01test/docroot --uri=acsf01test.trex-prod.acquia-sites.com  cr -v --ansi 2>&1' 2>&1);
```

What I am expecting is:
```
drush @acsf01.01test -l http://sitename.test-acsf.acsitefactory.com/ cr -v
Calling proc_open(ssh -o PasswordAuthentication=no -t acsf01.01test@acsf01test.ssh.trex-prod.acquia-sites.com 'drush9  --verbose --root=/var/www/html/acsf01.01test/docroot --uri=http://sitename.test-acsf.acsitefactory.com/  cr -v --ansi 2>&1' 2>&1);
```
Note the uri is behaving as expected in the second command. My problem is that the uri override is not working. Could this be the result of an improper alias file or something? I am currently investigating if there is a way I can do this with a hook.

Mramir-bounteous:
I was able to fix this locally without a drush change by using the following hook

```
  /**
   * @hook pre-init *
   */
  public function initUri(InputInterface $input, AnnotationData $annotationData){
    $preflight = $GLOBALS['preflight'];
    if($preflight && $uri = $preflight->preflightArgs($GLOBALS['argv'])->get('uri')){
      $input->setOption('uri', $uri);
    }
  }
```

PR #4175, opened 2019-09-12
Add option to exclude entities by ID to entity:delete command.

leymannx:
Following up on https://drupal.stackexchange.com/q/285037 and since I found myself already in a situation to need to delete every node *except* some very certain ones, I added an `--exclude` option to the `entity:delete` command.

Is it OK how it's written? Is it right to have the `$options` parameter array extended to also take `'exclude' => self::REQ` with `self::REQ`? I just followed the `'bundle' => self::REQ` pattern.

Finally everything else I added are the following few lines after the `$entities` are already loaded.

```php
// Don't delete excluded entities.
if ($exclude = StringUtils::csvToArray($options['exclude'])) {
    foreach ($exclude as $id) {
        unset($entities[$id]);
    }
}
```

Is there anything else I need to consider when adding this option? Do I need to extend existing tests for it? I wasn't able to find any specific tests for this command by searching the codebase.

### DIFF
--- a/src/Drupal/Commands/core/EntityCommands.php
+++ b/src/Drupal/Commands/core/EntityCommands.php
@@ -53,8 +53,8 @@ class EntityCommands extends DrushCommands
         } elseif ($options['bundle'] || $options['exclude']) {
             $query = $storage->getQuery();
             if ($exclude = StringUtils::csvToArray($options['exclude'])) {
-                $bundleId = $this->entityTypeManager->getDefinition($entity_type)->getKey('id');
-                $query = $query->condition($bundleId, $exclude, 'NOT IN');
+                $idKey = $this->entityTypeManager->getDefinition($entity_type)->getKey('id');
+                $query = $query->condition($idKey, $exclude, 'NOT IN');
             }
             if ($bundle = $options['bundle']) {
                 $bundleKey = $this->entityTypeManager->getDefinition($entity_type)->getKey('bundle');
--- a/src/Drupal/Commands/core/EntityCommands.php
+++ b/src/Drupal/Commands/core/EntityCommands.php
@@ -29,6 +29,7 @@ class EntityCommands extends DrushCommands
      * @param array $options
      *
      * @option bundle Restrict deletion to the specified bundle. Ignored when ids is specified.
+     * @option exclude Exclude certain entities from deletion. Ignored when ids is specified.
      * @usage drush entity:delete node --bundle=article
      *   Delete all article entities.
      * @usage drush entity:delete shortcut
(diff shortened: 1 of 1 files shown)


weitzman:
If we are going to do this, we might as well add a whitelist as well and switch to a regular entityQuery so we dont select more entities than necessary.

leymannx:
By switching to a regular entityQuery you mean not in general but only when the exclude option is passed, right?

leymannx:
I'm wondering how a regular entityQuery would be beneficial here. In the current draft the `--exclude` option has minimal impact on the current flow. It will even work as intended when calling `drush edel node 1,2,3,4,5 --exclude=1,2,3`.

leymannx:
Ah, maybe you mean something like the following with the `--exclude` option set.

    $exclude_ids = \Drupal::entityQuery('node')->condition('nid', $nids, 'NOT IN');

OK, I'll check if I get that to work in a way that the unrealistic but still possible `drush edel node 1,2,3,4,5 --exclude=1,2,3` still runs fine.

leymannx:
Am I right that for all other entities except for nodes I need to set a condition on the `id` and only for nodes on the `nid`? Let's try.

weitzman:
With Entity type manager you can lookup the name of the id field, I believe

leymannx:
Do you mean like that? Or could this be done more elegantly? Ready for feedback.

src/Drupal/Commands/core/EntityCommands.php
```diff
@@ -29,27 +29,39 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager)
...
         $storage = $this->entityTypeManager->getStorage($entity_type);
         if ($ids = StringUtils::csvToArray($ids)) {
             $entities = $storage->loadMultiple($ids);
-        } elseif ($bundle = $options['bundle']) {
-            $bundleKey = $this->entityTypeManager->getDefinition($entity_type)->getKey('bundle');
-            $entities = $storage->loadByProperties([$bundleKey => $bundle]);
+        } elseif ($options['bundle'] || $options['exclude']) {
+            $query = $storage->getQuery();
+            if ($exclude = StringUtils::csvToArray($options['exclude'])) {
+                $bundleId = $this->entityTypeManager->getDefinition($entity_type)->getKey('id');
```
weitzman: I think this is $idKey, but $bundleId.

leymannx:
Ah yes, that makes sense.

Tests are failing because of something different I think.

weitzman:
Merged ... If you have time to test coverage here, that would be appreciated.

leymannx:
I'd absolutely love to. You mean like in adding tests for the `edel` command, right? But can you maybe point me to some file in the code base I can have a look at for reference? When searching the codebase for `entity:delete` or `entity-delete` or `edel` or `EntityCommands` all I find is this very file. Where would I add the test or extend existing tests?

weitzman:
See the classes in the [tests/integration folder](https://github.com/drush-ops/drush/tree/master/tests/integration). Those are preferred over `tests/functional` tests.

PR #4178, opened 2019-09-16
Add maintenance-mode flag support to `drush updatedb` command

jrockowitz:
### Problem

When running certain database updates, Drupal does not have to be switch into maintenance mode.

### Solution

Provide an optional flag to switch Drupal into maintenance mode during database updates.  The --maintenance-mode should default to TRUE, since it is recommended to place a site in maintenance mode... unless you know what you are doing.

Here is an example of the new flag.

`drush --maintenance-mode=0 updatedb`

#### Tasks

- Decide if this feature should be added to Drush.
- Determine if this feature can be added to Drush 9 or 10.
- Create PR implementing this feature
- Write test coverage this feature

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -442,11 +442,11 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
         }
 
         if ($options['maintenance-mode']) {
-          $original_maint_mode = \Drupal::service('state')->get('system.maintenance_mode');
-          if (!$original_maint_mode) {
-              \Drupal::service('state')->set('system.maintenance_mode', true);
-              $operations[] = ['\Drush\Commands\core\UpdateDBCommands::restoreMaintMode', [false]];
-          }
+            $original_maint_mode = \Drupal::service('state')->get('system.maintenance_mode');
+            if (!$original_maint_mode) {
+                \Drupal::service('state')->set('system.maintenance_mode', true);
+                $operations[] = ['\Drush\Commands\core\UpdateDBCommands::restoreMaintMode', [false]];
+            }
         }
 
         $batch['operations'] = $operations;
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -27,13 +27,14 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
      *
      * @command updatedb
      * @option cache-clear Clear caches upon completion.
+     * @option maintenance-mode Switch site into maintenance mode during updates.
      * @option entity-updates Run automatic entity schema updates at the end of any update hooks. Not supported in Drupal >= 8.7.0.
      * @option post-updates Run post updates after hook_update_n and entity updates.
      * @bootstrap full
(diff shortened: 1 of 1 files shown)


src/Commands/core/UpdateDBCommands.php
```diff
@@ -187,15 +188,17 @@ public function process($batch_id, $options = ['format' => 'json'])
...
      */
-    public static function updateDoOne($module, $number, array $dependency_map, DrushBatchContext $context)
+    public static function updateDoOne($module, $number, array $dependency_map, array $options, DrushBatchContext $context)
     {
         $function = $module . '_update_' . $number;
 
         // Disable config entity overrides.
-        if (!defined('MAINTENANCE_MODE')) {
+        if ($options['maintenance-mode'] && !defined('MAINTENANCE_MODE')) {
             define('MAINTENANCE_MODE', 'update');
```
greg-1-anderson: Defining maintenance mode here does not have any effect on the Drupal site, as experienced through the web server. This definition only serves to make Drupal respond as if it were in maintenance mode during the update operation. By appearances, it looks like this code is designed specifically to operate when the site is NOT in maintenance model.

Perhaps this block, and the one like it below, should remain unmodified?  I am unsure about this. Do you have a specific reason or need to remove this?

jrockowitz: My concern is that once `define('MAINTENANCE_MODE', 'update');` is executed, all modules can assume that the site is in maintenance mode even in their *.install files which are loaded immediately after this code.

greg-1-anderson: Yes, all of the modules will think that the site is in maintenance mode, including when their update hooks run. This was added in #3606, and seems to still be necessary. I can't think of a bad effect of defining MAINTENANCE_MODE like this, but #3604 identifies situations where not defining it is problematic.

Chi-teck:
Maybe we can simply take the maintenance-mode out of `drush updb` command to make it follow single responsibility principle.
Users can still set the mode explicitly with Drush like follows.
`drush sset system.maintenance_mode 1 && drush updb && drush sset system.maintenance_mode 0` 

Worth noting that maintenance mode requires clearing cache for anonymous users.
https://www.drupal.org/project/drupal/issues/2804387

greg-1-anderson:
We could stop setting maintenance mode in a new major version of Drush, but should not do it in Drush 9, since that would be a backwards compatibility break. It might be late to make this sort of change in Drush 10, since there is already an rc1.

Having the flag as a service seems reasonable.

jrockowitz:
So far my research is showing to be safe, we should automatically switch a site to maintenance mode during database updates but give developers, who know that they are doing, the option to not switch to maintenance mode.  This approach would be backwards compatible.

Ironically, I just realized the `drush config-import` does not switch a site to  maintenance mode when some fairly major databases changes could be happening.

weitzman:
Does not seem like we can omit this as data integrity is important. Greg's link is instructive - https://github.com/drush-ops/drush/issues/3604 (and especially its drupal.org issue).

updatedb has always been a "full service" command that does a bunch of things (similar to sql:sync). i dont think we can change that at this point in drush's lifetime.

weitzman:
this got done in #3606

Issue #4158, opened 2019-08-22
Drush SQL commands not working with non standard MySQL socket


r7l:
**Describe the bug**

I have a Docker setup with a MySQL container only connected to my PHP container over a socket. So far so good, everything works and was working with older Drush versions (It works with Drupal 7 and Drush 8).

The socket is defined to be at `/var/run/mysql/mysql.sock` in Drupal settings.php and MySQL configuration. I can run any MySQL command from cli as well as any other Drush command (like `drush updb`). Drupal is running fine as well.

I can't run any Drush SQL commands. It will always end up with this error:

```Got error: 2002: "Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)" when trying to connect```

**To Reproduce**
Run a command like `drush sqlc`

**Expected behavior**
Connect to MySQL

**Actual behavior**
Error Message: 

```ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)```

**Workaround**
Use any MySQL command directly.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | Drush Commandline Tool 9.7.1
| Drupal version? | 8.7.6
| PHP version     | 7.3.6
| OS?             | Linux / Docker

**Additional information**
When debugging this issue i've found that Drush seems to add some default configuration. I am not entirely sure what is in that configuration since it will delete the files after using it.

This is what the debug gives me:

 [info] Executing: mysqldump **--defaults-file=/tmp/drush_Vf8J0Z** drupal --host=localhost --no-autocommit --single-transaction --opt -Q  [0.29 sec, 3.81 MB]

What exactly is in **/tmp/drush_Vf8J0Z** and how to prevent Drush from using it? How  to set it to the correct socket?

Suggested fix would be to not introduce a default when there is one set already in MySQL config or use the one from Drupal in settings.php when it's set like this: `'unix_socket' => '/var/run/mysql/mysql.sock',`

weitzman:
IIRC, there was some uproar about mysql client config overriding whats setup in Drupal. So now the Drupal config is used to populate tmp files like ` /tmp/drush_Vf8J0Z` and thats used unconditionally. 

[...]

PR #4180, opened 2019-09-17
add unix socket configuration to mysql credentials

r7l:
Hello, i've asked for this feature some time ago in #4158 and finally had the time to add it. 

It would leave an empty line if no socket is set but that shouldn't be a big issue as the file is temporary anyways. Would be nice to have this feature in Drush in order to be able to run MySQL commands on setups with socket connections.

If you merge it, thanks in advance.


weitzman:
Fixed up slightly and merged. Thanks.

PR #4183, opened 2019-09-18
Make check for test modules stricter in PM:LIST

### LINKED ISSUE #4182: Modules with "tests" in name are filtered from list of modules
**Describe the bug**
I have module with the name `fuzz_tests_on_prod` and I have noticed that it's not listed when I execute `drush pml`.

**To Reproduce**
Create an empty module with name `fuzz_tests_on_prod`, make `info.yml` file.

**Expected behavior**
When I run `drush pml | grep tests`, to have that module in the list.

**Actual behavior**
The module is not listed.

**Workaround**
The module is listed on extensions page - URL: `/admin/modules`

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.7.1
| Drupal version? | 8.7.7
| PHP version     | 7.3.9
| OS?             | Mac/Linux

mtodor:
Fixes #4182 

My proposal that explained problem would be to make check stricter.

Instead of checking for `tests`, we should check for `/tests/` in module/theme path.

### DIFF
--- a/src/Drupal/Commands/pm/PmCommands.php
+++ b/src/Drupal/Commands/pm/PmCommands.php
@@ -170,7 +170,7 @@ class PmCommands extends DrushCommands
             $extension->info += ['package' => ''];
 
             // Filter out test modules/themes.
-            if (strpos($extension->getPath(), 'tests')) {
+            if (strpos($extension->getPath(), DIRECTORY_SEPARATOR . 'tests' . DIRECTORY_SEPARATOR)) {
                 continue;
             }
 


weitzman:
Kinda worried that getPath() wont match DIRECTORY_SEPARATOR always, and we will have a worse problem than what this fixes.

mtodor:
@weitzman Would it be better to use regex? Something like `[/\\]tests[/\\]`?

greg-1-anderson:
I missed the review request on this PR last year. I am not sure, but I also suspect that DIRECTORY_SEPARATOR will never match getPath() on Windows. The regex suggestion seems reasonable.

src/Drupal/Commands/pm/PmCommands.php
```diff
@@ -176,7 +170,7 @@ public function pmList($options = ['format' => 'table', 'type' => 'module,theme'
             $extension->info += ['package' => ''];
 
             // Filter out test modules/themes.
-            if (strpos($extension->getPath(), 'tests')) {
+            if (strpos($extension->getPath(), DIRECTORY_SEPARATOR . 'tests' . DIRECTORY_SEPARATOR)) {
```
donquixote, line 173: This seems to be equivalent to the condition in Drupal core, which simply adds `'tests'` to `RecursiveExtensionFilterCallback::$skippedFolders`.

Although Drupal core only does this when `$settings['extension_discovery_scan_tests']` is not `TRUE`.

donquixote, line 173: Regarding `DIRECTORY_SEPARATOR`:
Will this always do what we expect it to do?
On a system where `DIRECTORY_SEPARATOR !== '/'`, would `$extension->getPath()` have the system directory separators, or would it just be `'/'` because it is already converted somewhere or the path is glued together with code, with hard-coded `'/'`?

donquixote, line 173: Also now we can do `str_contains()` instead of `strpos()`.

PR #4185, opened 2019-09-19
Better message when sql program doesnt exist and Drupal fails to install

weitzman:


src/Commands/core/SiteInstallCommands.php
```diff
@@ -142,7 +143,15 @@ public function install(array $profile, $options = ['db-url' => self::REQ, 'db-p
         require_once DRUSH_DRUPAL_CORE . '/includes/install.core.inc';
         // This can lead to an exit() in Drupal. See install_display_output() (e.g. config validation failure).
         // @todo Get Drupal to not call that function when on the CLI.
-        drush_op('install_drupal', $class_loader, $settings);
+        try {
+            drush_op('install_drupal', $class_loader, $settings);
+        } catch (AlreadyInstalledException $e) {
+            if (!$this->programExists($sql->command())) {
+                throw new \Exception(dt('Drush was unable to drop all tables because `@program` was not found. This may the reason that Drupal halted installation with a AlreadyInstalledException.', ['@program' => $sql->command()]));
```
danepowell: The user never sees the AlreadyInstalledException because it's caught, so it's probably better to explain that it was thrown, or just leave out any mention of it.

```suggestion
                throw new \Exception(dt('Drush was unable to drop all tables because `@program` was not found, and therefore Drupal threw an AlreadyInstalledException. Ensure `@program` is available in your PATH.', ['@program' => $sql->command()]));
```

PR #4187, opened 2019-09-19
Calculate memory_limit correctly when there is no suffix

kdebisschop:
Fixes drush_memory_limit to give correct value if memory limit is strictly numeric


weitzman:
When is there no suffix?

kdebisschop:
The current code handles things like 100K and 100M, but fails for 100000000.

You can add K, M, or G for kilobytes, etc., to PHPs memory_limit. But but is not required that you do.

weitzman:
Backported to 9.x as well in cad7fcf1]

Issue #3574, opened 2018-06-04
Drush becomes unusable if a command requires a new module


deviantintegral:
If a drush command depends on a service in a module that is not yet enabled, drush can become unusable.

For example, let's say `Custom Module A` adds a Drush command and in it's `drush.services.yml` file uses a service exposed by `Contrib Module B`. In a minor release, `Contrib Module B` alters it's service definition to use a class from a newly added dependency. Note it's not necessarily breaking API contracts; perhaps a new module has some nice implementation of a core-provided interface.

The problem arises when trying to use Drush to run update hooks. As a developer or site builder, I expect I can run `composer update && drush -y updatedb`. However, while discovering all implemented Drush commands, building the container fails because the newly-required class exists in a module that is not yet enabled. As an end user, the only workaround is to use the user interface to run update hooks, or to do two deployments.

There is a way around this [by using a service provider instead of the `drush.services.yml` file](https://github.com/Lullabot/media_mpx/pull/52/files). That lets code do basic checks like `class_exists(...)` before defining a service. However, the services are not as discoverable as a YAML file so the DX for new developers isn't great. As is, I think this is required if you depend on any services or classes outside of the command's module.

There is a related core issue at [Allow updating modules with new service dependencies](https://www.drupal.org/node/2863986).

weitzman:
My first suggestion for fixing this is to change `* @bootstrap full` to `* @bootstrap database`. [See updatedb annotation](https://github.com/drush-ops/drush/blob/44daf5422d3f1cba04bdc36df22a16890c7ba8bd/src/Commands/core/UpdateDBCommands.php#L26). This will prevent Drush from trying to load commandfiles from modules. The rest of the job is just dealing with fallout from that change (if any).

weitzman:
possibly related  https://www.drupal.org/project/drupal/issues/2863986

greg-1-anderson:
It seems to me that `Custom Module A` should only use classes from Drupal Core or from its own code base. I would expect that it would be possible (likely?) that loading classes from a non-enabled `Custom Module B` could cause a crash. I don't know that Drush can validate that the drush.services.yml file is conforming in this way. Would you ever put a dependency between two modules in your `services.yml` file without making the other module a dependency? I wouldn't think so -- and if the other module was a dependency, then it should always be enabled when your module is enabled.

I'd still be in favor of strengthening `drush updatedb` against edge cases like this, even if we need to invent a new bootstrap level (e.g. `full but without Drush extensions`).

[...]

PR #4196, opened 2019-09-26
Allow updating modules with new service dependencies

jcisio:
Add feature from Drupal 8.7.5.
@see https://www.drupal.org/node/2863986

This (partially?) fixes https://github.com/drush-ops/drush/issues/3574

### DIFF
--- a/lib/Drush/Drupal/DrushServiceModifier.php
+++ b/lib/Drush/Drupal/DrushServiceModifier.php
@@ -5,6 +5,8 @@ namespace Drush\Drupal;
 use Drush\Log\LogLevel;
 use Drupal\Core\DependencyInjection\ServiceModifierInterface;
 use Drupal\Core\DependencyInjection\ContainerBuilder;
+use Drupal\Core\Update\UpdateCompilerPass;
+use Symfony\Component\DependencyInjection\Compiler\PassConfig;
 
 class DrushServiceModifier implements ServiceModifierInterface
 {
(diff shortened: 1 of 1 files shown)


weitzman:
Perhaps @deviantintegral is up for a review here?

pfrenssen:
I think this approach is mostly harmless but not the "right way". This change will always invoke the `UpdateCompilerPass`, while the idea behind this is that it should only happen when using the `UpdateKernel`. So only when updates are being performed. When not doing updates core uses `DrupalKernel` which does not invoke `UpdateCompilerPass`.

With this change we will _always_ silently drop services from the container if they have missing dependencies, but this is not the desired behavior if we're not updating. It might even cause critical bugs if a typo is being made in a `*.services.yml` file and this is not noticed before pushing to production. Normally when a required dependency is missing (due to a module not being enabled, a file not being committed, or a typo in the yml file) an exception will be thrown when the container is compiled, and this alerts the developer that something is seriously wrong with the dependency chain.

The correct approach would be to use the `UpdateKernel` for all commands that are executed during a site update. This is already the case for all commands in `UpdateDBCommands` - they are tagged with the annotation `@kernel update`. They should already work fine on Drupal 8.7.x.

But I have discovered today that there is (at least) one more command that is used in the site update procedure which is not using the right kernel: `config:import`. It still crashes with this exception. I was not aware of this PR yet, I have proposed an approach in #4217.

deviantintegral:
TIL'ed about `@kernel update` - that's very nice.

weitzman:
#4217 fixed this AFAICT.

Issue #3636, opened 2018-07-31
Drush Sanitize fails on [error]  No entity bundle was specified


brooke-heaton:
**Describe the bug**
When running `drush sanitize`, it works through a series of fields but always fails at `[error]  No entity bundle was specified` - it is unclear on why this happens. 

**To Reproduce**
`drush sanitize` with no whitelist

**Expected behavior**
I expect a clean sanitization without error.

**Actual behavior**
It fails and triggers an error that seems entirely unrelated to users or text fields.

**Workaround**
No.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.4-dev-dev
| Drupal version? | 8.5
| PHP version     | 7.1
| OS?             | Mac

**Additional information**
I am using Acquia BLT so the sanitization is part of BLT.

brooke-heaton:
Fair enough. I probably won't have the time. I stepped through it a few weeks ago and did find that it was likely related to a commerce product variation type - that was the entity without a bundle. But how that would ever appear due to a user text file sanitization is beyond me.

ey-:
https://github.com/drush-ops/drush/blob/master/src/Drupal/Commands/sql/SanitizeUserFieldsCommands.php#L69

It looks like this is too early, as this generates first sample values for _all_ user fields, also entity references for example. And later on the generated values are only used by the listed field types in the switch.

PR #4198, opened 2019-10-02
Issue-3636: Generate sample values only for supported field types.

ey-:
As described in https://github.com/drush-ops/drush/issues/3636 drush sql-sanitize fails sometimes, because it generates sample values for all fields.

This PR adds a check, to generate the sample value only, when the field type is supported.

### DIFF
--- a/src/Drupal/Commands/sql/SanitizeUserFieldsCommands.php
+++ b/src/Drupal/Commands/sql/SanitizeUserFieldsCommands.php
@@ -66,8 +66,11 @@ class SanitizeUserFieldsCommands extends DrushCommands implements SanitizePlugin
             $query = $conn->update($table);
             $name = $def->getName();
             $field_type_class = \Drupal::service('plugin.manager.field.field_type')->getPluginClass($def->getType());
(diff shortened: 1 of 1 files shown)


weitzman:
Why is this needed? Its part of the field type interface that they must implement generateSampleValue() or whatever its called.

ey-:
Thats true, normally the sample value is also generated without issues for simple field types.

But if you have field types which are more complex, the sample value generation may fail.

This is the case for example if you have a field of type entity reference, which is referencing an entity which needs an API connection.

The EntityReferenceItem::generateSampleValue checks if there is an entity, which can be referenced and if not, tries to generate one.

In both situations, it potentially can try to connect to the API, which can fail, as that environment has no connection to the needed API. 
OR even this may be the wrong API which is called, as we are currently importing a database dump from another (i.e. production) environment and probably need/have other API settings for the current environment. As the config-import if potentially happens after the DB-dump is imported we are in a wrong state. 

Also, it feels wrong to generate a sample value (which can be a heavy operation), which is not used anyway.

PR #4202, opened 2019-10-03
Replace classy with stark in EnDisUnListInfoCase::testEnDisUnList

JTubex:
Since classy is no longer the default testing theme, the Unish\EnDisUnListInfoCase::testEnDisUnList is currently failing.

This test is checking if classy is installed. Since classy is no longer the default testing theme, the test will nog longer find classy in the installed modules/themes list.

Issue: https://www.drupal.org/project/drupal/issues/2352949#comment-13252667
CR: https://www.drupal.org/node/3083055

### DIFF
--- a/tests/functional/PmEnDisUnListInfoTest.php
+++ b/tests/functional/PmEnDisUnListInfoTest.php
@@ -31,8 +31,8 @@ class EnDisUnListInfoCase extends CommandUnishTestCase
         $out = $this->getOutput();
         $this->assertContains('devel', $out);
 
-        $this->drush('core:status', [], ['field' => 'drupal-version']);
-        $drupal_version = $this->getOutputRaw();
+        $drupal_root = Drush::bootstrapManager()->getRoot();
+        $drupal_version = Drush::bootstrap()->getVersion($drupal_root);
 
         // Test the testing install profile theme is installed.
         // Since Drupal 8.8, stark is the default testing theme.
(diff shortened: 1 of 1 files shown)


JTubex:
Currently PR #4198 and #4201 are failing because of this change in Drupal Core.

greg-1-anderson:
This looks fine. It would also be okay to test for classy OR stark, I think.

JTubex:
Allright, tests are passing again.

weitzman:
Thanks. I was wondering why this started failing.

The right way to get Drupal version is to call `\Drush\Boot\Boot::getVersion`. We have a helper in Drush.php for getting the current Boot object. But if you want to change the assertion to Stark or Classy thats fine with me.

JTubex:
Hmm, when using 
```        
$drupal_root = Drush::bootstrapManager()->getRoot();
$drupal_version = Drush::bootstrap()->getVersion($drupal_root);
```

I get the following error:

```
1) Unish\EnDisUnListInfoCase::testEnDisUnList
RuntimeException: Drush::$container is not initialized yet. \Drush::setContainer() must be called with a real container.
```

weitzman:
Sorry I did not realize we were in a test. Your prior call to $this->drush() is the best way.

JTubex:
Np. Reverted the changes. Tests should pass again.

weitzman:
Thanks for persisting. Merged into master and 9.x

PR #4204, opened 2019-10-04
Closes #4203 - In CREATE DATABASE statements, quote database name by default

### LINKED ISSUE #4203: Inconsistent quoting of the database name in CREATE DATABASE statements executed by sql:create and site:install
**Describe the bug**
There is inconsistent quoting of the database name in CREATE DATABASE statements executed by sql:create and site:install.  Both are hardcoded and not configurable.

- [sql:create passes TRUE to the $quoted param ](https://github.com/drush-ops/drush/blob/master/src/Commands/sql/SqlCommands.php#L103)
- [site:install calls SqlBase::dropOrCreate which does not pass a param so $quoted is FALSE by default](https://github.com/drush-ops/drush/blob/master/src/Sql/SqlBase.php#L463)

**To Reproduce**
Create databases with special characters, such as hyphens, in the names:

This succeeds: 
`drush sql:create --db-url="mysql://user:pass@127.0.0.1/name-with-hyphens"`

This fails when the database has not already been created: 
`drush site:install --db-url="mysql://user:pass@127.0.0.1/another-name-with-hyphens"`

**Expected behavior**
drush site:install should be able to complete with a supplied database name with hyphens that has not yet been created.

**Actual behavior**
drush site:install fails

**Workaround**
The workaround is to create the database first before doing a site install.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.x (master)
| Drupal version? | 8.x
| PHP version     | 7.1 & 7.3 tested
| OS?             | Mac & Linux

ericras:
The default behavior should be to quote database names and to provide an option to not do so if on a platform that has problems with it.

Closes #4203

### DIFF
--- a/src/Commands/core/SiteInstallCommands.php
+++ b/src/Commands/core/SiteInstallCommands.php
@@ -372,7 +372,7 @@ class SiteInstallCommands extends DrushCommands implements SiteAliasManagerAware
         $bootstrapManager = Drush::bootstrapManager();
         $bootstrapManager->doBootstrap(DRUSH_BOOTSTRAP_DRUPAL_SITE);
 
-        if ($program_exists && !$sql->dropOrCreate()) {
+        if ($program_exists && !$sql->dropOrCreate($sql->getOption('quoted'))) {
             throw new \Exception(dt('Failed to drop or create the database: @error', ['@error' => $sql->getProcess()->getOutput()]));
         }
     }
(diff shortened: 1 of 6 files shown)


weitzman:
I think there is a broader open issue in drupal and drush around quoting and reserved keywords. Cant find it right now.

weitzman:
I'd rather not add options. Can we just add quoting to site:install?

weitzman:
Lets discuss in #4204

weitzman:
This was fixed in #5167

PR #4209, opened 2019-10-07
Issue-4208: Fix User:login command incorrect url handling

hernani:
Issue-4208

### DIFF
--- a/src/Commands/core/LoginCommands.php
+++ b/src/Commands/core/LoginCommands.php
@@ -82,7 +82,7 @@ class LoginCommands extends DrushCommands implements SiteAliasManagerAwareInterf
                 ],
                 [
                   'absolute' => true,
-                  'language' => \Drupal::languageManager()->getCurrentLanguage(),
+                  'language' => \Drupal::languageManager()->getLanguage($account->getPreferredLangcode()),
                 ]
             )->toString();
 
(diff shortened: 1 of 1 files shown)


weitzman:
Doesn't `Url` use current language by default? I dont see this in other calls to Url, so I'm curious.

hernani:
I think the problem is in the way that user:login command is concatenating '/login' after the url has been properly constructed in https://github.com/drush-ops/drush/blob/9.x/src/Commands/core/LoginCommands.php#L74 , so even if the url was properly built, it can be wrong when we append the suffix manually.

hernani:
Yeap, when using a language suffix as language negotiation method.

Before PR:
- http://locahost/user/reset/1/1570442335/VUGw8gkJPUQZ3ZhdtzfSpo2DzptejtnDfj6Z1vp38q8_en/login

Which is not recognized.

After PR:
- http://localhost/user/reset/1/1570442335/VUGw8gkJPUQZ3ZhdtzfSpo2DzptejtnDfj6Z1vp38q8/login_en

Which works correctly. Drupal just recognize the one time login link built directly from the route.

weitzman:
We should do the destination= querystring inside `Url` as well, IMO

weitzman:
Fixed that, and merged to 9.x and master. Thanks.

hernani:
Thanks Moshe!

PR #4217, opened 2019-10-15
Use the UpdateKernel for config:import.

### LINKED ISSUE #3574: Drush becomes unusable if a command requires a new module

### LINKED ISSUE #4216: `config:import` should use the `UpdateKernel`
**Describe the bug**
I am trying to deploy a new feature that introduces a new custom module in my project. In this module there is a new service which will be used by some of the existing services in the existing modules.

So in the production database this new module with the new service dependency is not yet enabled. The new module will be enabled when I run `drush config:import`, but unfortunately when I run this command I get the following error:

```
    [drush] Executing command: ./vendor/bin/drush --root="./web" --verbose --yes config:import

In CheckExceptionOnInvalidReferenceBehaviorPass.php line 31:

  [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]
  The service "asset_release.fulfillment_guard" has a dependency on a non-existent service "workflow_state_permission.workflow_state_permission".
```

This is because the dependency container is built using the standard `DrupalKernel` which detects if a non-existing dependency is used and refuses to build. We should instead use the `UpdateKernel` which will skip all unmet dependencies automatically (ref. `UpdateCompilerPass::process()`). This has been introduced recently in Drupal 8.7.x, ref [Change record #3067480: The UpdateKernel removes services with unmet dependencies from the container](https://www.drupal.org/node/3067480).

I naively tried to add the annotation `@kernel update` to the `ConfigImportCommands::import()` command but this doesn't work. The `config:import` command is itself a service, so it is only discovered _after_ Drupal is already fully bootstrapped in `\Drush\Application::bootstrapAndFind()`. Unfortunately this will bootstrap the default `DrupalKernel` and will crash before it even has the chance to discover the command.

**To Reproduce**
* Add a new module that introduces a new service dependency.
* Run `drush config:import`

**Expected behavior**
* The config is imported.

**Actual behavior**
* `Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The service "existing_module.existing_service" has a dependency on a non-existent service "new_module.new_service".`

**Workaround**
* Not tested but probably it is possible to enable the new module in a `hook_update_n()` hook before running `drush config:import`.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.x
| Drupal version? | 8.7.x\|8.8.x
| PHP version     | 5.x\|7.x
| OS?             | Mac\|Linux\|Windows\|Unix

pfrenssen:
Fixes #4216

edit: Fixes #3574

### DIFF
--- a/src/Drupal/Commands/config/ConfigImportCommands.php
+++ b/src/Drupal/Commands/config/ConfigImportCommands.php
@@ -290,30 +290,4 @@ class ConfigImportCommands extends DrushCommands
             }
         }
     }
-
-    /**
-     * @hook validate config-import
-     * @param \Consolidation\AnnotatedCommand\CommandData $commandData
-     * @return \Consolidation\AnnotatedCommand\CommandError|null
-     */
-    public function validate(CommandData $commandData)
-    {
-        $msgs = [];
-        if ($commandData->input()->getOption('partial') && !$this->getModuleHandler()->moduleExists('config')) {
-            $msgs[] = 'Enable the config module in order to use the --partial option.';
-        }
-
-        if ($source = $commandData->input()->getOption('source')) {
-            if (!file_exists($source)) {
-                $msgs[] = 'The source directory does not exist.';
-            }
-            if (!is_dir($source)) {
-                $msgs[] = 'The source is not a directory.';
-            }
-        }
-
-        if ($msgs) {
-            return new CommandError(implode(' ', $msgs));
-        }
-    }
 }
--- a/src/Drupal/Commands/config/ConfigImportCommands.php
+++ b/src/Drupal/Commands/config/ConfigImportCommands.php
@@ -177,14 +177,10 @@ class ConfigImportCommands extends DrushCommands
     /**
      * Import config from a config directory.
      *
-     * @command config:import
-     * @param $label A config directory label (i.e. a key in \$config_directories array in settings.php).
-     * @interact-config-label
-     * @option diff Show preview as a diff.
-     * @option preview Deprecated. Format for displaying proposed changes. Recognized values: list, diff.
-     * @option source An arbitrary directory that holds the configuration files. An alternative to label argument
-     * @option partial Allows for partial config imports from the source directory. Only updates and new configs will be processed with this flag (missing configs will not be deleted).
-     * @aliases cim,config-import
+     * This command is invoked through a wrapper command because it requires to
+     * be bootstrapped using the UpdateKernel.
+     *
+     * @see \Drush\Commands\config\ConfigImportCommands::import()
      */
     public function import($label = null, $options = ['preview' => 'list', 'source' => self::REQ, 'partial' => false, 'diff' => false])
     {
--- a/tests/functional/ConfigTest.php
+++ b/tests/functional/ConfigTest.php
@@ -3,6 +3,8 @@
 namespace Unish;
 
 use Composer\Semver\Comparator;
+use Drupal\Core\Serialization\Yaml;
+use Webmozart\PathUtil\Path;
 
 /**
  * Tests for Configuration Management commands for D8+.
@@ -11,6 +13,7 @@ use Composer\Semver\Comparator;
  */
 class ConfigCase extends CommandUnishTestCase
 {
+    use TestModuleHelperTrait;
 
     public function setUp()
     {
@@ -31,10 +34,7 @@ class ConfigCase extends CommandUnishTestCase
 
     public function testConfigExportImportStatusExistingConfig()
     {
-        // Get path to sync dir.
-        $this->drush('core:status', [], ['format' => 'json', 'fields' => 'config-sync']);
-        $sync = $this->webroot() . '/' . $this->getOutputFromJSON('config-sync');
-        $system_site_yml = $sync . '/system.site.yml';
+        $system_site_yml = $this->getConfigSyncDir() . '/system.site.yml';
 
         // Test export.
         $this->drush('config-export');
@@ -82,5 +82,53 @@ class ConfigCase extends CommandUnishTestCase
             $page = $this->getOutputFromJSON('system.site:page');
             $this->assertContains('unish existing', $page['front'], 'Existing config was successfully imported during site:install.');
         }
+    }
+
+    public function testConfigImport()
+    {
+        $options = [
+            'include' => __DIR__,
+        ];
+        $this->setupModulesForTests(['woot'], Path::join(__DIR__, 'resources/modules/d8'));
+        $this->drush('pm-enable', ['woot'], $options);
+
+        // Export the configuration.
+        $this->drush('config:export');
+
+        $root = $this->webroot();
+
+        // Introduce a new service in the Woot module that depends on a service
+        // in the Devel module (which is not yet enabled).
+        $filename = Path::join($root, 'modules/unish/woot/woot.services.yml');
+        $serviceDefinition = <<<YAML_FRAGMENT
+  woot.depending_service:
+    class: Drupal\woot\DependingService
+    arguments: ['@devel.dumper']
+YAML_FRAGMENT;
+        file_put_contents($filename, $serviceDefinition, FILE_APPEND);
+
+        $filename = Path::join($root, 'modules/unish/woot/woot.info.yml');
+        $moduleDependency = <<<YAML_FRAGMENT
+dependencies:
+  - devel
+YAML_FRAGMENT;
+        file_put_contents($filename, $moduleDependency, FILE_APPEND);
+
+        // Add the 'devel' module in core.extension.yml.
+        $extensionFile = $this->getConfigSyncDir() . '/core.extension.yml';
+        $this->assertFileExists($extensionFile);
+        $extension = Yaml::decode(file_get_contents($extensionFile));
+        $extension['module']['devel'] = 0;
+        require_once $root . "/core/includes/module.inc";
+        $extension['module'] = module_config_sort($extension['module']);
+        file_put_contents($extensionFile, Yaml::encode($extension));
+
+        $this->drush('config:import');
+    }
+
+    protected function getConfigSyncDir()
+    {
+        $this->drush('core:status', [], ['format' => 'json', 'fields' => 'config-sync']);
+        return $this->webroot().'/'.$this->getOutputFromJSON('config-sync');
     }
 }
--- a/includes/drush.inc
+++ b/includes/drush.inc
@@ -216,7 +216,7 @@ function drush_call_user_func_array($function, $args = []) {
 function drush_memory_limit() {
   $value = trim(ini_get('memory_limit'));
   $last = strtolower($value[strlen($value)-1]);
-  $size = (int) substr($value, 0, -1);
+  $size = (int) rtrim($value, 'GgMmKk');
   switch ($last) {
     case 'g':
       $size *= DRUSH_KILOBYTE;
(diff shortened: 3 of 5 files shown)


weitzman:
This looks good to me. Perhaps @greg-1-anderson has thoughts on whether this is the right pattern (registering a drupal service and calling it from a pre-bootstrap command command). Maybe it would be more better/explicit if the \Drush\Drupal\Commands\config\ConfigImportCommands::import were a hidden Drush command that was called by \Drush\Commands\config\ConfigImportCommands::import.

Also see #3574. Is that fixed or influenced by this PR?

pfrenssen:
I can confirm that #3574 is exactly the same issue. I will close mine and add the info there.

pfrenssen:
The test is green for all environments except "PHP 7.2 - Highest". `Unish\ConfigCase::testConfigExportImportStatusExistingConfig` is failing.

weitzman:
Thats a known issue, unrelated to the PR.

weitzman:
So UpdateKernel isnt really a reference to update.php? Its useful for config:import as well?

pfrenssen:
The `UpdateKernel` has some adaptations that allow it to bootstrap Drupal when it is in the middle of an update. A deployment happens in the following stages:

1. New code is deployed.
2. First round of database updates - `hook_update_n()`.
3. Import configuration
4. Perform post-config-import tasks (will be added in https://www.drupal.org/project/drupal/issues/2901418)
5. Clear the cache (depending on the project, either all of them or only specific bins).

During this procedure the site is considered to be in a "broken state" because there might be discrepancies between the code, database and cache that can cause fatal errors. At the end of these 5 steps everything is in sync again and the site is back in a working state. 

Here is how the `UpdateKernel` protects against fatal errors during the update:

-  It prevents services critical to the deployment process from using the cache (`\Drupal\Core\Update\UpdateServiceProvider::alter()`).
- It forces a container rebuild (`\Drupal\Core\Update\UpdateKernel::initializeContainer()`).
- It clears broken services from the dependency injection container (`\Drupal\Core\Update\UpdateCompilerPass::process()`).
- It prevents the modified container from being cached so that it will be regenerated in correct state after the update is finished (`\Drupal\Core\Update\UpdateKernel::cacheDrupalContainer()`).

This PR is fixing step 3 in the update procedure. At this point any new modules that are added during the deployment are not enabled yet (they will soon be enabled when the `core.extension` config item is updated) so the site is still in a broken state. More specifically, the container cannot be compiled if any previously existing service depends on a service from one of the not-yet-enabled modules.

So we need to use the `UpdateKernel` in this step.

pfrenssen:
I have in the meantime tried this solution in a test deployment of a branch that was exhibiting the problem in our "large" project. It now works perfectly.

src/Drupal/Commands/config/ConfigImportCommands.php
```diff
@@ -177,14 +177,10 @@ public function __construct(ConfigManagerInterface $configManager, StorageInterf
     /**
      * Import config from a config directory.
      *
-     * @command config:import
```
weitzman: Can you see if the validate() hook in this file still runs? Even if it does, I think it should move to the new file along with its command definition.

pfrenssen: It is still working:

```
$ ./vendor/bin/drush config:import --partial
 [error]  Enable the config module in order to use the --partial option.
```

I'll move it to the new file anyway.

claudiu-cristea:
I've provided a regression test in #4220. It's in a separate PR because I wanted to see the failure against master

claudiu-cristea:
Now the PR #4220 is against config-import-updatekernel branch. If is accepted will be merged here.

weitzman:
Thanks much. I feel this is a bit risky to backport to 9, but am happy to be convinced otherwise.

pfrenssen:
Thanks!

There is no pressing need for us to have this backported to Drush 9. The patch from this PR applies cleanly to 9 and can be used for now by us and others who are hit with this problem. And Drush 10 is just around the corner!

alexpott:
This needs to be rolled back. It does work for the most basic of configuration imports that enable a module. Fundamentally the config import operation is not an update operation.

Steps to reproduce:
1. Install standard.
2. Install media_library
3. export configuration
4. visit admin/config/media/media-library
5. uninstall media_library
6. use drush to import configuration
7. visit admin/config/media/media-library - it's broken
8. use drush cr but it is still broken but the module is showing as enabled and there's nothing to import.

This is happening because of https://www.drupal.org/node/3055443 but that doesn't make this change correct. The original issue behind this feels like it is the wrong way around. Delivering config updates and code updates shouldn't happen at the same time. If your updating modules to depend on a new service provided by a new module then you need to ship a code update that enables that module via hook_update_N. 

This change has made Drush's config import work in a different way that cores that is not supported. If we want to support that - and I'm not sure we do - then we need to change core's config ui as well and do this in a co-ordinated fashion.

BramDriesen:
Cross referencing a bit here: https://www.drupal.org/project/drupal/issues/3119373

tests/functional/ConfigTest.php
```diff
@@ -83,4 +83,52 @@ public function testConfigExportImportStatusExistingConfig()
...
+        $extensionFile = $this->getConfigSyncDir() . '/core.extension.yml';
+        $this->assertFileExists($extensionFile);
+        $extension = Yaml::decode(file_get_contents($extensionFile));
+        $extension['module']['devel'] = 0;
+        require_once $root . "/core/includes/module.inc";
+        $extension['module'] = module_config_sort($extension['module']);
+        file_put_contents($extensionFile, Yaml::encode($extension));
+
+        $this->drush('config:import');
+    }
```
alexpott: This is not actually testing that config import works. It's making an assumption that if the command is successful it's successful but it doesn't actually properly test that a module has been enabled and it in the container.modules parameter.

BramDriesen:
@weitzman  Please have a look at this (not sure if you notified on comments of closed PR's)

weitzman:
Thanks Bram. We are awaiting a reply from @claudiu-cristea  and @pfrenssen about this. Please join #drush on Slack as there may be conversation there as well.

pfrenssen:
> Delivering config updates and code updates shouldn't happen at the same time. If your updating modules to depend on a new service provided by a new module then you need to ship a code update that enables that module via hook_update_N.

I'm not sure how to handle this situation when using automated deployments. Isn't it the case that in normal circumstances the config import happens after the db updates but before the post updates? For example a typical post-update script will handle a data migration to a new field which is deployed through config. How are we supposed to enable a module in a post-update before deploying config without resorting to a manual deployment?

There is also the problem that you don't have real granular control over when and how users deploy your project, they might skip a bunch of releases and go e.g. from version 1.10 to 1.20 in one deployment. How are we supposed to deal with this if there is a post-update hook in version 1.12 that enables a module and should be run before any of the config from later versions is deployed?

BramDriesen:
@pfrenssen even just enabling a contributed/core module (that might even be on the server already for a long time) does not work...

pfrenssen:
@BramDriesen can you open an issue for this providing more information about the specific circumstances that you encounter?

This is definitely not a generic case, looking back in our project history we have enabled 21 modules and disabled 2 modules using this code since it was released and had not a single problem.

BramDriesen:
@pfrenssen see #4350 referenced above.

bartlangelaan:
I can confirm that I experienced this issue, even when only enabling modules. Applying the patch from issue [#3119373](https://www.drupal.org/project/drupal/issues/3119373#comment-13515436) fixed this problem.

I executed this command, once with patch and once without patch. 

```
ddev import-db < ../qfyruyad6alua--master-7rqtwti--drupal-database--main--dump.sql && \
ddev exec drush -y cache-rebuild && \
ddev exec drush -y updatedb && \
ddev exec drush -y config-import --diff && \
ddev exec drush -y cache-rebuild && \
curl https://[project]/graphql
```

In the end of the command result, you can see that without that patch the `graphql` module does not get properly installed, because the 403 path is loaded. With the patch, 403 is loaded (the path exists, just not the right permissions). This is expected behaviour.


<details><summary>Without patch</summary>

```
Successfully imported database 'db' for [project] 
Existing database 'db' was dropped before importing  
```

```
```

```
 [success] No pending updates.
```

``` 
diff --git a/tmp/drush_tmp_1585601413_5e825b85b68a7/core.extension.yml b/tmp/drush_tmp_1585601413_5e825b85c6e83/core.extension.yml
index 2403cd61d..87d84b10c 100644
--- a/tmp/drush_tmp_1585601413_5e825b85b68a7/core.extension.yml
+++ b/tmp/drush_tmp_1585601413_5e825b85c6e83/core.extension.yml
@@ -12,6 +12,9 @@ module:
   field_ui: 0
   file: 0
   filter: 0
+  graphql: 0
+  graphql_core: 0
+  graphql_export: 0
   link: 0
   menu_ui: 0
   node: 0


 // Import the listed configuration changes?: yes.                                                                      

 [notice] Synchronized extensions: install graphql.
 [notice] Synchronized extensions: install graphql_core.
 [notice] Synchronized extensions: install graphql_export.
 [notice] Finalizing configuration synchronization.
 [success] The configuration was imported successfully.
```

``` 
```

```
system.404
```
</details>

<details><summary>With patch</summary>

```
Successfully imported database 'db' for [project] 
Existing database 'db' was dropped before importing
```

``` 
```

```
 [success] No pending updates.
```

```
diff --git a/tmp/drush_tmp_1585601997_5e825dcdde7e6/core.extension.yml b/tmp/drush_tmp_1585601997_5e825dcdf2c8b/core.extension.yml
index 2403cd61d..87d84b10c 100644
--- a/tmp/drush_tmp_1585601997_5e825dcdde7e6/core.extension.yml
+++ b/tmp/drush_tmp_1585601997_5e825dcdf2c8b/core.extension.yml
@@ -12,6 +12,9 @@ module:
   field_ui: 0
   file: 0
   filter: 0
+  graphql: 0
+  graphql_core: 0
+  graphql_export: 0
   link: 0
   menu_ui: 0
   node: 0


 // Import the listed configuration changes?: yes.                                                                      

 [notice] Synchronized extensions: install graphql.
 [notice] Synchronized extensions: install graphql_core.
 [notice] Synchronized extensions: install graphql_export.
 [notice] Finalizing configuration synchronization.
 [success] The configuration was imported successfully.
```


```
```

```
system.403
```
</details>

So, the patch fixed it.

But when I applied the [Drush patch](https://www.drupal.org/project/drupal/issues/3119373#comment-13510850) from the same Drupal issue, the same (expected) result appeared.

So, it looks like both patches fix this problem.

Maybe the update kernel should only be used from 8.8.5 onwards, if at all?

PR #4223, opened 2019-10-21
Closes #4071 by setting -y to sql-sync command

### LINKED ISSUE #4071: sql-sync fails at core:rsync step
**Describe the bug**
In Drush 9.6.2, sql-sync now fails when attempting to run core:rsync. 

**To Reproduce**
1. Install Drush 9.6.2 in docker container `ten7/flight-deck-web`. This can either be globally via cgr or locally.
2. Set up SSH keys as expected.
3. Set up drush aliases as expected.
4. Run drush sql-sync @remote_source @self

**Expected behavior**
The sql-sync to complete.

**Actual behavior**
When running the command, the sync fails when attempting to rsync the file from the remote source. See verbose output:

https://gist.github.com/socketwench/287e8bf4e1aae90b4e5a7c8907738edc

**Workaround**
If the `core:rsync` command is run manually, it succeeds and the resultant database dump can be imported as expected.

Alternatively, revert back to drush 9.5.2, where the sql-sync completes without issue. 

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.6.2
| Drupal version? | 8.x
| PHP version     | 7.2
| OS?             | Alpine Linux via Docker.

**Additional information**

back-2-95:
Tested with

- `drush sql-sync @remote-site-alias @self`
- `drush -y sql-sync @remote-site-alias @self`

This fixes the problem described in #4071 run inside the Alpine container.

Please advice if this needs to be somewhere else.

### DIFF
--- a/src/Commands/sql/SqlSyncCommands.php
+++ b/src/Commands/sql/SqlSyncCommands.php
@@ -221,8 +221,7 @@ class SqlSyncCommands extends DrushCommands implements SiteAliasManagerAwareInte
                 $runner = $targetRecord;
             }
             $this->logger()->notice(dt('Copying dump file from source to target.'));
-            $options = ['yes' => true];
-            $process = $this->processManager()->drush($runner, 'core-rsync', [$sourceRecord->name() . ":$source_dump_path", $targetRecord->name() . ":$target_dump_path"], $options, $double_dash_options);
+            $process = $this->processManager()->drush($runner, 'core-rsync', [$sourceRecord->name() . ":$source_dump_path", $targetRecord->name() . ":$target_dump_path"], ['yes' => true], $double_dash_options);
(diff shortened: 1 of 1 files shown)


weitzman:
This change looks good to me. I'm puzzled why there arent more bug reports for this.

Could you inline the `['yes' => true]`? That way we dont redefine what $options means.

back-2-95:
I was wondering that too as we have the same Drush version in multiple other projects. Seems odd that currently this seems related to Alpine Docker image.

I can inline that. Was wondering also should possible global debug flag to be added to this array?
It would show rsync more verbose.

weitzman:
Lets omit the verbose flag for now.

back-2-95:
Ok, now it's inline.

PR #4227, opened 2019-10-22
Try out Drupal 9 support

greg-1-anderson:
Drupal 9 for Drush 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ matrix:
 
     # D8.7.x
     - php: 7.0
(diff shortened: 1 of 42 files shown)


weitzman:
Looks like Circle didnt run the full suite.

weitzman:
Oh this is for drush8. Lets chat.

greg-1-anderson:
It is not possible to make more progress here until updates.drupal.org is advertising Drupal 9 versions.

greg-1-anderson:
I hacked around in the `pm-*` code and sorta-kinda made it work with the new `all` endpoint, so tests are getting a little farther.

I skipped the tests that require a version of devel for Drupal 9. Now the test suite is down to just two failures.

PR #4231, opened 2019-10-24
Remove test dependency on devel/empty_theme

larowlan:
Removes test dependency on devel and empty_theme by adding mock module and themes for the same purposes

I tried to remove the dependency on `drupal/alink` too, but unfortunately SecurityUpdatesTest expects to find an insecure module in composer.lock.

I don't know if that can be gotten around.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,18 +55,16 @@
     "webmozart/path-util": "^2.1.0"
   },
   "require-dev": {
-    "lox/xhprof": "dev-master",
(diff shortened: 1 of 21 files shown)


greg-1-anderson:
This is great. Presuming that the tests all come back green, I think we can merge here and then just markTestSkipped if the Drupal version is 9.x and the test needs drupal/alink, until such a time as it is available.

greg-1-anderson:
Awesome! Great improvements to the tests, thanks!

PR #4232, opened 2019-10-25
Try running the Drush 10 test suite with Drupal 9.

greg-1-anderson:


.circleci/config.yml
```diff
@@ -111,6 +111,31 @@ jobs:
...
+    docker:
+      - image: wodby/php:7.2
+        environment:
+          - MYSQL_HOST=127.0.0.1
+          - UNISH_DB_URL=mysql://root:@127.0.0.1
+      - image: circleci/mysql:5.7.24
+    steps:
+      - checkout
+      - run: cp .docker/zz-php.ini /usr/local/etc/php/conf.d/
+      - run: composer remove --dev webflo/drupal-core-strict --no-update
```
greg-1-anderson: We can convert to drupal/core-recommended and drupal/dev-dependencies now, but let's do that as follow-on work.

greg-1-anderson: Once we convert, we won't have to just remove here.

.circleci/config.yml
```diff
@@ -126,6 +151,8 @@ workflows:
       - check_mergable
       - test_71_mysql:
           <<: *requires
+      - test_72_drupal88:
+          <<: *requires
       - test_72_highest:
```
greg-1-anderson: We could rename this "drupal90" if desired.

tests/functional/UserTest.php
```diff
@@ -115,6 +115,12 @@ public function testUserLogin()
 
     public function testUserCancel()
     {
+        // TODO: re-enable this test once the Drupal Code Generator is
+        // compatible with Drupal 9.
+        if (version_compare('9.0.0', \Drupal::VERSION, '>=')) {
+            $this->markTestSkipped('Drupal Code Generator not working on Drupal 9 yet.');
```
greg-1-anderson: @Chi-teck Could you update DCG to add `core_version_requirement` to generated module info.yml files please?

tests/integration/SecurityUpdatesTest.php
```diff
@@ -15,6 +15,12 @@ class SecurityUpdatesTest extends UnishIntegrationTestCase
    */
     public function testInsecurePackage()
     {
+        // TODO: re-enable this test once the pm:security command is
+        // compatible with Drupal 9.
+        if (version_compare('9.0.0', \Drupal::VERSION, '>=')) {
+            $this->markTestSkipped('pm:security not working on Drupal 9 yet.');
```
greg-1-anderson: I'm not really sure why pm:security is failing, but it might have something to do with lack of data for 9.x right now. I didn't look into it -- I figured it was better to just skip this test for now and circle back here later.

weitzman: Yeah, lack of 9.x update data blocks updating of https://github.com/drupal-composer/drupal-security-advisories upon which that command depends.

.circleci/config.yml
weitzman: If it downloads Druapl 9, that rename seems wise.

composer.lock
```diff
@@ -2517,16 +2516,16 @@
         },
         {
             "name": "behat/mink-selenium2-driver",
-            "version": "1.3.x-dev",
+            "version": "dev-master",
```
weitzman: Kinda surprised that is needed. Whats causing this?

greg-1-anderson: Good catch. I accidentally committed the wrong lock file. Pushed another commit that fixes it.

tests/functional/UserTest.php
Chi-teck: Added in 1.30.5.

PR #4244, opened 2019-11-01
Fix #4243. site-install doesn't use db connection in settings.php

### LINKED ISSUE #4243: site-install doesn't use db connection in settings.php
**Describe the bug**
I have my DB settings stored in settings.php like... well, pretty much every Drupal site on the planet :) In Drush 9, I could run `drush si` and it would install to whichever default db was configured via settings.php. In Drush 10, it seems to ignore these settings. If I pass no other parameters, it prompts for db connection info. If I pass `--no-interaction`, the install fails with `Failed to drop or create the database`

**To Reproduce**
Take an existing site with a default db defined in settings.php, upgrade to Drush 10, try running `drush si`.

**Expected behavior**
Site to be installed according to default db connection in settings.php.

**Actual behavior**
Drush seems to ignore the default db info. It prompts for db connection info, and provides inappropriate defaults (drupal/drupal/drupal) that don't match my configured db settings.

**Workaround**
Manually pass the db URL or fill in the prompts when running `drush si`

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 8.7.8
| PHP version     | 7.x
| OS?             | Linux

weitzman:
Affects a re-install.

### DIFF
--- a/src/Commands/core/SiteInstallCommands.php
+++ b/src/Commands/core/SiteInstallCommands.php
@@ -248,6 +248,7 @@ class SiteInstallCommands extends DrushCommands implements SiteAliasManagerAware
      */
     public function validate(CommandData $commandData)
     {
+        $bootstrapManager = Drush::bootstrapManager();
         if ($sites_subdir = $commandData->input()->getOption('sites-subdir')) {
             $lower = strtolower($sites_subdir);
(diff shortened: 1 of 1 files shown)


danepowell:
Works great!

PR #4245, opened 2019-11-04
Remove more —backend cruft

weitzman:
Also removes a few other unused options.


greg-1-anderson:
Maybe we should remove as much as possible such that `drush --backend` does not exit with an error.

weitzman:
Removed one more line. Now you try to use `--backend` and you get ` The "--backend" option does not exist.`. Is that adequate?

greg-1-anderson:
Sorry, I was trying to convey the opposite, but I didn't test. I think that this PR prior to 
c79a228 is what I was suggesting. If we cause `--backend` to exit with an error, then we might break sql-sync with older versions of Drush.

I'm not sure if that is a goal or not. Maybe it works with Drush 9 but not Drush 8, which might be sufficient.

weitzman:
Oh, now I see what you mean. I think we can't merge this until Drush 11 :(. I was considering the use case where Drush 10 calls into a remote Drush 9 but I had not yet considered Drush 9 calling into Drush 10, which is valid and would be broken by this PR.

Issue #4235, opened 2019-10-29
Import multiple custom translation po files


davidferlay:
**Is your feature request related to a problem? Please describe.**

- Drush already provides a command to import individual custom translation po files, such as :
`drush locale:import fr /var/www/html/translations/fr.po --type=customized --override=all`
- This command requires to be duplicated when you want to import files for multiples languages, making things [messy](https://github.com/skilld-labs/skilld-docker-container/blob/master/Makefile#L92)
- It also requires you to know for which languages you have translations files or not

**Describe the solution you'd like**
- Drush could provide a similar command to import multiple po files : 
   - It would allow to streamline the process of importing custom translations on projects, regardless of enabled languages and available po files

- Proposal : `drush locale:import:all $directory $options`
   - The command would require files to be named with valid 2 char language code
   - It would take a directory path as argument
   - It would offer the same options (type + override)
- Exemple : `drush locale:import:all /var/www/html/translations/ --type=customized --override=all`

**Describe alternatives you've considered**
- https://www.drupal.org/project/drush_language brings `drush locale-import-all` : 
   - The feature seem critical enough to be brought by Drush itself rather than a contrib module
   - Most likely code of this module will be a good source of inspiration

andypost:
On other hand it could be polished as "drush_language" and added to core as part of https://www.drupal.org/project/drupal/issues/2242947

davidferlay:
Tested locally, works perfectly with `drush locale:import:all /var/www/html/translations/ --type=customized --override=all` and similar commands, using po files like [this one ](https://github.com/drush-ops/drush/files/3840330/en.txt)(to rename as .po).

PR was done from Drush version : 10.0.3-dev

- [x] Drush provides new command `drush locale:import:all $directory $options`
- [x] The command requires files to be named with valid 2 char language code
- [x] It takes a directory path as argument
- [x] It offers the same options (type + override)
- [x] it imports only translations for installed modules and enabled languages
- [x] It imports only updated and new translations

![2019-11-12_13-47](https://user-images.githubusercontent.com/20738264/68750043-8ab2e280-05ff-11ea-92e0-143a680818aa.png)

davidferlay:
CI fixed!)

davidferlay:
Any feedback @geek-merlin ?

ptmkenny:
Is there any way or plan to export custom translations as well via drush?

PR #4251, opened 2019-11-13
Drush - Import multiple custom translation po files

Sergey-Orlov:
This adds new drush command to import multiple custom translation po files from directory, as described in #4235


weitzman:
We are lacking a locale maintainer at the moment so I apologize that nobody from Drush core team is reviewing this. We need a review from outside the team, as next step.

alexkras:
Reviewed and tested on real project, all works fine.
Can be merged as-is.

alexkras:
@weitzman what need to do to speed up merge of this changes? thx.

weitzman:
Thanks for the review.

I'm hesitant to add more locale functionality while we have no maintainer for it. Ideally someone maintains these commands outside of Drush core for now.

andypost:
I'd say it needs tests and then it could be a locale console command patch in core

davidferlay:
FYI we have been using this patch in production for years now without any issue

DieterHolvoet:
Maybe we can consider making the directory argument an option? That way, you could set a sitewide default directory through a Drush config file. If you guys are on board with this idea I wouldn't mind implementing it.

davidferlay:
Hi @DieterHolvoet ,
Making the directory argument an option would require to have a default value as fallback (because user can then omit this input, as it is optional). So i'd say, why not providing we figure out a meaningful default value for it.

WDYT ?

davidferlay:
Just made a bit of research and discovered that "option" may not necessarily be "optional" (https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02)

> We could add an extra check in code to make sure the option is present

++ as long the helper reflects it

> We should also support passing relative paths

I agree that this would be a welcome addition

weitzman:
The the decision about merging and maintaining this falls to the locale maintainer, @Sutharsan. This PR would need tests and needs to target 11.x branch.

Issue #4259, opened 2019-11-24
drush site-install fails with MySQL 8 (backport to drush8)


AdamPS:
**Describe the bug**
See title.

**To Reproduce**
Run `drush site-install` with no database or user existing specifying a V8 mysql server.

**Expected behavior**
Database and user created

**Actual behavior**

> Failed to create database: ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IDENTIFIED BY '<PASSWORD>'' at line 1

**Workaround**
Create the user by hand first.

### System Configuration
Should be the same on any versions.  I was using drush 8, but I checked and the code is the same in master.

**Additional information**
The problem is that starting with MySQL 8 you no longer can (implicitly) create a user using the GRANT command.  . Use CREATE USER instead, followed by the GRANT statement.  See this [question/answer](https://stackoverflow.com/questions/50177216/how-to-grant-all-privileges-to-root-user-in-mysql-8-0).

weitzman:
This has caused incompat with MariaDB in master branch so lets hold off on backport until thats resolved.

mstenta:
> This has caused incompat with MariaDB in master branch so lets hold off on backport until thats resolved.

I just ran into this issue attempting to use Drush 8.3.5 `site-install` with MariaDB 10.5.5.

The query changes made in PR #4261 appear to be the cause, specifically the change to: `CREATE USER % IDENTIFIED WITH mysql_native_password BY '%s';`

I can confirm that that query alone does not work in MariaDB 10.5.5:

```
MariaDB [mysql]> CREATE USER test IDENTIFIED WITH mysql_native_password BY 'test';
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'BY 'test'' at line 1
```

**So as of Drush 8.3.0 (when #4261 was merged), `drush site-install` is broken on MariaDB.**

@totten's comment in https://github.com/drush-ops/drush/pull/4261#pullrequestreview-418924852 (which actually points to ANOTHER issue these queries may cause) includes a suggestion that may be a good option for fixing this: conditionally building the queries based on MySQL (5.6 vs 5.7) vs MariaDB.

mstenta:
Oh just found https://github.com/drush-ops/drush/issues/4318 and https://github.com/drush-ops/drush/pull/4364 - looks like that may fix this issue?

[...]

PR #4261, opened 2019-11-24
drush site-install fails with MySQL 8 #4259

AdamPS:
Drush 8 version of fix for #4259 that has been tested

### DIFF
--- a/lib/Drush/Sql/Sqlmysql.php
+++ b/lib/Drush/Sql/Sqlmysql.php
@@ -93,8 +93,9 @@ EOT;
       // - If the database is on a remote server, create a wilcard user with %.
       //   We can't easily know what IP adderss or hostname would represent our server.
       $domain = ($this->db_spec['host'] == 'localhost') ? 'localhost' : '%';
-      $sql[] = sprintf('GRANT ALL PRIVILEGES ON %s.* TO \'%s\'@\'%s\'', $dbname, $this->db_spec['username'], $domain);
-      $sql[] = sprintf("IDENTIFIED BY '%s';", $this->db_spec['password']);
+      $user = sprintf("'%s'@'%s'", $this->db_spec['username'], $domain);
+      $sql[] = sprintf("CREATE USER %s IDENTIFIED WITH mysql_native_password BY '%s';", $user, $this->db_spec['password']);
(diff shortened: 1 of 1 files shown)


weitzman:
LGTM. Lets see what @greg-1-anderson  says as he is maintaining Drush8

greg-1-anderson:
Did not see this earlier. Basically looks good to me; let's let it live on the 8.x branch for a while and see how it does.

I'll look into fixing the Drupal 9 failures separately; they appear unrelated.

AdamPS:
Great thanks

lib/Drush/Sql/Sqlmysql.php
```diff
@@ -93,8 +93,10 @@ public function createdb_sql($dbname, $quoted = FALSE) {
       // - If the database is on a remote server, create a wilcard user with %.
       //   We can't easily know what IP adderss or hostname would represent our server.
       $domain = ($this->db_spec['host'] == 'localhost') ? 'localhost' : '%';
-      $sql[] = sprintf('GRANT ALL PRIVILEGES ON %s.* TO \'%s\'@\'%s\'', $dbname, $this->db_spec['username'], $domain);
-      $sql[] = sprintf("IDENTIFIED BY '%s';", $this->db_spec['password']);
+      $user = sprintf("'%s'@'%s'", $this->db_spec['username'], $domain);
+      $sql[] = sprintf("DROP USER IF EXISTS %s;", $user);
```
totten, line 97: Just a small aside (to save you time - if anyone else brings it up)... and maybe it doesn't matter if you don't need compat with older `mysqld`s... but this usage of `IF EXISTS` first appeared in MySQL 5.7 (compare: [5.6](https://dev.mysql.com/doc/refman/5.6/en/drop-user.html), [5.7](https://dev.mysql.com/doc/refman/5.7/en/drop-user.html)). So this probably doesn't work on MySQL 5.6 or earlier.

In another context (not specifically drush or drupal), I had a round of discussion with @seamuslee001 about this - and we couldn't find a single set of SQL statements that worked across versions. If it arises that you do need older `mysqld`s, then you may need more conditionals ([example](https://github.com/amp-cli/amp/pull/57/files)).

PR #4262, opened 2019-11-24
drush site-install fails with MySQL 8 #4259

AdamPS:
Drush 10 version of fix for #4259 that hasn't been tested


AdamPS:
Great thanks

PR #4267, opened 2019-12-03
Add cache:tags command to purge content by tag.

weitzman:


src/Commands/core/CacheCommands.php
```diff
@@ -58,6 +58,23 @@ public function get($cid, $bin = 'default', $options = ['format' => 'json'])
...
     }
 
+    /**
+     * Invalidate by cache tags.
+     *
+     * @command cache:tags
+     * @param string $tags A comma delimited list of cache tags to clear.
+     * @aliases ct
+     * @bootstrap full
+     * @usage drush cache:tag node-12,user-4
```
LionsAd: Would it not be:

`node:12,user:4`

?

LionsAd: Also is "cache:tag" the right command? Maybe:

`cache:invalidate-tags` (though that is very long)

src/Commands/core/CacheCommands.php
```diff
@@ -58,6 +58,23 @@ public function get($cid, $bin = 'default', $options = ['format' => 'json'])
...
+     *   Purge content associated with two cache tags.
+     */
+    public function tags($tags)
+    {
+        $tags = StringUtils::csvToArray($tags);
+        Cache::invalidateTags($tags);
+        $this->logger()->success(dt("Tnvalidated tag(s): '!list'.", ['!list' => implode(' ', $tags)]));
```
eporama: Are we using "T" instead of "I" in "Tnvalidated" for a reason?

src/Commands/core/CacheCommands.php
weitzman: Fixed tag names. Thanks.

About the command name, I think you are technically correct but I dont think there will be other operations that people want to do with tags besides invalidate. So I'm inclined to keep the briefer name.

src/Commands/core/CacheCommands.php
weitzman: Fixed. Thanks.

Issue #4290, opened 2020-01-07
How can I use a composer symlinked package which contains custom commands?


nvaken:
**Describe the bug**
When using a Composer repository of type `path` to symlink a local package, Drush will not pick up it's custom commands. 

**To Reproduce**
Create a local Composer package, add the `path` repository to your project and require the package and make sure it is being symlinked (not mirrored).

**Expected behavior**
Drush picking up the package.

**Actual behavior**
Drush ignores symlinked package.

**Workaround**
Yes, somewhat. I've found that when removing the if statement in the dependency `consolidation/annotated-command`  in file `src/CommandFileDiscovery.php` on line `402` my package is being picked up. This makes somewhat sense, since `followLinks` is `false` by default. Is drush able to set `followLinks` to `true`? 

**Edit**: Even better workaround seems to be to add `->followLinks(TRUE)` in Drush's: `/src/Application.php` in `commandDiscovery()` where `CommandFileDiscovery` is being instantiated. Is that maybe a acceptable permanent fix?

weitzman:
I don't think we want to follow symlinks. We used to do that and it caused as many problems as it helped. Maybe @greg-1-anderson has input on this ... The workaround is to add new paths for discovery https://github.com/drush-ops/drush/blob/master/examples/example.drush.yml#L52-L67

nvaken:
Would it then be an idea to make following symlinks optional? I'd love to be able to just install composer packages as if it was a normal package while still developing. This way I'll be sure that the package has been installed in the correct path etc...

greg-1-anderson:
I am ambivalent on adding configuration to follow symlinks in the general case. Configuring paths for discovery seems better. Is it really necessary to follow symlinks for module Drush commands, though? It seems that we should be able to find a module wherever it is through service discovery, and the command classes should be defined in the drush.services.yml file.

If more code is needed to make module Drush services / commandfiles discoverable for path repositories, I'd be in favor of a targeted PR that did that. (By targeted, I mean code that specifically identifies and finds modules relocated via a path repository rather than just turning on symlinks for all searching.)

The flip side of symlink following, though, is that Drush had tons of problems with this because in the old days, Drush often walked up to parent directories for the various locations it found. If we are sure that the scenarios that we are enabling symlink following for never does that, then we could consider allowing that, I guess.

nvaken:
[...]

PR #4292, opened 2020-01-09
Adding extra information about symlinked packaged commands

nvaken:
This pull request is in favor of drush-ops/drush#4290, which adds some information about symlinked composer packaged commands. I've created a new heading for this part since it is information that addresses all other headlines so I didn't really know where to put it but here. Everyone okay with that?


greg-1-anderson:
LGTM. Thanks.

Issue #1969, opened 2016-02-03
Twig compile is run without a render context.


rmuilwijk:
The twig-compile command can warmup the Twig Php storage cache. Because of some custom Twig extensions like attach_library it is possible render from the Renderer is called without a render context. This will result in errors.

```php
  // Some functions like attach_library call render without a context so errors
  // are thrown. Setup a dummy render context to avoid this.
  $context = new \Drupal\Core\Render\RenderContext();
  \Drupal::getContainer()->get('renderer')->executeInRenderContext($context, function() use ($searchpaths) {
    foreach ($searchpaths as $searchpath) {
      foreach ($file = drush_scan_directory($searchpath, '/\.html.twig/', array('tests')) as $file) {
        $relative = str_replace(drush_get_context('DRUSH_DRUPAL_ROOT'). '/', '', $file->filename);
        // @todo Dynamically disable twig debugging since there is no good info there anyway.
        twig_render_template($relative, array('theme_hook_original' => ''));
        drush_log(dt('Compiled twig template !path', array('!path' => $relative)), LogLevel::NOTICE);
      }
    }
  });
```

weitzman:
Any chance you can refactor to not use a closure and submit a PR? Thanks.

weitzman:
Or anyone else is welcome to do that change

mglaman:
Linking to a related Drupal core issue: https://www.drupal.org/project/drupal/issues/2950905

There needs to be more than just a request context, possible. When `attach_library` runs it executes a render and tries to get the context from the request.

PR #4294, opened 2020-01-15
Replace twig_render_template() with \Drupal::service('twig')->loadTemplate()

ceesgeene:
Loading the template ensures the compiled template is cached. It is not necessary to also render it.

drush-ops/drush#1969

### DIFF
--- a/src/Drupal/Commands/core/TwigCommands.php
+++ b/src/Drupal/Commands/core/TwigCommands.php
@@ -126,7 +126,7 @@ class TwigCommands extends DrushCommands
         foreach ($files as $file) {
             $relative = Path::makeRelative($file->getRealPath(), Drush::bootstrapManager()->getRoot());
             // Loading the template ensures the compiled template is cached.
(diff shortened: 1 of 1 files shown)


weitzman:
Makes sense. Lets use DI to bring in the twig service.

weitzman:
Looks like we already have $this->twig so DI not an issue. Just use that instead of Drupal:: call.

With and without this patch I'm seeing the following error in this command when using testing profile:

```
                                                                               
  [Twig\Error\LoaderError]                                                     
  Template "block.html.twig" is not defined in "core/modules/system/templates  
  /block--system-branding-block.html.twig" at line 1.      
```

Any idea whats wrong?

Also loadtemplate() has the following warning `* This method is for internal use only and should never be called directly.`. I'm OK ignoring the warning but it gives me pause.

ceesgeene:
@weitzman, not sure why you still encouter that error with the patch. The patch should fix that error.

weitzman:
I retried and this worked so in it goes. Sorry for the delay.

PR #4311, opened 2020-01-30
site:alias-convert does not work when aliases don't have unique keys

Sutharsan:
Converting Drush 8 style multiple aliases files does not work as expected. For example two files example1.* and example2.aliases.drushrc.php that both contains aliases for @example.prod and @example.dev

```
  example1.aliases.drushrc.php
   - prod alias of example1
   - dev alias of example1

  example2.aliases.drushrc.php
   - prod alias of example2
   - dev alias of example2
```

Will be converted to four files: 
```
  example1.site.yml
  # 'Empty except for placeholder comment

  example2.site.yml
  # 'Empty except for placeholder comment

  prod.site.yml
  - prod alias of example2

  dev.site.yml
  - dev alias of example2
```


Sutharsan:
The example1.aliases.drushrc.php file:
```
<?php

$aliases['dev'] = array(
  'root' => '/var/www/example1/htdocs',
  'uri' => 'example1.localhost',
);

$aliases['prod'] = array(
  'root' => '/var/www/example1/htdocs',
  'uri' => 'www.example1.com',
);
```

Sutharsan:
I noticed that the primary difference with the D8 alias files in the unit tests, is that those aliases has unique keys across all alias files, where our files do not have unique keys. When array keys are modified to be unique (example1.dev, example2.dev, etc.) the conversion works again.

Sutharsan:
I don't mind fixing the test if the maintainers agree this is a bug and should be fixed.

greg-1-anderson:
I don't remember how the alias converter decides which aliases are different environments of the same site. If it gets that right, then it should put all of the environments for one site into a single yml file.

If the converter is putting each environment in a separate yml file, then it is confused and does not realize that the aliases all belong to the same site.

The only reason I'd be wary of the suggested PR is that is might make the above determination worse. It seems likely that it should be at least as good as the current code (all of the environments for a single site should be in a single group file). If that's the case, then I'm +1.

If doing the tests reveal problems in this area, then I might request changes, though.

weitzman:
convert is pretty much unmaintained now and will be dropped in next version.

PR #4314, opened 2020-02-05
Unsupport Drush 9 as of May 2020

weitzman:
Drush 9 will be minimally maintained until May 1, 2020. Sites are encouraged to upgrade to Drush 10. The command APIs have not changed - just a few deprecated functions are removed. Most commandfiles will just work. Feel free to use [#drush on Drupal Slack](https://drupal.slack.com/archives/C62H9CWQM) for discussion.

Drush 8 is still supported for sites that are on older versions of Drupal or PHP. See bottom of [Install docs](http://docs.drush.org/en/master/install/).

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -39,7 +39,7 @@ Drupal Compatibility
     <td> Drush 9 </td>
     <td> 9.x </td>
     <td> 5.6+ </td>
-    <td></td> <td></td> <td></td> <td>✅</td> <td></td>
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
+1 on merging this in May 2020.

weitzman:
This went in already via another PR as people wanted early announcement of Drush 9 EOL on the docs site.

greg-1-anderson:
The checkmark is still green ATM, but there is now a column saying when it will go not-green.

This needs a re-roll if we're going to un-green the check.

PR #4322, opened 2020-02-13
Issue 4032 queue run lease time

### LINKED ISSUE #4032: Queue:run has a different behavior than the processQueues() section of Drupal cron.

kevin-dutra:
Resolves #4032. Builds on #4083. This includes the work from @lems3 and @m4olivei.

I've resolved the merge issues due to more recent changes. I think there's value to separating out the queue lease time, so I've done that here also. As an example, let's say we're using the same value both time limit and queue lease and I know queue worker X takes 30s on average, so I use `--time-limit=45` to be on the safe side. 

On one run, it processes item A in 40s. There are still 5s left, so it claims item B, but item B is claimed with an expiry that's only 5s in the future and we know that it's very unlikely to complete within that time frame, leaving the possibility that Drupal will reset item B and a second worker will claim it while it's still being processed by the first worker.

By defaulting to an hour (which the `QueueInterface` _says_ is the default) and allowing someone to specifically allocate the lease time, we can avoid this issue.


weitzman:
Hoping @juampynr, @damiankloip or @davereid can chime in here. Vaeguely recall you guys contributing to this command in the past.

weitzman:
Pity we couldn't get any reviews on this. I just added a bit of polish. I'll merge when it comes back green.

I don really follow the lease-time example in the OP but the code is simple to follow so in it goes.

Issue #4324, opened 2020-02-15
drush.yml isn't discovered in web/sites/default/drush.yml


gitressa:
According to [example.drush.yml](https://github.com/drush-ops/drush/blob/master/examples/example.drush.yml) it should be possible to add a `drush.yml` file these locations:
```
# Directories and Discovery
#
# Rename this file to drush.yml and copy it to one of the places listed below
# in order of precedence:
#
# 1.  Drupal site folder (e.g. sites/{default|example.com}/drush.yml).
# 2.  Drupal /drush and sites/all/drush folders, or the /drush folder
#       in the directory above the Drupal root.
# 3.  In any location, as specified by the --config (-c) option.
# 4.  User's .drush folder (i.e. ~/.drush/drush.yml).
# 5.  System wide configuration folder (e.g. /etc/drush/drush.yml or C:\ProgramData\Drush\drush.yml).
```
I would prefer to use the first option, by simply adding a file to an existing folder, but it doesn't seem to work:
- `web/sites/default/drush.yml`

These work:
- `drush/drush.yml`
- `web/sites/all/drush/drush.yml`

**To Reproduce**
Add a `web/sites/default/drush.yml` file, which isn't discovered.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version  | 9.7.2
| Drupal version | 8.8.2
| PHP version     | 7.2.24
| OS             | Ubuntu with Lando

gitressa:
The same happens with `Drush` 10.2.1.

weitzman:
Can you try a bit more nesting such as `sites/d8/drush/drush.yml`? Thats what I see at https://github.com/drush-ops/drush/blob/c7f062b2519242d5eb95302b854d106245d39480/tests/unit/ConfigLocatorTest.php#L37

gitressa:
I get the same result, i.e. it also doesn't register at `web/sites/default/drush/drush.yml` ...
Also, I should add that the content of my `drush.yml` is created dynamically via a script, so to keep it simple I would much prefer to not have to create a directory, which is why I prefer putting it in `web/sites/default`, since that folder already exists.

richardbporter:
I've tried both `default/drush.yml` and `default/drush/drush.yml` and neither seem to work.

richardbporter:
There is a comment in ConfigLocator related to multisite: https://github.com/drush-ops/drush/blob/master/src/Config/ConfigLocator.php#L109

Is that related?

[...]

PR #4325, opened 2020-02-15
Fixes drush.yml not discovered in sites/default

gitressa:
Fixes that drush.yml isn't discovered in for example web/sites/default/drush.yml, see #4324 .

### DIFF
--- a/src/Config/ConfigLocator.php
+++ b/src/Config/ConfigLocator.php
@@ -280,7 +280,7 @@ class ConfigLocator
         // Remember that we've seen this location.
         $this->siteRoots[] = $siteRoot;
 
-        $this->addConfigPaths(self::DRUPAL_CONTEXT, [ dirname($siteRoot) . '/drush', "$siteRoot/drush", "$siteRoot/sites/all/drush" ]);
+        $this->addConfigPaths(self::DRUPAL_CONTEXT, [ dirname($siteRoot) . '/drush', "$siteRoot/drush", "$siteRoot/sites/all/drush", "$siteRoot/sites/default" ]);
         return $this;
     }
 
@@ -486,7 +486,7 @@ class ConfigLocator
      */
     protected function getSiteCommandFilePaths($root)
     {
-        $directories = ["$root/drush", dirname($root) . '/drush', "$root/sites/all/drush"];
+        $directories = ["$root/drush", dirname($root) . '/drush', "$root/sites/all/drush", "$root/sites/default"];
 
         return array_filter($directories, 'is_dir');
     }
--- a/src/Config/ConfigLocator.php
+++ b/src/Config/ConfigLocator.php
@@ -486,7 +486,7 @@ class ConfigLocator
      */
     protected function getSiteCommandFilePaths($root)
     {
-        $directories = ["$root/drush", dirname($root) . '/drush', "$root/sites/all/drush", "$root/sites/default"];
+        $directories = ["$root/drush", dirname($root) . '/drush', "$root/sites/all/drush"];
 
         return array_filter($directories, 'is_dir');
     }
--- a/tests/fixtures/sites/default/drush.yml
+++ b/tests/fixtures/sites/default/drush.yml
@@ -1,0 +1,2 @@
+test:
+  default: 'A default setting'
--- a/tests/unit/ConfigLocatorTest.php
+++ b/tests/unit/ConfigLocatorTest.php
@@ -35,6 +35,7 @@ class ConfigLocatorTest extends TestCase
         $this->assertEquals($this->fixturesDir() . '/etc/drush/drushVARIANT.yml', Path::canonicalize($sources['test']['variant']));
         $this->assertEquals($this->fixturesDir() . '/home/.drush/drush.yml', Path::canonicalize($sources['test']['home']));
         $this->assertEquals($this->fixturesDir() . '/sites/d8/drush/drush.yml', Path::canonicalize($sources['test']['site']));
+        $this->assertEquals($this->fixturesDir() . '/sites/default/drush.yml', Path::canonicalize($sources['test']['default']));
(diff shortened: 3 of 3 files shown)


weitzman:
Sorry, this doesn't look like right. We discover site-specific config later in the bootstrap process (SITE phase).

gitressa:
Like do you mean it's wrong way of doing it, or it won't work? Because it works well on my machine ...

weitzman:
Wrong way of doing it.

gitressa:
You're right, I added it too many places, it isn't needed under getSiteCommandFilePaths. I have updated the patch.

weitzman:
Now this patch privileges the default directory despite the fact that it might not be in use (could be foo or bar). We have passing test coverage for this so some code is already adding the active site dir in (but perhaps not matching docs in example.drush.yml).

greg-1-anderson:
-1 on this. Should use `drush/drush.yml` at the Drupal root or project root for common configuration, or `sites/foo/drush/drush.yml` for configuration that is specific to the `foo` multisite.

weitzman:
FYI this is the line that tests site specific config loading https://github.com/drush-ops/drush/blob/c7f062b2519242d5eb95302b854d106245d39480/tests/unit/ConfigLocatorTest.php#L37

gitressa:
Ok, but then this line should be edited, and `default` removed in [example.drush.ym](https://github.com/drush-ops/drush/blob/master/examples/example.drush.yml), right?

> 1.  Drupal site folder (e.g. sites/{default|example.com}/drush.yml).

As I wrote earlier, I would much prefer to simply create a `drush.yml` somewhere, without having to create a folder, and `sites/default` is perfect for this usage.

weitzman:
No need to edit that documentation because default is still a valid site and the file should get loaded from there.

I suggest you create that top level folder in your repo or on your server.

gitressa:
I am not sure I fully understand what you're saying. If I put it at web/sites/default/drush.yml it doesn't get picked up ... which it should, right?

Have you verified that a drush.yml file in web/sites/default gets loaded?

weitzman:
I have not verified it, as I dont use multisite. Single sites should
be using /drush above the docroot.

In order to merge a fix, we are going to have to also fix the test so
it errors in the current (allegedly) buggy state.

gitressa:
Thanks for the feedback. I tried adding `tests/fixtures/sites/default/drush.yml` and a test for the presence of a `drush.yml` file at that location, see attached `add-test-for-default-sites.txt` below.

But when I run the test I get this error back:

```
$ composer unit
There was 1 error:
1) Drush\Config\ConfigLocatorTest::testLoadAll
Undefined index: default
/home/user_name/drush/tests/unit/ConfigLocatorTest.php:38
```
I tried looking for places where the existing `test.home` and `test.site` indexes were defined, with no luck ...

[add-test-for-default-sites.txt](https://github.com/drush-ops/drush/files/4292719/add-test-for-default-sites.txt)

gitressa:
By the way https://github.com/drush-ops/drush/tree/master/tests says that 

> 6. Run all test suites: `composer test`

But I had to run `composer unit` to trigger the `Undefined index: default` error above.

gitressa:
I updated the patch with the failing tests.

src/Config/ConfigLocator.php
```diff
@@ -280,7 +280,7 @@ public function addSitewideConfig($siteRoot)
         // Remember that we've seen this location.
         $this->siteRoots[] = $siteRoot;
 
-        $this->addConfigPaths(self::DRUPAL_CONTEXT, [ dirname($siteRoot) . '/drush', "$siteRoot/drush", "$siteRoot/sites/all/drush" ]);
+        $this->addConfigPaths(self::DRUPAL_CONTEXT, [ dirname($siteRoot) . '/drush', "$siteRoot/drush", "$siteRoot/sites/all/drush", "$siteRoot/sites/default" ]);
```
greg-1-anderson: Still -1 on this. We should not load config from sites/default when some other sites directory is bootstrapped. Use /drush/drush.yml at the project root or Drupal root.

greg-1-anderson:
Closing as a duplicate of #4345, which is closer to what we need.

gitressa:
I am fine with that decision. Still, the way I read the documentation (the original issue #4324) I would expect `sites/default/drush.yml` to work:

> 1.  Drupal site folder (e.g. sites/{default|example.com}/drush.yml).

... so I still think the documentation should be updated, except of course if #4345 adds this behavior.

As a side note, I have just discovered that my main purpose for using `drush.yml`, namely defining `Site URI` can quite easily be done in `.lando.yml`:

```
services:
  appserver:
    overrides:
      environment:
        DRUSH_OPTIONS_URI: "http://mysite.lndo.site"
```

https://docs.lando.dev/config/drupal8.html#default-url-setup

weitzman:
Right. They are taking advantage of env variables that set drush config. Thats documented at https://github.com/drush-ops/drush/blob/master/docs/using-drush-configuration.md

gitressa:
Cool. I was wondering where I could find a list of available variables, without luck. For fun I tried setting the number of column characters (see  [Possible to change COLUMNS environment variable?](https://github.com/lando/lando/issues/1847)) with the below and rebuilding Lando. It seems to get registered, but I still get the crazy long 256 character lines in the terminal:
```
services:
  appserver:
    overrides:
      environment:
        DRUSH_OPTIONS_URI: "http://mysite.lndo.site"
        DRUSH_OPTIONS_COLUMNS: 80

$ lando ssh -s appserver -c env | grep -i col
COLUMNS=256
DRUSH_OPTIONS_COLUMNS=80
```

gitressa:
I got the columns environment working in Lando, see https://github.com/lando/lando/issues/1847#issuecomment-602185785.

PR #4334, opened 2020-02-25
use defaults-extra-file instead of defaults-file

ofrzeta:
use defaults-extra-file for creds instead of defaults-file so mysqldump still reads the global defaults for setting stuff like "set-gtid-purged=OFF"

### DIFF
--- a/src/Sql/SqlMysql.php
+++ b/src/Sql/SqlMysql.php
@@ -31,7 +31,7 @@ password="{$dbSpec['password']}"
 EOT;
 
             $file = drush_save_data_to_temp_file($contents);
-            $parameters['defaults-file'] = $file;
+            $parameters['defaults-extra-file'] = $file;
         } else {
             // User is required. Drupal calls it 'username'. MySQL calls it 'user'.
             $parameters['user'] = $dbSpec['username'];


greg-1-anderson:
Seems okay to me. The only question I have is, what version of mysql was this option first supported? Do we risk breaking old mysql versions? If so, do we care?  I found this exists in mysql 5.6, 5.7 and 8.0.  It looks like Drupal 8 still supports mysql 5.5, so that implies Drush does too. Not sure if this option is available there.

If that's not an issue, then this is an improvement.

ofrzeta:
Thanks for the review. Looks like MySQL 5.5 has the option: http://download.nust.na/pub6/mysql/doc/refman/5.5/en/option-files.html (also found some references to it in the docu of 5.1).

weitzman:
Seems reasonable to me as well but something always breaks when we change mysql config stuff like this. Mulling it.

ofrzeta:
Any progress on this?

greg-1-anderson:
If we're concerned about unlikely edge cases causing a problem, perhaps we could introduce a Drush configuration option to revert to the old directive.

weitzman:
I did some research. We started out with --defaults-extra-file as proposed in this PR. See https://github.com/drush-ops/drush/issues/365. It got changed to --defaults-file in https://github.com/drush-ops/drush/pull/2387. Given the history, I'm uninclined to change it back. Maybe we could make this a config pref.

greg-1-anderson:
OK the history from  #2387 is that some users put sql credentials in their my.cnf file. If we go back to using --defaults-extra-file, then Drush breaks for all of those users, because credentials from the my.cnf file take precedence over credentials provided in --defaults-extra-file.

As disappointing as this is, I think it is unavoidable that we close this PR and look for another solution.

weitzman:
Seems like drush sql:dump --extra-dump="--set-gtid-purged=OFF" would fix this, right? This option can be added to a site's drush config if the typing gets error-prone or tedious.

weitzman:
This approach wont work. Lets continue in #4410

PR #4337, opened 2020-02-26
Run hook_requirements() on pm:enable and throw exception on errors

### LINKED ISSUE #4190: Module installation ignores `hook_requirements`.
**Describe the bug**
When installing a module through `drush en` the result of `hook_requirements` isn't taken into account. We've got a module that sets an requirement error if a certain environment variable is missing, when installing through the Drupal UI installation is correctly blocked, when installing through Drush the module is enabled no matter what. 

**To Reproduce**
Add a module that implements `hook_requirements` and return a requirement with the severity `REQUIREMENT_ERROR`.

**Expected behavior**
That the installation of the module would be blocked.

**Actual behavior**
The module is installed.

**Workaround**
N/A

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.7.1 
| Drupal version? | 8.7.7
| PHP version     | 7.2.20
| OS?             | Linux

davereid:
This resolves #4190 

I cannot call \drupal_check_module() as it doesn't throw exceptions, it just send \Drupal::messenger() error messages, which I couldn't figure how to get back into Drush. So I had to replicate the logic and throw my own exception. This also ensures it runs on any of the additional dependencies being enabled.

### DIFF
--- a/src/Drupal/Commands/pm/PmCommands.php
+++ b/src/Drupal/Commands/pm/PmCommands.php
@@ -71,6 +71,7 @@ class PmCommands extends DrushCommands
      * @command pm:enable
      * @param $modules A comma delimited list of modules.
      * @aliases en,pm-enable
+     * @bootstrap full
      */
     public function enable(array $modules)
     {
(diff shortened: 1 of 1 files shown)


weitzman:
Do we need requirements check on uninstall?

davereid:
No, the only times we need to check hook_requirements() are on install (covered here), runtime (covered already by running core:requirements), and update (covered already by running updatedb), not uninstall as per the [documentation](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Extension!module.api.php/function/hook_requirements/8.x):

> This hook has three closely related uses, determined by the $phase argument:
>
>    Checking installation requirements ($phase == 'install').
>    Checking update requirements ($phase == 'update').
>    Status reporting ($phase == 'runtime').

weitzman:
Thanks. Any chance we can add a test for this? Maybe add it to tests/functional/resources/modules/d8/woot module

markcarver:
Wow... this is a massive regression 😱https://www.drupal.org/project/drush/issues/593314

bserem:
I tested this on my systems and it works as advertised:
```
module_b: You are trying to enable a conflicting module. (Currently using MODULE version A) 
```

Quite ok and on par with what drupal does from the UI.

codebymikey:
Hi @weitzman,

The tests have been implemented in #4733

davereid-pfg:
This is now resolved with https://github.com/drush-ops/drush/pull/4733/files and can be closed.

PR #4342, opened 2020-03-02
Do not throw an exception when valid JSON data evaluates to empty PHP data

### LINKED ISSUE #4341: Exception thrown when JSON data is correct but empty
**Describe the bug**
`OutputUtilsTrait::getOutputFromJSON()` throws an exception when the JSON output is valid but evaluates to empty.

**To Reproduce**
Execute a command that returns `[]` which is valid JSON for a data object that represents an empty array, and call `OutputUtilsTrait::getOutputFromJSON()`.

**Expected behavior**
The method should return an empty array.

**Actual behavior**
```
There was 1 error:
Exception: No json output received.

Output:
[]
```

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 9.x/8.x
| PHP version     | 7.x,5.x

pfrenssen:
Fixes #4341

### DIFF
--- a/src/TestTraits/OutputUtilsTrait.php
+++ b/src/TestTraits/OutputUtilsTrait.php
@@ -154,7 +154,7 @@ trait OutputUtilsTrait
     {
         $output = $this->getOutput();
         $json = json_decode($output, true);
-        if (!$json) {
+        if ($json === NULL) {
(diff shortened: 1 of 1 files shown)


pfrenssen:
There is one case which is not covered here, the string value `null` is also valid JSON and evaluates to NULL. I can harden this by checking if `json_last_error()` contains an error value.

PR #4351, opened 2020-03-16
Allow command discovery to follow symlinks.

### LINKED ISSUE #4094: command discovery doesn't follow symlinks
**Describe the bug**

Drush command files aren't discovered if there are symlinks in the path.

This is important because, AFAIK, the recommended way to develop a Composer library such as a Drush command is to have Composer symlink it into the repo where you run it by declaring your local checkout of the library as a repository, e.g. I have a Drupal project in whose composer.json I have this:

```
        {
            "type": "path",
            "url": "/Users/joachim/Sites/_sandbox/drupal-code-builder-drush-9x"
        }
```

There's a PR that should have fixed this #2915, but I can't find a call to followLinks() anywhere in the drush codebase for drush 9.7.0

 consolidation/annotated-command calls followLinks() on the symfony file finder if it's had that call itself, but by default that's set to false.

**To Reproduce**

Install a Drush command as above.

**Expected behavior**

Symlinks get followed.

**Actual behavior**

They don't.

**Workaround**

Hack CommandFileDiscovery so it calls followLinks() on the symfony file finder.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.7.0
| Drupal version? | 8.x
| PHP version     | 7.3
| OS?             | Mac

**Additional information**

davereid:
Fixes #4094

I attempted to manually list a symlinked directory using my drush.yml file, but even then it doesn't work:
```
drush:
  paths:
    include:
      - '${env.PWD}/Commands/custom/symlinked-directory'
```

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -372,6 +372,7 @@ class Application extends SymfonyApplication implements LoggerAwareInterface, Co
         $discovery
             ->setIncludeFilesAtBase(true)
             ->setSearchDepth(3)
+            ->followLinks()
             ->ignoreNamespacePart('contrib', 'Commands')
             ->ignoreNamespacePart('custom', 'Commands')
             ->ignoreNamespacePart('src')


Leksat:
+1 for this one!

In our case Composer is [configured to install local versions of packages](https://github.com/AmazeeLabs/silverback-mono/blob/a181f906681fd58c6643128b1e3e13f540dfb0de/apps/silverback-drupal/composer.json#L11-L15) from the monorepo. It does it by symlinking. Without this patch commands from local packages cannot be found.

It took me quite some time to figure out what is going on 😅

weitzman:
Diccovery via config will be deprecated soon. Use psr4 discovery instead. See https://www.drush.org/latest/commands/#auto-discovered-commands

Leksat:
@weitzman I already use psr4 discovery. There are just two files in my package:
- `composer.json`
- `Commands.php`

Yet, the discovery does not work because the package is symlinked from the same monorepo.

weitzman:
Commandfiles in a Composer Path repository should be discovered and loaded as of #5455. Please share other use cases if needed. I will reopen this if needed.

PR #4359, opened 2020-03-21
Add new deploy command. Performs site deployment.

weitzman:
This command is intended to standardize how Drupal deployments work. The intent is that your deployment script updates the codebase for the target site and then this command performs the following:

```
drush updatedb --no-cache-clear # Runs updates and post-updates
drush cache:rebuild
drush config:import
drush cache:rebuild
drush deploy:hook
```

- The last step above runs pending [HOOK_deploy_NAME()](https://github.com/drush-ops/drush/blob/master/tests/functional/resources/modules/d8/woot/woot.deploy.php) implementations. This is a new hook introduced by Drush for deployments. Code which used to be in post_update hook should usually move to this hook now. This hook could alternatively be run via a web UI, should anyone care to write that.
- See the [Help topic](https://github.com/drush-ops/drush/blob/master/docs/deploy.md) about this feature.

Todo
--------
- [x] Tests
- [x] Add a topic
- [x] Update/Write Drupal docs about deployment.
- [x] The deploy:hook command still needs to be written.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -61,9 +61,9 @@ install:
   # - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
 
 test_script:
-  - vendor/bin/phpunit --colors=always --configuration tests --testsuite --debug functional
-  - vendor/bin/phpunit --colors=always --configuration tests --testsuite --debug integration
-  - vendor/bin/phpunit --colors=always --configuration tests --testsuite --debug unit
+  - vendor/bin/phpunit --colors=always --configuration tests --testsuite functional --debug
+  - vendor/bin/phpunit --colors=always --configuration tests --testsuite integration --debug
+  - vendor/bin/phpunit --colors=always --configuration tests --testsuite unit --debug
 
 on_finish:
   # Uncomment this and above line to enable RDP into build machine https://www.appveyor.com/docs/how-to/rdp-to-build-worker/
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ install:
   - SET PATH=C:\Program Files\MySql\MySQL Server 5.7\bin\;%PATH%
   - choco search php --exact --all-versions -r
   #Install PHP per https://blog.wyrihaximus.net/2016/11/running-php-unit-tests-on-windows-using-appveyor-and-chocolatey/
-  - ps: appveyor-retry cinst --ignore-checksums -y php --version 7.3.7
+  - ps: appveyor-retry cinst --limit-output --ignore-checksums -y php --version 7.3.7
   - cd c:\tools\php73
   - copy php.ini-production php.ini
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -61,9 +61,9 @@ install:
   # - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
 
 test_script:
-  - vendor/bin/phpunit --colors=always --configuration tests --testsuite functional --debug
-  - vendor/bin/phpunit --colors=always --configuration tests --testsuite integration --debug
-  - vendor/bin/phpunit --colors=always --configuration tests --testsuite unit --debug
+  - vendor/bin/phpunit --colors=always --configuration tests --testsuite --debug functional
+  - vendor/bin/phpunit --colors=always --configuration tests --testsuite --debug integration
+  - vendor/bin/phpunit --colors=always --configuration tests --testsuite --debug unit
 
 on_finish:
   # Uncomment this and above line to enable RDP into build machine https://www.appveyor.com/docs/how-to/rdp-to-build-worker/
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -7,8 +7,7 @@ performs the following:
 ```
 drush updatedb
 drush config:import
-drush deploy:hook
-drush cache:rebuild
+drush cache:rebuilddrush deploy:hook
 ```
 
 # Authoring update functions
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -11,6 +11,15 @@ drush deploy:hook
 drush cache:rebuild
 ```
 
+# Authoring update functions
+Below are the 3 types of update functions run by this command. Choose the most appropriate for your need. 
+
+| Function | Purpose |
+| --- | --- |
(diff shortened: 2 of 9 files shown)


Chi-teck:
I think `drush cache:rebuild` should be first on the list.

fgm:
One thing we do on all D8 projects is have different settings during build/deploy/upgrade (no extra drivers like memcached, mongodb, redis, rabbitmq...) and in production. Supporting this needs the ability to do some extra changes, like switching settings files, and flushing both in build and run modes.

weitzman:
@Chi-teck - Since https://www.drupal.org/project/drupal/issues/3055443, I don't think any early clear should be needed. mass.gov has always cleared plugin cache early on during deploy but I think thats vestigial. I do think cache:rebuild should move up a step, before the deploy:hook.I just pushed that.

@fgm See Drupal link above. Some of that is not needed anymore. Also you can still use conditional logic in settings.php. This command wont be able to handle everything. Please elaborate what you would like to see.

fgm:
In our case, part of the issue is precisely with having two different sets of settings (no logic in settings at this point, to keep them readable): it means the cache flush only works for the currently active settings, so one is needed on each set of settings. The reason for the early flush comes from the fact that the initial state of the build cache reflects the state of the last install step of the previous run, which may (been there, hurt by that) pollute installation. Now, maybe more recent versions are able to install flawlessly in such a situation, but it wasn't the case in the latest D8 projets we started. The most recent still started on 8.3, though, and we didn't try to do it otherwise on maintenance since it worked that way, so /maybe/ it has become a non-issue.

Punk-UnDeaD:
why hook?
why not event?

weitzman:
A hook is consistent with other one time use  registries like hook_update_n and post_update. There is no event in Drupal that works as one time use.

behzad1q:
drush updatedb sometimes fail and need to be run twice especially on large content websites.

weitzman:
@behzad1q. Nothing changes for you in that case. Just re-run the deploy command.

acbramley:
@weitzman one thing to keep in mind is that if there were no db updates run, then caches aren't cleared before config import. We have seen issues with this when (for example) adding configuration that depends on a new plugin. Our deployment process is now:

```
drush updatedb --no-cache-clear
drush cache:rebuild
drush cexy (using drush_cmi_tools, but will soon be replaced by config:import)
drush cache:rebuild
```

weitzman:
Yes, thats an issue. See #4375

claudiu-cristea:
With this feature, updates & post-updates are in fact the same thing. The only difference is the priority (i.e. the updates run first). But the priority was already something determinable inside the `hook_update_N()`. Now a developer, having, not two, but three hooks to implement, is more confused.

pfrenssen:
I'm also a bit intrigued by this change, is there a backstory to this, why it was decided to add a new hook instead of reusing what is already there?

I understand the motivation for this - it is needed to be able to run the updates in the order:

1. Database schema changes (hook_install())
2. Config updates
3. Any updates that affect content and depend on the new config

But it would be sufficient to follow these steps in the deploy command, it doesn't seem to be needed to introduce a new hook_deploy()

```
drush updatedb --no-cache-clear --no-post-updates
drush config:import
drush updatedb
```

weitzman:
Correct. Here is a chat from @bircher 

> I created it because I don't really want to mess with cores update hook system and use it in ways it was not designed. ie run update and post update hooks separately (since in the UI you can't do that) especially with regards to contrib modules (I didn't have that problem yet) but cores post_update functions assume nothing changes dramatically between the update hook and the post update hook

Drupal has no concept of a deployment. It leaves that for 3rd parties. Thats probably a good thing.

Also see the help topic which shows why each hook is unique and deserves to exist https://github.com/drush-ops/drush/blob/master/docs/deploy.md#authoring-update-functions

weitzman:
Also see long chat at https://drupal.slack.com/archives/C45342CDD/p1584547447143000.

leymannx:
I'm wondering if `drush deploy:hook` also triggers `HOOK_deploy_NAME()` hooks of module's that only have been enabled right before via `drush cim`. I have to try that out.

weitzman:
Yes, it should, because the hook running happens in a fresh new subprocess. I'm not sure the use case for that, but it should work.

leymannx:
`hook_install()` cannot be used for content imports in modules because the module cannot rely on default config to be there if they are installed during config import ([#2914213])(https://www.drupal.org/project/drupal/issues/2914213). But then this maybe works with `hook_deploy_NAME()`.

weitzman:
Yes thats a good use case for a given site. Its not a general purpose solution though because it only happens during a drush deployment. You can enable a module with the GUI or with pm:enable and no deploy hook runs.

leymannx:
> Yes thats a good use case for a given site. Its not a general purpose solution though because it only happens during a drush deployment. You can enable a module with the GUI or with pm:enable and no deploy hook runs.

Ah no that must not work. Imagine installing a contrib module for the first time while the module already contains `hook_deploy_NAME` implementations. Then the existing implementations shouldn't be picked up on `drush deploy` or `drush deploy:hook` , only the ones that get added when I update the module later. Same as for `hook_update_N`.

amaisano:
> Then the existing implementations shouldn't be picked up on drush deploy or drush deploy:hook

Are `hook_deploy_NAME` not tracked in the database like `hook_update_N`? That would avoid hitting included deploy hooks from modules. Basically tell the database (key_value schema) that on install of the module, we're already up to date with hook updates N through N+5 AND hook deploys N through N+X.

bircher:
Well I think we should probably update the documentation then.
Deploy hooks are not meant for contrib modules. Contrib modules can not rely on which config is imported when the module is updated as part of a deployment. In a contrib module you can use update hooks and post_update hooks to change the configuration. And if you want to weigh in on configuration synchronisation you can do so with the "new" API that is in core since 8.8 https://www.drupal.org/node/3066005
Deploy hooks are for custom modules so that you can deploy your site in a scripted way with code (hooks) that runs before (post_update) and after (deploy) a config import.

damienmckenna:
For anyone who comes across this page and can't find the hook_deploy help page, it was renamed:

* https://github.com/drush-ops/drush/blob/12.x/docs/deploycommand.md

macmladen:
@weitzman I'm not sure where this question fits here: is there a way to customise **which** commands constitute `deploy`?

I see that by default they are

```
drush updatedb --no-cache-clear # Runs updates and post-updates
drush cache:rebuild
drush config:import
drush cache:rebuild
drush deploy:hook
```

Is there a way to override that in `drush.yml` to, e.g.

```
drush updatedb --no-cache-clear # Runs updates and post-updates
drush cache:rebuild
drush config:import
drush cache:rebuild
drush locale:check
drush locale:update
drush cache:rebuild
drush cron
```

weitzman:
Your best bet is making your your own deploy script, and leaving the deloy command unused.

PR #4362, opened 2020-03-24
Drush 8 - PHP 7.4: Fix deprecation notices

sjerdo:
- Curly brace syntax for accessing array elements and string offsets has been deprecated in PHP 7.4
- Passing the $glue and $pieces parameters in reverse order to implode has been deprecated since PHP 7.4; $glue should be the first parameter and $pieces the second

### DIFF
--- a/commands/core/rsync.core.inc
+++ b/commands/core/rsync.core.inc
@@ -75,7 +75,7 @@ function drush_core_rsync($source, $destination, $additional_options = array()) 
   // Get all of the args and options that appear after the command name.
   $original_args = drush_get_original_cli_args_and_options();
   foreach ($original_args as $original_option) {
-    if ($original_option{0} == '-') {
+    if ($original_option[0] == '-') {
       $options .= ' ' . $original_option;
     }
   }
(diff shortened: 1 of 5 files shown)


weitzman:
Any chance you could please check master branch (i.e. drush 10)?

Ayesh:
@weitzman not the author, but I linted current master and see [no issues](https://github.com/Ayesh/drush/runs/585573485?check_suite_focus=true#step:6:6).

weitzman:
Hopefully @greg-1-anderson can make these changes on Drupal 8.

Thanks for checking Drush master. That link is my first look at GitHub Actions CI. Pretty neat. If they support SSH into builds, I'd consider switching from Circle.

Ayesh:
I'd love to contribute with GitHub actions for this repo if that decision is made! I'm moving from Travis these days, and it's a joy to work with. Unfortunately I don't think GitHub Actions has a feature that lets you login to the containers, as far as I see. It would be pretty awesome feature indeed though.

greg-1-anderson:
The problem with GitHub actions that I saw in the failed tests is that it doesn't seem to allocate a TTY, which is a problem for some of our tests. I don't know if any of our tests actually depend on the TTY; we might be able to get around this just by telling consolidation/process to not allocate one.

greg-1-anderson:
Test failures are unrelated; will need to spend some time cleaning those up in a separate issue.

Thanks for the cleanup here.

PR #4364, opened 2020-03-27
Fix SQL user creation with MariaDB

### LINKED ISSUE #4318: "drush site-install" fails to create the user because of a syntax error
**Describe the bug**
When trying to install Drupal with Drush and `--db-su`, it fails with an SQL syntax error.

**To Reproduce**
```bash
drush site-install config_installer -y --db-url='mysql://drupal_drush_t:drupal_drush_test@localhost/drupal_drush_test' --db-su=*** --db-su-pw='***'
```

**Expected behavior**
It creates the user, the database and then installs the site.

**Actual behavior**
Creating the user fails with this error:
```
 [info] sql:query: DROP DATABASE IF EXISTS drupal_drush_test; CREATE DATABASE drupal_drush_test /*!40100 DEFAULT CHARACTER SET utf8 */; DROP USER IF EXISTS 'drupal_drush_t'@'localhost'; CREATE USER 'drupal_drush_t'@'localhost' IDENTIFIED WITH mysql_native_password BY 'drupal_drush_test'; GRANT ALL PRIVILEGES ON drupal_drush_test.* TO 'drupal_drush_t'@'localhost'; FLUSH PRIVILEGES; [0.29 sec, 11.93 MB]
 [info] Executing: mysql --defaults-file=/tmp/drush_ZHdBsY --database=information_schema --host=localhost --silent -A < /tmp/drush_cGzB8y [0.29 sec, 11.94 MB]
 [warning] Failed to drop or create the database. Do it yourself before installing. ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'BY 'drupal_drush_test'' at line 1
```

**Workaround**
Create the user manually beforehand.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.2.1
| Drupal version? | 8.8.1
| PHP version     | 7.1.33
| OS?             | Linux

**Additional information**
We are using MariaDB 10.1.44.

It seems this syntax was introduced in https://github.com/drush-ops/drush/pull/4262.

prudloff-insite:
Fixes #4318

Tested with Mysql 5.6, Mysql 8.0 and MariaDB 10.1.

### DIFF
--- a/src/Sql/SqlMysql.php
+++ b/src/Sql/SqlMysql.php
@@ -106,7 +106,8 @@ EOT;
             $domain = ($dbSpec['host'] == 'localhost') ? 'localhost' : '%';
             $user = sprintf("'%s'@'%s'", $dbSpec['username'], $domain);
             $sql[] = sprintf("DROP USER IF EXISTS %s;", $user);
-            $sql[] = sprintf("CREATE USER %s IDENTIFIED WITH mysql_native_password BY '%s';", $user, $dbSpec['password']);
+            $sql[] = sprintf("CREATE USER %s IDENTIFIED WITH mysql_native_password;", $user);
+            $sql[] = sprintf("SET PASSWORD FOR %s = PASSWORD('%s');", $user, $dbSpec['password']);
             $sql[] = sprintf('GRANT ALL PRIVILEGES ON %s.* TO %s;', $dbname, $user);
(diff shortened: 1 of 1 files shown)


weitzman:
OK, lets go with this. Thanks for testing so broadly.

PR #4365, opened 2020-03-29
Add deploy:hook command for hook_deploy_NAME "update hooks"

bircher:
The deploy hook for #4359

### DIFF
--- a/src/Drupal/Commands/core/DeployHookCommands.php
+++ b/src/Drupal/Commands/core/DeployHookCommands.php
@@ -1,0 +1,185 @@
+<?php
+
+namespace Drush\Drupal\Commands\core;
+
+use Consolidation\Log\ConsoleLogLevel;
+use Consolidation\OutputFormatters\StructuredData\RowsOfFields;
+use Consolidation\SiteAlias\SiteAliasManagerAwareTrait;
+use Consolidation\SiteAlias\SiteAliasManagerAwareInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\Core\KeyValueStore\KeyValueFactoryInterface;
+use Drupal\Core\Update\UpdateRegistry;
+use Drupal\Core\Utility\Error;
+use Drush\Commands\DrushCommands;
+use Drush\Exceptions\UserAbortException;
+use Psr\Log\LogLevel;
+
+class DeployHookCommands extends DrushCommands implements SiteAliasManagerAwareInterface
+{
+    use SiteAliasManagerAwareTrait;
+
+    /**
+     * @var \Drupal\Core\Update\UpdateRegistry
+     */
+    protected $registry;
+
+    /**
+     * @var \Drupal\Core\KeyValueStore\KeyValueStoreInterface
+     */
+    protected $keyValue;
+
+    /**
+     * DeployHookCommands constructor.
+     *
+     * @param string $root
+     *   The app root.
+     * @param string $site_path
+     *   The site path.
+     * @param \Drupal\Core\Extension\ModuleHandlerInterface $moduleHandler
+     *   The module handler.
+     * @param \Drupal\Core\KeyValueStore\KeyValueFactoryInterface $keyValueFactory
+     *   The key value store factory.
+     */
+    public function __construct($root, $site_path, ModuleHandlerInterface $moduleHandler, KeyValueFactoryInterface $keyValueFactory)
+    {
+        parent::__construct();
+        $this->keyValue = $keyValueFactory->get('deploy_hook');
+        $this->registry = new class(
+            $root,
+            $site_path,
+            array_keys($moduleHandler->getModuleList()),
+            $this->keyValue
+        ) extends UpdateRegistry {
+            public function setUpdateType($type)
+            {
+                $this->updateType = $type;
+            }
+        };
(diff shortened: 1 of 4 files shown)


src/Drupal/Commands/core/DeployHookCommands.php
```diff
@@ -0,0 +1,185 @@
...
+                $this->updateType = $type;
+            }
+        };
+        $this->registry->setUpdateType('deploy');
+    }
+
+    /**
+     * Prints information about pending deploy update hooks.
+     *
+     * @usage deploy:hook-info
```
weitzman: How about hook-status instead of hook-info? Thats consistent with updatedb:status and pm-updatestatus (drush8)

src/Drupal/Commands/core/DeployHookCommands.php
```diff
@@ -0,0 +1,185 @@
...
+        $level = $success ? ConsoleLogLevel::SUCCESS : LogLevel::ERROR;
+        $this->logger()->log($level, dt('Finished performing deploy hooks.'));
+        return $success ? self::EXIT_SUCCESS : self::EXIT_FAILURE;
+    }
+
+    /**
+     * Marks a deploy update hook as not having run.
+     *
+     * During development one often wants to re-run the hook, so this helps
+     * re-setting it so that it can be run again.
```
weitzman:
```suggestion
     * resetting it so that it can be run again.
```

src/Drupal/Commands/core/DeployHookCommands.php
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * @usage deploy:hook-reset mymodule_deploy_runagain
+     *   Unregisters that a deploy hook has run so that it runs again.
+     *
+     * @param string $hook
+     *   The hook name to reset.
+     *
+     * @command deploy:hook-reset
```
weitzman: How hard would it to provide a reset for hook_update_n and post_update? I think we should do them all or do none.

tests/functional/DeployHookTest.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Unish;
+
+use Webmozart\PathUtil\Path;
+
+/**
+ *  @group slow
+ *  @group commands
+ */
+class DeployHookTest extends CommandUnishTestCase
```
weitzman: Thanks for writing a test! We prefer to extend UnishIntegrationTestCase when possible. Those tests are faster. If you run into problems, then this is fine.

tests/functional/resources/modules/d8/woot/woot.deploy.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+/**
+ * Successful deploy hook.
```
weitzman: Since this is our only docs of a new hook, lets add doxygen in all functions - `Implements HOOK_deploy_NAME`

tests/functional/resources/modules/d8/woot/woot.deploy.php
```diff
@@ -0,0 +1,31 @@
...
+    return t('This is the update message from woot_deploy_a');
+}
+
+/**
+ * Successful batched deploy hook.
+ */
+function woot_deploy_batch(array &$sandbox)
+{
+    module_load_install('woot');
+    return woot_update_8105($sandbox);
```
weitzman: This is a pretty weird example - running an update function. Any chance we use other code?

src/Drupal/Commands/core/DeployHookCommands.php
bircher: makes sense

src/Drupal/Commands/core/DeployHookCommands.php
bircher: It would not be hard, the code for post_update hooks would look almost the same (with a different key-value store) the hook_update_n would look like `drupal_set_installed_schema_version("woot", 8103)`.

We can also skip it.. I just found it to be convenient to use while working on the deploy hook, it would indeed also be convenient for the other hooks. So maybe we remove it here and then add it in a second step for all?

tests/functional/resources/modules/d8/woot/woot.deploy.php
bircher: hmm but as for post_update hooks the comment is used in the display of the pending update functions. I don't mind either way..

tests/functional/resources/modules/d8/woot/woot.deploy.php
bircher: yes sure, I just didn't want to copy the code but of course we can.

tests/functional/DeployHookTest.php
bircher: All tests that use the woot module use CommandUnishTestCase, I could not easily transform the test to UnishIntegrationTestCase.

src/Drupal/Commands/core/DeployHookCommands.php
bircher: I removed it for now. we can add a new PR later

tests/functional/DeployHookTest.php
weitzman: There is one such Integration case - \Unish\AnnotatedCommandCase. Lets explore using Integration in a future follow-up.

PR #4368, opened 2020-03-31
Don't use the UpdateKernel for config:import.

### LINKED ISSUE #4350: drush cim -> Configuration synchronisation that both enables & configures a module fails
**Describe the bug**
I noticed something strange today with configuration synchronisation. When we deploy configuration that contains both the enabling of a (contrib/core) module & the config from it it doesn’t seem to work. The module appears to get enabled, and the configuration gets imported, but in the end the module really isn’t enabled, although the status in the modules overview screen shows it is enabled. Uninstalling & re-enabling the module & re-importing the config fixes this, but it’s strange as it happened to me twice today.

Encountered on Drupal Core 8.8.3

**To Reproduce**
I encountered the issue twice today on 2 different sites for 2 different modules (jsonapi_extras / media_library), below are the 2 scenarios written out.

**jsonapi_extras**
- locally enabled the module (drush en)
- locally configured some overrides
- locally exported the configuration (drush cex -y)
- committed and merged on our test branch that triggers a build & deploy
- on the server the configuration is imported (should enable & configure the module)
- config import done & cache is cleared --> no errors
- checked the json:api confiugration --> FAIL: No json:api extras tab
- checked the module overview --> module is enabled --> not true...
- on the server drush pm:uninstall the module
- on the server drush en the module
- on the server drush cim -y (lists all the changes again for the module)
- now it works fine

**media_library**
- locally enabled the module (drush en)
- locally configured some fields to use the media widget
- locally exported the configuration (drush cex -y)
- committed and merged on our test branch that triggers a build & deploy
- on the server the configuration is imported (should enable & configure the module)
- config import done & cache is cleared --> no errors
- checked the field configuration --> FAIL: The fields don't have the media library widget selected, nor is it available
- checked the module overview --> module is enabled --> not true...
- on the server drush pm:uninstall the module
- on the server drush en the module
- on the server drush cim -y (lists all the changes again for the module)
- now it works fine

**Expected behavior**
Configuration should enable the module properly and then import the config.

**Actual behavior**
Module is fakely enabled and the config is imported but doesn't take effect.

**Workaround**
No

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.2
| Drupal version? | 8.8.3
| PHP version     | 7.4 & 7.3
| OS?             | Mac (local) & Linux (sever) & Docker (local)

Full drush status from my local
```
 Drupal version   : 8.8.3
 Site URI         : http://default
 DB driver        : mysql
 DB hostname      : database
 DB port          : 3306
 DB username      : drupal8
 DB name          : drupal8
 Database         : Connected
 Drupal bootstrap : Successful
 Default theme    : total_theme
 Admin theme      : seven
 PHP binary       : /usr/local/bin/php
 PHP config       :
 PHP OS           : Linux
 Drush script     : /app/vendor/drush/drush/drush
 Drush version    : 10.2.0
 Drush temp       : /tmp
 Drush configs    : /app/vendor/drush/drush/drush.yml
                    /app/drush/drush.yml
 Install profile  : standard
 Drupal root      : /app/web
 Site path        : sites/default
 Files, Public    : sites/default/files
 Files, Private   : sites/default/files/private
 Files, Temp      : /tmp
```

**Additional information**
See the issue as well on DO https://www.drupal.org/project/drupal/issues/3119373

bartlangelaan:
This removes the usage of UpdateKernel from the `config:import` command and thereby fixes #4350.

I expected to see an error in 'testConfigImport', but all tests pass (after fixing another minor test issue).

It also looks like the config validation event was not invoked anymore - that is now also fixed and being checked on in the test.

To showcase that this test would be failing without this change, please see #4369 .

### DIFF
--- a/src/Commands/config/ConfigImportCommands.php
+++ b/src/Commands/config/ConfigImportCommands.php
@@ -37,7 +37,6 @@ class ConfigImportCommands extends DrushCommands
      * @option source An arbitrary directory that holds the configuration files. An alternative to label argument
      * @option partial Allows for partial config imports from the source directory. Only updates and new configs will be processed with this flag (missing configs will not be deleted). No config transformation happens.
      * @aliases cim,config-import
-     * @kernel update
      * @bootstrap full
      */
     public function import($label = null, $options = ['preview' => 'list', 'source' => self::REQ, 'partial' => false, 'diff' => false])
--- a/tests/functional/ConfigTest.php
+++ b/tests/functional/ConfigTest.php
@@ -137,7 +137,17 @@ YAML_FRAGMENT;
         $extension['module'] = module_config_sort($extension['module']);
         file_put_contents($extensionFile, Yaml::encode($extension));
 
+        // When importing config, the 'woot' module should warn about a validation error.
+        $this->drush('config:import', [], [], null, null, CommandUnishTestCase::EXIT_ERROR);
+        $this->assertContains("woot config error", $this->getErrorOutput(), 'Woot returned an expected config validation error.');
+
+        // Now we disable the error, and retry the config import.
+        $this->drush('state:set', ['woot.shoud_not_fail_on_cim', 'true']);
         $this->drush('config:import');
+
+        // We make sure that the service inside the newly enabled module exists now.
+        $this->drush('php:eval', ['var_dump(\Drupal::hasService("drush_empty_module.service"));']);
+        $this->assertOutputEquals('bool(true)');
     }
 
     protected function getConfigSyncDir()
--- a/tests/functional/ConfigTest.php
+++ b/tests/functional/ConfigTest.php
@@ -145,9 +145,9 @@ YAML_FRAGMENT;
         $this->drush('state:set', ['woot.shoud_not_fail_on_cim', 'true']);
         $this->drush('config:import');
 
-        // We make sure that the service inside the newly enabled module exists now.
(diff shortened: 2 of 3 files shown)


weitzman:
LGTM. Lets give @pfrenssen a bit of time to comment.

@alexpott suggested that config import test(s) should check `\Drupal::container()->getParameter(‘container.modules’)` in order to assure that the module is really enabled. Any chance you can add that @bartlangelaan?

weitzman:
After this gets in, we can undo some of the complexity introduced in https://github.com/drush-ops/drush/pull/4217. That introduced a second file called ConfigImportCommands. I think we can go back to one now.

pfrenssen:
Now this is completely unexpected, but fantastic news. It appears that the problem that necessitated the usage of the `UpdateKernel` before has been resolved in the past 6 months. The test coverage that was proving the bug before is still passing. Tests are running on 8.8.x, 8.9.x and 9.0.x so all current core versions are working.

I agree with @weitzman that it would be nice to check the container if the modules are correctly enabled, but this is not a hard blocker for me, more like a nice bonus. If @bartlangelaan has availability for this in the next days this would be great to have.

In #4217 we moved out the `config:import` command into a separate Command file in order to make it possible to use the DrupalKernel. This is no longer needed and we can move it back into its former home. I will make a followup to tackle this once this is in.

Big thanks @bartlangelaan for tackling this problem!

PR #4369, opened 2020-03-31
Add failing test case for config import with UpdateKernel.

bartlangelaan:
Do not merge.

This is just to show that master fails on #4350 if this check is added, but doesn't fail on #4368.

### DIFF
--- a/tests/functional/ConfigTest.php
+++ b/tests/functional/ConfigTest.php
@@ -138,6 +138,10 @@ YAML_FRAGMENT;
         file_put_contents($extensionFile, Yaml::encode($extension));
 
         $this->drush('config:import');
+
+        // We make sure that the service inside the newly enabled module exists now. A fatal
+        // error will be thrown by Drupal if the service does not exist.
+        $this->drush('php:eval', ['Drupal::service("drush_empty_module.service");']);
     }
 
     protected function getConfigSyncDir()


BramDriesen:
I don't think this is correct. Services need to be updated by an update hook, not by drush cim magic as discussed here https://github.com/drush-ops/drush/issues/4350#issuecomment-606656560

Maybe have a look here at the test only patch that was added to Drupal core to check this on cache clear. https://www.drupal.org/project/drupal/issues/3119373#comment-13511113

bartlangelaan:
The service is not "updated", nor am I trying to update a service.

This check just makes sure that a Drupal service called `drush_empty_module.service` exists. If this service doesn't exist, the test fails. If this service exists, the test passes.

This service is part of the `drush_empty_module` module, as defined here:
https://github.com/drush-ops/drush/blob/cd6e21c11264c6946bfd4d8da285dea697206b76/sut/modules/unish/drush_empty_module/drush_empty_module.services.yml#L2

So, if the `drush_empty_module` is enabled, this service should exist. If the module is disabled, it should not.

The `config:ignore` command just before this added line _should_ enable the `drush_empty_module`. The command completes succesfully - it looks like the module is enabled. But it's not. If the module was really enabled, the service would exist.

If there is another (easy) way to check if a module is enabled and correctly working, please let me know. I'm not familiar with the Drush codebase.

BramDriesen:
aha yes that's right :-) if the service get's added by enabling the module it should indeed work! it's only when adding a new service to an existing & enabled module.

PR #4374, opened 2020-04-01
Fix #4371. Reunite ConfigImportCommands classes. They no longer need the UpdateKernel

### LINKED ISSUE #4371: Reunite ConfigImportCommands classes now that they no longer need the UpdateKernel
**Describe the task**
In #4217 the `config:import` command was modified to use the `UpdateKernel` in order to work around fatal errors when a new module is being enabled that contains a service that is depended on by existing modules. In order to facilitate this the `::import()` and `::validate()` methods have been factored out into a new `\Drush\Commands\config\ConfigImportCommands` class that would allow us to boot using this kernel.

In #4368 this change has been reverted because it was causing other problems and the original core bug has been resolved in the meantime.

We now no longer need the `ConfigImportCommands` as a separate class. Let's move `::import()` and `::validate()` back into `\Drush\Drupal\Commands\config\ConfigImportCommands` so they are reunited with their siblings.

weitzman:
Also tests `Drupal::getContainer()->getParameter('container.modules')`


tests/functional/ConfigTest.php
```diff
@@ -144,6 +144,12 @@ class: Drupal\woot\DependingService
         // Now we disable the error, and retry the config import.
         $this->drush('state:set', ['woot.shoud_not_fail_on_cim', 'true']);
         $this->drush('config:import');
+        $this->drush('php:eval', ["return Drupal::getContainer()->getParameter('container.modules')"], ['format' => 'json']);
+
+        // Assure that new modules are fully enabled.
+        $out = $this->getOutputFromJSON();
+        $this->assertArrayHasKey('woot', $out);
+        $this->assertArrayHasKey('drush_empty_module', $out);
```
pfrenssen, line 152: Thanks for adding this test! :+1:

PR #4381, opened 2020-04-07
Document how to rerun watchdog:show every 2 seconds.

weitzman:
Fixes #3523 (in a more generic way).

### DIFF
--- a/src/Drupal/Commands/core/WatchdogCommands.php
+++ b/src/Drupal/Commands/core/WatchdogCommands.php
@@ -24,7 +24,7 @@ class WatchdogCommands extends DrushCommands
      * @option severity Restrict to messages of a given severity level.
      * @option type Restrict to messages of a given type.
      * @option extended Return extended information about each message.
-     * @usage  drush watchdog-show
+     * @usage  drush watchdog:show
      *   Show a listing of most recent 10 messages.
      * @usage drush watchdog:show "cron run succesful"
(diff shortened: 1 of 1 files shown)


cspitzlay:
The merged example does not work. It is missing a "do"

weitzman:
Fixed. Thanks.

src/Drupal/Commands/core/WatchdogCommands.php
```diff
@@ -34,6 +34,8 @@ class WatchdogCommands extends DrushCommands
      *   Show a listing of most recent 10 messages with a severity of notice.
      * @usage drush watchdog:show --type=php
      *   Show a listing of most recent 10 messages of type php
+     * @usage  while sleep 2; drush watchdog:show; done
+     *   Every 2 seconds, show the most recent 10 messages.
```
amaisano, line 38: Doesn't validate or work (mac osx):

image

weitzman: That was pointed out above and then fixed https://github.com/drush-ops/drush/commit/a9840accc5554d2e628a323ad179e94d2555694e

amaisano: Ah sorry! Google pointing me to old posts.

dahousecat:
`watch -n2 drush ws` works better for me as doesn't show duplicated rows.

Issue #4382, opened 2020-04-08
rsync with exclude-paths inside site aliases not working


jcmartinez:
**Describe the bug**
In previous versions I used to define rsync exclude-paths inside the site aliases file. I'm trying to do the same but it doesn't work.

**To Reproduce**
According to https://github.com/drush-ops/drush/blob/master/examples/example.site.yml
```
#   command:
#     sql:
#       sync:
#         options:
#           no-dump: true
```
Following this logic, I have configured my site alias files with:

```
  command:
    rsync:
      mode: rlptz
      exclude-paths: 'css:imagecache:ctools:js:tmp:php:styles:'
```
I have also tried:
```
  command:
    rsync:
      mode: rlptz
      exclude-paths:
        - 'css:'
        - 'imagecache:'
        - 'ctools:'
        - 'js:'
        - 'tmp:'
        - 'php:'
        - 'styles:'
```

**Expected behavior**
I expected the mode and the exclude-paths to be used.

**Actual behavior**
Neither the mode or the exclude paths are picked up.

**Workaround**
I have been able to exclude the paths by typing them directly on the command line:
```
drush -v --exclude-paths=:css:imagecache:ctools:js:tmp:php:styles: rsync @drushaliases.sitename.com:%files/ @drushaliases.sitename.dev:%files
```
but this is very inconvenient because I have to remember to type them always.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.7.2
| Drupal version? | 8.8.5
| PHP version     | 7.2.29
| OS?             | Linux

weitzman:
try:

```
command:
   core: 
    rsync:
      mode: rlptz
      exclude-paths: 'css:imagecache:ctools:js:tmp:php:styles:'
```

jcmartinez:
Thanks @weitzman, for pointing me on the direction:.

This is how it actually works:

```
  command:
    core:
      rsync:
        options:
          mode: rlptz
          exclude-paths: 'css:imagecache:ctools:js:tmp:php:styles:'
```
Happy to create a PR to add this snippet to the documentation.

PR #4383, opened 2020-04-08
#4382: Documenting rsync exclude-paths in site aliases

jcmartinez:
Only adds a snippet of code to an example file.

### DIFF
--- a/examples/example.site.yml
+++ b/examples/example.site.yml
@@ -293,6 +293,14 @@
 
 # Example of rsync with exclude-paths
 #
+# Note that most options typically passed to rsync via `drush rsync` are
+# "passthrough options", which is to say they appear after the `--` separator
+# on the command line. Passthrough options are actually arguments, and
+# it is not possible to set default arguments in an alias record. The
+# `drush rsync` command does support two options, `--mode` and `--exclude-paths`,
+# which are interpreted directly by Drush. Default values for these options
+# may be specified in an alias record, as shown below.
+#
 # @code
 # dev:
 #   root: /path/to/docroot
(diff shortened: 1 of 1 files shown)


weitzman:
I'm not sure what this adds beyond the sql:sync example above it. Clearly we cant add examples for all commands.

jcmartinez:
Yes, you're right, it only adds a sql:sync example with the use of excluded paths.
I'd like to point out that following the existing example, it was impossible for me to get it right without coming here and asking. Sometimes, because we know the way that something works, it seems obvious to us, but not necessarily to everybody.
I'm of the opinion that it doesn't hurt to have open source projects abundantly documented.

weitzman:
Right. How can we expand the sql:sync example so it might have pointed you
in the right direction? We can add comments in the yml.

greg-1-anderson:
I think the thing about rsync that's confusing is that you can't add arguments to aliases, only options. Most rsync "options" are considered arguments to Drush (as they appear after the `--`), but `exclude-paths` is different, as it is processed directly by Drush, and ergo is an option, and ergo can be included in an alias. Same with `mode`.

sql:sync, on the other hand, is a native Drush command, and does not have this issue.

I am +1 on including this example with a comment on passthru options (arguments as options), because rsync is unusual in this respect. Feel free to close again if you still feel it is superfluous, though.

weitzman:
OK, lets add that comment about passthru options and merge,

greg-1-anderson:
Will do.

greg-1-anderson:
My explanation was a bit wordy, but there it is.

PR #4388, opened 2020-04-13
Replace deprecated core constant

Sutharsan:
The `locale:update` command uses core's deprecated constant `REQUEST_TIME`. This should be replaced.


src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -143,7 +143,7 @@ public function update($options = ['langcodes' => self::REQ])
         // translation updates. If the status is expired we clear it an run a batch to
         // update the status and then fetch the translation updates.
         $last_checked = $this->getState()->get('locale.translation_last_checked');
-        if ($last_checked < REQUEST_TIME - LOCALE_TRANSLATION_STATUS_TTL) {
+        if ($last_checked < $_SERVER['REQUEST_TIME'] - LOCALE_TRANSLATION_STATUS_TTL) {
```
greg-1-anderson, line 146: Although `$_SERVER['REQUEST_TIME']` is invariantly filled in by php, perhaps `time()` would be clearer, and sufficient for this use case?

Sutharsan: I did choose `$_SERVER['REQUEST_TIME']` because it is closest to the current, but both will do fine.

PR #4389, opened 2020-04-13
pm:enable option to skip translations

Sutharsan:
There are situations when a developer does not want to download and import interface translation files while enabling a module. For example:

* When enabling large number of modules
* When enabling a module to try it out. (Translation import can not be reverted)
* When working on a limited bandwith connection, e.g. while traveling

Usage example:

```
drush en devel --skip-translations
```

This PR replaces PR #3531 that could not be continued because I lost the repo.


Chi-teck:
+1
Imported translations in Drupal are never get deleted.

PR #4391, opened 2020-04-16
Allow passing config-dir for clean install

### LINKED ISSUE #4326: Specify config directory on site install
Am trying to install site from config, where settings.php does not exist, so I need to give config path in install command, on newer versions this parameter is no longer available, would be great if you could add or give default path like `config/sync` or something

JorisVanEijden:
A drush site:install with `--existing-config` option fails if there is no settings.php yet since in that case `drush_config_get_config_directory()` throws an exception with the rather unhelpful message "The configuration directory type 'sync' does not exist"

This solution takes the passed in `config-dir` option instead, allowing a completely fresh install of drupal from existing config.

Fixes #4326

### DIFF
--- a/src/Commands/core/SiteInstallCommands.php
+++ b/src/Commands/core/SiteInstallCommands.php
@@ -83,7 +83,7 @@ class SiteInstallCommands extends DrushCommands implements SiteAliasManagerAware
 
         // Was giving error during validate() so its here for now.
         if ($options['existing-config']) {
-            $existing_config_dir = drush_config_get_config_directory();
+            $existing_config_dir = $options['config-dir'] ?? drush_config_get_config_directory();
             if (!is_dir($existing_config_dir)) {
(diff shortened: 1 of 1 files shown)


JorisVanEijden:
In truth this PR changes no functionality at all. It simply prevents an exception being thrown by `config_get_config_directory()` due to the absence of a 'config_sync_directory' in the \Drupal\Core\Site\Settings at that point.

The `$existing_config_dir` variable is _only_ used to check if the directory exists and not for anything else.

In effect this PR fixes a tiny logic bug deep inside the broad pre-existing config install usecase.
If we want to implement proper and documented support for all the different scenarios involved in this usecase then I think quite a bit more work is required.
If there is a need for that I am willing to put in some work, but I'm not that well versed in the internals, history or roadmap of the Drush project.
What do you think?

PR #4023, opened 2019-03-31
Backported pg_dump passwordfile fix for drush 8.

kalpaitch:
Linked to #4022 

Not expecting this to be merged, but having the pr allows me to apply the patch for now.

Issue #4022, opened 2019-03-29
PGPASSFILE not getting passed to sql-dump


kalpaitch:
I am running drush sql-dump on a postgres 9.5 instance, and it's still asking for a password prompt.

```
Running: drush @xxxxx.xxxxx sql-dump --result-file='/tmp/db-dump-test-unsanitized.sql' --extra='-O -x' --verbose
Loaded alias @xxxxx.xxxxx from file /var/www/xxxxx/web/sites/all/drush/xxxxx.aliases.drushrc.php    [notice]
Executing: PGPASSFILE=/tmp/drush_bZy0Wm psql -q --dbname=par --host=localhost --port=5432 --username=xxxxx  --no-align --field-separator="	" --pset tuples_only=on --file /tmp/drush_2z1SbB
Calling system((pg_dump  par --host=localhost --port=5432 --username=par -O -x --clean && pg_dump --schema-only --table=flood --table=sessions --table=watchdog xxxx --host=localhost --port=5432 --username=xxxxx -O -x --clean) > /tmp/db-dump-test-unsanitized.sql);
Password: 
```

I was expecting drupal to pass the pwd to postgres by writing a temporary PGPASSFILE, which it seems to do for the psql command, but perhaps it's not passing that along to the pg_dump command??

### System Configuration
Drush 8.18
Drupal 8.6
PHP 7.1

weitzman:
Fixed in 9.6

kalpaitch:
See https://github.com/drush-ops/drush/issues/3330#issuecomment-478880276

jcnventura:
This needs a backport to Drush 8, please.

PR #4394, opened 2020-04-23
Set PGPASSFILE also in the dump command for PostgreSQL (#4022)

jcnventura:
The problem remains that drush sql-dump doesn't work on PostgreSQL on Drush 8.x, which we must use for Drupal 7. The solution, as fixed in drush 9+, is to set the PGPASSFILE for this command as well.

Taking some inspiration from #4023, but I ended up looking mostly at drush 9.x's version of Sqlpgsql.php to recreate the backport.


jcnventura:
It makes no sense that thes tests would fail for some versions of Drupal 8, but not for others. Can you rerun the Travis CI tests @weitzman ?

weitzman:
@greg-1-anderson is maintaining Drush 8. He could use some help getting tests to pass, as there are known failures.

jcnventura:
Amended the initial commit, as we also need to pass the PGPASSFILE variable before the "pg_dump --schema-only" command. #4023 was right about that change.

PR #4401, opened 2020-04-28
Note problems with having src in command path

neclimdul:
As discussed in Slack, including src in command path will break discovery and cause other(unclear) problems. :)


neclimdul:
browser edit seems to have touched the end of the file... cloning to touch up locally.

PR #4403, opened 2020-04-29
Fix #4402 by removing deprecated use of Process::inheritEnvironmentVariables

### LINKED ISSUE #4402: Deprecated use of Process::inheritEnvironmentVariables
**Describe the bug**
CliTestTrait uses deprecated Symfony  3.3 code. See https://www.drupal.org/pift-ci-job/1667378

**To Reproduce**
Run a test that leverages CliTestTrait.

**Expected behavior**
No deprecation warnings

**Actual behavior**
Deprecation warnings.

**Workaround**

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 9.x/8.x
| PHP version     | 7.x

**Additional information**

heddn:
Here's the message from Symfony's code

```php
        if (!$inheritEnv) {
            @trigger_error('Not inheriting environment variables is deprecated since Symfony 3.3 and will always happen in 4.0. Set "Process::inheritEnvironmentVariables()" to true instead.', E_USER_DEPRECATED);
        }
```

### DIFF
--- a/src/SiteAlias/ProcessManager.php
+++ b/src/SiteAlias/ProcessManager.php
@@ -153,7 +153,9 @@ class ProcessManager extends ConsolidationProcessManager
         // Handle BC method of making env variables inherited. The default in
         // later versions is always inherit and this method disappears.
         if (method_exists($process, 'inheritEnvironmentVariables')) {
+            set_error_handler();
             $process->inheritEnvironmentVariables();
+            restore_error_handler();
         }
         $process->setLogger(Drush::logger());
         $process->setRealtimeOutput(new DrushStyle(Drush::input(), Drush::output()));
--- a/src/SiteAlias/ProcessManager.php
+++ b/src/SiteAlias/ProcessManager.php
@@ -153,7 +153,7 @@ class ProcessManager extends ConsolidationProcessManager
         // Handle BC method of making env variables inherited. The default in
         // later versions is always inherit and this method disappears.
         if (method_exists($process, 'inheritEnvironmentVariables')) {
-            set_error_handler();
+            set_error_handler(null);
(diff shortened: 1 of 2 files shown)


weitzman:
Was the default ever false, and if so could a D8 site still be using that symfony version?

heddn:
It doesn't exist in https://github.com/symfony/process/blob/3.1/Process.php. It seems it was added only in 3.2 under https://github.com/symfony/process/commit/6be2b290df0f9001ca1b944d7f1f3b6619097d3d. And the default seems to be TRUE. Drush doesn't ever seem to use anything but the default, from what I can tell.

https://github.com/symfony/process/commit/a8c0522f1f99d9c2542fe282c51e5331b7ca1e9f seems to have immediately deprecated its use before a stable 3.3 release.

weitzman:
I think the error message is due to a missing HOME env variable at some point. That is, its causes by this PR. Not sure of it, but enough that I'm not inclined to merge until we are green.

heddn:
I tend to agree this change does break things too. I'm not sure how we can make the code support symfony 3.4|4.4|5.0 compatible. It will still work on 4.4 (with deprecations). but come 5.0, we'd be calling a missing method.

heddn:
We should probably add a TODO to remove this once we no longer support symfony 3. Not sure how those are denoted on this project.

weitzman:
Code looks good. A simple @todo code comment would suffice.

claudiu-cristea:
@weitzman I've moved the work in #4726 and added there the requested `@todo`s

weitzman:
Fixed in #4726

Issue #4366, opened 2020-03-30
pm:security-php only suggests a composer command for the first package


prudloff-insite:
**Describe the bug**
I am trying https://github.com/drush-ops/drush/commit/426ebed9c9ad2f1e50c46e244ae345cc9a38fb89 and it works great but it only suggests a `composer why` command for the first package.

**To Reproduce**
```bash
composer require spoonity/tcpdf 6.2.1
composer require zetacomponents/mail 1.8.1
drush pm:security-php
```

**Expected behavior**
Suggest multiple commands to run.

**Actual behavior**
```
 [warning] One or more of your dependencies has an outstanding security update.
 [notice] Run composer why spoonity/tcpdf to learn what  module requires the package.
spoonity/tcpdf:
  version: 6.2.1
  advisories:
    -
      title: 'Attackers can trigger deserialization of arbitrary data via the phar:// wrapper.'
      link: 'https://github.com/tecnickcom/TCPDF/commit/1861e33fe05f653b67d070f7c106463e7a5c26ed'
      cve: CVE-2018-17057
zetacomponents/mail:
  version: 1.8.1
  advisories:
    -
      title: 'Arbitrary code execution via a crafted email address'
      link: 'https://github.com/zetacomponents/Mail/issues/58'
      cve: CVE-2017-15806
```

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.2.2
| Drupal version? | 8.8.4
| PHP version     | 7.1.33
| OS?             | Linux

PR #4406, opened 2020-05-04
Display "composer why" suggestion for every package

prudloff-insite:
Fixes https://github.com/drush-ops/drush/issues/4366.

Unfortunately, it does not seem possible to use a single `composer why` call for multiple packages.

### DIFF
--- a/src/Commands/pm/SecurityUpdateCommands.php
+++ b/src/Commands/pm/SecurityUpdateCommands.php
@@ -177,7 +177,7 @@ class SecurityUpdateCommands extends DrushCommands
         $process->mustRun();
         $out = $process->getOutput();
         if ($packages = json_decode($out, true)) {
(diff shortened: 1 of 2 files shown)


src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -177,7 +177,7 @@ public function securityPhp($options = ['format' => 'yaml'])
         $process->mustRun();
         $out = $process->getOutput();
         if ($packages = json_decode($out, true)) {
-            $suggested_command = "composer why " . key($packages);
+            $suggested_command = "composer why " . implode('; composer why ', array_keys($packages));
```
weitzman, line 180:
```suggestion
            $suggested_command = "composer why " . implode('&& composer why ', array_keys($packages));
```
I think this is more Windows friendly.

weitzman:
Looks like one test needs adjusting as well.

PR #4416, opened 2020-05-20
Add Microsoft SQL Server support for Drush sql commands

starbloom:
This commit add Microsoft SQL Server support for Drush.  Drupal has been supporting SQL Server for a long time, but drush doesn't due to lack of interest.  It adds support for sql:cli, sql:query and sql:create commands.  Tested in both Linux and Windows.


weitzman:
Please contribute these to the existing drupal sql srv module or as a standalone package. See how I did that for Oracle drivers https://gitlab.com/weitzman/drush-oracle-driver/

PR #4431, opened 2020-06-05
Remove entity_load function in favour of Role::load to be Drupal 9 co…

seamuslee001:
…mpatible

@greg-1-anderson This is as per https://www.drupal.org/node/2266845.


greg-1-anderson:
Thanks. Looks like there are no tests for these; there are probably other calls that were also missed.

Issue #4432, opened 2020-06-07
Drush 8.x scanning recursively for alias file


pritamrungta:
**Describe the bug**
In _drush_sitealias_find_alias_files of includes/sitealias, the comment says that scan is not deep and done only in immediate directories.
However the last parameter of scan function for recursive is TRUE. This is taking forever in my system, containing lot of files in sites/default/private_files which is neither in blacklist as files.

**To Reproduce**
Fresh install drush 8.3.5

**Expected behavior**
As mentioned in comment, it should have only scanned the required directories non-recursively.

**Actual behavior**
According to the code, it went deep down for scanning.

**Workaround**
Changed that part for me to use drush.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.3.5
| Drupal version? | 7.x
| PHP version     | 7.x
| OS?             | Linux

**Additional information**
Not sure if recursive scan was intended here. If not, could you please fix this, so I wouldn't have to make any changes in future upgrades.

greg-1-anderson:
Yeah, the comment is wrong. This is fixed in Drush 10.

Changing Drush 8 would risk breaking configurations that required deeper searching, and it's hard to tell how deep folks expect to go. A PR that would allow a limit to be set in a drushrc.php file would be welcome, though.

pritamrungta:
Thanks!

PR #4436, opened 2020-06-08
Drush8 alias files search depth

pritamrungta:
#4432

added option 'alias-search-depth'(Default: TRUE) in drushrc.php


greg-1-anderson:
Awesome, thanks.

PR #4438, opened 2020-06-12
Fix #4134: Detects proxy settings before fetching drupal-security-adv…

### LINKED ISSUE #4134: Drush 9.x is not using http proxy.
**Describe the bug**
Drush 9 could not reach drupal-security-advisories if placed behind a proxy.
The `methodSecurityUpdateCommands::fetchAdvisoryComposerJson` is using raw `file_get_contents()` without any proxy settings.

**To Reproduce**
Execute `drush sec` behind a proxy

**Expected behavior**
Drush should be aware of proxy settings either from settings.php, parameters or env variables.

**Actual behavior**
Drush could not reach internet host if placed behind a proxy.

**Workaround**
Patch these method to pass proxy setting (there's also 21 other occurrences of  `file_get_contents` in drush 9 code base):
```
// TODO : get proxy settings
$context = stream_context_create([
    'http' => [
        'proxy' => 'tcp://PROXY_HOST:PROXY_PORT',
        'request_fulluri' => true,
    ],
]);
$response_body = file_get_contents('https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/8.x-v2/composer.json');
```

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.x
| Drupal version? | 8.x
| PHP version     | 7.1
| OS?             | Linux

obriat:
Hi,
here's a PR to fix  https://github.com/drush-ops/drush/issues/4134 
It use bootstrap configuration to detect proxy settings and use them to get drupal-security-advisories.

It could be improve by detecting proxy env variable.

### DIFF
--- a/src/Commands/pm/SecurityUpdateCommands.php
+++ b/src/Commands/pm/SecurityUpdateCommands.php
@@ -44,7 +44,7 @@ class SecurityUpdateCommands extends DrushCommands
      *
      * @command pm:security
      * @aliases sec,pm-security
-     * @bootstrap none
+     * @bootstrap configuration
      * @table-style default
      * @field-labels
      *   name: Name
@@ -94,8 +94,27 @@ class SecurityUpdateCommands extends DrushCommands
     protected function fetchAdvisoryComposerJson()
     {
         try {
+            // Detects proxy settings.
+            $http_client_config = \Drupal\Core\Site\Settings::get('http_client_config');
+            $cxContext = null;
+            if (!empty($http_client_config['proxy'])) {
+                $aContext = [];
+                if (!empty($http_client_config['proxy']['http'])) {
+                    $aContext['http'] = [
+                        'proxy' => $http_client_config['proxy']['http'],
+                        'request_fulluri' => true,
+                    ];
+                }
+                if (!empty($http_client_config['proxy']['https'])) {
+                    $aContext['https'] = [
+                        'proxy' => $http_client_config['proxy']['https'],
+                        'request_fulluri' => true,
+                    ];
+                }
+                $cxContext = stream_context_create($aContext);
+            }
             // We use the v2 branch for now, as per https://github.com/drupal-composer/drupal-security-advisories/pull/11.
-            $response_body = file_get_contents('https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/8.x-v2/composer.json');
+            $response_body = file_get_contents('https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/8.x-v2/composer.json', false, $cxContext);
             if ($response_body === false) {
                 throw new Exception("Unable to fetch drupal-security-advisories information.");
             }
--- a/src/Commands/pm/SecurityUpdateCommands.php
+++ b/src/Commands/pm/SecurityUpdateCommands.php
@@ -5,6 +5,7 @@ use Composer\Semver\Semver;
 use Consolidation\AnnotatedCommand\CommandResult;
 use Consolidation\OutputFormatters\StructuredData\RowsOfFields;
 use Consolidation\OutputFormatters\StructuredData\UnstructuredData;
+use Drupal\Core\Site\Settings;
 use Drush\Commands\DrushCommands;
 use Drush\Drush;
 use Exception;
@@ -85,6 +86,40 @@ class SecurityUpdateCommands extends DrushCommands
     }
 
     /**
+     * Return file_get_contents context if proxy is detected.
+     *
+     * Drupal settings override ENV ones.
+     *
+     * @return resource
+     *  A stream context resource.
+     */
+    protected function getProxySettings()
+    {
+        $settings = Settings::get('http_client_config');
+        // Check for env defined proxy then look at drupal settings one.
+        $http_proxy = getenv('http_proxy') ?: null;
+        $https_proxy = getenv('https_proxy') ?: null;
+
+        $http_proxy = !empty($settings['proxy']['http']) ? $settings['proxy']['http'] : $http_proxy;
+        $https_proxy = !empty($settings['proxy']['https']) ? $settings['proxy']['https'] : $https_proxy;
+
+        $aContext = [];
+        if (!empty($http_proxy)) {
+            $aContext['http'] = [
+                'proxy' => $http_proxy,
+                'request_fulluri' => true,
+            ];
+        }
+        if (!empty($https_proxy)) {
+            $aContext['https'] = [
+                'proxy' => $https_proxy,
+                'request_fulluri' => true,
+            ];
+        }
+        return stream_context_create($aContext);
+    }
+
+    /**
      * Fetches the generated composer.json from drupal-security-advisories.
      *
      * @return mixed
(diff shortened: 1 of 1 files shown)


src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -44,7 +45,7 @@ public static function composerLockPath(): string
      *
      * @command pm:security
      * @aliases sec,pm-security
-     * @bootstrap none
+     * @bootstrap configuration
```
weitzman, line 48: Unfortunately this is a regression. One use case is to run this in in cron/CI without an installed site present. The only prerequisite is composer.lock.

weitzman: I guess this only adds a settings.php requirement which isn't terrible but I'm not sure its a net positive.

obriat: I tested it with a single composer.lock and didn't get any errors or warnings, I guess it could go wrong if Drupal is broken and could not bootstrap (but at this point you got bigger problems).

If you still find a regression about the bootstrap part, this dependency could be removed, the `getenv('https_proxy)'` part is still working.

obriat: So, will you merge it or I have something to change ?

weitzman: You can change to `@bootstrap max`. That will go as far as possible, and then start running the command. Please test with an unconfigured site, just to make sure the command still work (of couse no proxy support in this case).

src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -84,6 +85,40 @@ public function suggestComposerCommand($updates)
...
 
+    /**
+     * Return file_get_contents context if proxy is detected.
+     *
+     * Drupal settings override ENV ones.
+     *
+     * @return resource
+     *  A stream context resource.
+     */
+    protected function getProxySettings()
```
weitzman, line 96: It might make sense to move this somewhere more central, so other commands can benefit.

src/Commands/pm/SecurityUpdateCommands.php
obriat: I tried with a composer.lock alone and it works, should I write a test ?

src/Commands/pm/SecurityUpdateCommands.php
obriat: Yes, as I said in the issue they are more than 20 uses of  `file_get_contents ` in drush, it should  be a good idea to replace them with a helper method that manage proxy. But it's a bit out of my league...

src/Commands/pm/SecurityUpdateCommands.php
obriat:
```bash
$ env |grep proxy
http_proxy=my-proxy:80
https_proxy=my-proxy:80
no_proxy=localhost,127.0.0.0,
$ mv  sites/default/settings.php sites/default/settings.old
$ drush sec
 [warning] One or more of your dependencies has an outstanding security update.
 [notice] Try running: composer require drupal/core --update-with-dependencies
 [notice] If that fails due to a conflict then you must update one or more root dependencies.
+-------------+-------------------+
| Name        | Installed Version |
+-------------+-------------------+
| drupal/core | 8.8.2             |
+-------------+-------------------+
$ unset http_proxy
$ unset https_proxy
$ drush sec
 [warning] file_get_contents(): php_network_getaddresses: getaddrinfo failed: Name or service not known SecurityUpdateCommands.php:136
 [warning] file_get_contents(https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/8.x-v2/composer.json): failed to open stream: php_network_getaddresses: getaddrinfo failed: Name or service not known SecurityUpdateCommands.php:136

In SecurityUpdateCommands.php line 141:

  Unable to fetch drupal-security-advisories information.


pm:security [--format FORMAT] [--fields FIELDS] [--field FIELD] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-d|--debug] [-y|--yes] [--no] [--remote-host REMOTE-HOST] [--remote-user REMOTE-USER] [-r|--root ROOT] [-l|--uri URI] [--simulate] [--pipe] [-D|--define DEFINE] [--filter [FILTER]] [--notify [NOTIFY]] [--druplicon] [--xh-link XH-LINK] [--] <command>
```

src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -84,6 +85,40 @@ public function suggestComposerCommand($updates)
...
+    {
+        $settings = Settings::get('http_client_config');
+        // Check for env defined proxy then look at drupal settings one.
+        $http_proxy = getenv('http_proxy') ?: null;
```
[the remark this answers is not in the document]
obriat: Could you merge it until you migrate to Guzzle ? 
Do you need any help on this migration ?
You are right about the case, since there is no standard, both cases :) should be tested (https://unix.stackexchange.com/questions/212894/whats-the-right-format-for-the-http-proxy-environment-variable-caps-or-no-ca)

src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -94,8 +129,11 @@ public function suggestComposerCommand($updates)
     protected function fetchAdvisoryComposerJson()
     {
         try {
+            // Detects proxy settings.
+            $cxContext = $this->getProxySettings();
```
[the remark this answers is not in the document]
obriat: And change the error " Unable to fetch drupal-security-advisories information." to " Unable to fetch drupal-security-advisories information: check you internet connection, firewall or proxy settings."

obriat:
Hi,

I think the best way is to move Drush from  `file_get_contents  `to `Guzzle `and only use the uppercase env parameters (so we don't need any bootstrapping or look into settings.php).

Trivia :
Composer use upper and lowercase, 
https://github.com/composer/composer/blob/master/src/Composer/Util/StreamContextFactory.php#L72
```php
        // Handle HTTP_PROXY/http_proxy on CLI only for security reasons
        if ((PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') && (!empty($_SERVER['HTTP_PROXY']) || !empty($_SERVER['http_proxy']))) {
            $proxy = parse_url(!empty($_SERVER['http_proxy']) ? $_SERVER['http_proxy'] : $_SERVER['HTTP_PROXY']);
        }
```
Guzzle only looks for the uppercase ones.
https://github.com/guzzle/guzzle/blob/master/src/Client.php#L239
```php
        if (\PHP_SAPI === 'cli' && ($proxy = Utils::getenv('HTTP_PROXY'))) {
            $defaults['proxy']['http'] = $proxy;
        }
```

weitzman:
That looks like a good suggestion.

obriat:
Should I just replace this  file_get_contents  and the MR is done  or should I just update the documentation and wait for another issue to take car of **all** file_get_contents  substitutions with Guzzle ?

weitzman:
250 file_get_content()? The ones that are reading from a local file don't need to change.

I'd be fine with this PR changing security and security-php commands.

weitzman:
Fixed in #4522

PR #4441, opened 2020-06-19
Declare the minimum supported version of Robo

### LINKED ISSUE #4352: Drush should require the correct minimum version of Robo
**Describe the bug**
After updating drush from version 9 to version 10 I get this error on each drush command,
like: `drush cr` or `drush status`
`[preflight] Alias (parameterInjection) is not being managed by the container`

**To Reproduce**
working website with Drupal core 8.8.2
drush and all drush commands were working fine with Drush 9.0.0-beta7
updated drush to Drush 10 with: `composer require drush/drush --update-with-dependencies`
update went fine without any problems
after update I get the error on every drush command I try
Important information: I am using Lando. I updated my Lando config to use drush 10 also and used `lando rebuild` which did also update the drush version of the lando container.

**Expected behavior**
Updating drush from v9 to v10 would still keep drush workinf

**Actual behavior**
Updating drush from v9 to v10 causes all drush commands to error: "[preflight] Alias (parameterInjection) is not being managed by the container."

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.2.2
| Drupal version? | 8.8.2
| PHP version     | 7.2.28
| OS?             | Windows

pfrenssen:
Fixes #4352

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     "composer/semver": "^1.4",
     "consolidation/config": "^1.2",
     "consolidation/filter-via-dot-access-data": "^1",
-    "consolidation/robo": "^1 || ^2",
+    "consolidation/robo": "^1.4.11 || ^2",
     "consolidation/site-alias": "^3.0.0@stable",
     "consolidation/site-process": "^2.1 || ^4",
     "grasmash/yaml-expander": "^1.1.1",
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Looks good, but any change to composer.json should also come with an updated composer.lock too.

pfrenssen:
Performed the following and rebased the PR:

```
$  composer update consolidation/robo --with-dependencies
```

Issue #4410, opened 2020-05-15
Missing COLUMN_STATISTICS with mysqldump 8


guix77:
**Describe the bug**
Drush now throws an error with mysqldump 8 for certain operations like drush sql-dump.

**To Reproduce**
Updated Ubuntu 18.04 LTS => 20.04 LTS
```
mysqldump --version
mysqldump  Ver 8.0.20-0ubuntu0.20.04.1 for Linux on x86_64 ((Ubuntu))
drush sql-dump > db.sql
```

**Expected behavior**
Dump the DB to a file.

**Actual behavior**
```
> mysqldump: Couldn't execute 'SELECT COLUMN_NAME,                       JSON_EXTRACT(HISTOGRAM, '$."number-of-buckets-specified"')                FROM information_schema.COLUMN_STATISTICS                WHERE SCHEMA_NAME = 'gd8' AND TABLE_NAME = 'batch';': Unknown table 'COLUMN_STATISTICS' in information_schema (1109)

In SqlCommands.php line 242:
                                                                         
  Unable to dump database. Rerun with --debug to see any error message
```

**Workaround**
Not found so far

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.2.2 and 8.3.2
| Drupal version? | 8.8.5 and 7.53
| PHP version     | 7.3.12
| OS?             | Linux (Ubuntu 20.04 LTS)

weitzman:
I dont think that query comes from drush.

We tested with mysql8 and it passed all tests https://github.com/drush-ops/drush/issues/3830

guix77:
Hi Moshe,

Very sorry for answering so late, also I should have checked before a manual mysqldump, the error was very explicit.

So I checked and indeed the error is here with a usual mysqldump so nothing caused by Drush indeed.

Thanks again!

guix77:
For reference, mysqldump 8 is now expecting a information_schema.COLUMN_STATISTICS table.

On MariaDB there is no such column in information_schema: https://jira.mariadb.org/browse/MDEV-16555


 Workaround: https://github.com/snipe/snipe-it/issues/6800#issuecomment-470990126

guix77:
Sorry to reopen this...

In fact, while I can now run mysqldump from the command line, form Drush I still have the same issue.

Strange fact with the local fix I mentionned above:
- now in a Drupal 7 with Drush 8.3.2 I get no errors
[...]

PR #4442, opened 2020-06-21
Initialize mysqldump without column-statistics

netw3rker:
Starting in version 8, mysqldump expects column statistics to be part of the database information schema. However mysqldump does not check the mysql version of the server it is dumping the data from, and therefore will create an SQL statement that fails when targeting dbservers < version 8.

When mysqldump is installed, the default configuration is to disable this behavior, and as such most my.cnf files will generally include the switch that disables this. However, drush creates it's own config file to start mysql and mysqldump. This file is not combined with the system my.cnf files, and therefore needs this switch added to it in order to ensure that when it attempts to dump the database from a target system, it doesn't care that it's not a mysql8 database.

### DIFF
--- a/src/Sql/SqlMysql.php
+++ b/src/Sql/SqlMysql.php
@@ -24,6 +24,9 @@ class SqlMysql extends SqlBase
             // EMPTY password is not the same as NO password, and is valid.
             $contents = <<<EOT
 #This file was written by Drush's Sqlmysql.php.
+[mysqldump]
+column-statistics=0
+
 [client]
 user="{$dbSpec['username']}"
 password="{$dbSpec['password']}"


netw3rker:
well, interesting. this gets a little more complicated in that mysqldump7 won't allow that config option to be there. not sure what the path is for version determinant config loading here...

weitzman:
And my mysqldump has different numbering scheme

```
$ mysqldump --version
mysqldump  Ver 10.17 Distrib 10.4.11-MariaDB, for osx10.14 (x86_64)
```

greg-1-anderson:
Yeah, the MariaDB version mapping to Mysql versions is nontrivial.

I provided a suggestion in https://github.com/drush-ops/drush/issues/4410. It requires us to be able to find any applicable user / system my.cnf file. Not sure if that's workable, but perhaps easier than the alternatives.

weitzman:
This PR wont work. Discussion is in #4410.

PR #4454, opened 2020-07-01
Feat: add executable as option for sql-dump.

Nico-uncinc:
I'd like to use a GDPR dump library (https://github.com/machbarmacher/gdpr-dump) for creating anonymized dumps. That library provides its own mysqldump executable for anonymizing the data. So for this to properly work with drush, we would need an option to choose their mysqldump executable instead of the default one. This PR adds that option.


weitzman:
The OS already provides a way to control what executable comes first. See PATH variable. On a remote  drush, your alias can specify a PATH  - https://github.com/drush-ops/drush/blob/master/examples/example.site.yml#L171-L188

PR #4456, opened 2020-07-03
Wrap mysql table names in backticks to avoid errors in MySQL 8 

mcdruid:
e.g. the `system` table in D7

Without this change, `site-install` and `sql-drop` hit errors in Drupal 7 with MySQL 8 - e.g:

```
sql-query: DROP TABLE actions, authmap, [...snip...], variable, watchdog [10.75 sec, 11.84 MB]

ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'system, taxonomy_index, taxonomy_term_data, taxonomy_term_hierarchy, taxonomy_vo' at line 1
```

### DIFF
--- a/commands/sql/sql.drush.inc
+++ b/commands/sql/sql.drush.inc
@@ -445,7 +445,7 @@ function drush_sql_drop() {
   if (!drush_confirm(dt('Do you really want to drop all tables in the database !db?', array('!db' => $db_spec['database'])))) {
     return drush_user_abort();
   }
-  $tables = $sql->listTables(TRUE);
+  $tables = $sql->listTablesQuoted();
   return $sql->drop($tables);
 }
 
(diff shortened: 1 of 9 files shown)


mcdruid:
Yup that makes sense; let's try again...

greg-1-anderson:
The first failure is probably because `sql:dump` shouldn't use quoted table names. Not sure why the sanitize test is failing, but must be a similar reason.

mcdruid:
Sorry, only saw the first part of your comment and pushed that change before reading the rest.

I can submit a PR against the master branch, but I'll wait to hear if (and how) you want changes made to SqlBase and the other classes that extend it first.

greg-1-anderson:
SqlBase should declare the quote parameter in the listTables method. Doing this will probably require a similar change in all derived classes.

mcdruid:
If this looks right, do you want a new PR against master referencing this one? (not sure I've switched branches in a PR before, or if it's possible)

Oh, and guess who used double quotes for the commit message so `$quoted` disappeared. D'oh.

greg-1-anderson:
Thanks for working through all of the issues and backporting.

mcdruid:
Thanks for the shepherding these home @greg-1-anderson and @weitzman!

PR #4458, opened 2020-07-07
Add $quoted param to SqlBase::listTables so table names can

mcdruid:
...be wrapped in backticks e.g. for MySQL 5.7 and 8

Hopefully same changes as #4456

### DIFF
--- a/src/Sql/SqlBase.php
+++ b/src/Sql/SqlBase.php
@@ -508,6 +508,7 @@ class SqlBase implements ConfigAwareInterface
      */
     public function listTablesQuoted()
     {
+        return $this->listTables();
     }
 
     /*
--- a/src/Sql/SqlBase.php
+++ b/src/Sql/SqlBase.php
@@ -453,7 +453,7 @@ class SqlBase implements ConfigAwareInterface
     public function dropOrCreate()
     {
         if ($this->dbExists()) {
-            return $this->drop($this->listTables(true));
+            return $this->drop($this->listTablesQuoted());
         } else {
             return $this->createdb();
         }
@@ -492,12 +492,21 @@ class SqlBase implements ConfigAwareInterface
     /**
      * Extract the name of all existing tables in the given database.
      *
-     * @param bool $quoted
-     *  If TRUE, each table name should be appropriately quoted for the RDMS.
      * @return array|null
      *   An array of table names which exist in the current database.
      */
-    public function listTables($quoted = false)
+    public function listTables()
+    {
+    }
+
+    /**
+     * Extract the name of all existing tables in the given database.
+     *
+     * @return array|null
+     *   An array of table names which exist in the current database,
+     *   appropriately quoted for the RDMS.
+     */
+    public function listTablesQuoted()
     {
     }
 
--- a/src/Sql/SqlBase.php
+++ b/src/Sql/SqlBase.php
@@ -453,7 +453,7 @@ class SqlBase implements ConfigAwareInterface
     public function dropOrCreate()
     {
         if ($this->dbExists()) {
-            return $this->drop($this->listTables());
+            return $this->drop($this->listTables(true));
         } else {
             return $this->createdb();
         }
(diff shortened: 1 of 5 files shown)


mcdruid:
@weitzman #4456 has been approved by @greg-1-anderson for the 8.x branch, but the code's very similar in master.

This could affect any Drupal versions on MySQL 8, although without D7's `system` table sites might not have any tables with reserved names.

mcdruid:
Seems this affects MySQL 5.7 too in some cases.

weitzman:
Code looks good I'm wondering if this is really the only place we deal with dynamic table names.

Does this merit a new major version? I think contrib db drivers would break (perhaps not such a big deal), and maybe some code using the output of `listTables()` will break as well. I'm reluctant to merge it into Drush8, but I'll defer to @greg-1-anderson .

Any thoughts on https://github.com/drush-ops/drush/issues/4410 are appreciated.

greg-1-anderson:
Contrib db drivers will break -- but are there any?

Code that calls `listTables()` will continue to work, because the default value for the `$quoted` parameter is `false`.

If contrib db drivers is an issue, we could increase backwards compatibility by adding a new method "escapedListTables" in the SQL base class. This could just return listTables() by default, and could be overridden in the mysql driver to escape each result.

weitzman:
I guess in Drush8 there aren't contrib drivers. In Drush 10, [Oracle](https://gitlab.com/weitzman/drush-oracle-driver/) and Sql server are contrib. Those are so lightly used that releasing this in a minor is good enough IMO.

greg-1-anderson:
I think that the `escapedListTables` alternative is easy enough to implement, and would not hurt the contrib drivers. Maybe that's better, to avoid semver confusion?

weitzman:
Yeah, that would be great if @mcdruid has any more time for this.

mcdruid:
Hopefully this is what you had in mind(s)?

If you could confirm that, I'll backport it to the drush 8 PR too.

Thanks!

src/Sql/SqlBase.php
```diff
@@ -492,12 +492,21 @@ public function scheme()
...
+    }
+
+    /**
+     * Extract the name of all existing tables in the given database.
+     *
+     * @return array|null
+     *   An array of table names which exist in the current database,
+     *   appropriately quoted for the RDMS.
+     */
+    public function listTablesQuoted()
```
greg-1-anderson, line 509: The base implementation could call through to `listTables()`. That way contrib drivers that did not implement this method would not break.

mcdruid: Thanks, I had a feeling I was missing something :)

src/Sql/SqlPgsql.php
```diff
@@ -118,6 +118,11 @@ public function listTables()
         return array_filter($tables);
     }
 
+    public function listTablesQuoted()
```
greg-1-anderson, line 121: You can remove this + the same method in Sqlite now that this is in the base class.

src/Sql/SqlMysql.php
```diff
@@ -133,6 +133,16 @@ public function listTables()
...
     }
 
+    public function listTablesQuoted()
+    {
+        $tables = $this->listTables();
+        foreach ($tables as &$table) {
+            $table = "`$table`";
+        }
+        return $tables;
+        return $tables;
```
greg-1-anderson, line 143: Typo double `return` statement

mcdruid: Argh... sorry. Fixed.

src/Sql/SqlMysql.php
```diff
@@ -123,7 +123,7 @@ public function dbExists()
         return $this->alwaysQuery("SELECT 1;");
     }
 
-    public function listTables()
+    public function listTables($quoted = false)
```
LionsAd, line 126: @mcdruid I am confused -- `$quoted` is not used at all - it seems? It's optional, so does not break anything, but that feels really weird to look at.

I think that was left over ...

cc @weitzman

PR #4469, opened 2020-07-20
Replace blacklist/whitelist with blocklist/allowlist

leymannx:
Following up on https://www.drupal.org/project/ideas/issues/3159647#comment-13755051 let's fix this now.

This PR replaces blacklist which only had one occurrence in a comment. And it deprecates whitelist which only appeared in an option to the `sql:sanitize` command.

### DIFF
--- a/src/Drupal/Commands/sql/SanitizeUserFieldsCommands.php
+++ b/src/Drupal/Commands/sql/SanitizeUserFieldsCommands.php
@@ -52,7 +52,11 @@ class SanitizeUserFieldsCommands extends DrushCommands implements SanitizePlugin
         $conn = $this->getDatabase();
         $field_definitions = $this->getEntityFieldManager()->getFieldDefinitions('user', 'user');
         $field_storage = $this->getEntityFieldManager()->getFieldStorageDefinitions('user');
+        /** @deprecated Use $options['allowlist-fields'] instead. */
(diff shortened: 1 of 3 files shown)


src/Drupal/Commands/sql/SanitizeUserFieldsCommands.php
```diff
@@ -52,9 +52,13 @@ public function sanitize($result, CommandData $commandData)
         $conn = $this->getDatabase();
         $field_definitions = $this->getEntityFieldManager()->getFieldDefinitions('user', 'user');
         $field_storage = $this->getEntityFieldManager()->getFieldStorageDefinitions('user');
+        /** @deprecated Use $options['allowlist-fields'] instead. */
```
damienmckenna, line 55: Does this need an indicator on when the legacy value will be removed?

greg-1-anderson: Usually we would remove in the next major version. There is no Drush 11 branch at the moment, and no convention (other than `@deprecated`) to explicitly flag something for removal in Drush 11.

Drush 11 is far off, and it we don't have the bandwidth to maintain a "future" branch right now. We could consider introducing a convention, but in theory we remove all `@deprecated` functions in each major version, so `@deprecated` fills that role at the moment.

PR #4471, opened 2020-07-21
Better error handling for update functions in wrong files 

jonathanjfshaw:
This PR fixes 3 problems:
1. Currently post update hooks appear successful if placed in my_module.install, but are not actually invoked
2. If you put your update or post update function in the wrong file, it would be helpful to tell you what the right file is
3. Post-update processing should use Drupal::moduleHandler()->loadInclude not module_load_include(), as update processing already does.

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -303,7 +303,7 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
         }
 
         list($module, $name) = explode('_post_update_', $function, 2);
-        $filename = $module . '.post_update';    
+        $filename = $module . '.post_update';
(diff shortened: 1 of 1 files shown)


src/Commands/core/UpdateDBCommands.php
```diff
@@ -238,7 +238,10 @@ public static function updateDoOne($module, $number, array $dependency_map, Drus
             }
         } else {
             $ret['#abort'] = ['success' => false];
-            Drush::logger()->warning(dt('Update function @function not found', ['@function' => $function]));
+            Drush::logger()->warning(dt('Update function @function not found in file @filename', [
+                '@function' => $function,
+                '@filename' => "$module.install.php",
```
weitzman, line 243: Copy/paste snafu - extra trailing `.php`

jonathanjfshaw: Fixed, thanks for reviewing!

PR #4477, opened 2020-07-25
@topic annotation can be a relative path to a file

weitzman:
I would like to hyperlink the topics on the new Commands web pages. For example, see 3 topics at bottom of https://www.drush.org/commands/10.x/config_pull/.

This PR adds a value to one `@topic` with a relative link to the file. The mk:docs command can programmatically use that value to build the hyperlink.

This PR duplicates the path `../../../README.md` in the body of the command method. Ideally it would use the value from its own annotation. However I do not see a way in a command method to learn what the current command name is. If I could do that, I would get load a $command object and fetch the annotationData. Hooks get the command info as a param but the command method does not.

This PR will change more methods once the above question is settled.

### DIFF
--- a/src/Commands/DrushCommands.php
+++ b/src/Commands/DrushCommands.php
@@ -1,6 +1,7 @@
 <?php
 namespace Drush\Commands;
 
+use Consolidation\AnnotatedCommand\CommandData;
 use Drush\Drush;
 use Drush\Style\DrushStyle;
 use Psr\Log\LoggerAwareInterface;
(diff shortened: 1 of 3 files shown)


greg-1-anderson:
The primary command method does not get a reference to the command info. 

The best way to do this would be to have some early hook that gets the command info copy the data you need into the command handler class.

The other option is that I could add a feature to annotated commands to inject the command info to the primary command method, the same way that the Input object is.

Issue #4485, opened 2020-07-29
Allow explicitly listed packages to skip security checks


lpeabody:
**Is your feature request related to a problem? Please describe.**

Sometimes the security conflict list provided by https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/8.x-v2/composer.json is not always accurate.

As an example, today's security advisory for the Group module (https://www.drupal.org/sa-contrib-2020-030) explicitly states that 8.x-1.0-rc5 is safe. However, it is flagged in the security conflict list. Our project is still secure, but our CI process is throwing false positives now because we check security advisories.

**Describe the solution you'd like**

It would be nice if Drush permitted an allowed list of packages to be passed to the command so they could be summarily excluded from security checks, in the event that the posted security conflict list is inaccurate.

**Describe alternatives you've considered**

Drupal.org should permit maintainers to explicitly flag releases as safe, and then hopefully the conflict list that gets generated does not include those flagged releases.

However, this is out of Drush's control.

weitzman:
Sorry, I'm not inclined to add yet another way to solve this edge case. I suggest [using composer,](https://github.com/drupal-composer/drupal-security-advisories/#handling-failures) or alter away the unwanted packages https://github.com/consolidation/annotated-command#alter-hook

lpeabody:
Fair enough. Thanks got the pointers @weitzman.

PR #4486, opened 2020-07-29
4485: Allowed packages in security update checks

lpeabody:
# Changes

- Adds an `allowed` option that can be passed to `pm:security` which is a comma separated list of package/version pairs:

```
drush pm:security --allowed='drupal/group:1.0.0-rc5'
```

Or optionally in drush.yml:

```
command:
  pm:
    security:
      options:
        allowed:
          - drupal/group:1.0.0-rc5
```

Ultimately, the `allowed` option internally is converted into an array for processing.


lpeabody:
Fixes issue #4485.

PR #4499, opened 2020-08-01
Allow the backup dir parent to be set in config.

greg-1-anderson:
Pantheon is rolling out new "Container-Optimized Environments" (COE). Most existing bindings and all new sites are one COE. This new environment puts $HOME at `/`, and of course `/` is unwritable. This causes problems with Drush sql:sync, which by default wants to put the dump file in the backup directory. The Drush backup directory is at $HOME/drush-backups, which is not writable under COE.

This PR makes the backup dir configurable, just as it was in Drush 8.

### DIFF
--- a/src/Utils/FsUtils.php
+++ b/src/Utils/FsUtils.php
@@ -73,7 +73,7 @@ class FsUtils
         ];
 
         // Return the first usable candidate
-        foreach ($cadidates as $dir) {
+        foreach ($candidates as $dir) {
             if (self::isUsableDirectory($dir)) {
                 return $dir;
             }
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
So the workaround is:
```
drush sql:sync @site.env @self --source-dump=/tmp/db.sql
```
I think there was some reason why I didn't want to just set a command-specific source-dump property. Need to look at that again.

greg-1-anderson:
Oh, I remember. The source-dump property is read on the local machine. If we wanted to set a default command-specific option, it would have to be on the sql:dump command, but we do not want to change the behavior of that command unless it's being called from sql:sync.

weitzman:
I didn't look closely, but maybe this dir should fall back to temp dir like cache does - https://github.com/drush-ops/drush/blob/21494be38146049e3000b68557f2dac34bd1aaa8/src/Cache/FileCache.php#L35-L44

greg-1-anderson:
I can add a fallback to tmp if the configured or default directory is unwritable.

PR #4503, opened 2020-08-02
Fixes #4468: Rename "master" branch to "10.x".

### LINKED ISSUE #4468: Rename Drush master branch
There has been a lot of attention lately on language use that is unwelcoming for some participants. Drush uses "master" for the default branch, which is the default for Git and GitHub.

Rename default branch to "10.x" or "main".

**Additional context**
I did this for all of the Consolidation projects; it requires some updates to the CI jobs, but isn't too difficult. Open PRs might need to be recreated, but we only have 42.

greg-1-anderson:
n.b. This branch should not be merged; instead, once it is ready, change the GitHub default branch to 10.x and close this PR.

This PR changes references to the Drush 'master' branch to '10.x', but leaves references to 'master' to external entities unchanged.

### DIFF
--- a/docs/overrides/partials/tabs.html
+++ b/docs/overrides/partials/tabs.html
@@ -13,7 +13,7 @@
       {% endfor %}
 
       <li class="md-tabs__item">
-        <a href="/api/master" class="md-tabs__link">API</a>
+        <a href="/api/10.x" class="md-tabs__link">API</a>
       </li>
 
     </ul>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ jobs:
       - run: composer cs
 
   # Mergeable test:
-  #   FAIL if merging test branch with master produces conflicts
+  #   FAIL if merging test branch with 10.x produces conflicts
   #   PASS if the test branch is out of date, but mergeable without conflicts
   check_mergable:
     <<: *defaults
--- a/.circleci/mergable.sh
+++ b/.circleci/mergable.sh
@@ -1,7 +1,7 @@
 set -e
 
 # No need to check for mergability on commits that are already merged.
-if [ "$CIRCLE_BRANCH" == "master" ] || [ "$CIRCLE_BRANCH" == "9.x" ] ; then
+if [ "$CIRCLE_BRANCH" == "10.x" ] || [ "$CIRCLE_BRANCH" == "9.x" ] ; then
 	exit 0
 fi
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,19 +1,19 @@
 name: Build static sites
 
 # Controls when the action will run. Triggers the workflow on push or pull request
-# events but only for the master branch
+# events but only for the 10.x branch
 on:
   push:
-    branches: [master]
+    branches: [10.x]
 #  pull_request:
-#    branches: [master]
+#    branches: [10.x]
 
 jobs:
   build:
     name: Build and Push
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout master
+      - name: Checkout 10.x
         uses: actions/checkout@v2
 
       - name: Checkout gh-pages branch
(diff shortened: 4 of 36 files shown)


greg-1-anderson:
Hm, I said this branch should not be merged, but I think that perhaps the easiest way to get the `master` branch to stop publishing (e.g. drush.org) would be to merge this PR into master once it becomes the default branch.

Note that the CI for this PR has already republished drush.org using `10.x` in place of `master` in the documentation.

weitzman:
>  CI for this PR has already republished drush.org using 10.x in place of master in the documentation.

I dont think thats true. The pull request directive is currently commented out in our workflow file.

This PR looks good. Thanks. I note that [Github has asked us to wait a bit until we make this change](https://github.com/github/renaming#now-supporting-early-movers-). I'm inclined to follow their guidance. I'm also open to hearing about why thats not needed.

greg-1-anderson:
> The pull request directive is currently commented out in our workflow file.

Yeah, but this pull request changes the CI's idea of what the main branch is from master to 10.x, and this PR is the 10.x branch. Therefore, drush.org was rebuilt from the main branch rules, not from the PR rules.

I don't think that we need to wait for GitHub's magic redirecting code to be perfected to switch. GitHub is trying to make things better for projects that delete the master branch. Instead, we can keep the master branch around for a while, and sync it with the 10.x branch periodically (e.g. on every stable release). Once GitHub releases their fancy code, we can go ahead and delete master.

greg-1-anderson:
If you take a look at drush.org, it's using 10.x tags right now. The "API" menu still says `master`, probably because I missed one reference.

docs/overrides/partials/tabs.html
```diff
@@ -13,7 +13,7 @@
       {% endfor %}
 
       <li class="md-tabs__item">
-        <a href="/api/master" class="md-tabs__link">API</a>
+        <a href="/api/10.x" class="md-tabs__link">API</a>
```
greg-1-anderson, line 16: Hm, yeah, I got the 10.x here, and in fact the build worked itself out, and the API link on drush.org is now /api/10.x after my last rebase with master.

weitzman:
OK, LGTM. Feel free to change default branch and merge if you think thats helpful.

weitzman:
Actually it looks like Sami is still publishing to api/master. See gh-pages branch at https://github.com/drush-ops/drush/tree/gh-pages/api/master. I caught this because the API site is currently not working right https://www.drush.org/api/10.x

I think the most expedient fix is to replace `%version` with `10.x` at https://github.com/drush-ops/drush/blob/master/sami-config.php.

greg-1-anderson:
Adjusted Sami to build from the 10.x branch. We'll see if I got it right; I think this should work.

We could also build separate docs for each semver tag that is published, if desired. I didn't turn that on yet, though.

greg-1-anderson:
The end of the Sami build says:
```
 Version    Updated C    Removed C 
     10.x          168            0

 Version    Updated C    Updated N    Removed C    Removed N 
     10.x          168           43            0            0

Script PATH=$HOME/bin:$PATH sami.phar --ansi update sami-config.php handling the api event returned with error code 64
```

The parsing phase has a count of errors that is > 0, but it does not report what they are. Not sure why it's not working.

greg-1-anderson:
So now it does look like we have Sami rendering to 10.x and only 10.x:

https://github.com/drush-ops/drush/tree/gh-pages/api

I don't understand why it's not rendering correctly. The index.html file looks like it has a correct relative link to the css file & etc.

Maybe I just need to wait for GitHub pages to catch up?

greg-1-anderson:
Yeah, I just waited a tick and the GitHub pages api/10.x path started rendering correctly.

leymannx:
https://docs.drush.org/ now always redirects to https://docs.drush.org/en/8.x/ when it was at master before I think. This is related, isn't it?

weitzman:
Not related, its just that another change happenned around same time  Only 8.x docs are now hosted separately on readthedocs which is CNAME to docs.drush.org. The 10.x docs are integrated into the new http://www.drush.org site, along with a command list.

PR #4510, opened 2020-08-05
Fix the backup-dir configuration setting

greg-1-anderson:
The drush backup-dir configuration setting should have been drush.paths.backup-dir, for consistency.

Also added some documentation.

### DIFF
--- a/examples/example.drush.yml
+++ b/examples/example.drush.yml
@@ -79,6 +79,11 @@ drush:
 
     # Specify a folder where Drush should store its file based caches. If unspecified, defaults to $HOME/.drush.
     #cache-directory: /tmp/.drush
+    
+    # Specify a folder where Drush should store backup files, including
+    # temporary sql dump files created during sql:sync. If unspecified,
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
I don't really care if we keep the undocumented deprecated location.

weitzman:
If it was only released for a day, lets remove the deprecated location and do a quick patch release.

Has Pantheon considered setting the drush config `env.home` to a writable path? Might be an alternate way to solve the original need.

greg-1-anderson:
Unfortunately $HOME is set to `/` and cannot go anywhere else. I don't think we can override $HOME with Drush configuration; environment variables override config, not the other way around.

It's not necessary to do a quick patch release, unless you really want to for other users of this feature. I just set both config variables in the Pantheon drush.yml, so we can work with 10.3.2 and 10.3.3+.

greg-1-anderson:
It was on my mind to decide where to put `backup-dir` when I first did this, and then I just forgot.

PR #4533, opened 2020-08-19
Change how we NULL the cache dir during deploy command.

weitzman:
We can't use /dev/null as a value because it fails the candidate check `$fs->mkdir($candidate)` in \Drush\Config\DrushConfig::cache. A temp dir works just as well.


greg-1-anderson:
Seems like it might be better to short-circuit the mkdir check for `/dev/null` so that the bitbucket works, but this is okay too.

weitzman:
Anyone is welcome to file a follow-up to add handling for the bit bucket.

weitzman:
Follow up at #4535

PR #4535, opened 2020-08-20
Avoid exception when clearing Drush’s own cache files

weitzman:
Yet another attempt to deal with #3592. This accepts the race condition and simply converts the Exception into a notice. Its pragmatic, having dealt with this for years.

### DIFF
--- a/src/Commands/core/CacheCommands.php
+++ b/src/Commands/core/CacheCommands.php
@@ -288,11 +288,11 @@ class CacheCommands extends DrushCommands implements CustomEventAwareInterface, 
     public static function clearDrush()
     {
         try {
-            drush_cache_clear_all(null, 'default');// commandfiles, etc.
-            drush_cache_clear_all(null, 'factory'); // command info from annotated-command library
+            drush_cache_clear_all(null, 'default');// No longer used by Drush core, but still cleared for backward compat.
+            drush_cache_clear_all(null, 'factory'); // command info from annotated-command library (i.e. parsed annotations)
         } catch (IOException $e) {
             // Sometimes another process writes files into a bin dir and \Drush\Cache\FileCache::clear fails.
-            // That is not considered an error.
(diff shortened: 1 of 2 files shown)


src/Commands/core/CacheCommands.php
```diff
@@ -286,8 +287,14 @@ public function getTypes($include_bootstrapped_types = false)
...
     {
-        drush_cache_clear_all(null, 'default'); // commandfiles, etc.
-        drush_cache_clear_all(null, 'factory'); // command info from annotated-command library
+        try {
+            drush_cache_clear_all(null, 'default');// commandfiles, etc.
+            drush_cache_clear_all(null, 'factory'); // command info from annotated-command library
+        } catch (IOException $e) {
+            // Sometimes another process writes files into a bin dir and \Drush\Cache\FileCache::clear fails.
+            // That is not considered an error.
+            Drush::logger()->info($e->getMessage());
```
andypost, line 296: maybe better to `logger->error()` instead of info?

andypost: or warning to return exit code 0

greg-1-anderson: We usually throw exceptions rather than log to error(), but maybe warning would be appropriate if we go this route.

greg-1-anderson:
Stylistically I think I'd prefer to use [flock](https://www.php.net/manual/en/function.flock.php), which isn't much more code than a try / catch. I suppose it doesn't matter, though.

I haven't really looked at the cache code in a while. If we have multiple Drush processes sharing the same cache, are some of these for separate sites? If the site is different, might the cache values be different? If the cache isn't already separated per docroot, maybe we should do that too.

greg-1-anderson:
I think that the docroot is already included in the hash of the cache filename, but maybe it should be included in the cache base dir path too?

Hm, probably not, still doesn't fix the two-processes-on-one-site problem, so there's probably no point.

PR #4542, opened 2020-08-27
Fix #4541: Add again uid and mail options for user commands

### LINKED ISSUE #4541: Re implement uid & mail options for user:* commands
As stated in #4020, `--uid` and `--mail options `have been remove form the `user` command in drush 9. It complicates some scripting actions, like block/unblock admin(1) user, specially if you already change its name for security reason.

Re implement these options as it have been done for `user:login`

obriat:
Fix #4541: Add again uid and mail options for user command.


q0rban:
This would be nice!

q0rban:
As a workaround, you can do things like this:

```
vendor/bin/drush user:block "$(vendor/bin/drush user:information --uid=1 --fields=name --format=string)"
```

deviantintegral:
We discussed the above, and one issue we have is that it's not very discoverable for junior developers or those unfamiliar with the shell. Making it an argument exposes it in the help docs, where the above requires understanding of shell string substitution.

@weitzman I see you closed the related issue. Is there anything you would like to see in order to get a feature like this merged in?

weitzman:
In general, this looks great. I am looking at our more strict validation in user:login and wondering if the warnings proposed here are sufficient. See https://github.com/drush-ops/drush/blob/2cb89999327e28efbd2e8854f48c4f537e418ae3/src/Commands/core/LoginCommands.php#L55-L70

weitzman:
Sorry for the notification noise while I coax CircleCI run its tests on this PR.

miromichalicka:
@weitzman sorry to bother you, but could this be backported to 10.x as well since it's still a stable release? Seems to be mergeable without any changes.

PR #4553, opened 2020-09-23
drush runserver with Drush >= 8.1.* and Drupal 7: doesn't decode URL paths the same as Apache or nginx would

dsnopek:
In Panopoly's automated tests, we've been pinned to Drush 8.0.* for years, because using Drush 8.1.* or newer had some mysterious issues with Panels, that I've finally gotten around to figuring out. :-)

This is important for us now, because we need to update to Drush 8.4.* in order to run our tests on PHP 7.4.

Anyway, the bug is that `drush runserver` won't decode the URL paths the same way that Apache and nginx does. This affects some Panels AJAX URLs, for example, with this URL from my local test site:

http://127.0.0.1:55432/panels/ajax/ipe/save_form/panel_context%3Apage-panopoly_demo_panel%3A%3Apage_panopoly_demo_panel_panel_context%3A%3A%3A%3A?destination=demo

When the AJAX request for the URL goes through Apache or nginx, then `panels_ajax_router()` (a page callback) will get its 3rd decoded as `'panel_context:page-panopoly_demo_panel::page_panopoly_demo_panel_panel_context::::'`.

However, when running via `drush runserver` that argument will still be encoded as `'panel_context%3Apage-panopoly_demo_panel%3A%3Apage_panopoly_demo_panel_panel_context%3A%3A%3A%3A'` and Panels won't work as expected.

This broke in Drush 8.1.* when d7-rs-router.php was added; in Drush 8.0.*, both Drush 6 & 7 used runserver-prepend.php. Looking at runserver-prepend.php, it has:

```
$_GET['q'] = substr(urldecode($request_path), $base_path_len + 1);
```

... whereas, d7-rs-router.php has:

```
$_GET['q'] = $_REQUEST['q'] = substr($url['path'], 1);
```

So, it was just that `urldecode()` that got lost, and this PR is adding that back. :-)

### DIFF
--- a/commands/runserver/d7-rs-router.php
+++ b/commands/runserver/d7-rs-router.php
@@ -52,7 +52,7 @@ if (file_exists('.' . urldecode($url['path']))) {
 }
 
 // Populate the "q" query key with the path, skip the leading slash.
-$_GET['q'] = $_REQUEST['q'] = substr($url['path'], 1);
+$_GET['q'] = $_REQUEST['q'] = urldecode(substr($url['path'], 1));
 
 // We set the base_url so that Drupal generates correct URLs for runserver
 // (e.g. http://127.0.0.1:8888/...), but can still select and serve a specific
--- a/tests/makefiles/lockfiles/git.lock.yml
+++ b/tests/makefiles/lockfiles/git.lock.yml
@@ -9,4 +9,4 @@ projects:
       url: 'https://git.drupalcode.org/project/atom.git'
       type: git
       branch: 8.x-1.x
-      revision: eef5044d467380991080cc6e6060de3b99832c27
+      revision: c6b3f3fe1edd24641c56d080bb94ad71824d9b42


greg-1-anderson:
LGTM. Thanks for tracking this down. The other reason folks are going to need to upgrade to Drush 8.4+ is to handle any contrib modules that adopt SemVer releases -- if anyone in D7 land decides to do that.

Will merge this once tests finish running.

dsnopek:
No rush to merge! This seems to work in my manual testing, but we still haven't gotten a full successful test run of Panopoly with this patch (and PHP 7.4 and Drush 8.4.*). The fails could be unrelated, but I'd feel more comfortable waiting for that to all be done. I've marked this issue as [WIP] in the meantime.

greg-1-anderson:
Well, I think it wouldn't hurt to merge this to 8.x and open follow-on issues for anything else you find, as this looks like an improvement. I'll keep it open as you wish, though.

While you're in here, could you either update the hash in the failing make tests, or mark them skipped so the test comes out green, though?

dsnopek:
> Well, I think it wouldn't hurt to merge this to 8.x and open follow-on issues for anything else you find, as this looks like an improvement.

Fair enough! Our tests appear to be almost passing - we've got just one test failure now, and it seems like it's not likely related to this issue. I've removed the "[WIP]" from the title.

> While you're in here, could you either update the hash in the failing make tests, or mark them skipped so the test comes out green, though?

Sure. Tests are almost finished running and it looks likely that they'll be all green this time. :-)

dsnopek:
@greg-1-anderson all green! merge if you like :-)

greg-1-anderson:
Great! Thanks for taking care of the unrelated failure.

Issue #4194, opened 2019-09-25
drush config:status does not respect excluded modules 


Chi-teck:
**Describe the bug**
"Config exclude" functionality [has been added](https://www.drupal.org/node/3079028) to Drupal Core recently. However, `drush cst` command still checks excluded configuration.

**To Reproduce**
Put the following code to settings.php:
`$settings['config_exclude_modules'] = ['field_ui']`
Check config status: `drush config:status`

**Expected behavior**
```
 [notice] No differences between DB and sync directory.
```

**Actual behavior**
```
 ------------------- ------------ 
  Name                State       
 ------------------- ------------ 
  core.extension      Different   
  field_ui.settings   Only in DB  
```

**System Configuration**

| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 8.8.x
| PHP version     | 7.3
| OS?             | Linux

**Additional information**
`drush cex` and `drush cim` commands work well.

claudiu-cristea:
I confirm the issue even I'm still using the contrib module ([Config Exclude](https://www.drupal.org/project/config_exclude)) which was the starting point of the core new feature.

claudiu-cristea:
And if you use the `config_update_ui` (latest from dev) and run:
```
drush config:diff core.extension
```
that shows no diff

JTubex:
Created #4201 that edits the config:status command so it uses the `config_exclude_modules` setting.

bircher:
:+1:

weitzman:
Merged #4201.

Chi-teck:
It might be useful to have an option to see the status with excluded configuration.

theteknocat:
Why is this still an issue almost a year later? What's the point of a feature like this in core when it just doesn't work? It's even documented in the settings.local.php example file you get with Drupal, yet it never works when using Drush, which devs use ALL THE TIME. Using config split is a poor solution to the problem.

PR #4554, opened 2020-09-25
Transform the storage to import instead of the active one for export

bircher:
in #4194 we added a export transformation to compare with a given storage.
We should have instead transformed that storage for import to do the comparison.


bircher:
See the bug report https://www.drupal.org/project/config_split/issues/3172440#comment-13836770

Chi-teck:
@bircher will this hide excluded modules from `drush cst` output?

bircher:
It should yes. The difference is that we compare what would be imported whereas now we compare what would be exported. I can see the use in both (or even just straight up comparing it without any transformation).

PR #4539, opened 2020-08-23
Replace abandoned Sami by Doctum

williamdes:
Replace Sami by [Doctum](https://github.com/code-lts/doctum)

PR #4556, opened 2020-09-25
Fix #4539, use  --ignore-parse-errors

williamdes:
Fix #4539


williamdes:
Anyway, thank you @weitzman for merging my work !
it is deployed now :tada: https://www.drush.org/api/10.x/

PR #4561, opened 2020-09-28
Remove composer branch-alias

### LINKED ISSUE #4560: Cannot require the dev release on 10.x
**Describe the bug**
There doesn't seem to be a way to install the 10.x branch release with composer. There was some discussion in follow-up of #4557 for this.

**To Reproduce**
Run `composer require drush/drush:"10.x-dev"`. It works but pulls in the `master` branch. However, the master branch has not been updated in a while and I think it won't be updated anymore.

**Expected behavior**
Pull the `10.x` branch

**Actual behavior**
It pulls the `master` branch

**Workaround**
I suppose a workaround could be to define a custom composer repository but that's an overkill.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 9.x
| PHP version     | 7.4
| OS?             | Mac

**Additional information**
There is a discussion in #4557 which would give additional context. As mentioned there, the fix is likely to remove the `branch-alias`. If we don't intend to keep the `master` branch updated, the alias is meaningless.

hussainweb:
Fixes #4560

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -117,9 +117,6 @@
       "sut/profiles/contrib/{$name}": ["type:drupal-profile"],
       "sut/themes/contrib/{$name}": ["type:drupal-theme"],
       "sut/drush/contrib/{$name}": ["type:drupal-drush"]
-    },
-    "branch-alias": {
-        "dev-master": "10.x-dev"
     }
   }
 }


weitzman:
Thanks. The Composer branch alias documentation is not too clear IMO.

PR #4562, opened 2020-09-30
allow to work with guzzle 7 also

mihaileu:


greg-1-anderson:
Seems fine to me. We should update composer.lock when composer.json changes.

mihaileu:
But why should we update the lock file, isn't it available as composer package ?

greg-1-anderson:
Try:
```
$ composer validate
```
This should give you an error if the lock file is not up-to-date with the composer.json file.

greg-1-anderson:
It might be a good idea to add `composer validate` to the Drush tests. Some projects do that.

greg-1-anderson:
Looks like this change would require us to bump up to PHP 7.2 for tests that run with guzzle 7. We could use composer-test-scenarios (or just a `composer require`) to pull back the guzzle version if we want to test with PHP 7.1.

I'm not sure how important it is to keep testing 7.1.

greg-1-anderson:
What did you do differently in d06776b than 38061ce? Drush's composer.json should generate [the same set of dependencies regardless of which version of PHP the update is executed with](https://github.com/drush-ops/drush/pull/4562/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R84) (in theory).

mihaileu:
the first composer was with --ignore-platform-reqs, the second one was launched on docker container with the correct dependencies

greg-1-anderson (approved):

greg-1-anderson:
+1 on this change, just want @weitzman to weigh in / merge. Since Drush 10 is typically only installed site-local, it probably doesn't matter what's in the lock file.

PR #4576, opened 2020-10-20
fix deploy hook documentation

bircher:
just noticed some confusing wording after I posted the link to the documentation (which in turn links to this file) on a bunch of drupal.org issues


weitzman:
The wording is still confusing IMO. Fixed in 5a81ed399

PR #4581, opened 2020-10-28
Fixes #4580: Allow Drush to work with composer/semver ^3.

### LINKED ISSUE #4580: Drupal core now depends on composer/semver:^3.0
**Describe the bug**
Drush depends on `composer/semver:^1.4`. Drupal 9.1.x just committed a patch to bump to `composer/semver:^3.0` in order to support Composer 2.

See https://www.drupal.org/project/drupal/issues/3128631

Consider allowing `composer/semver:^1.4|^3.0` depending on Drush's use of the semver library.

greg-1-anderson:
Add a Drupal 9.1 test. I'm not sure what our overall testing coverage strategy should be; just dropped this in so that 9.1 would be covered. Maybe scale back the 9.0 test to be stable and leave 9.1 as the HIGHEST test?  Maybe soon bump up so that we're testing just 9.1 and 9.2 (once we're at 9.1 stable)?


greg-1-anderson:
Reference: [drupal/#3128631](https://www.drupal.org/project/drupal/issues/3128631)

.circleci/config.yml
```diff
@@ -140,6 +140,36 @@ jobs:
...
+      - checkout
+      - run: sudo composer self-update --preview
+      - run: cp .docker/zz-php.ini /usr/local/etc/php/conf.d/
+      # We test with 9.0.0 so that \Unish\SecurityUpdatesTest can find an insecure drupal/core.
+      # Remove that assertion once testing with 9.0.0 becomes stale.
+      - run: composer require --dev drupal/core-recommended:~9.1.0 --no-update
+      # Replace alinks with semver for D9.
+      - run: composer remove --dev drupal/alinks --no-update
+      - run: composer require --dev drupal/semver_example:2.2.0 --no-update
+      - run: composer config platform.php 7.3.0
```
weitzman, line 164: Should be removed or changed to 7.4

.circleci/config.yml
```diff
@@ -140,6 +140,36 @@ jobs:
...
+      - run: composer update --no-scripts
```
weitzman, line 165: I think this line was added in https://github.com/drush-ops/drush/commit/e1f904cc49a676ad5240e83989d3dd86c7f26e20. Its not needed anymore in the 9.0 and 9.1 sections, I think?

weitzman (changes requested):
That test failure looks very related IMO. That test is where we use composer/semver. Prob 3.0 has breaking changes.

.circleci/config.yml
greg-1-anderson: We need the `composer update`, but we can remove `--no-scripts` now. composer-test-scenarios were in use for PHP 5 testing back when that was added.

greg-1-anderson:
I'll dig into the failure and see what it's doing.

.circleci/config.yml
weitzman: Oh right. Its the following `composer install` that's superfluous.

greg-1-anderson: The `composer install` was necessary after `composer update --no-scripts`, to run the scripts. I'll take it out along with the --no-scripts, because it is superfluous once that's gone.

.circleci/config.yml
andypost: I bet it still required as https://github.com/drush-ops/drush/blob/10.x/composer.json#L83-L84 sticking it on 7.1.3

.circleci/config.yml
```diff
@@ -124,17 +124,43 @@ jobs:
...
+      - run: php --version
+      - run: composer config platform.php 7.3.23
+      - run: composer update
+      - run: composer lint
+      - run: composer unit
+      - run: composer functional
+      - run: composer integration
+
+
+  # PHP 7.4 test with HIGHEST dependencies:
```
weitzman, line 142: 7.3

tests/integration/SecurityUpdatesTest.php
```diff
@@ -23,13 +23,6 @@ public function testInsecureDrupalPackage()
         $this->arrayHasKey($expected_package, $security_advisories);
         $this->assertEquals($expected_package, $security_advisories[$expected_package]['name']);
         $this->assertEquals($expected_version, $security_advisories[$expected_package]['version']);
-        // Remove this clause if we don't have an insecure release to use.
```
weitzman, line 26: I'm not too happy about losing this test coverage. If we have to lose it for Drupal 9.1 thats fine, but losing it for all our Druapals seems like a bad idea. This has caught bugs before. I'm not pacified by the fact that contrib remains tested.

weitzman (changes requested):

tests/integration/SecurityUpdatesTest.php
greg-1-anderson: It is bad, though, to pin all of our Drupal 9 tests to 9.0.0, as then we don't catch any changes to Drupal 9. 

I'll push one more iteration & see what you think.

greg-1-anderson:
Here's a MVP for Drupal 9.1 that maintains the existing pm:security test against drupal/core 9.0.0.  We could as follow-on work make pm:security more testable.

weitzman:
So whats the issue with cache bins? Is it just that the test is broken, or do cache:get and cache:set not work on 9.1? If its just a testing issue, I think this is mergeable as is. If not, we I feel like this is a pretty.

greg-1-anderson:
Yeah, it is a rushed effort. Maybe I can look at cache bins after Halloween. I thought it was better to have green tests and a Drush that works with Drupal 9.1.0-alpha sooner rather than later.

gitressa:
Thanks for fixing this so quickly @greg-1-anderson and @weitzman. Both Drupal `9.1.0-alpha1` and `9.1.x-dev` now works fine with Composer `2.0.3`.

PR #4583, opened 2020-10-30
entity:delete -> execute deletion of entities in small steps to avoid loading all at once

cgoffin:
When deleting lots of entities at once (+1000). It seems that the command hangs. I changed the command to process the deletions in small steps and added a progress bar to the output to inform the user.

### DIFF
--- a/src/Drupal/Commands/core/EntityCommands.php
+++ b/src/Drupal/Commands/core/EntityCommands.php
@@ -5,6 +5,8 @@ namespace Drush\Drupal\Commands\core;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drush\Commands\DrushCommands;
 use Drush\Utils\StringUtils;
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Output\OutputInterface;
 
 class EntityCommands extends DrushCommands
 {
@@ -24,12 +26,14 @@ class EntityCommands extends DrushCommands
      *
      * To delete configuration entities, see config:delete command.
      *
+     * @param \Symfony\Component\Console\Output\OutputInterface $output the symfony output interface.
      * @param string $entity_type An entity machine name.
      * @param string $ids A comma delimited list of Ids.
      * @param array $options
      *
      * @option bundle Restrict deletion to the specified bundle. Ignored when ids is specified.
      * @option exclude Exclude certain entities from deletion. Ignored when ids is specified.
+     * @option chunks Define how many entities will be deleted in the same step. Default is 10.
      * @usage drush entity:delete node --bundle=article
      *   Delete all article entities.
      * @usage drush entity:delete shortcut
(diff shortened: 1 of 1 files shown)


src/Drupal/Commands/core/EntityCommands.php
```diff
@@ -40,18 +44,22 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager)
...
      *   Delete all users except uid=1.
+     * @usage drush entity:delete node --chunks=5
+     *   Delete all node entities in steps of 5.
      *
      * @command entity:delete
      * @aliases edel,entity-delete
      * @throws \Exception
      */
-    public function delete($entity_type, $ids = null, $options = ['bundle' => self::REQ, 'exclude' => self::REQ])
+    public function delete(OutputInterface $output, $entity_type, $ids = null, $options = ['bundle' => self::REQ, 'exclude' => self::REQ, 'chunks' => self::REQ])
```
weitzman, line 54: I didn't know that OutputInterface would work here. Look kosher to you, @greg-1-anderson? 


Also, lets just default chunks to 50 or so. We get advantages to bulk deleting so I think 10 is too few.

weitzman (commented):
Great work.

src/Drupal/Commands/core/EntityCommands.php
cgoffin: I altered the default to 50 and fixed some coding standard issues. Why do you think the OutputInterface wouldn't work? I tested on my Drupal installation and it worked fine.

weitzman:
Thanks. This is our first progress bar in core.

I think its a. bit cleaner if we use $this->io()->createProgressBar() and $this->io()->progressAdvance(). Could you try that?

cgoffin:
I adjusted the code and on my local Drupal environment it worked.

src/Drupal/Commands/core/EntityCommands.php
```diff
@@ -60,23 +64,42 @@ public function delete($entity_type, $ids = null, $options = ['bundle' => self::
...
+            unset($result[0], $result[1]);
         }
 
-        if (empty($entities)) {
+        if (empty($result)) {
             $this->logger()->success(dt('No matching entities found.'));
         } else {
-            $storage->delete($entities);
-            $this->logger()->success(dt('Deleted !type entity Ids: !ids', ['!type' => $entity_type, '!ids' => implode(', ', array_keys($entities))]));
+            $chunks = $options['chunks'] ?? 50;
```
weitzman, line 78: Lets remove this line, and just default to 50 in the method params. That way the default is documented in help.

weitzman (changes requested):
Looking good. One minor comment below.

src/Drupal/Commands/core/EntityCommands.php
cgoffin: Did know that, nice. I did the adjustment.

weitzman:
Thanks @cgoffin.

PR #4586, opened 2020-11-07
Add the option to skip language autocreation on the command locale:import.

FlorentTorregrosa:
Hello,

I have recently discovered that now Drush has a built in command to import translations. I use (and contributed to) [Drush language](https://www.drupal.org/project/drush_language) to script import of custom translations on my projects.

I tried to switch to the Drush built-in command and I would like to add the possibility to not create automatically a language when importing translations. Option that exists in the Drush language commands.

Thanks for any feedbacks.


Sutharsan:
@FlorentTorregrosa Can you please describe a use case for this functionality to strengthen the case.

The code looks Ok, but I have doubts on the name/usage of the option. Instead of `--autocreate-language=0`, I propose `--not-create-language` which I think is more intuitive because it does not need an option value.

FlorentTorregrosa:
@Sutharsan thanks for the reply.

No problem for renaming the option. Waiting feedbacks if the use case is ok, before.

There are no command (to my knowledge) to test if a language exist or not and then handle the result of this command in scripts.

So here is an example of use case:

I need this option to simplify scripts, for example on multi-sites installation where custom translations are stored in the same folder for all websites, but websites have different languages enabled. I would like to be able to loop on translation files and if the language is not enabled, ignore it silently (or not) like done with Drush language https://gitlab.com/florenttorregrosa-drupal/docker-drupal-project/-/blob/8.x/scripts/tasks/update_translations.sh.

So no need to make a system to list which languages are enabled on which websites.

weitzman:
The negation option appears automatically when you default an option to `true`. See runserver's `--browser` option for an example.

That doesnt sound like a common use case to me but @Sutharsan would know better than I.

Sutharsan:
I understand the use case. I think it is not common because there is only a small number of multilingual multisites and only those with different languages per instance may suffer.

The alternatives are hard coding the language import per instance or writing a custom command.

I have second thoughts on the design decision to create a language when importing translations. It is like creating a role when adding it to a user, which we don't and we shouldn't. Creating a language is far reaching, it has an impact all over. Outside of development, it should be avoided. @weitzman Now that I'm giving it more thought, I wonder if we should change the default behavior for the locale:import command.

weitzman:
>  I wonder if we should change the default behavior for the locale:import command.

Makes sense to me. A PR sounds like a good next step.

I'm not sure when we could merge the PR though since we aren't contemplating a new major version at this time. The PR sounds like a backward compat break.

greg-1-anderson:
For commands that break bc, we could keep the existing implementation under the existing name, and create a new command with a different name using the desired implementation. The new name could have some defined postfix like ":experimental".  When the new major version of Drush comes out, then we can drop the old implementation and rename the command back to its original name, keeping the ":experimental" alias for one more major version.

Not sure on the general "naming things" end, but the strategy should work.

Sutharsan:
Sounds like a good compromise to me. The alternative is to add an option for now and remove it in the BC breaking release. But I let you decide, I have no experience with the kind of choices.

weitzman:
I'm about to release Drush 11, so we can do what we want here. I've updated the PR to what I think we want here. Please review. /cc @Sutharsan. Thanks.

FlorentTorregrosa:
Thanks @weitzman for updating the PR. Looks good to me.

PR #4589, opened 2020-11-14
Add ability for commands to show a Termynal on their docs page

weitzman:
image

I got the idea from the Termynals that are shown at https://fastapi.tiangolo.com/. Above is a screenshot of that page.

Unfortunately, these page elements wrap poorly on narrow screens (just the a narrow terminal window). I cant think of a way to make these responsive. So I'm abandoning this PR for now. Here is what this PR currently looks like for pm:list

image

### DIFF
--- a/docs/css/termynal.css
+++ b/docs/css/termynal.css
@@ -1,0 +1,101 @@
+/**
+ * termynal.js
+ *
+ * @author Ines Montani <ines@ines.io>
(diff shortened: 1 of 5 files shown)


weitzman:
Perhaps these would look better on the homepage of www.drush.org similar to https://cli.github.com/? Tabular output commands still would look off on narrow screens.

PR #4597, opened 2020-12-04
Fixes #4594: When installing Drupal6, properly respect an existing db…

### LINKED ISSUE #4594: Drupal 6 install issue with drush8
**Describe the bug**
The newer drush8 is handling Drupal 6 install differently and it is causing issues.

Specifically, if the database settings are already set in the settings.php and file is read-only, the site-install command will fail. 

Newer drush versions are trying to write to the 'sites/default/settings.php' and are failing to do so. There is a new function that does this (drush8/vendor/drush/drush/commands/core/site_install.drush.inc):
```
function _drush_core_site_install_add_db_url_to_settings($settingsfile, $db_spec) {
  $settingsContents = file_get_contents($settingsfile);
  if (preg_match('#^.db_url *=#', $settingsContents)) {
    return true;
  }
  // On D6, we have to write $db_url ourselves. In D7+, the installer does it.
  if (file_put_contents($settingsfile, "\n" . '$db_url = \'' . drush_convert_db_url(drush_get_option('db-url')) . "';\n", FILE_APPEND) === false) {
    return false;
```

I'd expect the preg_match() to find the existing db_url line but it apparently won't.

**To Reproduce**
Have a Drupal 6 settings.php file that sets $db_url through an include file (or have a Drupal 6 settings.php file that has $db_url setting already set).

**Expected behavior**
It seems that regexp ''#^.db_url *=#'' can be satisfied with an existing/fake line in settings.php e.g.
```
// The following line must be present to trick Drush >= 8.1.19 into thinking the db_url 
// is present in this file, otherwise drush site-install will try to write to this file:
#db_url =
```

**Actual behavior**
The `if (preg_match('#^.db_url *=#', $settingsContents)) {` condition will never return a TRUE value unless ?db_url is in the first line of settings.php.

**Workaround**
Manual intervention is needed to remove the added db_url setting (by the installer).

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.1.19+
| Drupal version? | 6.x
| PHP version     | 7.x

**Additional information**
n/a

attilafekete:
… setting (db_url) in settings.php.

### DIFF
--- a/commands/core/site_install.drush.inc
+++ b/commands/core/site_install.drush.inc
@@ -210,7 +210,11 @@ function drush_core_pre_site_install($profile = NULL) {
 
 function _drush_core_site_install_add_db_url_to_settings($settingsfile, $db_spec) {
   $settingsContents = file_get_contents($settingsfile);
-  if (preg_match('#^.db_url *=#', $settingsContents)) {
+  // There's no need to write $db_url if it's already set.
+  // The regexp below will allow one arbitrary character before $db_url.
+  // Allowing more/unlimited chars would be unwise as it would match $db_url
+  // occurrences inside comments.
+  if (preg_match('#^.?\$db_url *=#m', $settingsContents)) {
     return true;
   }
   // On D6, we have to write $db_url ourselves. In D7+, the installer does it.


greg-1-anderson:
I'm surprised that this regex has been wrong for so many years without anyone catching it. Thanks for the fix.

attilafekete:
@greg-1-anderson Thanks! Likely no one tried to install Drupal6 with drush8 :)

PR #4598, opened 2020-12-07
Try testing Windows with GitHub Actions.

greg-1-anderson:
This would let us drop Appveyor. Work-in-progress. Includes roave/bc and some duplicate functionality (linting, cs, code coverage) just as an investigation.  ssh is not as necessary for these jobs. Maintaining Circle for other things.


greg-1-anderson:
Oh yeah I dropped in a php 8 test just to see what happens; planning on taking that out here.

greg-1-anderson:
Relates to #4598

greg-1-anderson:
Looks like I still need to fix up some assumptions in composer.json here; will do that later.

greg-1-anderson:
The tests are at least running now, but I still need an sql server, and there are some phpunit upgrade issues e.g. stringContainsString that need to be addressed.

greg-1-anderson:
I don't plan on continuing here in the near future. There's still quite a bit of work to do to set up the GitHub actions environment to work with a database -- easy to do Linux-only, but the value-add here was supposed to be Windows, which is not as well supported at the moment. (The easiest way to get a database on Windows is to use an external service.) Since the Appveyor tests are passing at the moment, there isn't much motivation to do this work.

We can re-open if there is motivation to move to GitHub actions in the future.

PR #4599, opened 2020-12-08
Convert to phpunit 9 syntax.

greg-1-anderson:
Use phpunit 7 thru 9 with phpunit 9 syntax via phpunit-polyfills to cover php versions 7.1 through 8.0.  We won't be able to test with php 8.0 unless we upgrade to at least phpunit 8, so this is necessary for long-term maintenance.


weitzman (approved):
LGTM. Thanks.

PR #4603, opened 2020-12-14
Add site name and Drupal version to php:cli startup message

Chi-teck:
I think this would make more clear to users that PHP shell is running within Drupal context.
drush-php

### DIFF
--- a/src/Drupal/Commands/core/CliCommands.php
+++ b/src/Drupal/Commands/core/CliCommands.php
@@ -46,6 +46,14 @@ class CliCommands extends DrushCommands
         // Set the Drush specific history file path.
         $configuration->setHistoryFile($this->historyPath($options));
 
+        $configuration->setStartupMessage(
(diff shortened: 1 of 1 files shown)


Chi-teck:
The default color of startup message is not well readble, at least on my terminal. Maybe we should fix this too. Something like this `$configuration->setFormatterStyles(['aside' => ['cyan', 'black']]);`.

greg-1-anderson (approved):
LGTM. I'm open to alternate text colors too.

Chi-teck:
Color issue might be specific to my terminal color scheme.

damiankloip:
Thanks for the PR, this LGTM also. It's useful to have the version there, not sure why we didn't do that before to be honest :)

RE the colours; I think the defaults are fine (at least in most colour schemes I've ever used). Feel free to spin off a follow up PR though.

PR #4606, opened 2020-12-15
Move Migrate Run commands in Drush core

### LINKED ISSUE #2926: Add migrate_run instead of migrate_tools

claudiu-cristea:
Fixes #2926. Development started in #3402.

## Status of porting changes from Migrate Tools since #3402

* [Using '--limit' with '--feedback' on import does not work](https://www.drupal.org/project/migrate_tools/issues/2919108): Already fixed in #3402. However, I've added a regression test `::testImportingWithLimitAndFeedback()`.
* [Implement --idlist options on rollback and messages](https://www.drupal.org/project/migrate_tools/issues/2798363) and followups [#2955032](https://www.drupal.org/project/migrate_tools/issues/2955032), [#3038549](https://www.drupal.org/project/migrate_tools/issues/3038549): Implemented with the exception of messages command option, which doesn't work in Migrate Tools either. We can fill a followup if `--idlist` is needed for messages. Also, `--idlist-delimiter` was not added because: (1) the name is misleading, actually it refers to the delimiter of columns from the same ID but the name could make you think that is the delimiter between IDs and (2) I don't think there's a real use case to use something different than colon ':'. But if there is a real need, we need a better name and understand that edge case.
* [using drush mi --update flags all rows as needing an update, even if used in conjunction with an --idlist filter](https://www.drupal.org/project/migrate_tools/issues/3015386): Fixed, added regression test `::testImportingWithUpdateAndIdlist()`.
* [Unset the key_value "migrate_last_imported" on MigrateEvents::POST_ROLLBACK](https://www.drupal.org/project/migrate_tools/issues/3011996): Fixed, added test coverage.
* [Symfony Console Error "Cannot add a required argument after an optional one" on migrate:status](https://www.drupal.org/project/migrate_tools/issues/3021881): Already fixed in #3402.
* [Drush migrate-import should return a non-zero exit code when migration fails](https://www.drupal.org/project/migrate_tools/issues/2999812): Already fixed in #3402.
* [The "--limit" option does not accept a value](https://www.drupal.org/project/migrate_tools/issues/3024399): Fixed all commands options defaults.
* [Remove $sourceId workaround on prepareRow now that issue is fixed](https://www.drupal.org/project/migrate_tools/issues/3081731): Fixed.
* [use Symfony Console's progress bar](https://www.drupal.org/project/migrate_tools/issues/3075131): Fixed, added test coverage.
* [Migrate support for deleting items no longer in the incoming data](https://www.drupal.org/project/migrate_tools/issues/2809433): Implemented but used a slightly different approach and used the _rsync style_ `--delete` option instead of `--sync`.
* [If getMessageIterator() includes useful IDs, don't print the source_ids_hash column in drush mmsg command](https://www.drupal.org/project/migrate_tools/issues/2960221): Fixed.

## Follow-ups

* **[needs decision]** [Add continue-on-failure option](https://www.drupal.org/project/migrate_tools/issues/3092369): I'm not sure that such an option should exist. To be decided!
* **[needs scope clarification]** ['status' is not a log level](https://www.drupal.org/project/migrate_tools/issues/3073748): It wasn't clear for me the scope of this issue.
* #4622: Convert functional MigrateRunnerTest to integration tests


src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -0,0 +1,694 @@
...
+                $sourceIdValuesList = MigrateUtils::parseIdList($userData['options']['idlist']);
+                $keys = array_keys($migration->getSourcePlugin()->getIds());
+                foreach ($sourceIdValuesList as $sourceIdValues) {
+                    $migration->getIdMap()->setUpdate(array_combine($keys, $sourceIdValues));
+                }
+            }
+        }
+
+        // Initialize the Symfony Console progress bar, if case.
+        $this->initProgressBar($migration, $userData['options']);
```
weitzman, line 359: I didnt yet fully investigate this PR to see how we are doing progress. I just want o point out that we added one in #4583 and it used `$this->io()` which I dont see here.

weitzman (commented):
Quick initial review. Thanks for your hard work on this.

includes/migrate_runner.inc
```diff
@@ -0,0 +1,20 @@
+<?php
```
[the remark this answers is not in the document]
claudiu-cristea: @weitzman this is temporary and will be deprecated as soon as https://www.drupal.org/project/drupal/issues/2952291 lands. Migrate Tools is also offering something similar.

> Could we hide it a bit more?

I don't understand how to do that :)

src/Drupal/Commands/core/MigrateRunnerCommands.php
claudiu-cristea: @weitzman unfortunately, for testing purposes, I need to pass the 3rd argument to `\Symfony\Component\Console\Helper\ProgressBar::__construct()` (>= Symfony 4). Otherwise I cannot predict the progress bar output, thus I cannot properly test. But `$this->io->progressStart()` doesn't allow to pass that argument.

claudiu-cristea: @weitzman I have 2 options:

- Keep the current approach
- Switch to `$this->io->progressStart()` but dramatically simplify that test.

weitzman: Lets keep the current approach and keep the testing. Thanks.

includes/migrate_runner.inc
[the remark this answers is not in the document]
weitzman: What code includes this file?

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -0,0 +1,706 @@
...
+        $this->keyValue = $keyValueFactory->get('migrate_last_imported');
+        $this->commandProgress = $commandProgress;
+    }
+
+    /**
+     * List all migrations with current status.
+     *
+     * @command migrate:status
+     *
+     * @drupal-dependencies migrate
```
weitzman, line 80: what is `@drupal-dependencies`? Maybe its a dupe of @validate-module-enabled?

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -0,0 +1,706 @@
...
+     *   imported: Imported
+     *   needing_update: Needing update
+     *   unprocessed: Unprocessed
+     *   last_imported: Last Imported
+     * @default-fields id,status,total,imported,needing_update,unprocessed,last_imported
+     *
+     * @return \Consolidation\OutputFormatters\StructuredData\RowsOfFields
+     *   Migrations status formatted as table.
+     */
+    public function status(?string $migrationIds = null, array $options = ['tag' => self::REQ, 'names-only' => false]): RowsOfFields
```
weitzman, line 116: I like the nullable string type hint and return type declaration.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -0,0 +1,706 @@
...
+     * @option progress Show progress bar
+     *
+     * @usage migrate:import --all
+     *   Perform all migrations
+     * @usage migrate:import --all --no-progress
+     *   Perform all migrations but avoid the progress bar
+     * @usage migrate:import --tag=user,main_content
+     *   Import all migrations tagged with user and main_content tags
+     * @usage migrateimport classification,article
+     *   Import new terms and nodes using migration 'classification' and 'article'
```
weitzman, line 253: Words enclosed in apostophes like 'classification' should now be enclosed in`<info>`. This helps us show nicer docs on the web. The tag gets replaced when viewed on the CLI. For example see https://github.com/drush-ops/drush/blob/8fd1118cbb4bff069ade0f1aa89ed9d6df8ba496/src/Commands/core/SiteInstallCommands.php#L29 and https://www.drush.org/commands/10.x/site_install/#arguments.

There are numerous instances of this in this file.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -0,0 +1,706 @@
...
+    {
+        /** @var \Drupal\migrate\Plugin\MigrationInterface $migration */
+        $migration = $this->getMigrationPluginManager()->createInstance($migrationId);
+        if ($migration) {
+            $status = $migration->getStatus();
+            if ($status == MigrationInterface::STATUS_IDLE) {
+                $this->logger()->warning(dt('Migration @id is already Idle', ['@id' => $migrationId]));
+            } else {
+                $migration->setStatus(MigrationInterface::STATUS_IDLE);
+                $this->logger()->log('status', dt('Migration @id reset to Idle', ['@id' => $migrationId]));
```
weitzman, line 499: We usually use the success() method here instead of log(). In general we prefer the specific methods to rather than calling log($LEVEL)

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -0,0 +1,706 @@
...
+            }
+        } else {
+            $this->logger()->error(dt('Migration @id does not exist', ['@id' => $migrationId]));
```
weitzman, line 502: I think this should be an Exception

tests/functional/MigrateRunnerTest.php
```diff
@@ -0,0 +1,406 @@
...
+namespace Unish;
+
+use Webmozart\PathUtil\Path;
+
+/**
+ * @group slow
+ * @group commands
+ * @coversDefaultClass \Drush\Drupal\Commands\core\MigrateRunnerCommands
+ */
+class MigrateRunnerTest extends CommandUnishTestCase
```
weitzman, line 12: Any chance this can extend UnishIntegrationTestCase? See \Unish\AnnotatedCommandCase as an example that also brings in woot module. Integration tests dont reinstall drupal so they are faster.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -0,0 +1,706 @@
...
+     *
+     * @return \Drupal\migrate\Plugin\MigrationInterface[][]
+     *   An array keyed by migration tag, each value containing an array of
+     *   migrations or an empty array if no migrations match the input criteria.
+     *
+     * @throws \Drupal\Component\Plugin\Exception\PluginException
+     */
+    protected function getMigrationList(?string $migrationIds, ?string $tags): array
+    {
+        $migrationIds = array_filter(array_map('trim', explode(',', $migrationIds)));
```
weitzman, line 607: Perhaps use `\Drush\Utils\StringUtils::csvToArray` here.

weitzman (changes requested):
Should we add a composer conflict with migrate_tools module <=5? Would prevent some chaos around duplicate commands being declared.

claudiu-cristea:
> Should we add a composer conflict with migrate_tools module <=5? Would prevent some chaos around duplicate commands being declared.

Good point. I've added also `"drupal/migrate_run": "*"` as this is a full replacement.

includes/migrate_runner.inc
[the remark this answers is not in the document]
claudiu-cristea: See `\Drush\Drupal\Commands\core\MigrateRunnerCommands::import()`

tests/functional/MigrateRunnerTest.php
claudiu-cristea: Well, right now the conversion is not straight as many some tests require clean start but I see your point. As this PR is already big, is it a possibility that we're doing this in a followup?

src/Drupal/Commands/core/MigrateRunnerCommands.php
claudiu-cristea: I have no idea how these tags landed here. It's possible that they were copied from an early variant of `migrate_tools`, still not sure. Removed.

tests/functional/MigrateRunnerTest.php
weitzman: Yes, thats fine

includes/migrate_runner.inc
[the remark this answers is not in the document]
weitzman: OK, how about we move the .inc file to src/Drupal/Migrate? I just want to hide it a bit as its unimportant for 99% of drush users.

src/Drupal/Commands/core/MigrateRunnerCommands.php
claudiu-cristea: I think it works. However, `StringUtils::csvToArray()` might need an outer `array_values()` as the result shows index gaps on consecutive commas.

src/Commands/core/DocsCommands.php
```diff
@@ -196,6 +196,18 @@ public function exampleCommand()
         self::printFileTopic($this->commandData);
     }
 
+    /**
+     * Defining and running migrations.
+     *
+     * @command docs:migrate
+     * @hidden
+     * @topic TBD
```
weitzman, line 204: This seems to have reverted. You had a path here earlier.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -0,0 +1,710 @@
...
+
+        return $row;
+    }
+
+    /**
+     * Perform one or more migration processes.
+     *
+     * @command migrate:import
+     *
+     * @drupal-dependencies migrate
```
weitzman, line 223: Please remove

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -0,0 +1,710 @@
...
+                break;
+            case MigrationInterface::STATUS_DISABLED:
+                $this->logger()->warning(dt('Migration @id is disabled', ['@id' => $migrationId]));
+                break;
+            case MigrationInterface::STATUS_STOPPING:
+                $this->logger()->warning(dt('Migration @id is already stopping', ['@id' => $migrationId]));
+                break;
+            default:
+                $migration->interruptMigration(MigrationInterface::RESULT_STOPPED);
+                $this->logger()->log('status', dt('Migration @id requested to stop', ['@id' => $migrationId]));
```
weitzman, line 456: please use ->success() instead of ->log()

weitzman (approved):

weitzman:
Thanks @claudiu-cristea. Great work.

PR #4617, opened 2020-12-28
Mysql 8.4616

AdamPS:
Patch tested with MySql 8.  Also rough fake testing for MariaDB by hard-coding in a version string.

### DIFF
--- a/src/Sql/SqlMysql.php
+++ b/src/Sql/SqlMysql.php
@@ -111,9 +111,12 @@ EOT;
 
             // For MariaDB, ALTER USER was introduced in 10.2 so before that use SET PASSWORD.
             // @see Drupal\Core\Database\Driver\mysql\Connection::getMariaDbVersionMatch()
-            // in Drupal 9 (code copied here for Drupal 8 support).
+            // in Drupal 9.
+            $this->su();
+            $this->alwaysQuery('SELECT VERSION();');
+            $version = trim($this->getProcess()->getOutput());
             $regex = '/^(?:5\.5\.5-)?(\d+\.\d+\.\d+).*-mariadb.*/i';
-            preg_match($regex, Database::getConnection()->version(), $matches);
+            preg_match($regex, $version, $matches);
             if (!empty($matches[1]) && version_compare($matches[1], '10.2') < 0) {
                 $sql[] = sprintf("SET PASSWORD FOR %s = PASSWORD('%s');", $user, $dbSpec['password']);
             }
--- a/src/Sql/SqlMysql.php
+++ b/src/Sql/SqlMysql.php
@@ -2,7 +2,6 @@
 
 namespace Drush\Sql;
 
-use Drupal\Core\Database\Database;
 use PDO;
 
 class SqlMysql extends SqlBase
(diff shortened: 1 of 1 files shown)


src/Sql/SqlMysql.php
```diff
@@ -107,7 +108,21 @@ public function createdbSql($dbname, $quoted = false)
             $user = sprintf("'%s'@'%s'", $dbSpec['username'], $domain);
             $sql[] = sprintf("DROP USER IF EXISTS %s;", $user);
             $sql[] = sprintf("CREATE USER %s IDENTIFIED WITH mysql_native_password;", $user);
-            $sql[] = sprintf("SET PASSWORD FOR %s = PASSWORD('%s');", $user, $dbSpec['password']);
+
+            // For MariaDB, ALTER USER was introduced in 10.2 so before that use SET PASSWORD.
+            // @see Drupal\Core\Database\Driver\mysql\Connection::getMariaDbVersionMatch()
+            // in Drupal 9.
+            $this->su();
```
weitzman, line 115: This is already handled by the caller. See \Drush\Sql\SqlBase::createdb
```suggestion
```

src/Sql/SqlMysql.php
```diff
@@ -107,7 +108,21 @@ public function createdbSql($dbname, $quoted = false)
...
+            $this->alwaysQuery('SELECT VERSION();');
+            $version = trim($this->getProcess()->getOutput());
+            $regex = '/^(?:5\.5\.5-)?(\d+\.\d+\.\d+).*-mariadb.*/i';
+            preg_match($regex, $version, $matches);
+            if (!empty($matches[1]) && version_compare($matches[1], '10.2') < 0) {
```
weitzman, line 120: Lets use the 3rd argument of version_compare(). See https://www.php.net/version_compare. Untested suggestion:
```suggestion
            if (!empty($matches[1]) && version_compare($matches[1], '10.2', '<')) {
```

src/Sql/SqlMysql.php
AdamPS: Unfortunately that doesn't work - createdb does it too late.  We need to have su now for the query `SELECT VERSION();` that is run immediately.

weitzman: Is it possible to determine version without admin perms? That would be generally useful. We could then make a sql SqlMariaDb class that inherits from SqlMysql and specializes createDb accordingly. Very few sites have entered `su` creds because they dont need to create a DB.

AdamPS: In general yes we can determine version without admin perms and it would be much easier than the code here because we could call `Drupal\Core\Database\Driver\mysql\Connection::getMariaDbVersionMatch()` (as soon as D8 is no longer supported.)  The only reason we need su now is because the database and user haven't been created yet - we are about to create them.  This code is inside `if (isset($db_superuser))` so it won't trouble all the sites that have not entered `su` creds.

AdamPS: This whole complex branch of code is only needed until EOL of D8 - after that the minimum maria version of D9 is already above the problem point.  Seeing as hardly anyone runs this code (like you say they mostly don't enter DB creds) and it took months for anyone to notice it was broken, we could just forget about it.  We could stick with ALTER USER and it's hard luck for anyone on the very old maria version.

weitzman: ALTER USER is needed by recent versions of mariadb and we dont use ALTER USER anywhere, right? So we cant just ignore this issue, if I'm reading this right.

AdamPS: Correct we cannot ignore this issue.  The correct patch that works for all database versions supported in Drupal 9 is:

```
-            $sql[] = sprintf("SET PASSWORD FOR %s = PASSWORD('%s');", $user, $dbSpec['password']);
+           $sql[] = sprintf("ALTER USER %s IDENTIFIED BY '%s';", $user, $dbSpec['password']);
```

Everything else is only there for old Maria versions still supported in D8.  If we add it now we would presumably then remove it again in 10 months time.  Please let me know if you would prefer to commit the simple patch now.

weitzman:
Lets do the simple patch. Please add a code comment indicating that MariaDB >= 10.2 is required ... Thanks for your persistence on this.

#### Background
- MariaDB 10.1. [The first stable release was in October 2014, and it will be supported until October 2020.](https://mariadb.com/kb/en/changes-improvements-in-mariadb-101/)
- MariaDB 10.2. [The first stable release was in May 2017, and it will be supported until May 2022](https://mariadb.com/kb/en/changes-improvements-in-mariadb-102/)

AdamPS:
Good idea to check what versions are supported - here is the simpler version.

Issue #4620, opened 2020-12-30
Deprecated Symfony Process usage


Sweetchuck:
## Deprecated Symfony Process usage

Passing a command as string when creating a "Symfony\Component\Process\Process" instance is deprecated since Symfony 4.2, pass it as an array of its arguments instead, or use the "Process::fromShellCommandline()" constructor if you need features provided by the shell.

**call stack**
1. `\Drush\TestTraits\DrushTestTrait::drush` line: 81
2. `\Drush\TestTraits\CliTestTrait::execute` line: 96

## To Reproduce

1. Write tests with `\weitzman\DrupalTestTraits\ExistingSiteBase` to test custom Drush commands.
2. Call `$this->drush('my:custom', ...);`

## System Configuration
| Q               | A        |
| --------------- | --------:|
| Drush version   |  10.3.6  |
| Drupal version  |   9.1.0  |
| PHP version     |   7.3.24 |
| symfony/process |   4.4.16 |

PR #4621, opened 2020-12-30
Issue #4620 - Support both array and string $command in symfony/process

Sweetchuck:
Issue #4620 - Support both array and string $command in symfony/process


weitzman:
Thanks.

PR #4625, opened 2021-01-02
Re-enable cache command tests.

weitzman:


weitzman:
I'm not sure why CI is failing. The cache get/set commands are working locally and working when I ssh into the CI server.

PR #4629, opened 2021-01-12
Fixes #4628. Core:rsync should escape ssh.options if it contains sing…

### LINKED ISSUE #4628: core:rsync should escape ssh.options if it contains single quotes
**Describe the bug**
When a site alias' ssh options contains single quotes, these should be escaped in the ssh options passed to rsync. When these are not escaped, the rsync command fails.

I encountered this problem when connecting to a site using a ssh ProxyCommand.

**To Reproduce**
Add a site alias:

```
production:
  host: example.com
  user: username
  root: /home/username/domains/example.com/public_html
  uri: https://www.example.com
  ssh:
    options: -p 2918 -o "ProxyCommand ssh -o 'ForwardAgent yes' user@otherhost.com 'ssh-add && nc %h %p'"
```

**Solution**
Please see the patch below. I've done a simple str_replace to escape any single quotes within ssh.options:

```
diff --git a/src/Commands/core/RsyncCommands.php b/src/Commands/core/RsyncCommands.php
index e7e45b551..2280e328e 100644
--- a/src/Commands/core/RsyncCommands.php
+++ b/src/Commands/core/RsyncCommands.php
@@ -74,6 +74,10 @@ class RsyncCommands extends DrushCommands implements SiteAliasManagerAwareInterf
         $parameters[] = Escape::shellArg($this->targetEvaluatedPath->fullyQualifiedPath());
 
         $ssh_options = $this->getConfig()->get('ssh.options', '');
+
+        // Escape any single quotes in ssh.options.
+        $ssh_options = str_replace("'","'\''",$ssh_options);
+
         $exec = "rsync -e 'ssh $ssh_options'". ' '. implode(' ', array_filter($parameters));
         $process = $this->processManager()->shell($exec);
         $process->run($process->showRealtime());
```

ruuds:
This is a fix for https://github.com/drush-ops/drush/issues/4628


weitzman:
Lets leave this to the person editing the config.

ruuds:
@weitzman i tried, but i wasn't allowed to get the following configuration to work no matter what i tried:

` options: -p 2918 -o "ProxyCommand ssh -o 'ForwardAgent yes' user@otherhost.com 'ssh-add && nc %h %p'"`

Further, i don't think the person editing the config should have any knowledge of the drush internals where the ssh options are wrapped in single quotes.

When i look at https://github.com/consolidation/site-process/blob/main/src/Util/Escape.php, there is some custom way of escaping done for commands there. Maybe that should be used instead?

Issue #4632, opened 2021-01-15
drush make --force-gitinfofile doesn't write for modules with semantic versioning


jrglasgow:
**Describe the bug**
drush make --force-gitinfofile
doesn't work with Semantic versioning

**To Reproduce**
Run 
drush make --force-gitinfofile
and notice that the info file for modules using semantic versioning doesn't get adjjusted

**Expected behavior**
What did you expect would happen?
the info file would get the tag added as a version

**Actual behavior**
What happened instead?
modules using semantic versioning  don't get modified

**Workaround**
Is there another way to do the desired action?
manually editing the info file after make process

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.4.2 (please be specific, and try latest release)
| Drupal version? | 9.x/8.x
| PHP version     | 7.x
| OS?             | Mac/Linux

**Additional information**
I know in the days of Drupal 9 and composer why am I using "drush make"... I wish I didn't have to but because of some customer requirements I don't have a choice at the moment.

PR #4633, opened 2021-01-15
#4632 check to see if we are using semantic versioning before decidin…

jrglasgow:
…g that we are not using info.yml files

### DIFF
--- a/commands/pm/pm.drush.inc
+++ b/commands/pm/pm.drush.inc
@@ -2245,13 +2245,8 @@ function drush_find_empty_directories($dir, $exclude = array()) {
 function drush_pm_inject_info_file_metadata($project_dir, $project_name, $version, $datestamp) {
   // `drush_drupal_major_version()` cannot be used here because this may be running
   // outside of a Drupal context.
-  $yaml_format = TRUE;
-  if (
-      strpos($version, 'x') && // non-semantic versioning
-      substr($version, 0, 1) < 8 // core version >= 8
-    ){
-      $yaml_format = FALSE;
-  }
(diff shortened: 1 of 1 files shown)


commands/pm/pm.drush.inc
```diff
@@ -2245,7 +2245,13 @@ function drush_find_empty_directories($dir, $exclude = array()) {
 function drush_pm_inject_info_file_metadata($project_dir, $project_name, $version, $datestamp) {
   // `drush_drupal_major_version()` cannot be used here because this may be running
   // outside of a Drupal context.
-  $yaml_format = substr($version, 0, 1) >= 8;
+  $yaml_format = TRUE;
+  if (
+      strpos($version, 'x') && // non-semantic versioning
```
greg-1-anderson, line 2250: Semver versions might contain an x e.g. for dev branches; perhaps a regex to test for 8.x-1.* et. al. would be better.

jrglasgow:
@greg-1-anderson,
Even though I don't see any [valid release options listed](https://www.drupal.org/node/1015226) that include an 'x' in it your point is taken.

I have modified my fix to use a [regular expression](https://regex101.com/r/SWWp97/1/) which is working in my testing. I tested in Regex101.com  and using this [test script with various possible release versions](https://extendsclass.com/php-bin/a4a0abd) before testing it in Drush and committing it.

greg-1-anderson:
I haven't used Drush make in such a long time; does make not have the option to pull from a dev branch? In any event, thank you for using the regex anyway; it is good to be technically correct, even if it is not relevant for practical scenarios.

The 8.8 test is failing due to an uncaught status warning (Drupal version is no longer supported). I don't know why the Drupal 6 tests are failing. Neither is related to this PR though so I'm going to go ahead and merge.

PR #4644, opened 2021-01-19
"Redispatch to local" confused by symlinks

### LINKED ISSUE #4643: "Redispatch to local" confused by symlinks
**Describe the bug**
If your vendor paths contain symlinks somewhere along the way then `RedispatchToSiteLocal::redispatchIfSiteLocalDrush()` can fail to realise that "the site-local Drush is us".  This results in a re-dispatch with the exact same paths.

I admit it's an unusual case.  However the consequences of the bug are nasty and hard to debug; the fix is simple and safe.

**To Reproduce**
Symlink your vendor dir to that of another site.  (Also do it for a handful of other directories and you get a form of multisite without 

**Expected behavior**
It works.

**Actual behavior**
An infinite loop of process spawning until all system resources are used up. 

**Workaround**
None known.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10..3.6
| Drupal version? | Any
| PHP version     | 7.4.3
| OS?             | Linux

AdamPS:
Fixes  #4643

### DIFF
--- a/src/Preflight/RedispatchToSiteLocal.php
+++ b/src/Preflight/RedispatchToSiteLocal.php
@@ -35,7 +35,7 @@ class RedispatchToSiteLocal
         }
 
         // If the site-local Drush is us, then we do not need to redispatch.
-        if (Path::isBasePath($vendor, $siteLocalDrush)) {
+        if (Path::isBasePath(realpath($vendor), realpath($siteLocalDrush))) {
             return false;
         }
 


weitzman:
We've changed that line a few times and removed realpath() I think. I'm disinclined to change it for an edge case. Sorry.

AdamPS:
OK no problem I can keep a local patch

PR #4649, opened 2021-02-01
Fix #4039. Experiment with more terse renderException() and Collision

### LINKED ISSUE #4039: Make error output less verbose
When the error is caused by wrong input Drush should only print relevant message unless more verbose output was requested via `-v` option.

Example:
```
$ drush sql-query 'WRONG QUERY'

In SqlCommands.php line 194:
                 
  Query failed.  
                 

sql:query [--result-file [RESULT-FILE]] [--file FILE] [--file-delete] [--extra EXTRA] [--db-prefix] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-d|--debug] [-y|--yes] [--no] [--remote-host REMOTE-HOST] [--remote-user REMOTE-USER] [-r|--root ROOT] [-l|--uri URI] [--simulate] [--pipe] [-D|--define DEFINE] [--database [DATABASE]] [--target [TARGET]] [--db-url DB-URL] [--notify [NOTIFY]] [--xh-link XH-LINK] [--druplicon] [--] <command> [<query>]
```

In this case `In SqlCommands.php line 194:`  and `sql:query` signature are not related to the error. The only relevant information that should be printed is `Query failed`. The current output is confusing.

A user may provide `-v` option to increase the verbosity of messages. However in this case it does not work well. Drush prints stack trace instead of PDO exception message which would help the user to fix the query.

Worth noting, this works pretty well in Drush 8.

weitzman:
Here is a frankenPR which accumulates two experiments in better exception and error presentation. Lets discuss and implement the pieces that make sense.

1. Implement \Drush\Application::renderException so that Console does not print the command help at end of each exception (e.g. `sql:query [--result-file ...`). This is a minimal change, and fixes the part of #4039 that I agree with. 
1. Instructs Symfony Console runner not to handle Exceptions. Now that we are responsible for presenting errors and exceptions, we use [Collision](https://github.com/nunomaduro/collision). See screenshots below. This error handler is the default in Laravel. Its presentation is verbose but I think it shows helpful info. The main components

1. Error/Exception message.
1. A code excerpt with failing line and surrounding lines.
1. Stack trace. 2 frames are shown by default, or all frames when in Verbose mode. Arguments are also shown, noting their types. 

Some risks of Collision approach
- I tried adding the new rendering into our existing ErrorHandler. This keeps our error levels logic the same. Note that we dont currently take over from Drupal's error handler so after bootstrap - thats another decision to make.
- I also tried registering Collision to handle all errors and exceptions and that’s great but all of a sudden php notices stopped execution whereas in current just we log a Notice and move on even though I have E_ALL enabled. 
- Collision will be great when folks are authoring commands but a lot of Drush is running other people's commands. maybe this is too much info by default. We could look into configurable verbosity (i.e. omit code excerpt and stack trace by default) if needed.
- Color woes - my terminal had hidden line numbers and had some less readable text until I changed to a stock iTerm theme
- The actual code excerpt can be unhelpful but that’s life as a developer. The stack trace often points to the real fix point.

err

_Below is with Verbose. Note additional stack frames in the trace._

err_vebose

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,7 @@
     "grasmash/yaml-expander": "^1.1.1",
     "guzzlehttp/guzzle": "^6.3 || ^7.0",
     "league/container": "~2",
+    "nunomaduro/collision": "^3.0 || ^4.0 || ^5.0",
     "psr/log": "~1.0",
     "psy/psysh": "~0.6",
     "symfony/event-dispatcher": "^3.4 || ^4.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "2fe319553915c3f976a85824e1936555",
+    "content-hash": "dc1c5f428a06dfd19a4ddcb3a6c5a5da",
     "packages": [
         {
             "name": "chi-teck/drupal-code-generator",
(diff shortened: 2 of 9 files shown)


greg-1-anderson:
I like the direction here, but don't have any opinions on the best way to approach the challenges yet.

PR #4657, opened 2021-02-08
Validate that ext-zip is available in pm:security-php.

weitzman:


src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -154,14 +154,12 @@ protected function calculateSecurityUpdates($composer_lock_data, $security_advis
...
      *
-     * Thanks to https://github.com/FriendsOfPHP/security-advisories
-     * and Symfony for providing this service.
-     *
      * @param array $options
      *
      * @return UnstructuredData
      * @throws \Exception
      * @command pm:security-php
+     * @validate-php-extension ext-zip
```
joestewart: Composer won't allow installation at all if the zip extension isn't installed will it?

https://github.com/drush-ops/drush/blob/10.x/composer.lock#L1077

weitzman: True, but one can use --ignore-platform-reqs to bypass.

weitzman: Its too bad Enlightn bakes in this Zipfile cache. Would be better to factor that out as an optional dependency.

paras-malhotra:
@weitzman, the Enlightn security checker (from v1.8.0) no longer requires `ext-zip` via https://github.com/enlightn/security-checker/pull/19.

weitzman:
Thanks much. Drush PR at https://github.com/drush-ops/drush/pull/4732. Will merge soon.

PR #4670, opened 2021-02-22
Add a command to rebuild all display- and form-modes

ndf:
This is a command that let you recreate all display- and form-modes with `$ drush entity:rebuild-display-config`.

In https://www.drupal.org/project/drupal/issues/2915036 we introduces a functionality that fires on field_insert on a per entity_type and bundle bases.
This command does the same but then for all entity_types and bundles.

There is no regression or misconfiguration issue. The command fixes possible misconfigurations.


ndf:
Ok cool the PR is here :) Now it won't be lost.
Please give me some time to doublecheck everything.

weitzman:
Thanks. Some quick notes:
- Any error handling needed?
- A preview of what updates will be confirmed before confirmation
- Can this move to src/Drupal/commands. Then it won't need @bootstrap full since all commands there are full.
- Testing
- "Rebuild display configuration" should be "form and display configuration"

ndf:
Still work-in-progress; A preview of what updates will be confirmed before confirmation YAY that's done

ndf:
| Can this move to src/Drupal/commands. Then it won't need @bootstrap full since all commands there are full.
That's done too

src/Drupal/Commands/core/drush.services.yml
```diff
@@ -41,10 +45,10 @@ services:
     tags:
       -  { name: drush.command }
   messenger.commands:
-      class: \Drush\Drupal\Commands\core\MessengerCommands
-      arguments: ['@messenger']
-      tags:
-        -  { name: drush.command }
```
ndf, line 47: unrelated tab -> space conversion

src/Drupal/Commands/core/drush.services.yml
```diff
@@ -18,10 +18,14 @@ services:
     tags:
       -  { name: drush.command }
   entity.commands:
-      class: \Drush\Drupal\Commands\core\EntityCommands
-      arguments: ['@entity_type.manager']
-      tags:
-        -  { name: drush.command }
```
ndf, line 24: unrelated tab -> space conversion

ndf:
@weitzman I think all the needs-work things are fixed. The ticket was locked quit a while, because of an unrelated build-check failure.
Would be awesome if this issue could be merged :)

weitzman:
I changed target branch to 11.x. There is some unrelated changes showing in the PR. I'm still undecided about this in Drush core. When would users need to run this?

ndf:
Sorry I missed your reply.
Just now it's rebased to the 11.x branch.

>  I'm still undecided about this in Drush core. When would users need to run this?
The command is useful for sites where the exported config of display is different than the actual (in database stored) config.
What I have seen a lot, is that when a field is added to a bundle and then the config is exported, the actual export contains many more (unrelated) changes. This happens if a display is not re-saved every time the dependencies of the display changes.
The goal of this new command is mostly to get cleaner config-exports.

src/Drupal/Commands/core/EntityDisplayCommands.php
```diff
@@ -0,0 +1,81 @@
...
+class EntityDisplayCommands extends DrushCommands
+{
+    /**
+     * Rebuild form and display configurations for entities and bundles.
+     *
+     * See Drupal Core: \Drupal\field\EntityDisplayRebuilder::rebuildEntityTypeDisplays
+     *
+     * @command entity:rebuild-displays
+     *
+     * @param string $entity_type_option Entity-type machine-name.
```
weitzman, line 17: Lets remove 'option' from both param names. They are arguments not options. Also we can remove dash in entity-type. And lets omit 'machine-name' entirely. User can figure that out from Examples.

weitzman (changes requested):

src/Drupal/Commands/core/EntityDisplayCommands.php
ndf: Inside the loops in the actual `rebuildDisplays`  method, there are multiple usages of $bundle and $entity_type variables. Example:
```
if ($bundle !== $bundle_arg && $bundle_arg !== 'all') {}
```
and
```
if ($entity_type !== $entity_type_arg && $entity_type_arg !== 'all') {}
```

There is a bit of a trade-off to rename the arguments or the variables inside the loops.
What is your opinion about this?

ndf: [done] Also we can remove dash in entity-type.
[done] And lets omit 'machine-name' entirely. User can figure that out from Examples.

src/Drupal/Commands/core/EntityDisplayCommands.php
```diff
@@ -0,0 +1,81 @@
...
+     *   Rebuild form and display configurations for article bundle in all entity types.
+     *
+     * @usage drush entity:rebuild-displays all all
+     *   Alias for `drush entity:rebuild-displays`.
+     * @usage drush entity:rebuild-displays node all
+     *   Alias for `drush entity:rebuild-displays node`.
+     */
+    public function rebuildDisplays(string $entity_type_arg = 'all', string $bundle_arg = 'all')
+    {
+        $this->output()->writeln(dt('Form and display configurations will be rebuild for the following bundles:'));
```
weitzman, line 36:
```suggestion
        $this->output()->writeln(dt('Form and display configurations will be rebuilt for the following bundles:'));
```

src/Drupal/Commands/core/EntityDisplayCommands.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->io()->newLine();
+
+        /** @var \Drupal\Core\Entity\EntityTypeBundleInfoInterface $bundle_info */
+        $bundle_info = \Drupal::service('entity_type.bundle.info')->getAllBundleInfo();
```
weitzman, line 40: This service should be injected via drush.services.yml

src/Drupal/Commands/core/EntityDisplayCommands.php
```diff
@@ -0,0 +1,81 @@
...
+                }
+                $this->output()->writeln("$entity_type:$bundle");
+            }
+        }
+
+        // Simulation is not supported, because class EntityDisplayRebuilder does not support it.
+        // Therefore we cannot get a diff of specific configurations that will be changed until we
+        // have finished rebuilding all form and display configurations.
+        if ($this->getConfig()->simulate()) {
+            throw new \Exception(dt('rebuild-display-config command does not support --simulate option.'));
```
weitzman, line 60:
```suggestion
            throw new \Exception(dt('entity:rebuild-displays command does not support --simulate option.'));
```

src/Drupal/Commands/core/EntityDisplayCommands.php
```diff
@@ -0,0 +1,81 @@
...
+        }
+        if (!$this->io()->confirm(dt('Do you really want to continue?'))) {
+            throw new UserAbortException();
+        }
+
+        /* @var \Drupal\field\EntityDisplayRebuilder $entity_display_rebuilder */
+        $entity_display_rebuilder = \Drupal::classResolver(\Drupal\field\EntityDisplayRebuilder::class);
```
weitzman, line 67: This service should be injected via drush.services.yml if it is indeed a service.

src/Drupal/Commands/core/EntityDisplayCommands.php
```diff
@@ -0,0 +1,81 @@
...
+        foreach ($bundle_info as $entity_type => $bundles) {
+            foreach ($bundles as $bundle => $bundle_settings) {
+                $entity_display_rebuilder->rebuildEntityTypeDisplays($entity_type, $bundle);
+                $this->logger()->notice('Finished rebuilding form and display configurations for: ' . $entity_type . ':' . $bundle);
+            }
+        }
+
+        // Flush all caches at the end of the operation. This ensures that
+        // config-export shows the correct changes.
+        drupal_flush_all_caches();
```
weitzman, line 77: I think we can just flush the config cache bin. See https://github.com/drush-ops/drush/pull/3861/files for how to inject that bin into the instance and then flush it.

weitzman (commented):
Sorry for long delay. A few more changes are needed.

src/Drupal/Commands/core/EntityDisplayCommands.php
```diff
@@ -0,0 +1,81 @@
...
+
+class EntityDisplayCommands extends DrushCommands
+{
+    /**
+     * Rebuild form and display configurations for entities and bundles.
+     *
+     * See Drupal Core: \Drupal\field\EntityDisplayRebuilder::rebuildEntityTypeDisplays
+     *
+     * @command entity:rebuild-displays
+     *
```
weitzman, line 16:
```suggestion
     * @version 11.3
```

PR #4672, opened 2021-02-24
Fix #4671. Show DB connection error details during bootstrap.

### LINKED ISSUE #4671: 'command not found' errors need better error messages
**Describe the bug**

My MySQL server had died, but Drush's error messages did not show this.

**To Reproduce**

```
drush @my.site ms --tag=mytag --debug

Using the Drush script found at /Users/joachim/Sites/mysite/vendor/drush/drush/drush using proc_open
 [preflight] Config paths: /Users/joachim/Sites/mysite/vendor/drush/drush/drush.yml
 [preflight] Alias paths: /Users/joachim/Sites/mysite/web/drush/sites,/Users/joachim/Sites/mysite/drush/sites
 [preflight] Commandfile search paths: /Users/joachim/Sites/mysite/vendor/drush/drush/src,/Users/joachim/Sites/mysite/web/drush,/Users/joachim/Sites/mysite/drush
 [debug] Bootstrap further to find ms [2.14 sec, 8.19 MB]
 [debug] Trying to bootstrap as far as we can [2.14 sec, 8.19 MB]
 [debug] Drush bootstrap phase: bootstrapDrupalRoot() [2.14 sec, 8.19 MB]
 [debug] Change working directory to /Users/joachim/Sites/mysite/web [2.14 sec, 8.19 MB]
 [debug] Initialized Drupal 8.9.12 root directory at /Users/joachim/Sites/mysite/web [2.24 sec, 8.32 MB]
 [debug] Drush bootstrap phase: bootstrapDrupalSite() [2.28 sec, 8.62 MB]
 [debug] Initialized Drupal site localhost:8888 at sites/localhost.mysite.web.clean [2.52 sec, 8.85 MB]
 [debug] Drush bootstrap phase: bootstrapDrupalConfiguration() [2.52 sec, 8.85 MB]
 [debug] Add service modifier [2.75 sec, 9.03 MB]
 [debug] Unable to connect to database. More information may be available by running `drush status`. This may occur when Drush is trying to bootstrap a site that has not been installed or does not have a configured database. In this case you can select another site with a working database setup by specifying the URI to use with the --uri parameter on the command line. See `drush topic docs-aliases` for details. [2.84 sec, 9.3 MB]
 [debug] Bootstrap phase bootstrapDrupalDatabase() failed to validate; continuing at bootstrapDrupalConfiguration() [2.84 sec, 9.28 MB]
 [debug] Done with bootstrap max in Application::bootstrapAndFind(): trying to find ms again. [2.84 sec, 9.28 MB]

In Application.php line 239:

  [Symfony\Component\Console\Exception\CommandNotFoundException]
  Command ms was not found. Drush was unable to query the database. As a result, many commands are unavailable. Re-run your command
  with --debug to see relevant log messages.
```

(Aside: I already *was* running it with --debug...)

So I followed this instruction:

> Unable to connect to database. More information may be available by running `drush status`. But `drush @my.site  status` all looks perfect! DB is shown there, everything.

It was only when I tried cr that I got the actual explanation:

```
$ drush @my.site cr

In Connection.php line 416:

  SQLSTATE[HY000] [2006] MySQL server has gone away


In Connection.php line 416:

  PDO::__construct(): Error while reading greeting packet. PID=8075
```

**Expected behavior**

The '  SQLSTATE[HY000] [2006] MySQL server has gone away' error should have appeared when I tried to do 'drush mim'.

**Actual behavior**

The error messages did not reflect the actual problem.

**Workaround**

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.3.6
| Drupal version? | 8.9.12
| PHP version     | 7.3
| OS?             | Mac

**Additional information**

weitzman:
Thanks for the report. 

The ultimate error isn't easy to improve as Symfony Console is throwing the Exception about command not found and we'd have to override a lot to take control and report back the SQL error.

This PR enhances the debug log message to show the actual connection error.

`drush status` tells you Bootstrap Successful when all is OK and tells you nothing when it isn't. I think we should say Unsuccessful in that case. Might need a major version to change this unfortunately.

### DIFF
--- a/src/Boot/DrupalBoot8.php
+++ b/src/Boot/DrupalBoot8.php
@@ -189,7 +189,7 @@ class DrupalBoot8 extends DrupalBoot implements AutoloaderAwareInterface
             $connection_options = $connection->getConnectionOptions();
             $connection->open($connection_options);
         } catch (\Exception $e) {
-            $this->logger->log(LogLevel::BOOTSTRAP, 'Unable to connect to database. More information may be available by running `drush status`. This may occur when Drush is trying to bootstrap a site that has not been installed or does not have a configured database. In this case you can select another site with a working database setup by specifying the URI to use with the --uri parameter on the command line. See `drush topic docs-aliases` for details.');
+            $this->logger->log(LogLevel::BOOTSTRAP, 'Unable to connect to database with message: ' . $e->getMessage() . '. More debug information is available by running `drush status`. This may occur when Drush is trying to bootstrap a site that has not been installed or does not have a configured database. In this case you can select another site with a working database setup by specifying the URI to use with the --uri parameter on the command line. See `drush topic docs-aliases` for details.');
             return false;
         }
         if (!$connection->schema()->tableExists('key_value')) {


greg-1-anderson:
>  think we should say Unsuccessful in that case. Might need a major version to change this unfortunately.

This is a worthwhile change. The question is, is the output of `drush status` an API? I know folks use it as an API in some cases (and `sql:sync` does). Maybe using the absence of a bootstrap message to determine if there is a valid site is enough of an edge case that it wouldn't be too inconvenient to change the behavior in a minor version update.

weitzman:
I hear you, but I have a slightly bad feeling about this one. I can totally see sites changing behavior if that item is empty. Maybe even reinstall a DB.

greg-1-anderson:
Sure thing, it's your call. We'd be going from empty to non-empty, though, so maybe it's not as dangerous as a db re-install.

Maybe we need a Drush 11 anyway, just to go from "happens to work" to "supported" for PHP 8 though. If we do a Drush 11, then it'd probably also be time to consider hard-breaking global installations.

weitzman:
Good point about empty to non-empty.

Yeah, Drush 11 looms. I just created https://github.com/drush-ops/drush/milestone/6. Lets collect ideas there. Would be great to do a virtual sprint on it sometime. Or even a in-person sprint later in the year.

PR #4676, opened 2021-02-25
Instruct drupal-finder to use any provided drupal root.

weitzman:
This gives a bypass to sites that mistakenly or deliberately dont have `installer-paths.drupal-core` in their composer.json extra. In general, it’s a small robustness improvement IMO. 

WIP until the next release of drupal-finder. See https://github.com/webflo/drupal-finder/pull/55.


weitzman:
This PR doesn't help when a root is specified via alias.

In fact, I think this approach is flawed. Lets make composer.com installer-paths a requirement and give a good error message when not present.

Issue #1947, opened 2016-01-28
Drush Memory Limit Detection Does Not Account For "Infinite" memory


rvtraveller:
Drush includes a function `drush_memory_limit()` which is designed to get the memory limit of PHP.  However, the function assumes the memory limit will be a certain value (256M, 1G, etc).  While this is good in practice, many default installation of PHP (such as Ubuntu LTS versions) ship with a `memory_limit` for the CLI of -1.

For the most part, this is not a big deal.  For batch processing via `drush_backend_batch_process()`, this is a huge deal because it forces Drupal to bootstrap for every single item in the batch.  If you have 100,000 items to process then you have to bootstrap Drupal 100,000 times.  Setting the memory limit to a real value (2G, 512M, whatever) allows Drush to process multiple entries per Drupal bootstrap which significantly reduces processing time.

I'm not sure of the best approach for what to do if `memory_limit` is -1.  Maybe we just give the user some sort of warning?  My quick Google searches didn't yield an easy way for PHP to see how much free RAM is on the box so this value could be set intelligently.

weitzman:
Not sure how to proceed, but I want to share prior work at #1405

weitzman:
Maybe we should just act like value is 512M or something when php reports -1.

greg-1-anderson:
I like the solution in #1405. Let's just move that over to batch_7.inc as suggested.

greg-1-anderson:
The implication being, that Drush treats "unlimited" as "yes, I really have unlimited memory". If you want batch processing to fork a new process to account for memory limits, then set memory_limit.

alexpott:
This causes interesting issues for very long running processes like site map generation when memory limit is set to -1. I think it'd be useful if we provided a way to trigger spawning a new process. If seems like either having the ability to define a drush memory limit on the command line like `DRUSH_MEMORY_LIMIT=512M vendor/bin/drush` or having an arbitrary time limit that once exceeded we should start a new thread. I think I prefer a time limit that is settable via an environment variable.

weitzman:
@alexpott would this work for your use case `php -d memory_limit=2GB vendor/bin/drush ...`. That should fork new processes when you exceed 60% mof 2GB in memory usage.

alexpott:
[...]

PR #4684, opened 2021-03-09
Add ability to set a time limit

alexpott:
Adds the ability to set a time limit for a spawned batch process - DRUSH_BATCH_PROCESS_TIME_LIMIT - to something like "100s" - so after 100 seconds it'll spawn a new process. At the moment it is configured to default to 0 - ie. no time limit but maybe we should consider setting this to something like 10 minutes by default.

This is one way of addressing #1947

### DIFF
--- a/includes/batch.inc
+++ b/includes/batch.inc
@@ -217,8 +217,9 @@ function _drush_batch_worker() {
   $current_set =& _batch_current_set();
   $set_changed = TRUE;
 
+  $start = microtime(TRUE);
   if (empty($current_set['start'])) {
-    $current_set['start'] = microtime(TRUE);
+    $current_set['start'] = $start;
   }
   $queue = _batch_queue($current_set);
   while (!$current_set['success']) {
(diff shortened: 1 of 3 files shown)


weitzman:
I dont love our memory limit handling and this kinda adds to the stinkpile, but I do see how its useful.In fact I could use it right now on a project. So, I'll probably merge this if nobody objects. This should probably use drush config instead of a straight  environment variable, since drush config can also be set via an env variable. I can make that change ... Thanks for adding a test.

weitzman:
Actually, this wont help me since my command is queue:run which doesn’t use the batch system. I'm undecided about this PR.

weitzman:
FYI https://www.drupal.org/node/2577417 might help.

weitzman:
Closing this as won't fix for now.


To summarize, one can set a memory_limit as per https://github.com/drush-ops/drush/issues/1947#issuecomment-798527477. Also, if folks could get https://www.drupal.org/project/drupal/issues/1199866 over the finish line, it would help in situations like this.

Issue #3843, opened 2019-01-04
drush doesn't pick up current terminal width when run through a pipe


joachim-n:
**Describe the bug**

When I run a drush command on a remote site, the output is wrapped even though my local machine's terminal might not need it.

E.g. if I run 'drush pml' on the local site, where I've set my terminal super-wide to prevent wrapping, I might get:

```
  Purge                             Purge Tokens (purge_tokens)                                                         Disabled   8.x-3.0-beta8
  Purge                             Purge UI (purge_ui)                                                                 Disabled   8.x-3.0-beta8
```

But on a remote site:

```
Voting        Voting API (votingapi)           Disabled   8.x-3.0-beta1
votingapi     votingapi_widgets                Disabled   8.x-1.0-alpha3
              (votingapi_widgets)
Webform       Webform (webform)                Enabled    8.x-5.0-rc16
Webform       Webform Bootstrap                Disabled   8.x-5.0-rc16
              (webform_bootstrap)
Webform       Webform Demo:                    Disabled   8.x-5.0-rc16
```

**Expected behavior**

It would be nice if Drush used the local terminal window's size for output from the remote site.

For instance, with the 'pml' command, it's useful to pipe this through a filtering tool (grep/ack/ag), and if lines are wrapped, then the results aren't as good, as the module machine name is not always on the same line as its status and version number.

**Workaround**

None afaik.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.3.1
| Drupal version? | 8.x
| PHP version     | 7.1
| OS?             | Mac

greg-1-anderson:
I can't remember if we changed this code recently, but it would be helpful if you reported issues using the most recent version of Drush. We happen to be re-working the process code in the latest Drush 9.6.0 betas (stable release coming shortly).

greg-1-anderson:
I'll see if I can do some of my own investigation while I have the hood open on the process code.

weitzman:
We could send COLUMNS environment variable to the subprocess. We already do some of that in the postgres db driver - https://github.com/drush-ops/drush/blob/2b4f5cff460192d7c55514ef5d0776cd569bc25f/src/Commands/sql/SqlCommands.php#L143.

I've seen some oddness when you do that around inheriting the environment variables of the parent process. Not sure if that would matter here. There is a Process method for forcing that inheritance.

joachim-n:
[...]

PR #4688, opened 2021-03-16
Fix #4043,#3843 - Use Symfony Console to determine terminal width.

### LINKED ISSUE #4043: Problem with colors in console on Win10
**Describe the bug**
When I run drush commands in console - i get uncolored pieces with extra symbols. 
screenshot
This is cmder console in phpstorm. If I use conemu/cmder direct in folder (outside phpstorm) - no problem with colors.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version  | 9.6.2
| Drupal version | 8.x
| PHP version     | 7.2
| OS             | Windows

weitzman:
Hopefully this fixes both issues linked in the title, and removes a bit of custom code. See https://github.com/symfony/console/blob/5.x/Terminal.php#L25-L37

### DIFF
--- a/src/Config/Environment.php
+++ b/src/Config/Environment.php
@@ -5,6 +5,7 @@ use Composer\Autoload\ClassLoader;
 
 use Drush\Drush;
 use Drush\Utils\FsUtils;
+use Symfony\Component\Console\Terminal;
 use Webmozart\PathUtil\Path;
 
 /**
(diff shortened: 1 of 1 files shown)


wimleers:
Unfortunately I can't use this yet because this shipped with `drush 10.4.1` and I cannot update beyond `10.3.6` because of the `drush` vs `migrate_tools` situation.

PR #4694, opened 2021-03-20
Add "project" to pm:list

greg-1-anderson:
We have a need to get information about modules in a Drupal site. We're using a remote call, and the target might be running various versions of Drupal + Drush. We'd like to know which project a module is associated with, and since pm:info no longer exists in current versions of Drush, and whereas the project information is readily available in the extension information, this seemed like a low-impact solution that is generally useful.

Note that the default fields do not change in this PR, so the default output does not change either (you only get `project` if you request it via `--fields`).

### DIFF
--- a/src/Drupal/Commands/pm/PmCommands.php
+++ b/src/Drupal/Commands/pm/PmCommands.php
@@ -148,6 +148,7 @@ class PmCommands extends DrushCommands
      * @option package Only show extensions having a given project packages (e.g. Development).
      * @field-labels
      *   package: Package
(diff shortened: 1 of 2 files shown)


andypost:
Will it affect existing list? Scripts could use to parse output

weitzman:
Looks good. Add something to an existing test?

Backward compat is fine imo because scripts should use a machine output format.

greg-1-anderson:
It does not affect existing output unless you pass `--fields` and include project in the list. I will add tests.

Issue #4641, opened 2021-01-19
Namespace problems with Site-Wide Drush Commands


AdamPS:
**Describe the bug**
Docs for [Site-Wide Drush Commands](https://docs.drush.org/en/9.x/commands/#site-wide-drush-commands) state to use a namespace `Drush\Commands`.  However this namespace isn't registered as an auto-loader, so it' doesn't work to try and use files in this directory.

**To Reproduce**
Create a site-wide command in `$PROJECT_ROOT/drush/Commands/ExampleCommands.php` that extends from a base class in the same directory.

**Expected behavior**
It works.

**Actual behavior**
The base class cannot be found.

**Workaround**
Add a line to your own composer.json.

```
  "autoload": {
    "psr-4": {
      "Drush\\Commands\\": "web/drush/Commands/"
    }
  },
```

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10..3.6
| Drupal version? | Any
| PHP version     | 7.4.3
| OS?             | Linux

AdamPS:
Proposed fix is to alter drush composer.json as per the workaround.  I can create a PR if you would accept this change.

Sweetchuck:
I have been using site-wide Drush commands for a while, and definitely something wrong with the Command file detection.

But my workaround is the following:
File layout (relative from project root):
* `./drush/drush.yml`
* `./drush/drush.app.yml`
* `./drush/drush.local.yml`
* `./drush/Commands/contrib/foo/Commands/drush.yml`
* `./drush/Commands/contrib/foo/Commands/foo/FooBarCommands.php`

File content ./drush/drush.yml
```yaml
drush:
  paths:
    config:
      - '${drush.vendor-dir}/../drush/Commands/contrib/foo/Commands'
      - '${drush.vendor-dir}/../drush/drush.app.yml'
      - '${drush.vendor-dir}/../drush/drush.local.yml'
    include:
      - '${drush.vendor-dir}/../drush/Commands/contrib/foo'
```

File content ./drush/drush.app.yml
```yaml
options:
  root: '${drush.vendor-dir}/../docroot'
```

File content ./drush/Commands/contrib/foo/Commands/foo/FooBarCommands.php
```php
<?php

namespace Drush\Commands\foo;

class FooBarCommands {}
```

`alias d='vendor/bin/drush --config=drush @app.local'`
`d list --filter=foo`
> it works, but far from ideal.

Sweetchuck:
[...]

PR #4696, opened 2021-03-21
Auto-discover non-Drupal commands

claudiu-cristea:
True, there is a way to discover such commands by adding their paths in Drush configuration. But I want to be able to have such commands available automatically, when I `composer require` a 3rd party library Drush commands provider.

This PR uses the Robo `RelativeNamespaceDiscovery` to grab the Drush commands living in 3rd party libraries, loadable via PSR4. Such a class should comply with the following requirements:

* The commands class namespace ends with `*\Drush\Commands`, e.g. `My\Custom\Library\Drush\Commands`.
* The class and file name ends with `*DrushCommands`, e.g. `FooDrushCommands`.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1,23 +1,25 @@
 <?php
 namespace Drush;
 
+use Composer\Autoload\ClassLoader;
 use Consolidation\AnnotatedCommand\AnnotatedCommand;
 use Consolidation\AnnotatedCommand\CommandFileDiscovery;
-use Drush\Boot\BootstrapManager;
-use Drush\Runtime\TildeExpansionHook;
 use Consolidation\SiteAlias\SiteAliasManager;
-use Drush\Log\LogLevel;
+use Drush\Boot\BootstrapManager;
 use Drush\Command\RemoteCommandProxy;
-use Drush\Runtime\RedispatchHook;
 use Drush\Config\ConfigAwareTrait;
-use Robo\Contract\ConfigAwareInterface;
-use Symfony\Component\Console\Application as SymfonyApplication;
-use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Exception\CommandNotFoundException;
-use Symfony\Component\Console\Input\InputInterface;
-use Symfony\Component\Console\Output\OutputInterface;
+use Drush\Log\LogLevel;
+use Drush\Runtime\RedispatchHook;
+use Drush\Runtime\TildeExpansionHook;
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerAwareTrait;
+use Robo\ClassDiscovery\RelativeNamespaceDiscovery;
+use Robo\Contract\ConfigAwareInterface;
+use Symfony\Component\Console\Application as SymfonyApplication;
+use Symfony\Component\Console\Exception\CommandNotFoundException;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * Our application object
@@ -306,7 +308,7 @@ class Application extends SymfonyApplication implements LoggerAwareInterface, Co
      * Configure the application object and register all of the commandfiles
      * available in the search paths provided via Preflight
      */
-    public function configureAndRegisterCommands(InputInterface $input, OutputInterface $output, $commandfileSearchpath)
+    public function configureAndRegisterCommands(InputInterface $input, OutputInterface $output, $commandfileSearchpath, ClassLoader $classLoader)
     {
         // Symfony will call this method for us in run() (it will be
         // called again), but we want to call it up-front, here, so that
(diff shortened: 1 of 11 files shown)


claudiu-cristea:
We may want to deprecate global commands discoverable via config (?)

weitzman:
LGTM. @greg-1-anderson would be right person to review and merge this.

weitzman:
We should make sure Generators maybe discovered similarly, and edit docs at https://www.drush.org/latest/generators/ accordingly.

claudiu-cristea:
@weitzman I see global generators are already discovered by their namespace (`\Drush\Generators\*`). Should we add the same logic, e.g. `My\Custom\Library\Drush\Generators`?

What about deprecating the old way for commands (see my previous comment) and generators (`\Drush\Generators\*`)?

weitzman:
> Should we add the same logic, e.g. My\Custom\Library\Drush\Generators

Yes, I think so. 

> What about deprecating the old way for commands (see my previous comment) and generators 

Yes, IMO.

greg-1-anderson (approved):
This is spot on; thank you for submitting it.

I am 👍 on merging this, but holding off for a bit to see if we want to add deprecations first. Auto-discovering generators would be good for consistency, but could be follow-on work IMO.

claudiu-cristea:
Generators are postponed to a followup. This is not a straight task as `RelativeNamespaceDiscovery` needs the class loader to grab the classes.

Consider to postpone also deprecations. In PR I'm suggesting to deprecate `drush.commands` discovery, which seems not documented anywhere, but can be observed here `\Drush\Application::commandsFromConfiguration()`. There is also `drush.patch.include` and this is confusing me.

src/Application.php
```diff
@@ -318,7 +320,15 @@ public function configureAndRegisterCommands(InputInterface $input, OutputInterf
         $discovery = $this->commandDiscovery();
         $commandClasses = $discovery->discover($commandfileSearchpath, '\Drush');
         $commandClasses[] = \Consolidation\Filter\Hooks\FilterHooks::class;
-        $commandClasses = array_merge($this->commandsFromConfiguration(), $commandClasses);
+        $commandsFromConfig = $this->commandsFromConfiguration();
+        if ($commandsFromConfig) {
+            @trigger_error("Discovering global commands by specifying them in 'drush.commands' config is deprecated and will be removed in Drush 11.0.0. See docs/commands.md on how to create auto-discoverable global commands.", E_USER_DEPRECATED);
+        }
+        $commandClasses = array_merge(
```
greg-1-anderson, line 327: The other question about this PR is, what happens if the same commandfile is found twice by multiple methods? For example, the [example-drush-extension](https://github.com/drush-ops/example-drush-extension) project creates an autoloadable class that is discovered based on its location, and is probably also discovered by this PR. I am presuming that this array_merge will cause these two identical items to become a single entry in the resulting array, since they should have the same key.

src/Application.php
```diff
@@ -320,8 +320,12 @@ public function configureAndRegisterCommands(InputInterface $input, OutputInterf
         $discovery = $this->commandDiscovery();
         $commandClasses = $discovery->discover($commandfileSearchpath, '\Drush');
         $commandClasses[] = \Consolidation\Filter\Hooks\FilterHooks::class;
+        $commandsFromConfig = $this->commandsFromConfiguration();
+        if ($commandsFromConfig) {
+            @trigger_error("Discovering global commands by specifying them in 'drush.commands' config is deprecated and will be removed in Drush 11.0.0. See docs/commands.md on how to create auto-discoverable global commands.", E_USER_DEPRECATED);
```
greg-1-anderson, line 325: I am wondering if `trigger_error` is too noisy. Maybe we should deprecate with `trigger_error` in Drush 11, and remove in Drush 12? I know that's a long long ways off, but folks are very slow at upgrading Drush commands in modules.

greg-1-anderson:
My last comment is that maybe we should note in the documentation that this new loading mechanism is experimental, and subject to change until Drush 10.5 or 11.0 is tagged.

greg-1-anderson:
I realize that it's early for this PR to deprecate the `drush.commands` discovery method, as this PR does not provide dependency injection for discovered commands, whereas `drush.commands` does.

I think this PR is still good to merge; we can add DI as follow-on work. We should back out the deprecation, though.

We could potentially deprecate auto-discovery by location (`__DRUPAL_ROOT__ . /drush`), potentially.

src/Application.php
claudiu-cristea: Well, not all parts of the merge have keys. So, we should remove the keys > merge > array_unique(), assuming that there are no same two namespaces with different paths in this universe. That would be a nonsense and a bug somewhere.

src/Application.php
claudiu-cristea: >  I know that's a long long ways off, but folks are very slow at upgrading Drush commands in modules.

This is something that leads to exactly the opposite conclusion: If they're slow to adopt changes, then we should deprecate as soon as possible, giving them enough time to take action. We can change the removal version from 11.0.0 to 12.0.0 but the deprecation should occur now. Deprecating doesn't mean the command doesn't work. Also note that this is a muted deprecation ` @trigger_error(...)`, so there should be no noise in console. Normally code deprecation tests should catch this error but that's fine.

claudiu-cristea:
> I realize that it's early for this PR to deprecate the drush.commands discovery method, as this PR does not provide dependency injection for discovered commands, whereas drush.commands does.

Not sure it provides DI but, yes, let's drop the deprecation for now, it needs more analysis.

src/Application.php
claudiu-cristea: Anyway, deprecation is postponed.

greg-1-anderson:
I guess I'm not that opposed to AutoloaderAwareTrait.

weitzman (approved):

AdamPS:
Many thanks I confirm this satisfies my requirement mentioned in #4641

PR #4698, opened 2021-03-22
Fix container rebuilds on deployment identifier change

alexpott:
This fixes the most recent error described on https://www.drupal.org/project/drupal/issues/2985199 (see comment 70)


weitzman:
thanks @dpagini and @alexpott for the fix.

dpagini:
Thank YOU @weitzman and @alexpott for your help with this one, it was super tricky!

awm086:
I know this is closed but I am experiencing this issue after upgrading to 9.2. should I create an issue or is it tracked?

njt1982:
Just for reference (in case this come up in Google!).... This patch also fixed an error where a `drush deploy` broke the container cache so this error happened

    Module XXXX has an entry in the system.schema key/value storage, but is not installed.

If I did `drush cr` everything was fine, but it seemed that `drush cim` triggered this error.

I was using Drush 10.3.6 and cannot currently upgrade past this due to the migrate_tools 5.x restrictions.

PR #4700, opened 2021-03-22
Add tests for changing deployment identifier and then running a drush command.

alexpott:
First test run removes the fix added in #4698 - will fail - hopefully. Yep https://github.com/drush-ops/drush/pull/4700/commits/39abc447c5b93cd5c068eae7eba9eccd428d2768 failed and without reverting the fix tests are passing.


weitzman:
I'm a little hesitant about this. First I'd check to make sure the module and settings.php edits are cleaned up after a test run.

alexpott:
Locally I can run the test time and time again - if they weren't then the tests would fail. See \Unish\UnishTestCase::cleanDirs() - it removes the sites directories.

alexpott:
I think this bug is tricky enough and potentially severe enough to need it's own test. If the container has an incorrect list of modules all sorts of bad things can happen.

alexpott:
I've moved the test to its own test because then we know know for sure that it will be cleaned up properly by \Unish\UnishTestCase::cleanDirs()

Issue #4704, opened 2021-03-25
Out of box tideways profiling fails


andypost:
**Describe the bug**
Using tideways_xhprof extension I'm getting
```
 $ DRUSH_XH_LINK=1 vendor/bin/drush version
PHP Fatal error:  Uncaught TypeError: Drush\Commands\core\XhprofCommands::xhprofFlags(): Argument #1 ($config) must be of type array, null given, called in /var/www/html/web/vendor/drush/drush/src/Commands/core/XhprofCommands.php on line 68 and defined in /var/www/html/web/vendor/drush/drush/src/Commands/core/XhprofCommands.php:91
Stack trace:
#0 /var/www/html/web/vendor/drush/drush/src/Commands/core/XhprofCommands.php(68): Drush\Commands\core\XhprofCommands::xhprofFlags()
#1 /var/www/html/web/vendor/consolidation/annotated-command/src/Hooks/Dispatchers/InitializeHookDispatcher.php(47): Drush\Commands\core\XhprofCommands->xhprofInitialize()
#2 /var/www/html/web/vendor/consolidation/annotated-command/src/Hooks/Dispatchers/InitializeHookDispatcher.php(36): Consolidation\AnnotatedCommand\Hooks\Dispatchers\InitializeHookDispatcher->doInitializeHook()
#3 /var/www/html/web/vendor/consolidation/annotated-command/src/Hooks/Dispatchers/InitializeHookDispatcher.php(29): Consolidation\AnnotatedCommand\Hooks\Dispatchers\InitializeHookDispatcher->callInitializeHook()
#4 /var/www/html/web/vendor/consolidation/annotated-command/src/CommandProcessor.php(145): Consolidation\AnnotatedCommand\Hooks\Dispatchers\InitializeHookDispatcher->initialize()
#5 /var/www/html/web/vendor/consolidation/annotated-command/src/AnnotatedCommand.php(296): Consolidation\AnnotatedCommand\CommandProcessor->initializeHook()
#6 /var/www/html/web/vendor/symfony/console/Command/Command.php(221): Consolidation\AnnotatedCommand\AnnotatedCommand->initialize()
#7 /var/www/html/web/vendor/symfony/console/Application.php(1027): Symfony\Component\Console\Command\Command->run()
#8 /var/www/html/web/vendor/symfony/console/Application.php(273): Symfony\Component\Console\Application->doRunCommand()
[...]

PR #4705, opened 2021-03-25
Provide global defaults for profiler

andypost:
Defaults for `xh` config is missing but `\Drush\Commands\core\XhprofCommands::xhprofFlags()` expects array as argument

It prevents to fail command as #4704

### DIFF
--- a/examples/example.drush.yml
+++ b/examples/example.drush.yml
@@ -177,3 +177,6 @@ xh:
   # Start profiling via xhprof/tideways and show a link to the run report.
 # link: http://xhprof.local
   # See src/Commands/core/XhprofCommands.php for more configuration settings.
+  profile-builtins: true
+  profile-cpu: false
+  profile-memory: false
--- a/src/Commands/core/XhprofCommands.php
+++ b/src/Commands/core/XhprofCommands.php
@@ -4,7 +4,7 @@ namespace Drush\Commands\core;
 
 use Consolidation\AnnotatedCommand\AnnotationData;
 use Consolidation\AnnotatedCommand\CommandData;
-use Drush\Drush;
+use Drush\Config\DrushConfig;
 use Symfony\Component\Console\Input\InputInterface;
 use Drush\Commands\DrushCommands;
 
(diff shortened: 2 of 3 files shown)


andypost:
Initially I think about to provide it as default like diff bellow, but it still hard to override

```diff
--- a/src/Config/Environment.php
+++ b/src/Config/Environment.php
@@ -186,6 +186,12 @@ class Environment
                 'site-file-previous' => $this->getSiteSetAliasFilePath('drush-drupal-prev-site-'),
                 'site-file-current' => $this->getSiteSetAliasFilePath(),
             ],
+            // Defaults for \Drush\Commands\core\XhprofCommands profiling flags.
+            'xh' => [
+                'profile-builtins' => true,
+                'profile-cpu' => false,
+                'profile-memory' => false,
+            ],
         ];
     }
```

andypost:
Using ` DRUSH_XH_LINK=1 vendor/bin/drush -Dxh.profile-memory=1 -Dxh.profile-cpu=1 version` to test that now options getting overrides in a strange way, used to add `var_dump($config); var_dump($flags)` into `\Drush\Commands\core\XhprofCommands::xhprofInitialize()`

```
/var/www/html $  DRUSH_XH_LINK=1 vendor/bin/drush -Dxh.profile-memory=1 -Dxh.profile-cpu=1 version
array(2) {
  ["profile-memory"]=>
  string(1) "1"
  ["profile-cpu"]=>
  string(1) "1"
}
int(7)
 [notice] XHProf run saved. View report at 1/index.php?run=605c1acd6f902&source=Drush
 Drush version : 10.4.0 
```

src/Config/Environment.php
```diff
@@ -186,6 +186,13 @@ public function exportConfigData()
                 'site-file-previous' => $this->getSiteSetAliasFilePath('drush-drupal-prev-site-'),
                 'site-file-current' => $this->getSiteSetAliasFilePath(),
             ],
+            // Setup \Drush\Commands\core\XhprofCommands default flags, to allow
+            // override in other init hooks.
+            'xh' => [
```
andypost, line 191: Looks here's enough to assign an empty array to allow it overridable

greg-1-anderson:
The main problem is that `xhprofInitialize` calls `$this->getConfig()->get('xh');`, which does not give the config system a chance to override sub-components (env overrides are done at `get` time).

Rather than list the defaults in a different part of the code, could we instead rewrite `XhprofCommands::xhprofFlags` to take the config object instead of an array, and have it call `get('xh.foo')` on each individual flag?

Today I was finding that overrides are working for arbitrary config, e.g. 
```
DRUSH_FOO=bar ./drush ev 'var_export(\Drush\Drush::config()->get("foo"));'
```
Not sure what I did wrong yesterday when I thought it wasn't working.

andypost:
Used to rework the approach, the only downside that I found no way to keep `xh.profile-builtins` as `true` - there's no way to unset it so I did rename it to `xh.profile-nobuiltins`

andypost:
Reverted option back but set its default to `!self::XH_PROFILE_BUILTINS` to make work

andypost:
I'd like to add one more option - `xh.path` to store profiles and get rid of useless dependency https://github.com/drush-ops/drush/blob/10.x/composer.json#L61

weitzman:
Is that a useless dependency only when using the xhprof drupal module?

andypost:
@weitzman yes, it's useless even for the module, so here clean-up and workaround #4710

andypost:
Fixed test and squashed commits

PR #4710, opened 2021-03-26
Allow to override the path to save profile dump 

andypost:
- removes useless dependency
- it will not fail if extra source for xhprof will be missing
- allows to override directory where to save the profile dump

### DIFF
--- a/src/Commands/core/XhprofCommands.php
+++ b/src/Commands/core/XhprofCommands.php
@@ -136,14 +136,17 @@ class XhprofCommands extends DrushCommands
     {
         if (extension_loaded('tideways_xhprof')) {
             $data = \tideways_xhprof_disable();
-            $dir = $this->getConfig()->tmp();
-            $run_id = uniqid();
-            file_put_contents($dir . DIRECTORY_SEPARATOR . $run_id . '.' . $namespace . '.xhprof', serialize($data));
-            return $run_id;
         } else {
-            $xhprof_data = \xhprof_disable();
-            $xhprof_runs = new \XHProfRuns_Default();
-            return $xhprof_runs->save_run($xhprof_data, $namespace);
+            $data = \xhprof_disable();
+            if (class_exists('\XHProfRuns_Default')) {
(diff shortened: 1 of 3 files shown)


src/Commands/core/XhprofCommands.php
```diff
@@ -136,14 +136,17 @@ public function xhprofFinishRun($namespace)
...
-            $xhprof_runs = new \XHProfRuns_Default();
-            return $xhprof_runs->save_run($xhprof_data, $namespace);
+            $data = \xhprof_disable();
+            if (class_exists('\XHProfRuns_Default')) {
+                $xhprof_runs = new \XHProfRuns_Default($this->getConfig()->get('xh.path'));
+                return $xhprof_runs->save_run($data, $namespace);
+            }
         }
+        $config = $this->getConfig();
+        $dir = $config->get('xh.path', $config->tmp());
```
andypost, line 147: Maybe better to use `xh.output-dir` for option?

andypost: less to type sounds better

weitzman:
IMO _lox/xhprof_ is not a useless dependency. With it, sites can use XHProf UI without any more software. Removing it is a DX regression. One fewer dependency is always nice, though I note that this one has never caused a single bug report.

> it will not fail if extra source for xhprof will be missing

Can you elaborate?

> allows to override directory where to save the profile dump

Drush is currently out of the game if caring about output path. We delegate that to lox/xhprof so I dont love adding this.

andypost:
@weitzman this is require-dev dependency so it never installed except drush tests

weitzman:
Oh, that would explain the lack of bug reports :)

andypost:
> never caused a single bug report.

Because it never used to work( 

Moreover the only working profiler for php8 is tideways which has to put everything into tmp, and xhprof has the setting only for scripts https://github.com/longxinH/xhprof/blob/master/extension/xhprof.c#L110

The real problem with lox is it's not maintained (outdated for 5 years) over xhprof from pecl https://github.com/lox/xhprof

Issue #3391, opened 2018-02-16
cache-rebuild causes modules in "/sites/example.com/modules/" to stop working


noah-st-amand:
Whenever I run "drush cr", all the custom modules I have in "/sites/example.com/modules/" stop working, and I have to disable then re-enable them to get them working again. I have additional custom modules in /modules/custom that are unaffected. Clearing caches from the Drupal interface doesn't cause this problem.

There's nothing in the watchdog table to indicate why the modules stop working. This seems to be new as of 9.1—previously I used drush to rebuild the cache regularly without encountering this problem. It's happening in multiple server environments.

weitzman:
I don't know offhand why this would be happenning. Try passing `--uri=http://example.com` to see if it is resolved.

noah-st-amand:
Thanks—I tried passing in the uri, I'm still seeing the same problem.

I went through and cleared each cache with "drush cc", and the issue didn't occur, only rebuilding from drush appears to cause it. I also tested rebuilding via "/core/rebuild.php" and that also didn't trigger the problem.

I'm happy to hack around trying to isolate what's causing this, but I'm not sure where to start in /vendor/drush/...

A clarification: modules in "/sites/example.com/modules/" are still enabled after running "drush cr", they just don't work (e.g., any blocks they're generating appear as "This block is broken or missing. You may be missing content or you might need to enable the original module.", any hooks in the modules aren't run, etc.). Disabling them and re-enabling the modules restores their functionality.

jhuhta:
I'm suffering from the same problem I think. I see severe caching problems after upgrading to drush 9, mostly if not entirely related to custom code. I didn't realize before seeing this issue, that it might be the cr breaking everything, as it usually fixes everything, but yes, that's what it does. Everything works until cr. To be able to fix the broken caches, I needed to issue TRUNCATE statements to the cache_* tables, hitting button from the UI wasn't enough.

Using 9.2.1 currently (or at least before downgrading to drush8) with D8.4.5 and varnish. Any tips on how to debug this further are appreciated.

jhuhta:
It seems to me that drush is somehow assuming the site dir is sites/default, and fails when that is not the case. On the problematic site I had just some default files there, there real site dir is "example.com". It doesn't matter whether the "default" dir exists or not, though.

I can work around the problem by removing default "default" dir and symlinking "default" to "example.com". 

[...]

PR #4713, opened 2021-03-27
Fixed use of DRUPAL_ROOT in cache:rebuild command.

### LINKED ISSUE #4712: `drush cr` doesn't work if Drupal is symlinked in by Composer (Drush 12)
**Describe the bug**

If Drupal core is symlinked in by Composer (see https://github.com/joachim-n/drupal-core-composer) then `drush cr` fails with an error message about no hash salt.

This is because DrupalKernel and Settings get initialised with the wrong app root.

This is because CacheCommands::rebuild() is using the DRUPAL_ROOT constant that gets set by includes/bootstrap.inc, and it shouldn't, because that constant is assuming the Drupal package is in the standard place. (See https://www.drupal.org/project/drupal/issues/1792310 for more on this.)

The fix is actually very simple. By some magic I don't understand, the PHP current directory is already set to the app root at this point. So we can simply use getcwd().

MR coming.

**To Reproduce**

Installed Drupal with https://github.com/joachim-n/drupal-core-composer. Did `drush cr`.

**Expected behavior**

It works.

**Actual behavior**

Error message.

**Workaround**

Clear it in the UI, probably.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 9.2
| PHP version     | 7.4
| OS?             | Mac

**Additional information**

joachim-n:
Fixes #4712.

### DIFF
--- a/src/Commands/core/CacheCommands.php
+++ b/src/Commands/core/CacheCommands.php
@@ -212,7 +212,13 @@ class CacheCommands extends DrushCommands implements CustomEventAwareInterface, 
             $this->logger()->info(dt("Skipping cache-clear operation due to --no-cache-clear option."));
             return true;
         }
-        chdir(DRUPAL_ROOT);
+
+        // Through some magic I don't claim to understand, the current directory
+        // is already the web root.
+        // Use this, and NOT the DRUPAL_ROOT constant, as that is unreliable if
+        // the Drupal package is not installed in the usual location and
+        // symlinked in by Composer.
+        $root = getcwd();
 
         // We no longer clear APC and similar caches as they are useless on CLI.
         // See https://github.com/drush-ops/drush/pull/2450
@@ -226,7 +232,6 @@ class CacheCommands extends DrushCommands implements CustomEventAwareInterface, 
         DrupalKernel::bootEnvironment();
 
         // Avoid 'Only variables should be passed by reference'
-        $root  = DRUPAL_ROOT;
         $site_path = DrupalKernel::findSitePath($request);
         Settings::initialize($root, $site_path, $autoloader);
 


weitzman:
Thanks for the PR. If the constant is unreliable, that’s a core bug and I'd rather not work around it in Drush. We use DRUPAL_ROOT elsewhere as well.

joachim-n:
Well in the relevant core issue, https://www.drupal.org/project/drupal/issues/1792310, there is discussion about deprecating DRUPAL_ROOT. So I think removing it from Drush would be a good idea.

weitzman:
If that constant goes away, we'll use whatever its replacement is.

shaal:
@joachim-n Thank you for point me to this PR.
I'll utilize this trick in DrupalPod.

Without this, when using the special structure of `/repos/drupal/core` that is  symlinked via composer into `/web/core`, it is impossible to run the most popular and important drush command of all time - `drush cr`.

Even if this does get fixed in Drupal core, older Drupal versions (8.9.x, 9.1.x, 9.2.x) won't be able to take advantage of this fix.

joachim-n:
@shaal I think I've fixed the problem this PR had with newer versions of Drush.
I've pushed it to the this MR's branch, but it's not showing up here. Can you figure it out?

joachim-n:
Ah nope, drush cr still kills off contrib modules.

weitzman:
Can you try out https://github.com/drush-ops/drush/pull/5147 and see if it helps? You can prob just hack your Drush directly since only a few lines changed.

> Ah nope, drush cr still kills off contrib modules.

That sounds like https://github.com/drush-ops/drush/issues/3391 (which is thought to be a drupal core problem)

shaal:
@weitzman I tried applying patch #5147 it doesn't solve the problem; it has the same issue as in @joachim-n's patch.

I tested it locally (https://github.com/joachim-n/drupal-core-development-project + patch in this issue, as well as patch in #5147), and I tested it using DrupalPod.
The problem remains the same, on a special Drupal core setup, using `repos/drupal` and symlink `core` and `vendor` directories - running `drush cr` hides any contrib modules installed. Clearing cache in UI restore expected functionality.

Steps to replicate:
1. [Open Drupal core in DrupalPod](https://gitpod.io/#DP_PROJECT_NAME=drupal,,DP_PROJECT_TYPE=project_core,DP_MODULE_VERSION=9.4.x,DP_CORE_VERSION=9.4.x,DP_INSTALL_PROFILE=standard/https://github.com/shaal/drupalpod)
1. Log in as admin (admin / admin)
1. Go to `/admin/modules`
1. Confirm `admin_toolbar` module is installed
1. Go to `/admin/config/development/performance`
1. Press the button `Clear all caches`
1. Confirm `admin_toolbar` is still installed.
1. In terminal, type `ddev drush cr`
1. Notice that now `admin_toolbar` is no longer installed.
1. Clear cache through the UI
1. `admin_toolbar` is installed as expected.

weitzman:
Would be good if someone could see if https://github.com/drush-ops/drush/issues/3391 is at fault. the symptom is the same.

joachim-n:
I don't think this is easily fixable in Drush.

```
function drupal_rebuild($class_loader, Request $request) {
```

This doesn't allow us to pass in anything about the app root, and so it will go to DrupalKernel::guessApplicationRoot(), which just works with __DIR__ and so will get it wrong.

The only thing to do would be for Drush to make a copy of drupal_rebuild() which passes in the correct app root when instantiating DrupalKernel. I doubt @weitzman would be happy with that!

weitzman:
Thanks for the analysis.

You are right that I don't relish maintaining a copy of drupal_rebuild(). I will however pass $root to it (see #5149) , and then interested folks can petition Drupal core to add an optional param to it. I don't see why anyone would object to that. That patch could be safely backported to 9.x and 8.x. 

> can a package alter an existing command to switch its callback?

Yes, see https://github.com/consolidation/annotated-command#replace-command-hook. Perhaps https://github.com/consolidation/annotated-command#commandinfo-alterers would work as well.

joachim-n:
>  I will however pass $root to it

I don't think that's needed -- when the core DRUPAL_ROOT issue is fixed, this all should Just Work (famous last words!!)

weitzman:
> Though a fix to drupal_rebuild() might get in sooner, OTOH.

Right.

shaal:
@joachim with #5149, do you think it can fix the problem we're seeing (before drupal core fixes it)

joachim-n:
@shaal I've pushed a Drush command replacement to the composer project. Can you have a look?

weitzman:
Code looks good. If you verified that your code is getting called instead drush's, then its good. If you have trouble with that, see https://www.drush.org/latest/commands/#auto-discovered-commands

shaal:
@joachim-n I tested your latest update in DrupalPod https://github.com/joachim-n/drupal-core-development-project/commit/e3f5c7f362039dce851a4b8da53abe443d423d35

It works! Thank you!

PR #4721, opened 2021-04-06
Add the possibility to add custom mysql and mysqldump commands. This …

katsukawa:
…could be necessary when you have in your server multiple versions of mysql, maria or etc.

### DIFF
--- a/src/Sql/SqlMysql.php
+++ b/src/Sql/SqlMysql.php
@@ -11,7 +11,8 @@ class SqlMysql extends SqlBase
 
     public function command()
     {
-        return 'mysql';
+        $dbSpec = $this->getDbSpec();
+        return empty($dbSpec['mysql_command']) ? 'mysql' : $dbSpec['mysql_command'];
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks for the PR. I'm not inclined to take it unfortunately. I recommend you make a `mysql` anywhere on your system which points to `mysql5` or whatever you need. Then customize PATH so that your `mysql` comes before the system one. You can do that for a single request with `PATH=/path/to/custom_mysql:$PATH drush sql;dump ...`. If using a remote system, see env-vars key in [site.yml ](https://www.drush.org/latest/site-aliases/).

greg-1-anderson:
If we were to take config options here, the databases array wouldn't be the place for them; it would have to be in drush.yml.

Adjusting the $PATH is the "Drush way", but this is sometimes difficult for users. In this particular use case, the system mysql only exists as mysql5 so any mysql anywhere in the path should do the trick.

PR #4213, opened 2019-10-09
Uid mail sanitisation whitelist

miloskroulik:
This adds mail and uid based whitelist for SQL sanitization, as described in #3756 The test are running fine, but it's likely that PR still has some issues, because it's my first Drush patch.

PR #4725, opened 2021-04-08
Add sql:sanitize option to ignore admins in users_field_data

elissaheller:
This change adds a sql:sanitize option to ignore admins in users_field_data table when sanitizing. This is useful, for example, if sanitizing a database of customer data and importing to a QA environment where multiple Drupal administrators need to test. 

In the particular use case where this requirement came about, the site uses a third-party auth system which matches to Drupal users on email address, so if the emails are wiped, admins can't get in.

### DIFF
--- a/src/Drupal/Commands/sql/SanitizeUserTableCommands.php
+++ b/src/Drupal/Commands/sql/SanitizeUserTableCommands.php
@@ -74,7 +74,7 @@ class SanitizeUserTableCommands extends DrushCommands implements SanitizePluginI
             $messages[] = dt('User emails sanitized.');
         }
 
-        if ($this->isEnabled($options['ignore-admins']) && $options['ignore-admins'] === 'yes') {
+        if (in_array('ignore-admins', $options)) {
             $admins = $this->database->query("SELECT entity_id FROM user__roles WHERE roles_target_id='administrator'")->fetchCol();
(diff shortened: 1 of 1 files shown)


weitzman:
I'm a little torn on this. I do see its utility. The PR as it stands might be incomplete as it still sanitizes the User fields and Sessions for admins. Sometimes that’s desired and sometimes it it isn't. And even if we made that consistent, would all sanitize plugins be expected to skip processing for admins? I'm thinking of TFA sanitizing, for example.

weitzman:
If we had a uid whitelist, then your feature would be as as imple as making a script that grabbed all admins and passing it it to sql:sanitize. See https://github.com/drush-ops/drush/pull/4213

elissaheller:
Just made a change to allow multiple user roles to be ignored in sanitization, not just administrators. A cursory glance at #4213 suggests that the option added in that PR would meet the same need however: `--whitelist-mails=*@mycompany.org`.

Berdir:
FWIW, I think this is still useful even with the other issue merged. Our use case is that client admins (which aren't using our mail domain obviously for their users) can still log in on test environments.

dropping session seems fine, I don't think there are many valid scenarios where you could actually still use them. Also fine to keep user field values, all we care about is that clients can still log in with their e-mail and password.

weitzman:
#5808 has merged.

PR #4728, opened 2021-04-15
Support idlist option on migrate:messages

marvil07:
In the context of trying to make `migrate_tools` compatible with drush again, this change is bringing `--idlist` option for the `migrate:messages` command.
More context at [drupalspoons/migrate_tools#118: Migrate commands are now part of Drush core](https://gitlab.com/drupalspoons/migrate_tools/-/issues/118).

The current approach is just walking the full set.
This can be improved by using a custom \FilterIterator class to filter
the id map, as in migrate_tools implementation.
For now, only using a simple approach to avoid introducing a new class.


src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -493,8 +494,14 @@ public function resetStatus(string $migrationId): void
      * @param string $migrationId
      *   The ID of the migration.
      *
+     * @option idlist Comma-separated list of IDs to import. As an ID may have more than one column, concatenate the columns with the colon ':' separator.
+     *
      * @usage migrate:messages article
      *   Show all messages for the <info>article</info> migration
+     * @usage migrate:messages article --idlist=5
+     *   Import the article record with source ID 5.
+     * @usage migrate:messages node_revision --idlist=1:2,2:3,3:5
```
claudiu-cristea, line 503: Can we convert one of the IDs as strings and show how to properly quote an ID containing `:`. Also we can explain this is usage description.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -523,6 +530,18 @@ public function messages(string $migrationId): RowsOfFields
 
         $idMap = $migration->getIdMap();
         $sourceIdKeys = $this->getSourceIdKeys($idMap);
+        $idlist_source_id_values = [];
+        if (!empty($options['idlist'])) {
+            $keys = array_keys($migration->getSourcePlugin()->getIds());
+            array_walk($keys, function (&$value, $key) {
```
claudiu-cristea, line 536: Even it's a closure, we can strict type the params and the function return.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -543,6 +562,25 @@ public function messages(string $migrationId): RowsOfFields
             $row['source_ids'] = implode(' : ', $sourceIds);
             $row['destination_ids'] = implode(' : ', $destinationIds);
 
+            // Only show relevant messages.
+            // TODO This logic can be simplified if a new \FilterIterator child
```
claudiu-cristea, line 566: We have something similar I guess. Isn't `MigrateIdMapFilter` something that we can use here?

claudiu-cristea (dismissed):
Thank you for contribution. Apart from the inline remarks, this also needs test coverage.

src/Drupal/Commands/core/MigrateRunnerCommands.php
marvil07: Right, different name but similar. I will change the code to use it.

marvil07:
@claudiu-cristea, thanks for the feedback!

> Can we convert one of the IDs as strings and show how to properly quote an ID containing :. Also we can explain this is usage description.

During my testing of the feature, I think I discovered a bug, I opened https://github.com/drush-ops/drush/issues/4730 about it.


I made the changes related to the closure, and the documentation.

I have also added test coverage.
It was a nice and smooth experience to go over the tests :+1:
I liked the `woot` module migration flexibility, but took me a bit to realize it was dynamic nature, and failing cases are known and just a known set hard-coded on the `TestFailProcess` migrate process plugin.

Next steps: use the existing filter class.

marvil07:
That last item was less trivial than I was expecting :sweat_smile: 

Long story short:
- The filter iterator on the drush codebase does its job OK, but it sadly does not apply for message retrieval from an id map.
- There is a way in core to retrieve messages from one source ids set, but not for multiple.
- To work-around the problem, I am just iterating over the passed values on the CLI.

More details on the inline comment.

In the process I also fixed one related error on mainline: prevent error output if the id map is empty, e.g. before migrations has been executed; and moved the pre-processing of the message item into its own method for easier re-use.

@claudiu-cristea, this is now ready for a second review.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -524,30 +533,74 @@ public function messages(string $migrationId): RowsOfFields
...
+     * Preprocessing for message rows.
+     *
+     * Given an item inside the list generated by
+     * \Drupal\migrate\Plugin\MigrateIdMapInterface::getMessages(), prepare it
+     * for display.
+     *
+     * @param \StdClass $row
+     *   A messages to process.
+     * @param array $sourceIdKeys
+     *   The source ids keys, for reference.
```
claudiu-cristea, line 580: Same, s/ids/IDs

tests/functional/MigrateRunnerTest.php
```diff
@@ -271,6 +271,35 @@ public function testMigrateMessagesAndFieldSource(): void
         $this->assertSame('17', $output[2]['source_ids']);
         $this->assertEmpty($output[0]['destination_ids']);
 
+        // Only one invalid id.
```
claudiu-cristea, line 274: s/id/ID

tests/functional/MigrateRunnerTest.php
```diff
@@ -271,6 +271,35 @@ public function testMigrateMessagesAndFieldSource(): void
...
+        $this->drush('migrate:messages', ['test_migration'], ['format' => 'json', 'idlist' => '100']);
+        $output = $this->getOutputFromJSON();
+        $this->assertCount(0, $output);
+
+        // Only one valid id.
```
claudiu-cristea, line 279: s/id/ID

tests/functional/MigrateRunnerTest.php
```diff
@@ -271,6 +271,35 @@ public function testMigrateMessagesAndFieldSource(): void
...
+        $this->drush('migrate:messages', ['test_migration'], ['format' => 'json', 'idlist' => '2']);
+        $output = $this->getOutputFromJSON();
+        // @see \Drupal\woot\Plugin\migrate\process\TestFailProcess::transform()
+        $this->assertCount(1, $output);
+        $this->assertEquals(1, $output[0]['level']);
+        $this->assertSame('2', $output[0]['source_ids']);
+        $this->assertEmpty($output[0]['destination_ids']);
+        $this->assertSame('ID 2 should fail', $output[0]['message']);
+
+        // Three valid ids, two with data, one without, and one invalid id.
```
claudiu-cristea, line 289: s/ids/IDs and s/id/ID

claudiu-cristea (changes requested):
Thank you. There are only few nits left, otherwise this is good to go.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -696,6 +749,11 @@ protected function getMigrationPluginManager(): MigrationPluginManagerInterface
      */
     protected function getSourceIdKeys(MigrateIdMapInterface $idMap): array
     {
+        if (iterator_count($idMap) == 0) {
```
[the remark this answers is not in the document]
marvil07: Sure, it not strictly needed, since it behaves similar to `count()`, but the reason behind it may be just to make it easier to read (and avoid depending remembering details), so I added the change.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -524,30 +533,74 @@ public function messages(string $migrationId): RowsOfFields
...
+            // right thing filtering the data on the id map, sadly its
+            // getMessages() method does not take it into account the iterator,
+            // and retrieves data directly, e.g. at sql id map plugin.
+            // On the other side Drupal Core's
+            // \Drupal\migrate\Plugin\MigrateIdMapInterface only allows to
+            // filter by one source ids set, and not by multiple, on
+            // getMessages().
+            // For now, go over known ids passed directly, one at a time a
+            // work-around, at the cost of more queries in the usual sql id map,
+            // which is likely OK for its use, to show only few source ids
```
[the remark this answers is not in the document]
marvil07: I went with line breaks, just to help readability.
As you mention, it is just a big text, so making it continuous using periods without line breaks may make it harder to read.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -524,30 +533,74 @@ public function messages(string $migrationId): RowsOfFields
...
+            // messages.
+            foreach (MigrateUtils::parseIdList($options['idlist']) as $source_ids_value) {
```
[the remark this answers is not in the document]
marvil07: Instead of moving from `$source_ids_value` to `$sourceIdsValue`, I moved to `$sourceIdValues`, since that is used already on other part of the class, ad `executeMigration()`.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -524,30 +533,74 @@ public function messages(string $migrationId): RowsOfFields
...
                 }
             }
-            $row['source_ids'] = implode(' : ', $sourceIds);
-            $row['destination_ids'] = implode(' : ', $destinationIds);
-
-            $table[] = $row;
+            return new RowsOfFields($table);
+        }
+        $table = [];
+        foreach ($idMap->getMessages() as $soure_key => $row) {
```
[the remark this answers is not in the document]
marvil07: In this case, I removed the key from the loop, it is not used.

marvil07:
@claudiu-cristea, thanks for the feedback and the attention to detail.

I have added most of the requested changes, details in context in previous review thread, went I took a slightly different path.

marvil07:
About the in-ability to filter messages on an ID map by multiple items, I created a new Drupal core issue at [#3210257: Improve migrate messages retrieval](https://www.drupal.org/project/drupal/issues/3210257).

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -524,30 +533,74 @@ public function messages(string $migrationId): RowsOfFields
...
+            $table[] = $this->preprocessMessageRow($row, $sourceIdKeys);
         }
         return new RowsOfFields($table);
     }
 
+    /**
+     * Preprocessing for message rows.
```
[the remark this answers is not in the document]
claudiu-cristea: The change wasn't quite third person singular present tense verb but I've fixed myself.

claudiu-cristea (approved):
Thank you for your contribution.

PR #4731, opened 2021-04-18
Add return types in Drush class.

weitzman:


greg-1-anderson (approved):
Sure; it's safe to start doing this for classes that won't be overridden.

PR #4733, opened 2021-04-20
Run hook_requirements() on pm:enable

### LINKED ISSUE #4190: Module installation ignores `hook_requirements`.

codebymikey:
This resolves #4190 and supersedes the #4337 PR to include functional tests as well as read all requirement errors.

### DIFF
--- a/src/Drupal/Commands/pm/PmCommands.php
+++ b/src/Drupal/Commands/pm/PmCommands.php
@@ -71,6 +71,7 @@ class PmCommands extends DrushCommands
      * @command pm:enable
      * @param $modules A comma delimited list of modules.
      * @aliases en,pm-enable
+     * @bootstrap full
      */
     public function enable(array $modules)
     {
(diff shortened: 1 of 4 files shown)


src/Drupal/Commands/pm/PmCommands.php
```diff
@@ -87,6 +99,8 @@ public function enable(array $modules)
             }
         }
 
+        $this->validateInstallModules($todo);
```
weitzman, line 102: - You can remove this line and instead add `@hook validate pm:enable` at top of validateInstallModules method. 
- I think this method should be called `validateEnableModules`

src/Drupal/Commands/pm/PmCommands.php
```diff
@@ -279,6 +293,39 @@ public function addInstallDependencies($modules)
...
+                    if (isset($requirement['severity']) && $requirement['severity'] == REQUIREMENT_ERROR) {
+                        $message = $requirement['description'];
+                        if (isset($requirement['value']) && $requirement['value']) {
+                            $message = dt('@requirements_message (Currently using @item version @version)', ['@requirements_message' => $requirement['description'], '@item' => $requirement['title'], '@version' => $requirement['value']]);
+                        }
+                        $reasons[$id] = $message;
+                    }
+                }
+
+                throw new \Exception(sprintf("Unable to install module '%s' due to unmet requirement(s):%s", $module, "\n  - " . implode("\n  - ", $reasons)));
```
weitzman, line 324: Thanks. Given our long history of not validating this, lets give the user the option to proceed despite the validation error. See https://github.com/drush-ops/drush/blob/10.x/src/Commands/core/UpdateDBCommands.php#L54-L59 for an example from updatedb.

weitzman (changes requested):

src/Drupal/Commands/pm/PmCommands.php
codebymikey: All done, it'll default to `false` in order to fail by default.

weitzman: Lets default to continue like updatedb does (i.e. no exception). Otherwise we will have to wait for a major version for this.

codebymikey: I've updated the logic and test to reflect that.

And yes, that makes sense since we want to maintain backwards compatibility and avoid breaking any existing integrations/tests which rely on it (but one could also argue that if the module being installed doesn't meet the platform requirements in the first place, there's something wrong with their integration, and the failure _should_ be the appropriate default behaviour, and should lead to them fixing it on their end, and opt out with `--no`).

PR #4735, opened 2021-04-20
Add new core:route command

weitzman:
Not sure its really needed. Welcome feedback.


weitzman:
Hmm. Maybe Druapl 9 changed its update function descriptions because those tests are spontaneously failing.

PR #4737, opened 2021-04-23
Temporarily allow skip-progress-bar option on migrate commands

marvil07:
`migrate_tools` had a `--skip-progress-bar` option, that is equivalent to drush core `--no-progress` option, in both `migrate:import and `migrate:rollback`.

Temporarily allowing `--skip-progress-bar` increases compatibility on the CLI while both sets of commands are tried to be unified.

Supporting this option during drush `10.x`, can help improve the change over drush core migrate commands.
Later on `11.x`, this compatibility commit can be reverted.

### DIFF
--- a/src/Drupal/Commands/core/MigrateRunnerCommands.php
+++ b/src/Drupal/Commands/core/MigrateRunnerCommands.php
@@ -233,6 +233,7 @@ class MigrateRunnerCommands extends DrushCommands
      * @option timestamp Show progress ending timestamp in progress messages
      * @option total Show total processed item number in progress messages
      * @option progress Show progress bar
+     * @option skip-progress-bar Deprecated, please use --no-progress.
      * @option delete Delete destination records missed from the source
      *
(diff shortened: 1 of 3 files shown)


weitzman:
I dont see much benefit in putting in immediately deprecated code. This can be handled with a docs page. Or migrate tools can add in the option if that module still intends to be enabled.

claudiu-cristea:
Thank you. However, I don't see the practical value of this addition. This will only confuse the user. I’ve deliberately converted `--skip-progress-bar` to a logical `--option` / `--no-option` model which is an annotated commands standard on/off approach with the option defaulting to `true`.

claudiu-cristea:
Closing as "won't fix"

marvil07:
@weitzman, @claudiu-cristea: Thanks for the feedback here.

I will try to add this on `migrate_tools` new version as a compatibility layer on top of drush commands, where it probably makes more sense.

PR #4751, opened 2021-05-07
Fix handling of missing source rows in migrate:import

### LINKED ISSUE #4750: handleMissingSourceRows() forces --update on each migration run
**Describe the bug**
`handleMissingSourceRows()` sets `MigrateIdMapInterface::STATUS_NEEDS_UPDATE` to all source rows. This causes each migration to run as if `--update` option has been set. It also breaks `track_changes` source option by forcing update on items that were not modified.

**To Reproduce**
Run any migration (without any options) multiple times. 

```
drush mim example_migration
drush mim example_migration
```

**Expected behavior**

`track_changes=false`
On second run I expect to see:
```
Processed 0 items (0 created, 0 updated, 0 failed, 0 ignored)
```

If `track_changes=true` I expect to see only modified items to be updated during migration

E.g. if 2 items out of 1000 got modified I expect to see

```
Processed 2 items (0 created, 2 updated, 0 failed, 0 ignored)
```

**Actual behavior**

On second run (and all subsequent migration runs) all the items are updated. E.g. if source has 1000 rows

```
Processed 1000 items (0 created, 1000 updated, 0 failed, 0 ignored)
```

If `track_changes=true` all source items are still updated on each migration even if none has been modified.

```
Processed 1000 items (0 created, 1000 updated, 0 failed, 0 ignored)
```

**Workaround**
None that I'm aware of.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 9.1.7
| PHP version     | 7.3
| OS?             | Linux

**Additional information**

I suspect that highwater mark may also be affected by this bug but didn't verify.

vaish:
With this PR:
1. Update option is not forced on all migrations. It has to be explicitly specified on command line.
2. `track_changes` works as expected: only modified items will be updated
3. Time to complete migration is significantly reduced in some cases (especially for larger sources)

Changes introduced in this PR:

1. Handling of missing source rows is moved post import so that we get a chance to collect source rows without having to perform another dedicated iteration through source.
1. By collecting existing source rows during the import we:
    - Remove the need for dedicated iteration through all source items reducing the time required to complete migration.
    - Avoid modifying status in migrate map table which was forcing unnecessary updates of all source rows. This again reduces the time required to complete migration.
1. Deletion of missing rows is limited only to migrations that process all source rows. Options that limit number of source rows are not compatible with deleting missing rows because it's not possible to reliably determine which items still exist in the source and should not be deleted. Examples:
    - Running `--idlist` along with `--delete` would cause all items not explicitly listed with `--idlist` option to be deleted. 
    - Running `--limit` along with `--delete` may in some cases cause all items above the limit to be deleted.
    - If `highwater_mark_property` is used with any source that extends `SqlBase` plugin, it will fetch from database only items above the highwater mark. Rows below highwater mark will be considered missing and marked for deletion.

Resolves #4750

### DIFF
--- a/src/Drupal/Commands/core/MigrateRunnerCommands.php
+++ b/src/Drupal/Commands/core/MigrateRunnerCommands.php
@@ -233,7 +233,7 @@ class MigrateRunnerCommands extends DrushCommands
      * @option timestamp Show progress ending timestamp in progress messages
      * @option total Show total processed item number in progress messages
      * @option progress Show progress bar
-     * @option delete Delete destination records missed from the source
+     * @option delete Delete destination records missed from the source. Not compatible with --limit and --idlist options, and high_water_property source configuration key.
      *
      * @usage migrate:import --all
      *   Perform all migrations
--- a/src/Drupal/Migrate/MigrateExecutable.php
+++ b/src/Drupal/Migrate/MigrateExecutable.php
@@ -85,6 +85,13 @@ class MigrateExecutable extends MigrateExecutableBase
     protected $idlist;
 
     /**
+     * List of all source IDs that are found in source during this migration.
+     *
+     * @var array
+     */
+    protected $allSourceIdValues = [];
+
+    /**
      * Count of number of items processed so far in this migration.
      *
      * @var int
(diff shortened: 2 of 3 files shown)


jcisio:
I can confirm that with this PR it works as expected. No more handling of missing source rows when there is either --limit or --idlist is a good trade-off (not to say it's more logical) and a performance gain.

The only problem I've found is when high_water_property is set. This is more a technical limit than logical. Given the performance gain, should we leave it and add a new command to delete missing source rows in all cases?

claudiu-cristea (commented):
Thank you for tracking and fixing the problem, @vaish. Nice work! I'm just thinking that this deserves a test coverage. Looking to #4750 issue, I see a nice description of expected & actual behavior. Could you "translate" the expected behavior into a test?

claudiu-cristea:
Closing in favour of #4778

claudiu-cristea:
Actually, looking at `MigrateRunnerTest::testMissingSourceRows()` it seems that the scenario described in #4750 is already covered.

PR #4761, opened 2021-05-20
Allow comparing config snapshot with active config via config:status

tgifford-webfirst:
For some sites we don't install the Configuration Manager (config) module, which is currently the only place I know of that compares the snapshot and active config. This PR gives drush that ability.

There is a bit of an API change for the --state option to config:status, since the existing labels don't accurately describe the new behavior. This could be addressed without an API change, but then the help text would probably get a little more confusing.

I've also refactored the code a bit to delay the switch from a token (e.g. identical) to the label (e.g. Identical). It would probably be a good idea to add tests for the --state filter, but none appear to exist at the moment.


weitzman:
Not much demand for this.

Issue #4762, opened 2021-05-26
PHP 8.0 + drush8 TypeError array_key_exists() arg must be of type array, SimpleXMLElement given in pm:download


mcdruid:
**Describe the bug**
In PHP 8.0 drush8 hits a TypeError when running pm:download (and maybe other pm commands)

**To Reproduce**

```
$ drush pm:download seckit
WD php: TypeError: array_key_exists(): Argument #2 ($array) must be of type array, SimpleXMLElement given in array_key_exists() (line 205 of phar:///path/to/drush.phar/vendor/composer/../../lib/Drush/UpdateService/Project.php).                                  [error]
Drush command terminated abnormally due to an unrecoverable error.
```

**Expected behavior**
The project is downloaded.

**Actual behavior**
An error thwarts efforts to download the project.

**Workaround**
Don't use PHP 8.0 :)

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.4.8 (phar)
| Drupal version? | 7.x
| PHP version     | 8.0
| OS?             | Linux

**Additional information**
Not tried other pm: commands yet but they may also be affected if the problem happens parsing release feeds from d.o ?

seamuslee001:
now that #4763 has been merged this seems to work locally for me on php8.0 with a locally built phar of drush8 is there any chance for a release @greg-1-anderson ?

PR #4763, opened 2021-05-26
Workaround for array vs SimpleXMLElement in PHP 8 (#4762)

mcdruid:
I'm sure there are more elegant ways of fixing the problem, but this seems to get the pm: commands working again in PHP 8.0


greg-1-anderson:
LGTM - thanks!

rgpublic:
In case anyone desperately needs a PHAR-release with this fix for PHP 8, here is how I built mine. I created an empty folder somewhere, cd'd into it and then entered the following commands:

* `git clone https://github.com/drush-ops/drush.git`
* `git checkout remotes/origin/8.x`
* `composer install`
* `composer-phar:install-tools`
* `composer-global phar:build`
* `cp drush.phar /usr/bin/drush-legacy`

PR #4764, opened 2021-05-31
New no-dev option for security commands

prudloff-insite:
We use `drush sec` and `drush sec-php` to check for vulnerabilities on our websites automatically, but we don't want to be notified for vulnerabilities in development dependencies because these are never installed on the server.


weitzman:
Code looks good except for a minor code style violation.

Ideally we exercise this new option in SecurityUpdatesTest and assert that the expected packages are considered.

prudloff-insite:
I added a basic test but I'm afraid it might break if someday there is indeed a security vulnerability in one of the production dependencies.

weitzman:
Thanks. Failing tests are unrelated. Prob a chance in Drupal 9 updatedb-status formatting.

PR #4766, opened 2021-06-03
Fix install link for drush 8.

YesCT:
# Motivation

On https://github.com/drush-ops/drush/tree/8.x
the readme link for "Installing (and Upgrading)" goes to https://docs.drush.org/en/master/install/ which does not work as intended.

## Before

Screen Shot 2021-06-03 at 9 21 13 AM

Screen Shot 2021-06-03 at 9 21 22 AM

## After

Link should go to http://docs.drush.org/en/8.x/install/

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Drush is a command line shell and Unix scripting interface for Drupal. Drush cor
 
 Resources
 -----------
-* [Installing (and Upgrading)](http://docs.drush.org/en/master/install/)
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Thanks. I note that the 8.x installation docs say that the most recent version is 10.x, but fixing up that page does not have to be part of this PR.

YesCT:
@weitzman @greg-1-anderson Thanks!

PR #4767, opened 2021-06-04
Add a watchdog tail command

beejeebus:
WIP, feedback welcome.

Only tested with `composer functional -- --filter testWatchdogTail`.

### DIFF
--- a/src/Drupal/Commands/core/WatchdogCommands.php
+++ b/src/Drupal/Commands/core/WatchdogCommands.php
@@ -11,6 +11,7 @@ use Drupal\Component\Utility\Unicode;
 use Drupal\Component\Utility\Html;
 use Drush\Drupal\DrupalUtil;
 use Drush\Exceptions\UserAbortException;
+use Symfony\Component\Console\Output\OutputInterface;
 
 class WatchdogCommands extends DrushCommands
 {
@@ -34,8 +35,6 @@ class WatchdogCommands extends DrushCommands
      *   Show a listing of most recent 10 messages with a severity of notice.
      * @usage drush watchdog:show --type=php
      *   Show a listing of most recent 10 messages of type php
-     * @usage  while sleep 2; do drush watchdog:show; done
-     *   Every 2 seconds, show the most recent 10 messages.
      * @aliases wd-show,ws,watchdog-show
      * @validate-module-enabled dblog
      * @field-labels
@@ -68,6 +67,7 @@ class WatchdogCommands extends DrushCommands
         }
         if (empty($table)) {
             $this->logger()->notice(dt('No log messages available.'));
+            return;
         } else {
             return new RowsOfFields($table);
         }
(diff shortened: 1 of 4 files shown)


src/TestTraits/CliTestTrait.php
```diff
@@ -75,6 +75,42 @@ public function getErrorOutputRaw()
...
+     * @param string $command
+     *   The actual command line to run.
+     * @param sting cd
+     *   The directory to run the command in.
+     * @param array $env
+     *  Extra environment variables.
+     * @param string $input
+     *   A string representing the STDIN that is piped to the command.
+     */
+    public function startExecute($command, $cd = null, $env = null, $input = null)
```
weitzman, line 90: Is there no way to use $this->execute()? Would save a lot of duplicated code.

beejeebus: there is a lot of duplicated code, and i'm open to ways to reduce it.

the main new capability needed is 'run child process in the background, without blocking on it's completion, with access to output for tests'.

my first thought was that should be a separate method, rather than another flag and return type from execute(). so i was planning to reduce duplicate code by factoring out the shared bits in execute() and startExecute(). i also don't love the name startExecute(), happy to get better suggestions.

make sense?

src/Drupal/Commands/core/WatchdogCommands.php
```diff
@@ -51,25 +50,63 @@ class WatchdogCommands extends DrushCommands
...
+
+            $rsc = $query->execute();
+            while ($result = $rsc->fetchObject()) {
+                if ($result->wid > $last_seen_wid) {
+                  $last_seen_wid = $result->wid;
+                }
+                // TODO: how should the output string be built?
+                // Just printing stuff to show it doing something.
+                $row = $this->formatResult($result, $options['extended']);
+                print "\n  " . $row->wid . "\t";
```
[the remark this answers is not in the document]
beejeebus: ack. @greg-1-anderson [gave me some pointers](https://app.slack.com/client/T06GX3JTS/C62H9CWQM/thread/C62H9CWQM-1622834281.067000) re. output in Slack. one idea was to create another command for the tail case. thoughts?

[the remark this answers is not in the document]
greg-1-anderson: It's super weird when a command returns RowsOfFields sometimes, and nothing (or some completely different type) at other times, but since `--tail` never returns at all, I guess it's not so bad that the behavior is inconsistent.

I'm trying to deprecated `$this->io()` and `$this->writeln()` et. al. in Robo; the better way is to add an OutputInterface as the first parameter. Then you'll get an object you can `$writeln` on.

src/TestTraits/CliTestTrait.php
weitzman: How about we sadd a $timeout param and stop hardocding `0` as timeout [in this line](https://github.com/drush-ops/drush/blob/e343e01486a740eace4b47a58a849cada8fb6a6b/src/TestTraits/CliTestTrait.php#L102). And we add the same timeout param to drush() method which thencalls execute(). Then the test can just call something like `$this->drush(watchdog:show --tail, $timeout)`

beejeebus: I think that's a useful but slightly different improvement.

Too feel confident that this works, I'd like to create multiple watchdog entries while the process running `watchdog:show --tail` is running. That requires interactivity in the test process, so we can't block on the `$this->drush()` call.

src/Drupal/Commands/core/WatchdogCommands.php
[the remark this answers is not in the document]
beejeebus: I've update the PR to use an injected OutputInterface object.

src/TestTraits/CliTestTrait.php
beejeebus: updated the PR to factor out much of the duplicated code.

beejeebus:
I get the same failures on master locally as in CI. Could use some pointers for figuring out the tests.

weitzman:
The test failure is unrelated and is being fixed elsewhere.

maskedjellybean:
Is this available in any release yet, and if so, which versions?

weitzman:
Its available since 10.6. FYI this info is now viewable on the command's doc page https://www.drush.org/latest/commands/watchdog_tail/

PR #4773, opened 2021-06-14
Use full path to drush instead of composer symlink

TheGuit:
Hi,

I think it's safer to use the real path to drush script instead of composer's symlink. The symlink could be broken, during for example packaging in a format that don't support symlink (eg. zip).

We have take the habits to use full path to script when using drush, but it fail on some projet because drush use the symlink not the full path.

Open to discussion if there is some cons that I haven't anticipate.

Regards

### DIFF
--- a/src/SiteAlias/ProcessManager.php
+++ b/src/SiteAlias/ProcessManager.php
@@ -81,10 +81,10 @@ class ProcessManager extends ConsolidationProcessManager
             return $defaultDrushScript;
         }
 
-        // If the target is a local Drupal site that has a vendor/bin/drush,
-        // then use that.
+        // If the target is a local Drupal site that has a
+        // vendor/drush/drush/drush, then use that.
         if ($siteAlias->hasRoot()) {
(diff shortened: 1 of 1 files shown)


weitzman:
This is more brittle IMO. Drush could change its filenames. Also some folks use the bin/drush as a custom wrapper.

TheGuit:
> This is more brittle IMO. Drush could change its filenames.

As `vendor/bin/drush` is a symlink to vendor/drush/drush/drush`, as you can see in `composer.json` : 

```json
  "bin": [
    "drush"
  ],
```

Changing drush script filename would change both. So I don't think it's risky.

> Also some folks use the bin/drush as a custom wrapper.

`bin/drush` would always works as usual, only thing is drush package don't relies on the composer symlink.

weitzman:
> bin/drush would always works as usual, 

The wrapper would not run, and presumably would be needed.

TheGuit:
Hi,

> The wrapper would not run, and presumably would be needed.

Why are you saying that the wrapper would not run ?

Issue #4675, opened 2021-02-24
Add param and return type hints, OO interfaces, etc.


weitzman:
Major moderning and leveraging of recent PHP improvements. Some may break backward compat for commands.

greg-1-anderson:
If we break bc for module commands, maybe we should also change the way commandfile discovery works. I regret putting Drush commands into the Symfony DI container. We should switch to using a static factory method that takes the DI container as a parameter.

beejeebus:
this looks like fun.

can I get some suggestions for chunking this up?

weitzman:
I'm undecided about breaking backward compat for commands that ship with modules. Some approaches are discussed at https://www.drupal.org/project/drupal/issues/3050720

beejeebus:
ack, thanks, that's good context.

to get a feel for the tools around, I'll try some experiments with [Rector](https://github.com/rectorphp/rector).

weitzman:
I think internal classes like `src/Boot`, `src/Preflight`, `src/Runtime`, `src/SiteAlias`, `src/Symfony` are good places to start.

beejeebus:
Ok, #4775 is worth another look, might be good enough now.

What next?

weitzman:
Done.

PR #4775, opened 2021-06-16
run Rector on some src dirs, add types in function signatures and returns

beejeebus:
This is part 1 of n from #4675.

It might be good enough to merge now, and we can discuss next bits.

### DIFF
--- a/src/Boot/BootstrapManager.php
+++ b/src/Boot/BootstrapManager.php
@@ -53,7 +53,7 @@ class BootstrapManager implements LoggerAwareInterface, AutoloaderAwareInterface
     /**
      * @param int $phase
      */
-    protected function setPhase($phase)
+    protected function setPhase(int $phase)
     {
         if ($this->bootstrap) {
             $this->bootstrap()->setPhase($phase);
@@ -73,7 +73,7 @@ class BootstrapManager implements LoggerAwareInterface, AutoloaderAwareInterface
         }
     }
 
-    public function drupalFinder()
+    public function drupalFinder(): \DrupalFinder\DrupalFinder
     {
         if (!isset($this->drupalFinder)) {
             $this->drupalFinder = new DrupalFinder();
@@ -149,7 +149,7 @@ class BootstrapManager implements LoggerAwareInterface, AutoloaderAwareInterface
      *
      * @return \Drush\Boot\Boot
      */
-    public function bootstrap()
+    public function bootstrap(): \Drush\Boot\Boot
     {
         if (!$this->bootstrap) {
             $this->bootstrap = $this->selectBootstrapClass();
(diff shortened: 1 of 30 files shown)


src/Boot/BootstrapManager.php
```diff
@@ -410,15 +390,12 @@ protected function maxPhaseLimit($bootstrap_str)
     /**
      * Bootstrap to the specified phase.
      *
-     * @param int $max_phase_index
```
weitzman, line 413: Shouldn’t we only remove the type hint and not the whole line? The description lacks context without the `@param` line above.

src/Config/ConfigLocator.php
```diff
@@ -160,15 +158,15 @@ public function sources()
      *
      * @return string[]
      */
-    public function configFilePaths()
+    public function configFilePaths(): array
```
weitzman, line 161: Should we change these array hints to Iterable per https://steemit.com/php/@crell/php-never-type-hint-on-arrays?

beejeebus: Can we do that as a follow up? My intent is to keep the non-scripted changes low, and to do multiple passes with multiple PRs.

src/Boot/BootstrapManager.php
beejeebus: I'll see if there's a way to teach rector to leave those alone.

src/Config/ConfigLocator.php
weitzman: Yes, makes sense.

src/Boot/BootstrapManager.php
beejeebus: I found a way to make this stop, with a couple of one-line hacks to Rector.

src/Boot/DrupalBoot8.php
```diff
@@ -36,23 +36,23 @@ class DrupalBoot8 extends DrupalBoot implements AutoloaderAwareInterface
     /**
      * @return \Symfony\Component\HttpFoundation\Request
      */
-    public function getRequest()
+    public function getRequest(): \Symfony\Component\HttpFoundation\Request
```
weitzman, line 39: When we add a return hint like this, we should remove the Doxygen when the doxygen has no description (e.g. this method). Is that possible?

beejeebus: Let me poke at it. The "fix" for your initial request was to comment out the calls to remove docblock tags inside Rector, which is a very brute-force approach.

To be smahhter, I think we'll need to override these steps with our own, or contribute some patches upstream.

I'm ok to work on the longer steps if you think it's a worthwhile investment for Drush.

weitzman: Yeah, I think its worthwhile ... Would be great if you could put in a few hours and see how far you get. Thanks!

weitzman: If you arent available for more work on this, I'm OK with merging what we have.

weitzman:
We did it in a couple PRs. #4959  and #4963

PR #4782, opened 2021-06-21
Fix "Migrate status with empty --tag doesn't work"

claudiu-cristea:


weitzman:
I dont think we do grouping within a table anywhere else in Drush. We used to, and it got awkward. What we do is use a default output format of yaml instead of table. Maybe it is less scannable, but it supports grouping well. For example see `drush role:list`, `drush site:alias`. I'm not 100% against using tables with if everyone wants that here.

PR #4792, opened 2021-07-15
Fix locale aliases (drush-ops#4791)

jcnventura:
This fixes the incorrectly set drush-9 style locale aliases.


greg-1-anderson:
Thanks for catching this mistake!

PR #4793, opened 2021-07-19
Remove pm:info usage example from role:perm:add annotation

richardbporter:
The `role:perm:add --help` text has an example of listing permissions with `pm:info`. This doesn't work anymore and should probably be removed or updated. This PR just removes it.


weitzman:
Thanks. No replacement that I can think of.

Issue #4794, opened 2021-07-19
queue:run command without specifying lease-time does not fall back to 30 


akasake:
**Describe the bug**
Using drush queue:run without specifying lease-time does not fall back to 30 and throws InvalidArgumentException in DatabaseQueue::claimItem 
 https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/lib/Drupal/Core/Queue/DatabaseQueue.php#L116

**To Reproduce**
Run queue without lease-time

**Expected behavior**
When no lease-time is given in command, it will behave the same way cron processes the queue and thus fallback to time set in cron. https://git.drupalcode.org/project/drupal/-/blob/8.9.x/core/lib/Drupal/Core/Cron.php#L177

**Actual behavior**
`$lease_time = null` and DatabaseQueue::claimItem() throws InvalidArgumentException

**Workaround**
Is there another way to do the desired action?
specify lease-time in command every time (`drush queue:run 'random-index' --lease-time=30 -v`)

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 9.x
| PHP version     | 8.x
| OS?             | Mac

**Additional information**

akasake:
```
diff --git a/src/Drupal/Commands/core/QueueCommands.php b/src/Drupal/Commands/core/QueueCommands.php
index a9a9914c2..db2fae835 100644
--- a/src/Drupal/Commands/core/QueueCommands.php
+++ b/src/Drupal/Commands/core/QueueCommands.php
@@ -66,12 +66,14 @@ class QueueCommands extends DrushCommands
         $items_limit = (int) $options['items-limit'];
         $start = microtime(true);
         $worker = $this->getWorkerManager()->createInstance($name);
+        $info = $this->getWorkerManager()->getDefinition($name);
         $end = time() + $time_limit;
         $queue = $this->getQueue($name);
         $count = 0;
         $remaining = $time_limit;
+        $lease_time = $options['lease-time'] ?? $info['cron']['time'] ?? 30;

-        while ((!$time_limit || $remaining > 0) && (!$items_limit || $count < $items_limit) && ($item = $queue->claimItem($options['lease-time']))) {
+        while ((!$time_limit || $remaining > 0) && (!$items_limit || $count < $items_limit) && ($item = $queue->claimItem($lease_time))) {
             try {
[...]

PR #4795, opened 2021-07-19
#4794 Add fallback for lease_time when running queue

akasake:


weitzman:
Any thoughts @juampynr or @damiankloip or @davereid or @kevin-dutra? FYI we made relevant changes in #4322

kevin-dutra:
Makes sense to me. There definitely should be a fallback lease time so that we don't end up with a NULL value falling through to `claimItem()`, otherwise an item will basically be immediately available for processing again and your queue could get stuck.

weitzman:
thanks @akasake and @kevin-dutra

PR #4798, opened 2021-07-24
Add support for providing command/hook information via php8 Attributes.

weitzman:
A start on bringing the new [annotated-command feature](https://github.com/consolidation/annotated-command/pull/221/files) into Drush. For now I just copied the test commandfile from annotated-command

1. The first commit in this PR is a minor composer thrash to bring in `dev-main` of consolidation/annotated-command. It is possible that I updated more than needed. In any case, this PR raises our dependencies such that we need a new Drush major version (Drupal 9+, PHP 7.4+, etc. Drush 11 might even require Drupal 10+). Note that this PR targets opens a new Drush 11.x branch.
2. Got rid of `core:init` command as it was failing tests with new Robo and its a relic of old global drush days.

Todo
-----------------
- [x] Upgrade PHPCS to get rid of test fail
- [ ] This PR depends on https://github.com/consolidation/annotated-command/pull/223
- [ ] Need Robo to tolerate new annotated-command v5 [See this line](https://github.com/drush-ops/drush/pull/4798/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R39).

/cc @beejeebus @greg-1-anderson


greg-1-anderson:
You should rebase this with #4814 once that lands for better backwards compatibility. Regarding annotated-command 5.x, I think we should just backport the attributes stuff to 4.x once we feel it's stable, and have Drush / Robo keep using 4.x.  So far the attributes code is b/c with 4.x, and I have no other need to make a 5.x branch.

src/Boot/BootstrapManager.php
```diff
@@ -289,7 +289,7 @@ public function doBootstrap($phase, $phase_max = false, AnnotationData $annotati
             if ($phase_index > $bootstrapped_phase) {
                 if ($result = $this->bootstrapValidate($phase_index)) {
                     if (method_exists($bootstrap, $current_phase)) {
-                        $this->logger->log(LogLevel::BOOTSTRAP, 'Drush bootstrap phase: {function}()', ['function' => $current_phase]);
+                        $this->logger->log(LogLevel::INFO, 'Drush bootstrap phase: {function}()', ['function' => $current_phase]);
```
greg-1-anderson, line 292: Once #4814 lands, it won't be necessary to switch all of these to the standard log levels. You might still want to do this; it would probably be allowable to make a change like this in a minor version, but it might be better to wait until Drush 11.

weitzman:
Closed in favor of #4821

PR #4800, opened 2021-07-28
Update deploy:hook command to use batch system

bircher:
The original deploy:hook implementation passed the $sandbox variable to the deploy hooks to have the same signature as the post_update hooks. But we faked the batch process. However after using this for a while we found that it is necessary to do the batching in order to reduce the memory consumption that is unavoidable due to lots of static caches not being able to be cleared.


weitzman (approved):
LGTM. Lets give folks a day or two to comment, and then we merge.

rafatwork:
Just wanted to mention that I tested this with a deploy hook of mine that was causing out of memory issue. This PR fixed it :) Thanks!

Issue #2352, opened 2016-09-21
Support utf8mb4 on drush site-install


nvaken:
Since Drupal 7.50 it's possible to [setup Drupal with UTF8MB4 support](https://www.drupal.org/node/2754539). 

I would love to see support for UTF8MB4 installs. Would it be possible to add a `--utf8mb4` flag to make sure the site is being installed with the right settings? 
## Example command

```
drush site-install --db-url=mysql://root:pass@localhost:port/dbname --utf8mb4=true
```
### Test the database service settings (out of scope?)

Possibly out of scope, but a nice additional step might be to check if the database is set up properly to allow UTF8MB4. In that case, Drush should check for the following MySQL settings:

```
[mysqld]
innodb_large_prefix=true
innodb_file_format=barracuda
innodb_file_per_table=true
```
### Settings.php

I think the only thing on site installation is that it will need to setup the correct settings for the database in `settings.php`:

```
$databases['default']['default'] = array(
  'driver' => 'mysql',
  'database' => 'databasename',
  'username' => 'username',
  'password' => 'password',
  'host' => 'localhost',
  'charset' => 'utf8mb4',     // Add these automatically
  'collation' => 'utf8mb4_general_ci',     // Add these automatically
);
```
### Setup database with default utf8mb4_general_ci collation

Make sure the database has the correct collation selected as default.

```
ALTER DATABASE database_name CHARACTER SET = utf8mb4 COLLATE = utf8mb4_general_ci;
```

weitzman:
Whats needed is for someone to research what parameters should be added when we call install_drupal() such Drupal writes the proper collation and charset when it appends to settings.php. One a quick glance, I dont see a way to pass that to install_drupal() yet.

gitressa:
It would be really awesome to be able to define another collation as well, like utf8mb4_danish_ci.

ahwebd:
Would be very helpful to specify the desired character set and collation when creating the db,

as a temporary solution I'm hacking file "drush/lib/Drush/Sql/Sqlmysql.php", I changed line 88:

`    $sql[] = sprintf('CREATE DATABASE %s /*!40100 DEFAULT CHARACTER SET utf8 */;', $dbname);
`

`    $sql[] = sprintf('CREATE DATABASE %s /*!40100 DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci */;', $dbname);
`

rwalkerands:
[...]

PR #4810, opened 2021-08-20
Support utf8mb4 on drush site-install #2352

rwalkerands:
Support utf8mb4 character set on drush site-install. See #2352


weitzman:
Thanks ... The CR at https://www.drupal.org/node/2754539 suggests that there are php.ini prerequisites for this to work. Further, this wont help get 2 values into settings.php. I dont know how important that is. Needs research.

Issue #4813, opened 2021-08-28
drush sset not working in drush 10.6.0


holybiber:
**Describe the bug**
As of drush 10.6.0, the following commands don't work as expected for me:
`drush sset system.maintenance_mode 1 --input-format=integer`
`drush state:set system.cron_last 1406682882 --input-format=integer`
Both fail with
> Too many arguments, expected arguments "command" "key"

Also replacing `sset` with `state:set` and/or removing `--input-format=integer` doesn't make a difference.

I tried it on two systems. When I revert back to drush 10.5.0, everything works as expected.

**To Reproduce**
Run the commands above with drush 10.6.0

**Expected behavior**
Put site to maintenance mode / update timestamp for last cron run

**Actual behavior**
drush fails with the error mentioned above

**Workaround**
Using drush 10.5.0

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.6.0 (installed via composer)
| Drupal version? | 9.2.4
| PHP version     | 7.4.3 / 7.3.29
| OS?             | Linux

shean42:
The cset command is affected by this too.

ThomWilhelm:
I'm hitting this same issue. I too assumed it was the recent upgrade to drush 10.6.0 that broke this. But reverting to 10.5.0 I still receive the same error. Any workarounds would be greatly appreciated!

ThomWilhelm:
Workaround, pin the **consolidation/annotated-command** package to a version earlier than 4.3.0, eg. so in your composer.json:

**"consolidation/annotated-command": "4.2.4",**

greg-1-anderson:
Seems to be related to the release of consolidation/annotated-command:4.3.0. Pin to ~4.2.0 as a workaround. Investigating.

greg-1-anderson:
I've identified the issue, and think I can probably resolve it with a new annotated-command release without requiring any changes to Drush.

greg-1-anderson:
https://github.com/consolidation/annotated-command/pull/231 seems to clear this up. Will release as 4.3.1.

shean42:
Looks good, thanks much!!

greg-1-anderson:
Fixed in consolidation/annotated-command 4.3.1; run `composer update` to resolve.

ThomWilhelm:
Thanks for the quick fix!

holybiber:
For some reason on the one system I first had to do a `drush cr`.
But now it's working fine again, thank you very much!

PR #4814, opened 2021-08-30
Fixes #4812: Support Robo 3 while maintaining compatibility with Robo 2 and Robo 1.

### LINKED ISSUE #4812: Can't update robo
**Describe the bug**
We are currently on drush/drush 10.6.0 per the recommendation of Drupal 9 we are on PHP 8
But drush is requiring consolidation/robo: ^1.4.11 || ^2 which IS NOT compatible with PHP 8
Robo has a version 3 that is but we can't update because Drush is blocking that.

greg-1-anderson:
Going to take a swing at #4812. First, though, let's run the tests with latest dependencies and no other changes to see if maybe #4813 is related to the new Annotated Command library release.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     "composer/semver": "^1.4 || ^3",
     "consolidation/config": "^1.2",
     "consolidation/filter-via-dot-access-data": "^1",
-    "consolidation/robo": "^1.4.11 || ^2",
+    "consolidation/robo": "^1.4.11 || ^2 || ^3",
     "consolidation/site-alias": "^3.0.0@stable",
     "consolidation/site-process": "^2.1 || ^4",
     "enlightn/security-checker": "^1",
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
     "enlightn/security-checker": "^1",
     "grasmash/yaml-expander": "^1.1.1",
     "guzzlehttp/guzzle": "^6.3 || ^7.0",
-    "league/container": "~2",
+    "league/container": "^2.5 || ^3.4",
     "psr/log": "~1.0",
     "psy/psysh": "~0.6",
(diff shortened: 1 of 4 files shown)


greg-1-anderson:
First test failed due to bc break in Annotated Command 4.3.0.  After the release of Annotated Command 4.3.1, I re-ran the tests, and now the same commit passes.

Time to move on to updating to Robo 3.

greg-1-anderson:
OK, we have a small problem.

In league/container 2.x, a container is a `League\Container\ContainerInterface`

In league/container 3.x, a container is a `Psr\Container\ContainerInterface`

This isn't a problem in DependencyInjection.php, as all of the references are on protected methods, so I can just change those. However, in Drush.php we have:
```
public static function setContainer(ContainerInterface $container)
public static function getContainer(): ?ContainerInterface
```
I **could** just drop these two typehints. Usually, changing a method signature is considered a b/c break, but the Drush class is just a collection of static methods, so no one will be overriding it. If I took the typehints off, any existing module / extension code that used the Drush class should continue to work the same way.

There is the other issue that if someone uses `getContainer()` and then tries to extend Drush's container by adding more things to it, that code will break if it runs with a Robo 3 / League 3 site. No one should be extending Drush's container, though.

Close enough to continue on?

weitzman:
That’s fine yes. FYI I did some of this work in #4798. It makes sense to break it out separately here.

greg-1-anderson:
Yay, we are green here already. I'm going to make a couple minor code style changes, then this will be ready for review.

greg-1-anderson:
This turned out to be easier than I thought; `alterServicesForDrush` and `injectApplicationServices` do not need to vary based on league/container version, and the two methods to add in the Drush services are very regular between the 2.x and 3.x version.

weitzman (approved):
LGTM. Thanks.

PR #4821, opened 2021-09-05
Add support for providing command/hook information via php8 Attributes.

weitzman:
A start on bringing the new [annotated-command feature](https://github.com/consolidation/annotated-command/pull/221/files) into Drush. This replaces #4798 .

1. Depends on https://github.com/consolidation/annotated-command/pull/223. Will need a reroll when that merges.
1. This opens us up for more niceness. For example, see how the Bootstrap.php attribute class adds completion for the allowable bootstrap levels. This can be made more native in PHP 8.1 with [Enum](https://stitcher.io/blog/php-enums) variable type. 
1. The 8 attribute classes in AC (e.g. Usage) are extended in Drush without changes just so there is one namespace (i.e `Drush/Attributes as CLI`) This seems like a DX nicety. Alternative valid authoring approaches are [just use classes from either project directly](https://github.com/drush-ops/drush/blob/274e6ba20381f02488d11798bf658008e6dbe463/examples/Commands/XkcdCommands.php#L45-L54) or [authors pick from AC or DR aliases](https://github.com/drush-ops/drush/blob/274e6ba20381f02488d11798bf658008e6dbe463/tests/integration/resources/Commands/ExampleAttributesCommands.php#L65-L77).

/cc @beejeebus @greg-1-anderson


greg-1-anderson (approved):
LGTM, but might want to wait for a stable annotated-commands release before merging.

Issue #4820, opened 2021-09-05
Provide a way to alter the list of migration IDs in migrate commands.


benjifisher:
**Is your feature request related to a problem? Please describe.**

In the `migrate_tools` module, we are adding the `--group` option to the `migrate:status`, `migrate:import`, and `migrate:rollback` commands. It is an alternative way to select which migrations for the command.

**Describe the solution you'd like**

Add an event hook to these commands so that other modules can alter the list of migrations.

**Describe alternatives you've considered**

The current implementation is to override the commands. Fall back to the core implementation when the `--group` option is not provided.

**Additional context**

See https://www.drupal.org/project/migrate_tools/issues/3213947.

benjifisher:
A quick look at the code in `MigrateRunnerCommands` shows that all three commands call the `getMigrationList()` method.

If I were extending the `MigrateRunnerCommands` class, then I would overwrite that method and add a `group` parameter. Very simple.

As it is, I think it would be very easy to add a custom event to let other code modify the list after calling `getMigrationList()`. This would require almost no refactoring. (Just move `$list = $this->getMigrationList(...)` out of `if(...)`.)

I wonder if it makes more sense to do a little more refactoring in order to avoid a custom event. For example, the first few lines of `migrate:import` and `migrate:rollback` are identical:

```php
        $tags = $options['tag'];
        $all = $options['all'];

        if (!$all && !$migrationIds && !$tags) {
            throw new \Exception(dt('You must specify --all, --tag or one or more migration names separated by commas'));
        }

        if (!$list = $this->getMigrationList($migrationIds, $options['tag'])) {
            throw new \Exception(dt('No migrations found.'));
        }
```

If we move those lines to an `init` hook in Drush, then `migrate_tools` can add `pre-init` and/or `post-init` hooks to insert code before and after those lines.

Whether we use a custom event or a standard hook, `migrate_tools` will want to implement `pre-init` so that the lines above do not throw an exception when the `group` option is provided.

I appreciate any advice on which approach to take.

PR #4826, opened 2021-09-09
Fixes 4820: add init hooks for migrate:status, migrate:import, migrate:rollback

benjifisher:
Add `init` hooks for the `migrate:status`, `migrate:import`, and `migrate:rollback` commands.

Call `$this->getMigrationList()` in these hooks. Save the result as the new `migrationList` option.

Also move some validation to the `init` functions.

For `migrate:status`, also process the `fields` option.

### DIFF
--- a/src/Drupal/Commands/core/MigrateRunnerCommands.php
+++ b/src/Drupal/Commands/core/MigrateRunnerCommands.php
@@ -495,6 +495,17 @@ class MigrateRunnerCommands extends DrushCommands
     }
 
     /**
+     * Process options for rollback().
+     *
+     * @hook init migrate:rollback
+     * @option migrationList The array of migrations to process, used internally.
+     */
+    public function initRollback(InputInterface $input, AnnotationData $annotationData)
+    {
+        $this->initImport($input, $annotationData);
+    }
+
+    /**
      * Rollback one or more migrations.
      *
      * @command migrate:rollback
@@ -530,16 +541,7 @@ class MigrateRunnerCommands extends DrushCommands
      */
     public function rollback(?string $migrationIds = null, array $options = ['all' => false, 'tag' => self::REQ, 'feedback' => self::REQ, 'idlist' => self::REQ, 'progress' => true]): void
     {
-        $tags = $options['tag'];
-        $all = $options['all'];
-
-        if (!$all && !$migrationIds && !$tags) {
-            throw new \Exception(dt('You must specify --all, --tag, or one or more migration names separated by commas'));
-        }
-
-        if (!$list = $this->getMigrationList($migrationIds, $options['tag'])) {
-            $this->logger()->error(dt('No migrations found.'));
-        }
+        $list = $options['migrationList'];
 
         $executableOptions = array_intersect_key(
             $options,
--- a/src/Drupal/Commands/core/MigrateRunnerCommands.php
+++ b/src/Drupal/Commands/core/MigrateRunnerCommands.php
@@ -2,6 +2,7 @@
 
 namespace Drush\Drupal\Commands\core;
 
+use Consolidation\AnnotatedCommand\AnnotationData;
 use Consolidation\AnnotatedCommand\CommandData;
 use Consolidation\AnnotatedCommand\CommandError;
 use Consolidation\OutputFormatters\StructuredData\RowsOfFields;
@@ -18,6 +19,7 @@ use Drush\Drupal\Migrate\MigrateExecutable;
 use Drush\Drupal\Migrate\MigrateMessage;
 use Drush\Drupal\Migrate\MigrateUtils;
 use Drush\Utils\StringUtils;
+use Symfony\Component\Console\Input\InputInterface;
 use Webmozart\PathUtil\Path;
 
 /**
@@ -66,6 +68,37 @@ class MigrateRunnerCommands extends DrushCommands
         parent::__construct();
         $this->dateFormatter = $dateFormatter;
         $this->keyValue = $keyValueFactory->get('migrate_last_imported');
+    }
+
+    /**
+     * Process options for status().
+     *
+     * @hook init migrate:status
+     * @option migrationList The array of migrations to process, used internally.
+     */
+    public function initStatus(InputInterface $input, AnnotationData $annotationData)
+    {
+        $options = $input->getOptions();
+        $migrationIds = $input->getArgument('migrationIds');
+
+        // The --names-only option takes precedence over --fields.
+        if ($options['names-only']) {
+            if ($options['field'] && $options['field'] !== 'id') {
+                throw new \Exception("Cannot use --names-only with --field={$options['field']}.");
+            }
+            $deprecationMessage = 'The --names-only option is deprecated in Drush 10.5.1 and is removed from Drush 11.0.0. Use --field=id instead.';
+            $this->logger()->warning($deprecationMessage);
+            @trigger_error($deprecationMessage, E_USER_DEPRECATED);
+            $fields = ['id'];
+        } elseif ($options['field']) {
+            $fields = [$options['field']];
+        } elseif ($options['fields']) {
+            $fields = StringUtils::csvToArray($options['fields']);
+        }
+
+        $input->setOption('fields', $fields);
+        $list = $this->getMigrationList($migrationIds, $options['tag']);
+        $input->setOption('migrationList', $list);
     }
 
     /**
@@ -122,22 +155,8 @@ class MigrateRunnerCommands extends DrushCommands
       'names-only' => false,
     ]): RowsOfFields
     {
-        // The --names-only option takes precedence over --fields.
-        if ($options['names-only']) {
-            if ($options['field'] && $options['field'] !== 'id') {
-                throw new \Exception("Cannot use --names-only with --field={$options['field']}.");
-            }
-            $deprecationMessage = 'The --names-only option is deprecated in Drush 10.5.1 and is removed from Drush 11.0.0. Use --field=id instead.';
-            $this->logger()->warning($deprecationMessage);
-            @trigger_error($deprecationMessage, E_USER_DEPRECATED);
-            $fields = ['id'];
-        } elseif ($options['field']) {
-            $fields = [$options['field']];
-        } elseif ($options['fields']) {
-            $fields = StringUtils::csvToArray($options['fields']);
-        }
-
-        $list = $this->getMigrationList($migrationIds, $options['tag']);
+        $fields = $options['fields'];
+        $list = $options['migrationList'];
 
         $table = [];
         // Take it one tag at a time, listing the migrations within each tag.
--- a/src/Drupal/Commands/core/MigrateRunnerCommands.php
+++ b/src/Drupal/Commands/core/MigrateRunnerCommands.php
@@ -338,6 +338,32 @@ class MigrateRunnerCommands extends DrushCommands
     }
 
     /**
+     * Process options for import().
+     *
+     * @hook init migrate:import
+     * @option migrationList The array of migrations to process, used internally.
+     */
+    public function initImport(InputInterface $input, AnnotationData $annotationData)
+    {
+        $options = $input->getOptions();
+        $migrationIds = $input->getArgument('migrationIds');
+
+        $tags = $options['tag'];
+        $all = $options['all'];
+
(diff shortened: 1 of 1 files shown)


benjifisher:
I marked this PR as a draft for a few reasons.

* The commits so far are meant to be easy to review, moving code from primary command functions to `init` hooks. We can simplify the code a bit in later commits.
* I am not sure that `init` is the right hook to use.
* I have not yet tested that this achieves the goal of #4820. My hope is that the `migrate_tools` module can implement `pre-init` and `post-init` hooks in order to modify the list of migrations.
* Is there a better way to save the list of migration plugins between the `init` hook and the main command? I added the `migrationList` option, noting that it is for internal use.

benjifisher:
I tried setting `$userData` as an option for the `migrate:import` and `migrate:rollback` commands. That does not work. I am pretty sure that the problem is that `--no-progress` has not been converted to `--progress=false` when the `init` hook runs. (I hacked the test, replacing `'no-progress' => null` with `--progress=false`, and it passed.)

weitzman:
Thanks for the contribution. This change is unlike any other command and thus harder to read and maintain. So my first question is why is this needed. Specifically, why do we need --group in addition to --tag?

mdlutz24:
Off the top of my head, when you use migrate_upgrade to generate migration configs from the core migrations, they become separated out from the core migrations by group, i.e. all the core migrations end up in the default group, and the newly generated migrations that you are then going to tweak end up in the 'upgrage_drupal_7' group, and you use the --group option to run just those. 

Which isn't to say that can't be changed, but the --group option is pretty important for the current customized d7 migration workflow that people are currently used to.

weitzman:
Thanks for the reply. We need a stronger justification for this pr. Tags sound sufficient, and migrate_upgrade could  
use them.

benjifisher:
> This change is unlike any other command and thus harder to read and maintain.

I am surprised by that. Isn't the point of all these hooks so that you can separate argument validation from the real business of the command, just like Drupal provides validate and submit handlers for forms? If you tell me I should be using `@hook validate` instead of `@hook init`, I am happy to make that change.

> Specifically, why do we need --group in addition to --tag?

That is a good question. I have been asking around, and I have two more answers, but I think the one already offered by @mdlutz24 is the strongest.

### 1. The `migrate_upgrade` module creates the `upgrade_drupal_7` migration group.

Besides what @mdlutz24 already said, there is interaction with the next point.

> Tags sound sufficient, and migrate_upgrade could use them.

I do not like that solution. The migration group has other uses, such as shared configuration. So the `migrate_upgrade` module would have to add both a group and a tag. In effect, you are suggesting that we make the `migrate_upgrade` module less DRY.

### 2. The old behavior lets you select migrations with specified group and specified tag.

I admit that I do not know anyone who takes advantage of this. On the other hand, there are so many things you never hear about until you break them.

I might have migration groups based on the source of the migration, with shared configuration related to that source. (This much is common and very useful.) For example, I could have groups `atom` and `rss`, or if I am running a university system I might have a group for each campus.

Then I might have tags related to the migration destination, say `blog`, `article`, and `event`.

In the old system, I can use `drush mim --group=atom --tag=blog` to run the migrations in the `atom` group AND having the `blog` tag. If everything is a tag, then `drush mim --tag=atom,blog` runs the migrations with the `atom` tag OR the `blog` tag (inclusive OR).

If I no longer have the `--group` option, then I have to add tags like `atom-blog` to every migration. This is less flexible. In a more complex situation, it might be unmanageable.

Again, the `migrate_upgrade` module provides a migration group. Specifying `--group=upgrade_drupal_7 --tag=Content` runs just the migrations created by this module that have the `Content` tag.

### 3. Backwards compatibility

The is  the obvious answer to the question. If we are going to remove the `--group` option, then we would like a chance to deprecate it first.

Not every user has a collection of YAML files for their migrations and a collection of bash scripts to run their drush commands.

Many migrations are provided by other modules, such as `migrate_upgrade`, `commerce_migrate`, and `media_migrate`. Adding tags to a migration created by a deriver in a contrib module is a lot harder than adding tags to a YAML file.

Scripts are often embedded in CI systems. It is not always easy to update them.

benjifisher:
In the mean time, I have had a chance to test this PR along with a custom branch of the `migrate_tools` module. It works, so I am no longer marking this PR as a draft.

benjifisher:
@weitzman: I am still surprised by

> This change is unlike any other command ...

Is my proposed snippet,

```php
    /**
     * Process options for import().
     *
     * @hook init migrate:import
     * @option migrationList The array of migrations to process, used internally.
     */
    public function initImport(InputInterface $input, AnnotationData $annotationData)
    {
        $options = $input->getOptions();
        $migrationIds = $input->getArgument('migrationIds');

        $tags = $options['tag'];
        $all = $options['all'];

        if (!$all && !$migrationIds && !$tags) {
            throw new \Exception(dt('You must specify --all, --tag or one or more migration names separated by commas'));
        }

        $list = $this->getMigrationList($migrationIds, $options['tag']);
        if (!$list) {
            throw new \Exception(dt('No migrations found.'));
        }

        $input->setOption('migrationList', $list);
    }
```

so different from this, from `Drush\Drupal\Commands\core\ImageCommands`?

```php
    /**
     * @hook init image-flush
     */
    public function initFlush(InputInterface $input, AnnotationData $annotationData)
    {
        // Needed for non-interactive calls.
        if ($input->getOption('all')) {
            $styles = array_keys(ImageStyle::loadMultiple());
            $input->setArgument('style_names', implode(",", $styles));
        }
    }
```

Perhaps I was not clear in #4820. The only change I am proposing in Drush core is a little refactoring. Basically, move some argument/option handling out of the main command and into a hook. I may have muddied the waters by describing how `migrate_tools` plans to take advantage of that refactoring.

To be clear, the code in my proposed snippet is not new. I am just moving it out of `import()` into `initImport()`.

weitzman:
I have read the replies and am still not convinced that we need a grouping mechanism beyond tags.

benjifisher:
This MR does not introduce a grouping mechanism beyond tags. It moves some argument/option handling into hook implementations.

weitzman:
Thats true, the PR does not introduce another grouping mechanism. I mean that the justification for the PR is "we need another grouping mechanism because of A B and C". In my opinion A B and C are not compelling enough. I acknowledge that others disagree, and I apologize for the inconvenience this causes to folks still using the contrib migrate modules.

mdlutz24:
So clearly, moving the migrate commands into drush core was a huge mistake. With the core migrate commands and namespace now outside of the migration maintainers control, we now have to rebuild them under a new namespace, and document all new commands and try to explain to people that the core drush commands aren't the officially supported and maintained ones, and that all the command names are changing. It's a bit more than an inconvenience...

mstenta:
> Specifically, why do we need --group in addition to --tag?

If I may, I'd love to share how we are using `--group` for the [farmOS](https://drupal.org/project/farm) distribution [migrations from Drupal 7 to Drupal 9](https://docs.farmos.org/hosting/migration/).

(Notably, we've been forced to [pin our Drush version to 10.3](https://github.com/farmOS/farmOS/commit/999af3157886f5ee7b29b4d63672a7f33d4080f9) until https://www.drupal.org/project/migrate_tools/issues/3213947 is resolved.  @benjifisher's [comment](https://www.drupal.org/project/migrate_tools/issues/3213947#comment-14209673) in that issue is what led me to this PR.)

With farmOS, we have a unique challenge when it comes to migration. farmOS is a Drupal distribution that provides features for farm record keeping. It provides a set of custom entity types (eg: `asset` to represent real-world things that are being managed), and some standard bundles (example `asset` bundles include: `land`, `animal`, `plant`, etc), but we also encourage the farmOS community to build their own contrib modules to add additional bundles (for example, the [farm_bee](https://www.drupal.org/project/farm_bee) module adds a "Bee colony" bundle for hive management).

With the "migration groups" concept provided by `migrate_tools`, we can make it easier for our community module maintainers to upgrade to farmOS 2.x (Drupal 9) by providing shared configuration that covers all the base fields of `asset` entities. So the `farm_bee` migration just needs to declare that it is in the `farm_migrate_asset` group, and it's configuration can be very simple, only concerning itself with migration logic for bundle-specific fields, while farmOS core migration groups cover the base/common fields (the bulk of the logic): https://git.drupalcode.org/project/farm_bee/-/blob/2.x/config/optional/migrate_plus.migration.farm_migrate_asset_bee_colony.yml

Here is the shared config: https://github.com/farmOS/farmOS/blob/2.x/modules/core/migrate/config/install/migrate_plus.migration_group.farm_migrate_asset.yml

Our migration instructions are simple: "run these groups in this order" - and everyone can be pretty confident that it will work smoothly, even with contrib modules (it's important for all `asset` entities to be imported before `log` assets, for instance, because `logs` reference `assets`). We even made our own `farm_migrate:import` Drush command, which simply runs the migration groups in the correct order: https://github.com/farmOS/farmOS/blob/c7d2fce98e2a5e78fef409a6f7eef927d74349e0/modules/core/migrate/src/Commands/FarmMigrateCommands.php#L66

Another step we've taken is create a migrate `process` plugin that extends `migration_lookup` to enable looking up migrated entities across ALL migrations within a group. Imagine the following scenario: a farmer on farmOS 1.x has the `farm_bee` module installed, and has some `observation` logs (a bundle provided by farmOS core) that reference `bee_colony` assets. It isn't possible for farmOS core to provide a migration for `observation` logs that reference `bee_colony` assets without being aware of the `farm_migrate_bee_colony` migration, which is contrib (`migrate_lookup` would need to list ALL migrations that import `asset` bundles). But with a `migration_group_lookup` plugin, we can look up entities that were migrated by ANY migration in the `farm_migrate_asset` group, including the `farm_migrate_asset_bee_colony` migration (and other contrib migrations).

To be fair, tags can also be used to "run these tags in this order", and perhaps a similar `migration_tag_lookup` plugin could be written to accomplish the same result - but migration groups give it a more explicit structure, and the shared configuration capabilities makes it much easier for migrations to inherit common logic.

*AND* we can still use tags at the same time (all our migrations are tagged with "farmOS 1.x" to make it easier to run `drush migrate:status --tag="farmOS 1.x"`, for example), so I think there is value in using both. Just 2 cents from a distro maintainer climbing the mountain of D7->D9 migration... :-)

claudiu-cristea:
Assigning for review

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -68,6 +70,37 @@ public function __construct(DateFormatter $dateFormatter, KeyValueFactoryInterfa
         $this->keyValue = $keyValueFactory->get('migrate_last_imported');
     }
 
+    /**
+     * Process options for status().
```
claudiu-cristea, line 74: Let's use the command `migrate:status` instead of `status()`.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -68,6 +70,37 @@ public function __construct(DateFormatter $dateFormatter, KeyValueFactoryInterfa
...
+     *
+     * @hook init migrate:status
+     * @option migrationList The array of migrations to process, used internally.
```
claudiu-cristea, line 77: I'm not sure why this annotation is here (?)

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -459,6 +494,18 @@ protected function executeMigration(MigrationInterface $migration, string $migra
...
+    /**
+     * Process options for rollback().
+     *
+     * @hook init migrate:rollback
+     * @option migrationList The array of migrations to process, used internally.
+     */
+    public function initRollback(InputInterface $input, AnnotationData $annotationData)
+    {
+        // Do the same as hook init migrate:import.
+        $this->initImport($input, $annotationData);
```
claudiu-cristea, line 506: This is ugly. I wonder if is not possible to have a single method, properly renamed, for both commands. I'm reading from https://github.com/consolidation/annotated-command#hooks

> An annotation (e.g. @foo) will attach the hook to any command that is annotated with the given label.

Could we explore this option?

claudiu-cristea (changes requested):
Overall, I wonder if dispatching a Symfony event inside `MigrateRunnerCommands::getMigrationList()` would not be the preferable way. To me these init hooks look a little bit forced.

PR #4830, opened 2021-09-13
Updating to usable php help examples.

eporama:
The `php:eval` examples in the `@usage` section were still geared to Drupal 7. 

This pull request updates them to be successful with Drupal 9.2.5

I do get a warning about the format description option not being provided to `HelpCLIFormatter.php` but I believe that is just a relic of how I am set up locally.


weitzman:
Thanks for the modernizing, @eporama

PR #4835, opened 2021-09-19
Debug info preserved in stderr

claudiu-cristea:


weitzman:
Merged. @claudiu-cristea I finally figured out whats going on here. Console 3.4+ persists verbosity in an env variable. See https://github.com/symfony/console/blob/3.4/Application.php#L970-L972

PR #4844, opened 2021-10-03
Add EOL dates for Drush 8 and Drush 10

weitzman:
I propose we EOL Drush 8 [in concert with Drupal 8](https://www.drupal.org/psa-2021-2021-06-29), which is Nov 2, 2021. ﻿We might still accept bug fixes after that date, but the diligence level goes down. 

I also propose an EOL for Drush 10 at same time, and asking folks to upgrade to a nearly identical Drush 11.

This gets us to just 1 supported version of Drush :1st_place_medal:

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -50,8 +50,8 @@ Drupal Compatibility
     <td> Drush 8 </td>
     <td> 5.4.5+ </td>
     <!-- Released Nov 2015 -->
-    <td> Nov 2021 </td>
-    <td>✓</td> <td>✓</td> <td><b>✓️</b></td> <td></td>
+    <td> Nov 2022 </td>
+    <td>✅</td> <td>✅</td> <td><b>⚠</b></td> <td></td>
   </tr>
   <tr>
     <td> Drush 7 </td>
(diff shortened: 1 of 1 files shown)


andypost:
Drupal 7 will be maintained one more year, I bet everyone using legacy version for it can still use it

greg-1-anderson:
Yeah, the Drush documentation currently says that the EOL for Drush 8 is TBD, but probably to be in concert with the Drush 7 EOL. I will probably remain on the hook for (minimal) post-EOL Drush 8 support for Pantheon customers on Drupal 7 LTS.

I'm fine with calling Drush 10 EOL with the Drupal 8 EOL, but again, will need to continue providing minimal post-EOL support for Pantheon customers still on D8.

andypost:
I found that having drush8.phar is very handy for code-sprints when you mostly need `drush cr` ref https://www.drupal.org/project/drupal/issues/2894476

In long run this gonna be fixed via https://www.drupal.org/project/ideas/issues/3089277 but everything stuck on https://www.drupal.org/project/drupal/issues/2242947

PR #4839, opened 2021-09-26
Fix PHP 8.1 compatibility for drush 8.x

andypost:
The only notice I get using 8.4.8 with PHP 8.1rc2

```
$ drush st
PHP Deprecated:  Return type of DrushBatchContext::offsetSet($name, $value) should either be compatible with ArrayObject::offsetSet(mixed $key, mixed $value): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in phar:///usr/bin/drush/includes/batch.inc on line 39

Deprecated: Return type of DrushBatchContext::offsetSet($name, $value) should either be compatible with ArrayObject::offsetSet(mixed $key, mixed $value): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in phar:///usr/bin/drush/includes/batch.inc on line 39
 Drupal version         :  9.3.0-dev             
...
Drush version          :  8.4.8  
```

PR #4845, opened 2021-10-04
Fix PHP 8.1 - allow empty yaml

andypost:
One more fix for 10.x after  https://github.com/drush-ops/drush/pull/4840

It also needs https://github.com/consolidation/site-alias/pull/44

PS: 8.x branch is different https://github.com/drush-ops/drush/pull/4839


weitzman:
This is needed on 11.x also, right?

andypost:
Yes, just did not check on 11

weitzman:
Cherry picked to 11.x as well. Thanks.

andypost:
Thank you, I did not check gh-workflows but I bet it needs to add 8.1 to prevent regressions in future

weitzman:
Good idea. See https://github.com/drush-ops/drush/pull/4846

PR #4846, opened 2021-10-04
Always use latest PHP release in CI for tip test

weitzman:
This will move us to PHP 8.1 as soon as it is released, and so on. I think that’s the best option given what I see at https://github.com/wodby/php#docker-images /cc @andypost

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,7 +114,7 @@ jobs:
       - checkout
       - run: cp .docker/zz-php.ini /usr/local/etc/php/conf.d/
       - run: php --version
-      - run: composer config platform.php 8.0.0
+      - run: composer config platform.php --unset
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -105,7 +105,7 @@ jobs:
   test_80_drupal93_highest:
```
andypost, line 105: Core polishing last bits of compatibility so I filed https://github.com/wodby/php/issues/156

andypost: Meantime I do maintain PHP in Alpinelinux and there's "drupal-php" image based on it (edge repo, testing)

https://hub.docker.com/r/skilldlabs/php/tags?page=1&ordering=last_updated&name=81 you could use it meantime to test locally

PR #4849, opened 2021-10-05
Suppress bash process substitution warnings

chx:
As shown in https://bugs.php.net/bug.php?id=53465 PHP has special syntax for [bash process substitution](https://mywiki.wooledge.org/ProcessSubstitution). Now, since we pass the filename to shell utilities to do the actual work, it's okay(-ish) to just suppress all the warnings in PHP.

It also suppresses the natural fopen errors for all nonexisting files but since there's an explicit logger call for that, that's nicer in my opinion.


weitzman:
PRs should now target 11.x branch. Also, how do i test this? what command?

chx:
While it's pointless to do this, the easiest way to do is to replace `drush sqlq --file=somefile.sql` with `drush sqlq--file=<(cat somefile.sql)`. You might want to run with `--debug` too.

weitzman:
Does this look OK to folks? Any unwanted side effects? This area isnt exactly my strength.

greg-1-anderson:
Seems okay to me. My only reservation is whether preg_match is the best way to detect a named file descriptor. Since getting this wrong would only amount to an extra log message (worst case), though, I think it's probably okay to let it go.

Still needs rebasing on 11.x.

PR #4851, opened 2021-10-05
DCG 2

Chi-teck:
That's a draft for preview. Not ready to merge yet. Though the `drush generate` should already work.


Chi-teck:
DCG 2.1 dropped support for PHP 7.3. All class properties got typehints.

PR #4874, opened 2021-10-30
pm:enable should fail by default if hook_requirements() is not met

### LINKED ISSUE #4754: Next milestone: pm:enable should fail by default if hook_requirements() is not met

The solution implemented in #4733 currently attempts to maintain backwards compatibility by allowing invalid module installs to pass by default. Ideally, this shouldn't be the case.

Switch the default behaviour so that the install fails by default, and a conscious decision must be made to bypass the installation requirements.

This change in behaviour should only be applied in the next major release.

**Additional context**
Further discussion is available in the original PR.

weitzman:
Fixes #4754

### DIFF
--- a/src/Drupal/Commands/pm/PmCommands.php
+++ b/src/Drupal/Commands/pm/PmCommands.php
@@ -150,8 +150,8 @@ class PmCommands extends DrushCommands
         }
 
         if ($error) {
-            // Let the user confirm the installation if the requirements are unmet.
-            if (!$this->io()->confirm(dt('The module install requirements failed. Do you wish to continue?'))) {
+            // Allow the user to bypass the install requirements.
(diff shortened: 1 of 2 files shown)


simesy:
Just noting that drush will throw an error, but the caller can bypass the requirement (it's not possible to ignore the requirement in the UI).

I have a module that should never be installed on production (it creates test content) and so I throw an exception in hook_requirements 'install' phase if it's CLI on production.

wombatbuddy:
If we run the following command: drush en -y my_module and the [hook_requirements returns](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_requirements/10
) 'severity' => REQUIREMENT_ERROR then Drush display the error message but still enables the module.
Drush 12.5.2.0
Drupal 10.2.4

PR #4883, opened 2021-10-30
Remove custom log levels.

weitzman:
Any affected contrib or custom commands should use methods on the logger like debug(), info(), etc. 

Before: `$this->logger()->log(LogLevel::DEBUG, "Foo);`
After: `$this->logger()->debug("Foo));`


greg-1-anderson (approved):
These were for b/c; good time to remove them.

PR #4893, opened 2021-11-05
I-3773: Fixed drush_backend_batch_process() not clearing batch_get().…

damienmckenna:
… (#4886)

Co-authored-by: Francesco Placella <plach@183211.no-reply.drupal.org>


damienmckenna:
Thanks for merging the fix!

Issue #4894, opened 2021-11-05
Run deploy with config:import --diff


gitressa:
**Is your feature request related to a problem? Please describe.**
I normally run `drush config:import` with `--diff` to see the details of a configuration import. I just started using the `deploy` command which is great, but it doesn't allow showing the diff.

**Describe the solution you'd like**
Use `config:import --diff` in `deploy` command by default.

**Now**
```
drush updatedb --no-cache-clear
drush cache:rebuild
drush config:import
drush cache:rebuild
drush deploy:hook
```

**With `--diff` parameter**
```
drush updatedb --no-cache-clear
drush cache:rebuild
drush config:import --diff
drush cache:rebuild
drush deploy:hook
```

**Describe alternatives you've considered**
Alternatively, add it as a parameter:

`drush deploy -v --diff -y`

weitzman:
I think the changes that drupal reports today are compact and sufficient. Please run a diff before deploy if you want this info. The deploy command isnt meant to be run interactively, where you review the diff and cancel if you dislike.

gitressa:
> The deploy command isnt meant to be run interactively, where you review the diff and cancel if you dislike.

I agree that normally, just listing config items about to be updated is enough. But sometimes being able to search in the actual changes and seeing the `diff` right there in the terminal is nice, for example when you use `deploy` to create a local instance.

Another place where it would be useful is for post-mortems for CI builds, where something "odd" is happening. Usually in CI, the build process is logged, so it is possible to search and check for abnormalities. Also, it could be a log of configuration changes over time.

My original reason for suggesting adding `--diff` is that I have previously tried and failed at viewing but also stopping an import probably because I was using `-n`. But I just did some digging:

**Works, the import is rejected**
`drush config:import --diff --no`

**Doesn't work, the import goes through**
`drush config:import --diff -n`

... and in the documentation it does outline that `-n` is `--no-interaction`, so to stop an import, `--no` is required:

`drush topic` > select 1:

```
+----------------------+----------------------------------------------------------------------------------------------+
| Name                 | Description                                                                                  |
+----------------------+----------------------------------------------------------------------------------------------+
| --alias-path         | Specifies the list of paths where drush will search for alias files.                         |
| --ansi               | Force ANSI output                                                                            |
| --config             | Specify an additional config file to load. See example.drush.yml                             |
| --include            | A list of additional directory paths to search for Drush commands. Commandfiles should be    |
[...]

PR #4895, opened 2021-11-05
Adds --diff parameter to config:import command

gitressa:
See #4894.


weitzman:
I think the changes that drupal reports today are compact and sufficient. Please run a diff before deploy if you want this info.

PR #4896, opened 2021-11-07
Add conflict against drupal 8 and lower.

weitzman:


greg-1-anderson (approved):
Drupal 8 is EOL, so I am fine with the conflict. Also, global Drush 11 will work with Drupal 9, but fail with dependency conflicts with Drupal 8, so I think this conflict is appropriate.

PR #4903, opened 2021-11-17
Add bundle classes generator

weitzman:
- Generates [bundle classes](https://www.drupal.org/node/3191609) for one or more entity types (e.g. node, media).
- Generates an abstract base class that bundle classes extend
- Generates a `hook_entity_bundle_info_alter` implementation

#### Demo 
- Below is what `node,media` generation looks like, on the CLI
- [This gist](https://gist.github.com/689882efc1bbec7aa9155e53724603b8) shows the generated classes
- Devel screenshot below showing that Drupal is handing out Page class instead of Node

```
$ ./drush gen entity:bundle-classes

 Welcome to entity:bundle-classes generator!
–––––––––––––––––––––––––––

 Module machine name [sut]:
 ➤ drush_empty_module

 Entity type(s). Use comma to delimit. [node]:
  [ 0] block_content
  [ 1] comment
  [ 2] contact_message
  [ 3] file
  [ 4] media
  [ 5] menu_link_content
  [ 6] node
  [ 7] path_alias
  [ 8] shortcut
  [ 9] taxonomy_term
  [10] user
 ➤ 6,4
[warning] Run `drush cache:rebuild` so the bundle classes are recognized.

 The following directories and files have been created or updated:
–––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––
 • /var/www/html/sut/modules/unish/drush_empty_module/drush_empty_module.module
 • /var/www/html/sut/modules/unish/drush_empty_module/src/Bundle/media/Audio.php
 • /var/www/html/sut/modules/unish/drush_empty_module/src/Bundle/media/Document.php
 • /var/www/html/sut/modules/unish/drush_empty_module/src/Bundle/media/Image.php
 • /var/www/html/sut/modules/unish/drush_empty_module/src/Bundle/media/MediaBundleBase.php
 • /var/www/html/sut/modules/unish/drush_empty_module/src/Bundle/media/RemoteVideo.php
 • /var/www/html/sut/modules/unish/drush_empty_module/src/Bundle/media/Video.php
 • /var/www/html/sut/modules/unish/drush_empty_module/src/Bundle/node/Article.php
 • /var/www/html/sut/modules/unish/drush_empty_module/src/Bundle/node/NodeBundleBase.php
 • /var/www/html/sut/modules/unish/drush_empty_module/src/Bundle/node/Page.php
```
image

#### Todo
- [ ] @Chi-teck Any interest in using https://github.com/eddiriarte/console-select in DCG? Would be helpful for selecting multiple entity types, among other things. Maybe helpful for Drush as well.
- [ ] @Chi-teck I'm OK with this moving to DCG if you think thats better.


src/Drupal/Generators/entity/hook_bundle_info.php.twig
```diff
@@ -0,0 +1,10 @@
+/**
+* Implements hook_entity_bundle_info_alter().
+*/
+function drush_empty_module_entity_bundle_info_alter(array &$bundles): void {
+{% for id in entity_type_ids %}
+  {% for bundle in infos[id]|keys %}
+    $bundles['{{ id }}']['{{ bundle }}']['class'] = \Drupal\{{ machine_name }}\Bundle\{{ id }}\{{ bundle|camelize }}::class;
```
tgifford-webfirst, line 7: The default PHPCS lints will want this moved into a use statement. Given that it can also automatically fix the lint, I'd say no change is warranted here.

Chi-teck:
@weitzman I think this command should generate classes for just one entity type at a time. I would let user to specify a bundle.

Chi-teck:
Also I think generating abstract class should be optional. For instance, for custom entities there is no point to have a base bundle class as everything can be implemented in entity class.

src/Drupal/Generators/entity/hook_bundle_info.php.twig
Chi-teck: That's supposed to be updated manually. So far generators can only append or replace files. Importing `use` statement would require parsing PHP code which is a way more complex.

Chi-teck (commented):
> src/Bundle

How about `src/Entity` or `src/Entity/Bundle`? Though, I am not sure which path is better.

Chi-teck:
Aren't bundle classses only applicable to content entity types?

weitzman:
1. Oh right, I will filter to just content entities
1. As for generating just one entity type at a time, how would you handle the hook, which needs info from all entities and bundles. You wouldnt want to remove whats already there. Would you only write that hook if it already exists? Seems a bit weird for the generator to work the first time you ran it but not for subsequent entity types.

weitzman:
I am now only showing content entities (OP is updated), and I added the option to skip base class.

Chi-teck:
> how would you handle the hook, which needs info from all entities and bundles. You wouldnt want to remove whats already there. Would you only write that hook if it already exists? Seems a bit weird for the generator to work the first time you ran it but not for subsequent entity types.

I know that's tricky. The current approach is just generate as a second instance of the hook and let user to merge it manually. That's how it works in `template` generator. I am trying to avoid parsing PHP code to keep the things simple. So far hooks are
 the only place where it's really needed. Note that generating classes for multiple entity types will not eleminate the need to run the generator again later because over time new entity types and bundles may be added on a site.

weitzman:
Double printing the alter hook sounds reasonable to me ... I think I've responded to all your feedback except for allowing for bundle(s) to be specified. Should be simple.

Chi-teck:
> I'm OK with this moving to DCG if you think thats better.
I've just pushed it to DCG 2.x.
https://github.com/Chi-teck/drupal-code-generator/blob/master/src/Command/EntityBundleClass.php
<details><summary>Output</summary>
<p>

```
$ /var/www/d9/docroot/vendor/bin/dcg entity-bundle-class --dry-run

 Welcome to entity-bundle-class generator!
–––––––––––––––––––––––––––––––––––––––––––

 Module machine name [bar]:
 ➤ 

 Entity type:
  [ 1] Custom block
  [ 2] Comment
  [ 3] Contact message
  [ 4] File
  [ 5] Custom menu link
  [ 6] Content
  [ 7] URL alias
  [ 8] Shortcut link
  [ 9] Taxonomy term
  [10] User
 ➤ 6

 Bundle:
  [1] Article
  [2] Basic page
 ➤ 1

 Class [ArticleBundle]:
 ➤  

 Would you like to generate a base class? [No]:
 ➤ Yes

 Base class [NodeBundle]:
 ➤ 

 src/Entity/Bundle/NodeBundle.php
––––––––––––––––––––––––––––––––––
<?php

namespace Drupal\bar\Entity\Bundle;

use Drupal\node\Entity\Node;

/**
 * A bundle class for node entities.
 */
class NodeBundle extends Node {

}


 src/Entity/Bundle/ArticleBundle.php
–––––––––––––––––––––––––––––––––––––
<?php

namespace Drupal\bar\Entity\Bundle;

/**
 * A bundle class for node entities.
 */
class ArticleBundle extends NodeBundle {

}


 bar.module
––––––––––––
<?php

/**
 * @file
 * Primary module hooks for Bar module.
 */

/**
 * Implements hook_entity_bundle_info_alter().
 */
function bar_entity_bundle_info_alter(array &$bundles): void {
  $bundles['node']['article']['class'] = \Drupal\bar\Entity\Bundle\ArticleBundle::class;
}

```
</p>
</details>

Feel free to submit bug reports and improvements. Thank you!

src/Drupal/Generators/entity/bundle_class.php.twig
```diff
@@ -0,0 +1,13 @@
...
+namespace Drupal\{{ machine_name }}\Entity\Bundle\{{ entity_type_id }};
+{% if not use_base_class %}
+
+use {{ parent_class }};
+{% endif %}
+
+/**
+ * A bundle class for {{ entity_type_id }} entities.
+ */
+class {{ bundle_class }} extends {{ parent_class|split('\\')|last }} {
```
claudiu-cristea, line 11: IMHO, we should encourage bundle classes to extend bundle interfaces. A `Page` bundle class could implement ` `PageInterface`. This is very useful, as a bundle class can implement multiple interfaces:

```php

interface PageInterface {
  public function getContent();
}

interface ArchivableInterface {
  public function getArchivedDate();
}

class Page implements PageInterface, ArchivableInterface {
}
```

also this allows to do some more semantically checks:

```php
if ($node instanceof PageInterface) {...}
```

PR #4908, opened 2021-11-25
Stop prepending the Cancel option to choice lists

### LINKED ISSUE #4907: Stop prepending the Cancel option to choice lists
## Problem
A _Cancel_ option is prepended to every choice list, why is this necessary if you can just Ctrl+C to exit like with every other command?

CleanShot 2021-11-25 at 12 44 58@2x

@weitzman had this to say in https://github.com/drush-ops/drush/pull/4853#issuecomment-978041138:

> That cancel choice has been there so long I don't recall its justification. I think you are the first person to actually call it out. I bet others have silently grumbled. Lets discuss in a different PR.

## Solution
Stop adding the extra option in `DrushStyle::choice`.

https://github.com/drush-ops/drush/blob/b90445c2cdd54000a82958274ba0125f5c93f8dc/src/Style/DrushStyle.php#L34-L44

We cannot completely remove the method since the _Cancel_ option isn't the only change that override does: it also displays the choices as an indexed list instead of a keyed list:

CleanShot 2021-11-25 at 13 03 51@2x

CleanShot 2021-11-25 at 13 04 14@2x

This might be a BC break: if you pass an indexed array as options and you hardcoded the default value to a certain index, you will now have to subtract 1 from that index.

DieterHolvoet:
Fixes #4907.


greg-1-anderson:
Cancel is explicit. If you press [RETURN], the default option will be cancel. Existing Drush users are used to seeing the cancel option. I don't see the benefit of removing this.

weitzman:
I do see some benefit. It gets tedious on to see these on a long wizard like field:create. Also, drush generate command does not do this. Still mulling this.

greg-1-anderson:
Sure, go ahead if you prefer. Maybe reserve 0 for cancel but don't print it as a compromise? Still slightly prefer keeping cancel in place.

PR #4910, opened 2021-12-02
Add memory-efficient option to site:install

bartlangelaan:


weitzman:
Sorry, this is too hard to maintain in core drush (high complexity).

PR #4915, opened 2021-12-07
Add min drush version  to the docs pages where new stuff is added.

weitzman:
For example, see the 10.3+ below:

image


leymannx:
@weitzman – There's somebody wondering why the docker file got removed: https://drupal.stackexchange.com/q/308648/15055. Can you clarify?

Issue #4916, opened 2021-12-09
Incompatible declaration of Psysh\Shell::hasCommand


felmab:
I had to upgrade Drush from version 9 to version 10 (10.6.1) in order to upgrade a site from Drupal 8 to Drupal 9. I noticed this error when trying to launch the `core:cli` command:
> PHP Fatal error:  Declaration of Drush\Psysh\Shell::hasCommand($input) must be compatible with Psy\Shell::hasCommand(string $input): bool in /var/www/html/vendor/drush/drush/src/Psysh/Shell.php on line 37

On another site using the same version of Drush, there is no such issue. The difference appears to be that the former uses psy/psysh:0.11.0 while the latter uses psy/psysh:0.10.9 (with different method signatures indeed). Provided I did nothing wrong while updating, there might be a constraint missing somewhere in order to avoid newer versions of psy/psysh.

weitzman:
Looks like it. We can just require the newer psysh in Drush 11. PRs welcome.

felmab:
Tests fail in #4918 but errors look unrelated.

MatthieuScarset:
Same issue since today for me. 

This is a blocker because it prevents developer from using `drush php` comand.

felmab:
@MatthieuScarset while waiting for a new release you can manually adapt the signature of `Shell::hasCommand` like [this](https://github.com/drush-ops/drush/pull/4918/files) (or maybe force a dev version).

alexpott:
It might be handy to add a =< 0.10 constraint to Drush 10 so people can run composer update and composer can do the correct thing.

weitzman:
Sure I'd merge that PR if anyone gets inspired.

felmab:
> It might be handy to add a =< 0.10 constraint to Drush 10 so people can run composer update and composer can do the correct thing.

I think that's essentially what was done in #4917, only that it hasn't been released yet.

weitzman:
Thanks for clarifying. Released at https://github.com/drush-ops/drush/releases/tag/10.6.2

PR #4917, opened 2021-12-09
psysh 0.11 does not work with Drush 10

felmab:
See #4916


bobthecow:
That `~` sure is a pain 😕

I would recommend [`^` operator based constraints](https://getcomposer.org/doc/articles/versions.md#caret-version-range-) in general, but especially here, as they respect SemVer better. Caret knows what to do with 0.x versions, but tilde does not. In this case, something like

```
^0.6 || ^0.7 || ^0.8 || ^0.9 || ^0.10
```

… And update it with each 0.x version bump once it's been determined not to cause any backwards compatibility issues.

PR #4918, opened 2021-12-09
Make Drush compatible with newer versions of psysh

felmab:
See #4916.


weitzman:
lgtm. could you also run `composer update psy/psysh` so our .lock file gets updated? or give me perms to push to the pr. thx.

weitzman:
I was able to merge locally and push. thx.

PR #4919, opened 2021-12-09
psysh 0.11 does not work with Drush 8

Kazanir:
See #4916 :)


greg-1-anderson:
Do we need the older versions of psysh for older versions of Symfony?

Tests are now green on 8.x branch; please rebase.

Issue #4887, opened 2021-11-03
[warning] Undefined array key 1 Application.php:199


arvodia:
**Describe the bug**
in : drush/drush/src/Application.php
````
        // @todo Find a cleaner solution.
        if (Drush::config()->get('runtime.argv')[1] !== 'help') {
            $this->checkObsolete($command);
        }
````
to 
````
        if ((Drush::config()->get('runtime.argv')[1]??null) !== 'help') {
            $this->checkObsolete($command);
        }
````

PR #4921, opened 2021-12-11
Check for runtime.argv array size before use (#4887)

diegoe:
Otherwise we run into warnings when `drush` is run with no commands:

```shell
$ drush
 [warning] Undefined array key 1 Application.php:199
(...)
```

Closes: https://github.com/drush-ops/drush/issues/4887

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -195,7 +195,8 @@ class Application extends SymfonyApplication implements LoggerAwareInterface, Co
         $command = $this->bootstrapAndFind($name);
         // Avoid exception when help is being built by https://github.com/bamarni/symfony-console-autocomplete.
         // @todo Find a cleaner solution.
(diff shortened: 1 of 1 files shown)


diegoe:
FWIW this should be immediately backport-able to 10.x (I originally noticed this on `10.6.1`)

weitzman:
I didnt get that warning but your code looks good. Merged into 11.x and 10.x. Thanks.

diegoe:
> I didnt get that warning but your code looks good. Merged into 11.x and 10.x. Thanks.

I believe this is because of the default `error_reporting` level you have set. I noticed this warning on environments where I have `E_ALL` -- I think you'd need at least `E_WARNING` to see undefined key errors.

Thanks for the merge 🚀

PR #4926, opened 2021-12-12
Add field:delete command

DieterHolvoet:
This PR adds a command to delete fields, similar to the [field-delete command that was present in Drush 8](https://drushcommands.com/drush-8x/field/field-delete/). It has already been used in multiple projects as part of the [wmscaffold module](https://github.com/wieni/wmscaffold) and is in my opinion stable enough to be considered to be merged in Drush.

## Remarks
- `AskBundleTrait` and `ValidateEntityTypeTrait` are also included in https://github.com/drush-ops/drush/pull/4928, https://github.com/drush-ops/drush/pull/4929 and https://github.com/drush-ops/drush/pull/4930. 

## Test scenarios
- `./vendor/bin/drush fd` => _Not enough arguments (missing: "entityType")._
- `./vendor/bin/drush fd node` => Asks for bundle
- `./vendor/bin/drush fd node --no-interaction` => _The bundle argument is required._
- `./vendor/bin/drush fd node article` => Asks for field name
- `./vendor/bin/drush fd node article --no-interaction` => _The field-name option is required._
- `./vendor/bin/drush fd node article --no-interaction --field-name=field_test` => Deletes _field_test_. If that field does not exist: _Field with name 'field_test' does not exist on bundle 'article'._

## TODO
- [ ] Add tests

### DIFF
--- a/src/Drupal/Commands/core/FieldCreateCommands.php
+++ b/src/Drupal/Commands/core/FieldCreateCommands.php
@@ -26,7 +26,9 @@ use Symfony\Component\Console\Question\ChoiceQuestion;
 
 class FieldCreateCommands extends DrushCommands implements CustomEventAwareInterface
 {
+    use AskBundleTrait;
     use CustomEventAwareTrait;
+    use ValidateEntityTypeTrait;
 
     /** @var FieldTypePluginManagerInterface */
     protected $fieldTypePluginManager;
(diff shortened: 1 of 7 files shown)


DieterHolvoet:
I'm not sure whether we should place the traits in the same namespace as the command, or whether we just shouldn't use any traits. Opinions are welcome.

src/Drupal/Commands/core/FieldCreateCommands.php
```diff
@@ -26,7 +26,9 @@
 
 class FieldCreateCommands extends DrushCommands implements CustomEventAwareInterface
 {
+    use AskBundleTrait;
     use CustomEventAwareTrait;
```
weitzman, line 30: Looks like we don't use a CustomEvent directly.

src/Drupal/Commands/core/FieldDeleteCommands.php
```diff
@@ -0,0 +1,150 @@
...
+     * @param string $bundle
+     *      The machine name of the bundle
+     *
+     * @option field-name
+     *      The machine name of the field
+     *
+     * @option show-machine-names
+     *      Show machine names instead of labels in option lists.
+     *
+     * @usage drush field:delete
```
weitzman, line 47: Looks like we dont yet support omitting entityType. Same for field:info. I think we should. Any chance you can add this @DieterHolvoet? I've just pushed some tests.

src/Drupal/Commands/core/FieldDeleteCommands.php
```diff
@@ -0,0 +1,150 @@
...
+
+            // If there is only one bundle left for this field storage, it will be
+            // deleted too, notify the user about dependencies.
+            if (count($fieldStorage->getBundles()) <= 1) {
+                $fieldStorage->delete();
+            }
+
+            $message = 'The field :field has been deleted from the :type content type.';
+        } else {
+            $message = 'There was a problem removing the :field from the :type content type.';
```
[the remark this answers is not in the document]
DieterHolvoet: I used the exact same error message as shown in the UI: https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/modules/field_ui/src/Form/FieldConfigDeleteForm.php#L104

Isn't it possible that the field storage being locked can have other causes? Because if that's the case, we should keep the error message generic, just like `field_ui` does. Btw, I just tested it and I can delete fields with data in them using `field:delete`.

src/Drupal/Commands/core/FieldCreateCommands.php
DieterHolvoet: We do call the `getCustomEventHandlers` method in `FieldCreateCommands`, right?

src/Drupal/Commands/core/FieldDeleteCommands.php
DieterHolvoet: Good idea, I just added this to `field:delete`, `field:create` and `field:info`.

PR #4937, opened 2021-12-17
Add nodetype:create & vocabulary:create commands

DieterHolvoet:
This PR adds two commands, one to create a node type (bundle) and another to create a vocabulary (taxonomy bundle). It has already been used in multiple projects as part of the [wmscaffold module](https://github.com/wieni/wmscaffold) and is in my opinion stable enough to be considered to be merged in Drush.

### `nodetype:create`
- `./vendor/bin/drush nodetype:create` => Asks for _Human-readable name_.
- `./vendor/bin/drush nodetype:create --no-interaction` => _The label option is required._
- `./vendor/bin/drush nodetype:create --no-interaction --label=Article --machine-name=article` => _Successfully created node type with bundle 'article'_, if bundle already exists _'node_type' entity with ID 'article' already exists._
- `./vendor/bin/drush nodetype:create --no-interaction --label=Article --machine-name=article --menus-available=admin,main` => _Successfully created node type with bundle 'article'_, if menu_ui module is not installed _The "--menus-available" option does not exist._
- `./vendor/bin/drush nodetype:create --no-interaction --label=Article --machine-name=article --default-language=site_default` => _Successfully created node type with bundle 'article'_, if language module is not installed _The "--default-language" option does not exist._

### `vocabulary:create`
- `./vendor/bin/drush vocabulary:create` => Asks for _Human-readable name_.
- `./vendor/bin/drush vocabulary:create --no-interaction` => _The label option is required._
- `./vendor/bin/drush vocabulary:create --no-interaction --label=Tag --machine-name=tag` => _Successfully created vocabulary with bundle 'tag'_, if bundle already exists _'taxonomy_vocabulary' entity with ID 'tag' already exists._
- `./vendor/bin/drush vocabulary:create --no-interaction --label=Tag --machine-name=tag --default-language=site_default` => _Successfully created vocabulary with bundle 'tag'_, if language module is not installed _The "--default-language" option does not exist._

## TODO
- [ ] Add tests

### DIFF
--- a/src/Drupal/Commands/core/BundleMachineNameAskTrait.php
+++ b/src/Drupal/Commands/core/BundleMachineNameAskTrait.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Drush\Drupal\Commands\core;
+
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Symfony\Component\Console\Input\InputInterface;
+
+/**
+ * @property InputInterface $input
+ * @property EntityTypeManagerInterface $entityTypeManager
+ */
+trait BundleMachineNameAskTrait
+{
+    protected function askMachineName(string $entityTypeId): string
+    {
+        $label = $this->input->getOption('label');
+        $suggestion = null;
+        $machineName = null;
+
+        if ($label) {
+            $suggestion = $this->generateMachineName($label);
+        }
+
+        while (!$machineName) {
+            $answer = $this->io()->ask('Machine-readable name', $suggestion);
+
+            if (preg_match('/[^a-z0-9_]+/', $answer)) {
+                $this->logger()->error('The machine-readable name must contain only lowercase letters, numbers, and underscores.');
+                continue;
+            }
+
+            if (strlen($answer) > EntityTypeInterface::BUNDLE_MAX_LENGTH) {
+                $this->logger()->error('Field name must not be longer than :maxLength characters.', [':maxLength' => EntityTypeInterface::BUNDLE_MAX_LENGTH]);
+                continue;
+            }
+
+            if ($this->bundleExists($entityTypeId, $answer)) {
+                $this->logger()->error('A bundle with this name already exists.');
+                continue;
+            }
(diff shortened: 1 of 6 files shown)


weitzman:
The code here is high quality, its just more code than I was expecting. Am considering if it merits inclusion in drush core. Also, I dont think we have a way to add a bundle for a non-node entity type.

DieterHolvoet:
> Also, I dont think we have a way to add a bundle for a non-node entity type.

I have another command for creating taxonomy vocabularies I was going to make a merge request for. It would reuse `BundleMachineNameAskTrait` and possibly also `LanguageHooks`. I personally don't think `NodeTypeCreateCommands` is very complex or high maintenance, most of what it does is collecting some parameters and creating the node type.

DieterHolvoet:
I also added a `vocabulary:create` command to this PR, because they share a lot of code.

weitzman:
I have an alternative idea - not sure how good it is. It would be much less code. We put in a super simple bundle creation command in core. It its arguments are entity type and bundle name and its code is like [this script](https://github.com/drush-ops/drush/blob/11.x/tests/functional/resources/create_unish_article_bundles.php). It then create the bundle and suggests make further customizations via `drush config:edit  node.type.page.yml` (for example). This command would work for all entity types.

DieterHolvoet:
We could do that, it would be better than nothing, but IMO it would be a shame to discard the existing commands in this PR because they have the same functionality as the admin forms. What part of the code are you concerned about? Most of it is just collecting different options, I don't have the feeling that's the kind of code that's really high maintenance.

weitzman:
Yes, most of the code here is asking questions, and validating answers. That code requires maint as well. Keeping Drush as small as possible is one way we have been able to maintain its quality for so long.

DieterHolvoet:
I understand. It's your call, no hard feelings if you don't want to merge. We can still add the command you're suggesting at some point.

weitzman:
Lets go with the [alternative idea](https://github.com/drush-ops/drush/pull/4937#issuecomment-997919925) I mentioned above, whenever someone gets the urge to contribute it.

PR #4947, opened 2021-12-25
Fix #4933. Dont let Console propogate --quiet to subprocesses by default.

### LINKED ISSUE #4933: drush en some_module_name --quiet -y is broken
**Describe the bug**

`drush en some_module_name --quiet -y` is broken because of the --quiet option

**To Reproduce**
Run this for a not installed module:

drush en some_module_name --quiet -y

**Expected behavior**
enable the module without error

**Actual behavior**
the module gets enabled but you get an error

```
In ProcessBase.php line 155:

  Output is empty.

```

**Workaround**
don't use quiet
### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.6.1
| Drupal version? | 9.3
| PHP version     | 7.4
| OS?             | /Linux/

**Additional information**

weitzman:


weitzman:
Console was setting an env variable in the parent SHELL_VERBOSITY=-1 and that was auto-propogating to the child. See https://github.com/symfony/console/blob/3.4/Application.php#L970-L972. Think this is new in Symfony 3.4

PR #4949, opened 2021-12-26
Fix #4373. The command output does not check verbosity correctly.

### LINKED ISSUE #4373: The command output does not check verbosity correctly.
**Describe the bug**
[OutputInterface::isDebug()](https://github.com/symfony/console/blob/v4.4.7/Output/OutputInterface.php#L85) should return `true` only when `-vvv` flag was provided.

**To Reproduce**
Put the following code to some annotated command.
```
var_dump($this->io()->isDebug());
```
Run the command with `-vv` flags (very verbose).

**Expected output**
```
false
```

**Actual output**
```
true
```

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  |10.2.2
| Drupal version? | 9.0.0-dev
| PHP version     | 7.4.3
| OS?             | Linux

**Additional information**
Other verbosity methods (`isVerbose` and `isVeryVerbose`) work correctly.

weitzman:


greg-1-anderson (approved):
Seems good.

PR #4962, opened 2021-12-30
Test on Drupal 9.3, stop testing on Drupal 8 (EOL)

greg-1-anderson:
More test permutations.


greg-1-anderson:
Can't test on PHP 8+ until we upgrade phpunit.

PR #4965, opened 2021-12-31
Remove Drush's cache API. We no longer need to cache annotation parsing.

weitzman:


weitzman:
This is green now, and the benchmarks look good. In fact, we are now faster without the cache for some reason. Its a mystery but I'm not gonna sweat it.

| Branch | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| 11.x | 99.3 ± 4.2 | 92.2 | 112.6 | 1.05 ± 0.05 |
| this PR | 94.7 ± 2.1 | 91.7 | 99.8 | 1.00 |

This results from running [hyperfine](https://github.com/sharkdp/hyperfine).

```
hyperfine --prepare "git checkout 11.x"  "./drush" \  
          --prepare "git checkout remove-cache-api" "./drush" --export-markdown results.md
```

PR #4966, opened 2021-12-31
Try switching to phpunit-polyfills

greg-1-anderson:


greg-1-anderson:
Drush 8 tests passing on phpunit 9 with PHP 8.

PR #4974, opened 2022-01-03
Force failure in all builds 

greg-1-anderson:
Checking accuracy of Circle exit code reporting.


greg-1-anderson:
This is working in Drush 11.

PR #4977, opened 2022-01-03
Add missing .php8 file

andypost:
by default box packaging only *.php files

Ref https://gist.github.com/andypost/1fa0be5c9231bab3e8e124c77598e171


andypost:
The way to validate is

```
$ phar list -f drush.phar |grep php8
  | | | |-phar:///home/andypost/www/core9/modules_/drush/drush.phar/vendor/symfony/polyfill-mbstring/Resources/mb_convert_variables.php8
```

PR #4980, opened 2022-01-04
Inject migration plugin manager

claudiu-cristea:
Includes tightening typing and some cosmetic changes

### DIFF
--- a/src/Drupal/Commands/core/drush.services.yml
+++ b/src/Drupal/Commands/core/drush.services.yml
@@ -52,7 +52,7 @@ services:
         -  { name: drush.command }
   migrate_runner.commands:
     class: Drush\Drupal\Commands\core\MigrateRunnerCommands
-    arguments: ['@date.formatter', '@keyvalue']
+    arguments: ['@date.formatter', '@keyvalue', '@?plugin.manager.migration']
     tags:
       -  { name: drush.command }
   queue.commands:
--- a/src/Drupal/Migrate/MigrateMissingSourceRowsEvent.php
+++ b/src/Drupal/Migrate/MigrateMissingSourceRowsEvent.php
@@ -3,7 +3,7 @@
 namespace Drush\Drupal\Migrate;
 
 use Drupal\migrate\Plugin\MigrationInterface;
-use Symfony\Component\EventDispatcher\Event;
+use Symfony\Contracts\EventDispatcher\Event;
 
 /**
  * Missing source rows event.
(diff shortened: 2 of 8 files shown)


src/Drupal/Commands/core/drush.services.yml
```diff
@@ -52,7 +52,7 @@ services:
         -  { name: drush.command }
   migrate_runner.commands:
     class: Drush\Drupal\Commands\core\MigrateRunnerCommands
-    arguments: ['@date.formatter', '@keyvalue']
+    arguments: ['@date.formatter', '@keyvalue', '@?plugin.manager.migration']
```
weitzman, line 55: I forget what this `@?` syntax means. Please add a comment.

src/Drupal/Migrate/MigrateMissingSourceRowsEvent.php
```diff
@@ -3,7 +3,7 @@
 namespace Drush\Drupal\Migrate;
 
 use Drupal\migrate\Plugin\MigrationInterface;
-use Symfony\Component\EventDispatcher\Event;
+use Symfony\Contracts\EventDispatcher\Event;
```
weitzman, line 6: Is 'contracts' OK for Symfony 4 and 5?

weitzman (approved):

src/Drupal/Commands/core/drush.services.yml
claudiu-cristea: If the service doesn't exists (e.g. the service module is not installed) will pass `null`. I would not add a comment in the YAML as this a very known syntax and we're already using it in Drush, see https://github.com/drush-ops/drush/blob/11.x/src/Drupal/Commands/field/drush.services.yml

weitzman: Ah right, sounds good

src/Drupal/Migrate/MigrateMissingSourceRowsEvent.php
claudiu-cristea: Symfony 5.4 requires `"symfony/event-dispatcher-contracts": "^2|^3"`
Symfony 4.4 requires `"symfony/event-dispatcher-contracts": "^1.1"`

All `symfony/event-dispatcher-contracts` versions are exposing the Event class with the same namespace. See:

* https://github.com/symfony/event-dispatcher-contracts/blob/1.1/Event.php
* https://github.com/symfony/event-dispatcher-contracts/blob/2.3/Event.php
* https://github.com/symfony/event-dispatcher-contracts/blob/3.0/Event.php

Right now Drupal 9 doesn't allow to bump Symfony to 5, so we can't test. Anyway, maybe we should consider to run at least one test with `composer update --prefer-lowest` to check the lowest dependency versions scenario,

PR #4994, opened 2022-01-08
Move from PSR2 to PSR12

weitzman:
- Move to PSR12 from PSR2 for coding standard..
- A few more fixes.


DieterHolvoet (approved):
All good for me!

PR #4995, opened 2022-01-08
A few more rector-ish improvements.

weitzman:


weitzman:
Failing, and not worththe trouble right now.

PR #5000, opened 2022-01-09
Added catching of InstallerException. Fixes #4999.

### LINKED ISSUE #4999: 'drush si' should catch InstallerException
**Describe the bug**

If installation fails because Drupal throws a InstallerException, this shows as Drush crashing.

It would be better UX to catch this and show an error.

**To Reproduce**
What did you do?

**Expected behavior**
What did you expect would happen?

**Actual behavior**
What happened instead?

**Workaround**

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 11.x/10.x/8.x (please be specific, and try latest release)
| Drupal version? | 9.x/8.x/7.x
| PHP version     | 8.x/7.x

**Additional information**

joachim-n:


weitzman:
As it stands the PR will actually change the error code to success when the exception happens. The method needs to return a status code (int).

There are other exceptions that are not caught, all around Drush. And catching them arguably increases our coupling with Drupal. And makes the code less readable IMO. Happy to listen to other opinions on this.

Issue #5013, opened 2022-01-12
Add flag to filter for migration status (complete, incomplete, in progress)


brooke-heaton:
**Is your feature request related to a problem? Please describe.**

I'm running a D7 to D9 migration and am constantly updating the source DB as we approach our 'Go Live' date. This requires me to rerun migrations and occasionally I hit snags and need to rerun them. So I'm then having to look at the VERY LONG and very verbose output of `drush migrate-status` to see what needs updating.  Sometimes a migration 'get stuck' on 'In Progres' so I've been grepping the output of `drush migrate-status` but all of this experience suggests there's a better way.

**Describe the solution you'd like**

Create a `--status` parameter that can be passed to the `migrate-status` command that will look for migrations by status. 
`'in-progress` would show any migrations current in progress
`complete` would look for migrations at Idle that have 0 unprocessed entities
`incomplete` would look for migrations that have >0 unprocessed entities

**Describe alternatives you've considered**
Alternately if there were a better way to grep the output of `migrate-status` for a status output of the table, that might work, however the table only outputs `In Progress` as a status. `Idle` is not that helpful.

weitzman:
grep is a very good approach IMO. 

If you want to be more precise, you can usually use the --filter option on table output but for some reason that option is not present in this command. I'll open a PR for that in a sec. See https://www.drush.org/latest/output-formats-filters/#output-filters

weitzman:
I see now that you want to add a new type of status. I'll leave this open as a feature request

weitzman:
That option is added in https://github.com/drush-ops/drush/pull/5014

PR #5014, opened 2022-01-12
Add --filter option to migrate:status

weitzman:


weitzman:
See https://github.com/drush-ops/drush/issues/5013

PR #5021, opened 2022-01-15
Handle Throwables as well.

weitzman:


weitzman:
Upstream hasnt changed so we dont either.

Issue #4422, opened 2020-05-26
Drush proxy log object


jasonpartyka:
**Is your feature request related to a problem? Please describe.**
When running Drush commands via cron (or any other scheduler), it would be very useful to be able to redirect the logging output. 

**Describe the solution you'd like**
To be able to redirect a the logs for an arbitrary command to the specified location. For example, I may want to collect any output related to running `drush core:cron` to a specific location to log all cron data in a specific location.

**Describe alternatives you've considered**
Unix-level redirects can be used to redirect all output to a specific location. However, this only works for users who have direct shell access, or there is some other mechanism to forward on the logs.

**Additional context**
I had a conversation with `greg.1.anderson` on the d.o slack regarding this, and he suggested that: 

> I think the way to do this would be to make a Drush proxy log object.
That would just be a simple PSR-3 interface that called through to a Drush\Log\Logger instance.

The Drupal module [monolog](https://www.drupal.org/project/monolog) which integrates with the [same-named library ](https://github.com/Seldaek/monolog) is a possible implementation of this interface. This would then allow the client code to redirect the output to wherever desired -- a specific file, api endpoint, etc.

greg-1-anderson:
Drush maintainers are unlikely to work on this, but PRs are welcome.

Adding monolog to Drush core is probably not the right thing to do, but we could accept a PR that added the ability for a Drush extension to plug in a custom logger.

PR #5022, opened 2022-01-15
Enable custom Logger(s).

weitzman:
Refs #4422 and implements the suggestion at https://github.com/consolidation/log/issues/20#issuecomment-1008483330. DrushLoggerManager is a slightly customized version of the Consolidation/Log/LoggerManager.

The changes in Runtime.php illustrate my guess as to when and how a site would swap a custom Logger for Drush's. I think that we could refactor our front controller (i.e. [$runtime->run()](https://github.com/drush-ops/drush/blob/77cb2ef52fcc58c6c9ce248c3655a5b4c9f8c06a/drush.php#L72)) such that a site could replace that controller with one that swaps in own Logger. Is that a promising way to go?

### DIFF
--- a/src/Commands/DrushCommands.php
+++ b/src/Commands/DrushCommands.php
@@ -2,6 +2,7 @@
 
 namespace Drush\Commands;
 
+use Drush\Log\DrushLoggerManager;
 use Drush\Log\Logger;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Consolidation\AnnotatedCommand\CommandData;
@@ -63,7 +64,7 @@ abstract class DrushCommands implements IOAwareInterface, LoggerAwareInterface, 
     /**
      * Returns a logger object.
      */
-    protected function logger(): ?Logger
+    protected function logger(): ?DrushLoggerManager
     {
         return $this->logger;
     }
--- a/src/Log/DrushLoggerManager.php
+++ b/src/Log/DrushLoggerManager.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Drush\Log;
+
+use Consolidation\Log\ConsoleLogLevel;
+use Consolidation\Log\LoggerManager;
+
+class DrushLoggerManager extends LoggerManager implements SuccessInterface
+{
+    public function success(string $message, array $context = array())
+    {
+        $this->log(ConsoleLogLevel::SUCCESS, $message, $context);
+    }
+}
--- a/src/Log/Logger.php
+++ b/src/Log/Logger.php
@@ -20,12 +20,11 @@
 
 namespace Drush\Log;
 
-use Consolidation\Log\ConsoleLogLevel;
 use Drush\Drush;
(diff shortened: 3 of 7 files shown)


greg-1-anderson:
There are a number of options, depending on how much flexibility / complexity we want to bring in.

Robo provides a RelativeNamespaceDiscovery mechanism for finding well-known class names (matching a pattern) via the autoloader tables. An example of this is the [Robo plugin class discovery method](https://github.com/consolidation/robo/blob/5dd9012765aaf0ce37bc9611f8500b13d81445f8/src/Runner.php#L330,L334). One option would be to find specific class names using this discovery method; a site could then hook into Drush behavior by declaring a class that matched the pattern. If we did this, rather than having a plurality of classes discovered this way (over time), I'd recommend we discover a single class which then implements specific interfaces to override behavior. e.g. if we discovered a `DrushController` class it might override an interface that declared a method for configuring the logger. If the controller implements that interface, then Drush could call the logger configuration method, passing in the logger manager.

An alternate plan wold be to allow sites to simply declare a command file or a hook file via the existing annotated command mechanism, and use a `@hook init` to ask for the logger manager directly from the Drush logger manager. Doing that has the advantage of not requiring a new hook / extension mechanism, but has the disadvantage that the logger would not be swapped in until very late in the bootstrap. Finding the DrushController could happen pretty early.

I didn't review the Drush initialization code to figure out where this could happen. The place you have selected already has a reference to all of the command files; this is late enough that we could check each command file for a specific interface, e.g. to configure the logger, and not wait for `@hook init`. I think that finding the DrushController via discovery could happen much earlier. If we want to try to allow this to happen as early as possible, we should probably make this part of Drush 12, and get rid of the secondary autoloader (`$loader = $this->preflight->loadSiteAutoloader();`) and prohibit Drush from operating on any Drupal site other than the one it shares an autoloader / vendor with.

There is also the question of what to do with log messages that happen before the logger is fully configured. We could ignore this problem, and just let early log messages go to the "wrong" logger, or formalize the wrongness by having a simple "early log" instance that just prints log messages. Alternately, we could have the logger manager cache all of the early logs, and dump them all to the custom logger after the configuration step is done. If we did this, we'd also want the shutdown handler to dump all of the early log messages via `print` should Drush exit before the logger is configured.

weitzman:
I went with `@hook init` and its working well except the ConsoleLogger that gets swapped in is logging to stdout instead of stderr. If someone really does that with Drush, our redispatch commands like updatedb will break. Maybe our recommendation should be to add a logger instead of and not replace Drush's?

greg-1-anderson:
Note that Drush by default does not use the Symfony ConsoleLogger directly, but instead uses the logger from consolidation/log, which replaces it. The Symfony logger only uses stderr for log levels that are "errors"; this is a common misunderstanding vis-a-vis the purpose of the stderr stream.

See https://github.com/symfony/console/blob/5.4/Logger/ConsoleLogger.php#L75

PR #5035, opened 2022-01-19
Simplify compatiblity matrix now that Drupal 8 is EOL.

greg-1-anderson:
Since Drupal 8 is EOL and Drupal 8.0 through 8.3 are now very very old, there seems to be little reason to make fine distinctions between Symfony 2 and Symfony 3 in the support matrix. Unless we are concerned about frustrating users trying to reconfigure Drupal 8.2 installations to work with unsupported Drush versions, we might as well clean up the table and make it a bit easier to read.


weitzman:
Good idea

PR #5038, opened 2022-01-20
Dont define entity-updates

weitzman:
Fixes https://www.drupal.org/project/devel_entity_updates/issues/3259523


deakjahn:
Thanks.

PR #5039, opened 2022-01-21
Return values instead of keys in discoverCommandsFromConfiguration.

kporras07:
discoverCommandsFromConfiguration function was returning keys from $commandList instead of values. This caused that the returned values were the path to the command file instead of the class and therefore the classes (and commands) are not loaded.

The fix included in this PR makes the commands to load again.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -351,7 +351,7 @@ class Application extends SymfonyApplication implements LoggerAwareInterface, Co
             }
         }
         $this->loadCommandClasses($commandList);
-        return array_keys($commandList);
+        return array_values($commandList);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


weitzman:
Odd that tests did not cover this. Looks like that line was introduced in #4696 so ping to @claudiu-cristea.

greg-1-anderson:
@kporras07 could you add a simple test to Drush using this configuration technique, to guard against future breakage?

greg-1-anderson:
FYIO, @kporras07 works for Pantheon, and I asked him to take over investigating this bug that I was looking at earlier this morning. #4696 broke various tests, but it wasn't immediately obvious because Drush 10.3 was in the lock file for that project, and I didn't have a "highest" test.

greg-1-anderson (approved):
Tests look really good to me; thanks.

IDK if anyone else wants to see a separate PR with just the tests, for the failing case.

weitzman:
LGTM. Thanks.

PR #5043, opened 2022-01-24
Update endpoint in SecurityUpdateCommands

webflo:
I did some maintenance work on  drupal-composer/drupal-security-advisories and published security advisories for Drupal 9 on 9.x. The same advisories are published on 8.x-v2 for a while, this should not break older Drush releases.


weitzman:
This looks good to me. I will say that I was fine with leaving the 8.x name. Even though Drupal core is at 9.x, the module API is still at 8.x. ANd it will remain there even for Drupal 10. Or at least thats how I think of it.

PR #5048, opened 2022-01-28
Add `archive:dump` command

jabxjab:
Related PR https://github.com/drush-ops/drush/pull/5095


src/Commands/core/ArchiveCommands.php
```diff
@@ -0,0 +1,581 @@
...
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Yaml\Yaml;
+use Traversable;
+use Webmozart\PathUtil\Path;
+
+/**
+ * Class ArchiveCommands.
+ */
+class ArchiveCommands extends DrushCommands implements SiteAliasManagerAwareInterface
```
weitzman, line 29: Lets call the file ArchiveDumpCommands. Its big enough already.

weitzman (commented):
Thanks for adding a test. I'll review more once this is out of draft.

src/Commands/core/ArchiveCommands.php
```diff
@@ -0,0 +1,460 @@
...
+     * Validates the requirements for the site.
+     *
+     * The site must be a Composer-managed one with "web" docroot.
+     *
+     * @throws \Exception
+     */
+    private function validateSite(): void
+    {
+        if (
+            !is_file($this->getProjectPath() . DIRECTORY_SEPARATOR . 'composer.json')
```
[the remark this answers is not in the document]
jabxjab: @weitzman this is outdated. `Webmozart\PathUtil\Path::join()` is used

src/Commands/core/ArchiveCommands.php
```diff
@@ -0,0 +1,460 @@
...
+            || !is_dir($this->getProjectPath() . self::ARCHIVE_DIR_NAME . 'web')
```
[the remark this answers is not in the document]
jabxjab: @weitzman this was reworked to support standard and `web/`-docroot based sites

appveyor.yml
```diff
@@ -48,6 +48,7 @@ install:
   - echo extension=php_pgsql.dll >> php.ini
   - echo extension=php_gd2.dll >> php.ini
   - echo extension=php_fileinfo.dll >> php.ini
+  - echo memory_limit=256M >> php.ini
```
greg-1-anderson, line 51: Should we just use -1 here?

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,591 @@
...
+            [],
+            ['fields' => 'files', 'format' => 'json']
+        );
+        $process->mustRun();
+        $status = $process->getOutputAsJson();
+        if (!isset($status['files'])) {
+            throw new Exception('Failed to get path to Drupal files directory');
+        }
+
+        $drupalFilesPath = Path::join($this->siteAliasManager()->getSelf()->root(), $status['files']);
```
greg-1-anderson, line 367: Use Drupal API directly to get path to files. `\Drupal::service('file_system')->realpath('public://');`

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,591 @@
...
+use Symfony\Component\Yaml\Yaml;
+use Traversable;
+use Webmozart\PathUtil\Path;
+
+/**
+ * Class ArchiveDumpCommands.
+ */
+class ArchiveDumpCommands extends DrushCommands implements SiteAliasManagerAwareInterface
+{
+    use SiteAliasManagerAwareTrait;
```
greg-1-anderson, line 29: Use BootstrapManager instead of SiteAliasManager to get the root. Use a direct Drupal API to get the path to the files.

appveyor.yml
jabxjab: Not sure. The default was 128M. `PharData` ate all memory on compression

src/Commands/core/ArchiveDumpCommands.php
jabxjab: I took this (`core-status`-based) approach from other core commands

src/Commands/core/ArchiveDumpCommands.php
jabxjab: Ok, will do

src/Commands/core/ArchiveDumpCommands.php
greg-1-anderson: `core-status` needs the site alias manager for other things; archive dump does not. Using bootstrap manager instead of alias manager decreases the scope of the command, which is good.

jabxjab: Sure, will use Drupal API directly. I was referring to things like https://github.com/drush-ops/drush/blob/11.x/src/Commands/sql/SqlSyncCommands.php#L117

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,582 @@
...
+     *  The command options.
+     *
+     * @return string
+     *  The full path to the code archive component directory.
+     *
+     * @throws \Exception
+     */
+    private function getCodeComponentPath(array $options): string
+    {
+        $codePath = $this->isWebRootSite()
```
greg-1-anderson, line 294: Once you switch to the bootstrap manager, the method `getComposerRoot()` will return the project root (code path)

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    /**
+     * Returns the list of regular expressions to match Drupal contrib projects paths (modules, themes and profiles).
+     *
+     * @return array
+     *
+     * @throws \Exception
+     */
+    private function getWebDocrootDrupalExcludes(): array
```
greg-1-anderson, line 509: I've been thinking about this for a while, but I wasn't sure what to advise that would be better. There are two problems with this technique:

- Sites without a relocated docroot could still be Composer managed
- Sites might place Composer dependencies in locations other than the standard place

I discovered that `composer info --path --format=json` will return the installation paths for all packages placed by Composer, so we could use that here regardless of docroot value to find the locations to exclude from the archive.

jabxjab: @greg-1-anderson ok, will consider refactoring this

src/Commands/core/ArchiveDumpCommands.php
jabxjab: Updated

src/Commands/core/ArchiveDumpCommands.php
jabxjab: Done

src/Commands/core/ArchiveDumpCommands.php
jabxjab: @greg-1-anderson Update to use bootstrapped Drupal core

tests/functional/ArchiveTest.php
```diff
@@ -0,0 +1,135 @@
...
+        $this->assertStringContainsString(
+            'Found database connection settings',
+            $this->getErrorOutput()
+        );
+    }
+
+    public function testArchiveRestoreCommand(): void
+    {
+        // @todo: remove once archive:restore command has added.
+        if (!class_exists('\Drush\Commands\core\ArchiveRestoreCommands')) {
```
jabxjab, line 109: `archive:restore` command is here - https://github.com/drush-ops/drush/pull/5095

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,615 @@
...
+     * Creates a temporary directory for the archive.
+     *
+     * @throws \Exception
+     */
+    protected function prepareArchiveDir(): void
+    {
+        $this->filesystem = new Filesystem();
+        $this->archiveDir = Path::join(FsUtils::prepareBackupDir(self::ARCHIVES_DIR_NAME), self::ARCHIVE_SUBDIR_NAME);
+        $this->filesystem->mkdir($this->archiveDir);
+        register_shutdown_function([$this, 'cleanUp']);
```
greg-1-anderson, line 156: @weitzman: Drush 8 had a number of convenience methods related to temporary files and directories which were automatically cleaned up in Drush's shutdown handler. These were primarily used by make and pm commands, which are no longer part of Drush. Config import and rsync used to use them in Drush 8, but do not in Drush 11. What are your thoughts on direct use of `register_shutdown_function()` for cleanup in light of the lack of this service in Drush 11? Is this the best practice for commands that want cleanup on termination?

greg-1-anderson:
Drush tests are green again if 11.x branch is merged in, as was done in #5148. The problem with the appveyor build started happening recently; tried to find a fix, but couldn't. Downloading the cacert works locally, the problem seems to be with the supported TLS methods in the appveyor image not being up-to-date enough to talk to the providing server as of a week or two ago.

weitzman:
Fixed in #5148

Issue #5047, opened 2022-01-28
Drush 8.4.10 ClassNotFound when outputting yaml format


reece-oliver:
**Describe the bug**
When outputting in yaml formal you get a ClassNotFound Error when it tries to create a new Dumper

Worked fine in 8.4.5, upgraded to 8.4.10 for Full PHP 8 support and its seems something is wrong with it now. I downgraded and confirmed its working fine in 8.4.5 still

**To Reproduce**
```
drush -y pm-list --format=yaml
WATCHDOG: [ERROR] [php] Error: Class &quot;Drush\Internal\Symfony\Yaml\Dumper&quot; not found in drush_outputformat_yaml-&gt;format() (line 17 of phar:///opt/drush/drush8/commands/core/outputformat/yaml.inc) #0 phar:///opt/drush/drush8/commands/core/outputformat.drush.inc(457): drush_outputformat_yaml-&gt;format()
#1 phar:///opt/drush/drush8/includes/command.inc(583): drush_outputformat-&gt;process()
#2 phar:///opt/drush/drush8/includes/command.inc(453): drush_handle_command_output()
#3 phar:///opt/drush/drush8/includes/command.inc(231): _drush_invoke_hooks()
#4 phar:///opt/drush/drush8/includes/command.inc(199): drush_command()
#5 phar:///opt/drush/drush8/lib/Drush/Boot/BaseBoot.php(67): drush_dispatch()
#6 phar:///opt/drush/drush8/includes/preflight.inc(67): Drush\Boot\BaseBoot-&gt;bootstrap_and_dispatch()
#7 phar:///opt/drush/drush8/includes/startup.inc(465): drush_main()
#8 phar:///opt/drush/drush8/includes/startup.inc(369): drush_run_main()
#9 phar:///opt/drush/drush8/drush(114): drush_startup()
#10 /opt/drush/drush8(14): require(&#039;...&#039;)
```

**Expected behavior**
What did you expect would happen?
Output in yaml format

**Actual behavior**
What happened instead?
Errors with the above message

**Workaround**
Is there another way to do the desired action?
Nope

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.4.10 (please be specific, and try latest release)
| Drupal version? | 9.3.3
| PHP version     | 8.0.14
| OS?             | Linux

**Additional information**

reece-oliver:
[...]

PR #5050, opened 2022-01-28
Don't blacklist internal/symfony

reece-oliver:
Don't blacklist internal/symfony as its used by the yaml dumper

https://github.com/drush-ops/drush/issues/5047


greg-1-anderson:
LGTM; thanks.

Issue #5051, opened 2022-01-29
Migrations requirements are not visible in Drush logs


db-Matroskeen:
**Describe the bug**
If migration requirements are not met, then the following message is visible in Drush logs:
`[debug] Migration 'foo' is skipped as its source plugin has missed requirements: . [0.13 sec, 24.65 MB]`

Requirements details are missing. Let's see why.

```
foreach ($migrations as $migrationId => $migration) {
    try {
        $sourcePlugin = $migration->getSourcePlugin();
        if ($sourcePlugin instanceof RequirementsInterface) {
            $sourcePlugin->checkRequirements();
        }
    } catch (RequirementsException $exception) {
        $this->logger()->debug("Migration '{$migrationId}' is skipped as its source plugin has missed requirements: " . $exception->getRequirementsString());
        unset($migrations[$migrationId]);
    }
}
```

The problem is at `$exception->getRequirementsString()`, which returns an information about requirements array passed into the exception. This array is supposed to provide additional metadata and probably shouldn't be visible for developers. At least, I don't see any value in this information. That's why I opened a drupal.org issue to adjust the message format as well: https://www.drupal.org/project/drupal/issues/3254986

Speaking of Drush commands, if requirements were passed as 2nd argument, the message is the following:
`Migration 'd6_comment' is skipped as its source plugin has missed requirements: source_module: comment.`

Otherwise, the message is:
`Migration 'user_profile_field_instance' is skipped as its source plugin has missed requirements:  .`

**To Reproduce**
1) Install latest Drupal core (9.4.x);
2) Install Migrate Drupal module;
3) Run `drush ms user_profile_field_instance -vvv`

**Expected behavior**
I would expect the following message (with or without the requirements array):
`Migration 'd6_comment' is skipped as its source plugin has missed requirements: The module comment is not enabled in the source site.`

**Actual behavior**
`Migration 'd6_comment' is skipped as its source plugin has missed requirements: source_module: comment.`

PR #5052, opened 2022-01-29
Change the format of missing requirements message for migrate source plugins (#5051).

Matroskeen:


Matroskeen:
I don't see how to add **migrate** labels and connect #5051 issue here, sorry about that.
I suggest to wait until the proposed resolution is accepted by migrate maintainers in https://www.drupal.org/project/drupal/issues/3254986

PR #5053, opened 2022-01-30
Fix usage example in core:route command

Chi-teck:
Currently the example causes the following error.

> The user-entered string 'user/1' must begin with a '/', '?', or '#'.


weitzman:
Thanks. I will deal with "highest" failures in a new PR.

PR #5054, opened 2022-01-30
Fix tests on Drupal 10

weitzman:


greg-1-anderson:
I fixed one problem, but couldn't push it to your branch. You can see the change at https://github.com/drush-ops/drush/commit/e3acd1c6c34f24535099ebc11436ab2c2a74fdd0.

The next error after this one might be in the Annotated Command library; null is being passed to `setHelp()` where a string is required. Don't have time to continue rn.

weitzman:
I tried the same fix you made but I was seeing that $this->getStream() and $this->consoleSectionOutputs are not defined. Needs more research. I'm gonna merge what I have in a few mins.

src/Symfony/BufferedConsoleOutput.php
```diff
@@ -41,4 +43,9 @@ public function setErrorOutput(OutputInterface $error): void
     {
         $this->stderr = $error;
     }
+
+    public function section(): ConsoleSectionOutput
+    {
+        // @todo
```
greg-1-anderson, line 49: You can copy the implementation from Symfony, as I did in https://github.com/drush-ops/drush/commit/e3acd1c6c34f24535099ebc11436ab2c2a74fdd0

weitzman:
PR for setHelp() call - https://github.com/consolidation/annotated-command/pull/261

PR #5060, opened 2022-02-01
Add a default value for the field widget choice in field:create

DieterHolvoet:


DieterHolvoet:
The failing tests seem to be unrelated to this PR, so merging.

PR #5061, opened 2022-02-02
Fix #4588 pm:enable renamed to pm:install (backward compat preserved)

### LINKED ISSUE #4588: pm:enable should be pm:install
**Describe the bug**
In Drupal 8.x and above, there is no longer a way to "enable/disable" modules. Instead we "install/uninstall" them.

However, Drush still has a pm:enable command along with pm:uninstall.

This is confusing (enable is the opposite of uninstall?) and inconsistent with the Drupal UI.

**To Reproduce**
List commands in Drush and you'll see that pm:enable turns a module on, and pm:uninstall turns it off (which also involves adding/removing configuration, database tables, and/or content).

**Expected behavior**
The pm:enable command should be named pm:install for consistency with the Drupal UI and with the pm:uninstall command.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 9.x/8.x
| PHP version     | any
| OS?             | any

**Additional information**
Drupal core issue about making the Drupal UI more consistent in this area:
https://www.drupal.org/project/drupal/issues/2891294
which is part of a meta issue:
https://www.drupal.org/project/drupal/issues/2888657

rajeshreeputra:


src/Drupal/Commands/pm/PmCommands.php
```diff
@@ -65,12 +65,12 @@ public function getExtensionListModule(): ModuleExtensionList
     /**
      * Enable one or more modules.
      *
-     * @command pm:enable
+     * @command pm:install
      * @param $modules A comma delimited list of modules.
-     * @aliases en,pm-enable
+     * @aliases install, pm-install, en, pm-enable, pm:enable
```
weitzman, line 70: lets add `in` as an alias

tests/unit/ArgsRemapperTest.php
```diff
@@ -35,8 +35,8 @@ public function argsProvider()
     {
         return [
             [
-                ['en', 'en'],
-                ['pm:enable', 'en'],
+                ['install', 'install'],
+                ['pm:install', 'install'],
```
weitzman, line 39: Lets test all aliases of pm:install

weitzman (changes requested):

src/Drupal/Commands/pm/PmCommands.php
rajeshreeputra: added in alias for pm:install

src/Preflight/Preflight.php
```diff
@@ -141,7 +141,7 @@ protected function remapCommandAliases(): array
     {
         return [
             'si' => 'site:install',
-            'en' => 'pm:enable',
+            'install' => 'pm:install',
```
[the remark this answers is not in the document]
weitzman: remapCommandAliases is a different thing than adding aliases in the commandfile.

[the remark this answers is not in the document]
rajeshreeputra: added mapping for all aliases for pm:install

rajeshreeputra (commented):
updated the requested changes, please review.

PR #5063, opened 2022-02-03
Allow migrate:import's --delete option to function

### LINKED ISSUE #5062: migrate:import --delete flag not functioning in 11.x
**Describe the bug**
After upgrading to Drush 11.x, `migrate:import`'s `--delete` flag seems to have no effect and missing source rows are never rolled back.

**To Reproduce**
Upgrade from Drush 10.x to 11.x.
Remove some rows from source.
Run `migrate:import` with the `--delete` flag
Missing records are not rolled back.

**Expected behavior**
When running `migrate:import` with the `--delete` flag, I expected missing source rows to be deleted.

**Actual behavior**
Missing source rows were not deleted.  Runs as if the `--delete` flag is not present.

**Workaround**
Downgrade to Drush 10.x.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 11.0.4
| Drupal version? | 9.3.3
| PHP version     | 7.4
| OS?             | Linux

chrisolof:
Dispatch the event name onMissingSourceRows is listening for.

Resolves #5062.

### DIFF
--- a/src/Drupal/Migrate/MigrateExecutable.php
+++ b/src/Drupal/Migrate/MigrateExecutable.php
@@ -259,7 +259,7 @@ class MigrateExecutable extends MigrateExecutableBase
 
         if ($destinationIds) {
             $missingSourceEvent = new MigrateMissingSourceRowsEvent($migration, $destinationIds);
-            $this->getEventDispatcher()->dispatch($missingSourceEvent);
+            $this->getEventDispatcher()->dispatch($missingSourceEvent, MigrateMissingSourceRowsEvent::class);
         }
     }
 


src/Drupal/Migrate/MigrateExecutable.php
```diff
@@ -259,7 +259,7 @@ protected function handleMissingSourceRows(MigrationInterface $migration): void
 
         if ($destinationIds) {
             $missingSourceEvent = new MigrateMissingSourceRowsEvent($migration, $destinationIds);
-            $this->getEventDispatcher()->dispatch($missingSourceEvent);
+            $this->getEventDispatcher()->dispatch($missingSourceEvent, MigrateMissingSourceRowsEvent::class);
```
claudiu-cristea, line 262: @chrisolof, thank you for this catch. I think it should work also without the 2nd parameter. Could you try?

claudiu-cristea: Hm, I think I’m wrong. Will look during the weekend

gitressa: Thanks for looking into this @chrisolof and @claudiu-cristea, it seems like this feature should ideally work.

PR #5070, opened 2022-02-12
Allow Drupal install with contrib database drivers

### LINKED ISSUE #5069: Allow Drupal install with contrib database drivers
**Allow Drupal install with contrib database drivers**
At the moment, it's only possible to install Drupal via Drush with one of the database drivers available in Drupal core.

This is because in SqlBase we have this

>     public static function getInstance($db_spec, $options): ?self
>     {
>         $driver = $db_spec['driver'];
>         $class_name = 'Drush\Sql\Sql' . ucfirst($driver);
>         if (class_exists($class_name)) {
>             $instance = new $class_name($db_spec, $options);
>             // Inject config
>             $instance->setConfig(Drush::config());
>             return $instance;
>         }
>         return null;
>     }
> 

i.e. a class named like the driver must exist in Drush codebase.

As of Drupal 9.4, contrib drivers must be constructed like modules. My suggestion would be to agree on a convention, whereby adding a specific class in the driver's namespace - say, `Drupal\db_module\Driver\Database\db_driver\Drush\Sql`, `getInstance()` would also look if such a class exists and if available instantiate it, the proceed installation with that SQL setup.

mondrake:
Fixes #5069

### DIFF
--- a/src/Sql/SqlBase.php
+++ b/src/Sql/SqlBase.php
@@ -121,7 +121,7 @@ abstract class SqlBase implements ConfigAwareInterface
         }
     }
 
-    public static function getInstance($db_spec, $options): ?static
+    public static function getInstance($db_spec, $options): ?self
     {
         $driver = $db_spec['driver'];
         $class_name = 'Drush\Sql\Sql' . ucfirst($driver);


mondrake:
Thank you @weitzman

PR #5078, opened 2022-02-18
Move enlightn/security-checker to "suggest"

greg-1-anderson:
We can see how this works vis-a-vis tests. Not a great solution, but might be better than some of the alternatives.


greg-1-anderson:
Looks like the one failing test is caused by updating dependencies. The test passes locally with the composer.lock from the 11.x branch, but fails after running `composer update` on the 11.x branch.

greg-1-anderson:
Failing test is caused by:
```
  - Upgrading dflydev/dot-access-data (v1.1.0 => v3.0.1): Extracting archive
  - Upgrading grasmash/expander (1.0.0 => 2.0.0): Extracting archive
```
This combination breaks expansion of `${env.FIXTURES_DIR}` in configuration paths.

greg-1-anderson:
Failing test fixed by https://github.com/grasmash/expander/pull/13

greg-1-anderson:
Most tests green with the update to grasmash/expander 2.0.1.

Now, it looks like the latest Drush 10.0.x dev requires symfony/process ^6, and consolidation/site-process supports ^5 at the most. Will need some more releases on consolidation/*, it looks like.

greg-1-anderson:
The Composer dependencies resolve with Drupal 10 now; however, it seems that there have been some changes in the Drupal bootstrap to adjust to Symfony 6; similar changes will be needed in Drush. The error we are getting now is:
```
PHP Fatal error:  Uncaught Error: Call to undefined method Symfony\Component\HttpFoundation\Response::create() in /home/wodby/drush/src/Boot/DrupalBoot8.php:293
```

greg-1-anderson:
Related drupal.org Symfony 6 issue: https://www.drupal.org/project/drupal/issues/3252757#comment-14417393

tests/functional/CoreTest.php
```diff
@@ -10,7 +10,7 @@
  *
  * @group commands
  */
-class CoreTest extends CommandUnishTestCase
+class FunctionalCoreTest extends CommandUnishTestCase
```
andypost, line 13: is the difference in file name and class name required?

greg-1-anderson: I thought I renamed the file, must have missed it. Thanks for pointing that out.

greg-1-anderson:
Now it looks like we're just down to the `module_load_install()` stuff, which is being tracked in another issue. I think we need to combine efforts in order to move forward.

greg-1-anderson:
Getting really close. The integration tests have several errors. CircleCI does not report anything; running locally, I get:
```
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 40960 bytes) in /Users/ga/Code/open-source/drupal/drush11/vendor/symfony/dependency-injection/Compiler/ResolveParameterPlaceHoldersPass.php on line 91
```
Not sure why it's running out of memory; infinite recursion or something? Needs further investigation. Functional tests are all passing.

tests/functional/UpdateDBTest.php
```diff
@@ -70,7 +77,7 @@ public function testFailedUpdate($last_successful_update, $expected_status_repor
         $this->drush('php-script', ['updatedb_script'], ['script-path' => __DIR__ . '/resources']);
 
         // Force re-run of woot_update_8101().
-        $this->drush('php:eval', array('drupal_set_installed_schema_version("woot", ' . $last_successful_update . ')'), $options);
+        $this->drush('php:eval', array('\Drupal::service("update.update_hook_registry")->setInstalledVersion("woot", ' . $last_successful_update . ')'), $options);
```
greg-1-anderson, line 80: Does anyone know what's up with backslashes and Windows PHP? This line works in Circle, but fails in Appveyor.

claudiu-cristea: @greg-1-anderson just remove the leading `\` from `'\Drupal::service(...)`

greg-1-anderson: That might help. I think that the issue might be that backslashes are getting doubled up in a place where that's not necessary. But I'm in favor of avoidance at this stage. 😄

greg-1-anderson:
Hey, much better -- Appvayor is green, and we're down to a dozen integration test failures.

I didn't realize there was going to be so much to fix here.

greg-1-anderson:
Down to just some MigrateRunnerTest failures in the integration tests.

weitzman:
Migrate tests have been failing for a while in d10. Feel free to merge without them. 

See https://github.com/consolidation/site-process/issues/56 for windows bug.

weitzman (approved):

greg-1-anderson:
Great, thanks.

This will be better for Drupal 10 users, but Drupal 9 users lose enlightn/security-checker. Maybe that package will become compatible & we can add it back in before the next release. Or maybe temporarily add it in just for the release tag, and then take it out again so that Drupal 10 users can continue using the dev release of Drush.

paras-malhotra:
@greg-1-anderson @weitzman Symfony 6 is now supported via https://github.com/enlightn/security-checker/pull/27, so can we revert this PR?

greg-1-anderson:
We'll need a new PR to re-add enlightn/security-checker; this PR fixed quite a few other issues at the same time, so we shouldn't do a literal revert.

Re-enabling the security checks will be easy, though.

PR #5082, opened 2022-02-23
Remove psr/log as direct dependency

weitzman:
See https://www.drupal.org/node/3266017. Various consolidation components need psr/log v2 or v3 compat before drush's `test_81_drupal10_highest` tests will pass.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,6 @@
     "enlightn/security-checker": "^1",
     "guzzlehttp/guzzle": "^6.3 || ^7.0",
     "league/container": "^3.4 || ^4",
-    "psr/log": "~1.0",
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
chi-teck/drupal-code-generator is still holding us back to psr/log ^1.  Consolidation components + grasmash/expander will now allow psr/log ^2.  I didn't notice that there was a prs/log ^3, despite the comment "or v3" above. 🤦

greg-1-anderson:
It looks like we'll need a major version bump of at least consolidation/log to get to psr/log ^3. Maybe that's the only component. Drupal is only at ^2 at the moment, so it's not critical to allow ^3, but it would be good to do ahead of Drupal if possible.

PR #5097, opened 2022-03-19
Support command completion (and adjust our DDEV config accordingly)

weitzman:
It turns out that Symfony Console 5.4+ enables completion by default and its working in Drush. This PR configures our DDEV to demonstrate that. You must have DDEV 1.19+ (recently released).

1. When I type `drush core:` and then TAB, I get no results on the first TAB and then good results on a second TAB. Anyone know whats up?
1. I had to disable Drush's `--debug` option because Console ships with that option on its `completion` command. Can anyone find a way to harmonize this without Drush globally removing the option?
1. I've added completion to the `topic` command. Anyone available to help add completion for arguments and option values for other commands? 
image

This PR brings in a small change to annotated-command - https://github.com/consolidation/annotated-command/pull/275. (already merged is initial work at https://github.com/consolidation/annotated-command/pull/274)

Todo
---------
- [x] Drop composer.lock changes
- [x] Decide how to handle the --debug option conflict.
- [x] Make sure tests are passing


greg-1-anderson:
Using this natively on MacOS:
```
brew install bash-completion
eval "$(drush completion bash)"
```
However, it doesn't work like this. Typing `drush topic <TAB>` results in only two newlines. The `topicComplete` method doesn't seem to be called. Does the ddev environment work substantially differently than the MacOS / bsd bash environment? Do you know what command the _sf_drush bash function is calling to get the generated candidates?

weitzman:
Just looped back to this PR and the associated annotated-command PR. Its working well except for the fact that I have to hit tab twice. Dunno why.

I havent tried testing with on OSX since I use ZSH there and that shell is not available until Drupal 10 is on Symfony 6.2. The DDEV environment in this repo is setup for testing.

PR #5100, opened 2022-03-24
Drush 7 for PHP 5

greg-1-anderson:
Add config.platform.php so that "composer install" will produce a compatible vendor.

Pantheon is still installing Drush 7 globally for a few Really Old Sites that are still running PHP 5. Figured a trivial post-EOL Drush release was simpler than adding conditionals to the Pantheon Dockerfile.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,6 +47,11 @@
       "Unish":        "tests/"
     }
   },
+  "config": {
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Tests are on Travis, but there are no code changes here.

PR #5106, opened 2022-03-30
Issue 5105: Process batches generated during drush config:import.

sonfd:
A module may choose to initiate a batch process from hook_install(). The batch process is executed when a module is enabled in UI, via /admin/modules, or via config sync via /admin/config/development/configuration. Additionally, `drush pm:enable mymodule` also executes the batch process. Update the `drush config:import` command to use the same technique as `drush pm:enable` to get and process batches after every sync step.

### DIFF
--- a/src/Drupal/Commands/config/ConfigImportCommands.php
+++ b/src/Drupal/Commands/config/ConfigImportCommands.php
@@ -292,6 +292,12 @@ class ConfigImportCommands extends DrushCommands
                         $context = [];
                         do {
                             $config_importer->doSyncStep($step, $context);
+                            // Manually get and process any batches that may
+                            // have been created, e.g. via hook_install() when a
+                            // module is enabled.
+                            if (batch_get()) {
+                                drush_backend_batch_process();
+                            }
                             if (isset($context['message'])) {
                                 $this->logger()->notice(str_replace('Synchronizing', 'Synchronized', (string)$context['message']));
                             }


weitzman:
Its a little weird that config:import has to care about code in hook_install. What is the backtrace in the core /admin/config/development/configuration request when the batch gets processed?

sonfd:
I'm not sure what the backtrace is. I'm not sure how to generate that for you.

But I'm assuming this has to do with how drush handles batch operations. You manually execute this same code during `drush pm:enable` (`Drush\Drupal\Commands\pm\PmCommands::enable()`), there's a commit specifically for this addition with message: 

> D9 Run batch process during pm-enable when needed. (#3446)

I would ask the same question about pm:enable - why would it need to care about the contents of a module's hook_install()?

weitzman:
Because hook_install is about happens when enabling modules. config:import can result in enabling a module but thats more of a byproduct. Your patch may be right. My question is more about where to add the new code, and less about whether to add it.

FYI A backtrace can be viewed with by putting a call to print_r(debug_backtrace()) in the code where you want to get the info. It sometimes works to add a die() after that in order to actually see the printed info. Otherwise a debugger can stop and show you same info.

chx:
1.  config import has a core UI which runs a batch to import items and I must admit I have no clue how batch API behaves if you add more items to a batch while the UI is processing a batch. 
1. Search API `Index::save` -- which is a config entity -- has an explicit `// Do not try to track items when config is syncing.` in it. If this PR were merged then there would be a chance it could work -- but it might need a core patch to solve the previous point first.

PR #5120, opened 2022-04-08
Run theme post updates when supported by core

alexpott:
Core is likely to add support for post updates in themes. See https://www.drupal.org/project/drupal/issues/3259188

This PR changes drush to support this in a BC compatible way.

I think this should go back to 10.x too depending on which branches https://www.drupal.org/project/drupal/issues/3259188 ends up in.


weitzman:
LGTM. Lets merge once the drupal core issue is merged.

alexpott:
The core issue is merged so this will need to go into all Drush versions that support Drupal 9.4

weitzman:
Thanks ... Drush 10 is EOL per https://www.drush.org/latest/install/#drupal-compatibility. Still, if someone makes a PR for it I will merge.

PR #5137, opened 2022-04-27
Psr log 3

greg-1-anderson:
Using Robo ^4 should allow Drush to use psr/log ^3 in contexts where Drupal 10 allows psr/log ^3.

Once confirmed, we can make a stable release of Robo 4.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,6 +124,8 @@ jobs:
       - run: composer -n config platform.php --unset
       - run: composer -n require --dev drupal/core-recommended:10.0.x-dev --no-update
       - run: composer -n update
+      - run: composer info consolidation/*
+      - run: composer info psr/*
       - run: mkdir -p /tmp/results
       - run: composer -n lint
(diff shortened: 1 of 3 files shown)


weitzman (approved):
LGTM. You prob noticed that our Highest test is still using Log2 - https://app.circleci.com/pipelines/github/drush-ops/drush/3868/workflows/6d7b5744-9924-445e-9891-2cc41b708f10/jobs/23550?invite=true#step-108-9

greg-1-anderson:
Yeah; to test, we need a Drupal 10 tag, branch or patch that allows psr/log ^3. With that, in theory, "highest" will install that version.

greg-1-anderson:
Tested locally with https://github.com/greg-1-anderson/drupal-10-core-patched-psr-log-3.

greg-1-anderson:
Drupal will hold us back to Robo ^3 until psr/log ^3 is allowed, so I will just merge this with the Robo ^4.0.0-alpha1 optional dependency until the Drupal psr/log ^3 patch is merged. Once that is confirmed working, I'll release a stable Robo 4.

PR #5147, opened 2022-05-14
Dont use DRUPAL_ROOT in cache:rebuild

weitzman:


shaal:
@weitzman thank you!
I tested this in DrupalPod and locally, it doesn't solve the problem.
See details here https://github.com/drush-ops/drush/pull/4713#issuecomment-1126861626

weitzman:
Merged. Is useful even if it doesn't solve the mentioned issue.

PR #5148, opened 2022-05-18
Jabxjab archive:dump and archive:restore commands

greg-1-anderson:
Merging latest from 11.x into #5048


weitzman (commented):
It doesn't look like we check that there is a DB check after an archive:dump. Also we will need a full cycle test in the archive:restore tests. Is archive:restore command in a separate PR? Its odd to have its tests here but not its code.

greg-1-anderson:
I just pushed this branch to run the tests. The main archive:dump PR is #5048. Yes, archive:restore is a separate PR. We could combine them if you wanted round-trip tests.

src/Commands/DrushCommands.php
```diff
@@ -18,17 +18,20 @@
...
 use Robo\Common\IO;
+use Robo\LoadAllTasks;
+use Robo\Contract\BuilderAwareInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Consolidation\SiteProcess\ProcessManagerAwareTrait;
 use Consolidation\SiteProcess\ProcessManagerAwareInterface;
 use Webmozart\PathUtil\Path;
 
-abstract class DrushCommands implements IOAwareInterface, LoggerAwareInterface, ConfigAwareInterface, ProcessManagerAwareInterface
+abstract class DrushCommands implements IOAwareInterface, LoggerAwareInterface, ConfigAwareInterface, ProcessManagerAwareInterface, BuilderAwareInterface
```
greg-1-anderson, line 28: I did wonder if adding BuilderAwareInterface here might break existing commands already using Robo tasks, but it does work in PHP 7.4 and 8.1 to declare BuilderAwareInterface and LoadAllTasks both here and in the implementing Drush command class.

tests/functional/ArchiveTest.php
```diff
@@ -0,0 +1,181 @@
...
+            [],
+            array_merge($this->archiveDumpOptions, [
+                'destination' => $this->archivePath,
+                'overwrite' => null,
+            ])
+        );
+        $actualArchivePath = $this->getOutput();
+        $this->assertEquals($this->archivePath, $actualArchivePath);
+
+        // Validate database credentials are present in settings.php file.
```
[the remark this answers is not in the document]
greg-1-anderson: I think the intention here is to test the `overwrite` option.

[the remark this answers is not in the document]
greg-1-anderson: I reworked the test to use `file_put_contents` instead of the first `archive:dump` so that we can still test `--overwrite` and not call dump so many times.

greg-1-anderson:
> It doesn't look like we check that there is a DB check after an archive:dump. 

If this means what I think it does, I believe there is such a test. The last call to `archive:dump` asserts that the command comes back with an error, and that the error message contains the message "Found database connection settings". The other commands pass in an `exclude-code-paths` option that excludes settings.php, and therefore do not fail with the same error.

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,593 @@
...
+
+    /**
+     * Creates a temporary directory for the archive.
+     *
+     * @throws \Exception
+     */
+    protected function prepareArchiveDir(): void
+    {
+        $this->filesystem = new Filesystem();
+        $this->archiveDir = FsUtils::tmpDir(self::ARCHIVES_DIR_NAME);
```
greg-1-anderson, line 154: @weitzman: What do you think about wrapping the call to `register_shutdown_path` in a general-purpose Drush tmpDir method, available to commands that need it? Seems better to me than making each command implement bespoke cleanup.

greg-1-anderson:
Woot! Got Appveyor tests green again. See https://github.com/appveyor/ci/issues/3821 for details.

greg-1-anderson:
Combining with the archive:restore command, and testing that now.

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,593 @@
...
+            return;
+        }
+
+        // Lookup for non-empty $databases value in a site/*/settings.php file.
+        // Remove all PHP "include" and "require" directives from settings.php file before evaluating
+        // the $databases variable.
+        $settingPhpWithoutIncludes = preg_replace('/\n*\s*(include|require).+?;|<\?php/m', '', file_get_contents($file));
+
+        try {
+            eval($settingPhpWithoutIncludes);
```
greg-1-anderson, line 571: Maybe it would be better to just do a preg_match and try to find not-commented-out lines that contain `$databases =`. This might not be perfectly reliable, but we can't predict what `eval` might do on an arbitrary settings.php file.

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,593 @@
...
+                            $this->logger()->info(dt(
+                                'Path excluded (!exclude): !path',
+                                ['!exclude' => $exclude, '!path' => $localFileName]
+                            ));
+
+                            return false;
+                        }
+                    }
+
+                    $this->validateSensitiveData($file, $localFileName);
```
greg-1-anderson, line 461: Are we calling `validateSensitiveData` on every file copied? This is, perhaps, too much work, especially if that function calls `eval`.

src/Commands/core/ArchiveDumpCommands.php
greg-1-anderson: In Drush 8, archive:restore provides the option to restore to the database credentials stored in a settings.php file in an archive dump. This doesn't seem like a good idea to me (storing credentials in the archive), but I suppose there are use cases for it (e.g. backup and restore).

@weitzman: Do you have an opinion on whether or not we allow database credentials in the archive? Maybe we should just remove the seatbelts.

src/Commands/core/ArchiveRestoreCommands.php
```diff
@@ -0,0 +1,489 @@
...
+
+        if ($siteAlias->isLocal()) {
+            $bootstrapManager = Drush::bootstrapManager();
+            $destination = $bootstrapManager->getComposerRoot();
+            $this->rsyncFiles($source, $destination);
+
+            return;
+        }
+
+        $siteStatus = $this->getSiteStatus($siteAlias);
```
greg-1-anderson, line 236: I don't think we should support doing an archive:restore to a remote site.

src/Commands/core/ArchiveRestoreCommands.php
```diff
@@ -0,0 +1,489 @@
...
+     */
+    protected function importCode(string $source, ?string $destinationSite): void
+    {
+        $this->logger()->info('Importing code...');
+
+        if (!is_dir($source)) {
+            throw new Exception(dt('Directory !path not found.', ['!path' => $source]));
+        }
+
+        $siteAlias = $this->getSiteAlias($destinationSite);
```
greg-1-anderson, line 226: If the cwd is not within a Drupal site (e.g. cwd is inside an empty directory), then we should create a new site and extract the files there.

src/Commands/core/ArchiveRestoreCommands.php
```diff
@@ -0,0 +1,489 @@
...
+     */
+    protected function importDatabase(string $databaseDumpPath, array $options): void
+    {
+        $this->logger()->info('Importing database...');
+
+        if (!is_file($databaseDumpPath)) {
+            throw new Exception(dt('Database dump file !path not found.', ['!path' => $databaseDumpPath]));
+        }
+
+        $sql = SqlBase::create($options);
```
greg-1-anderson, line 441: We might not be able to rely on database credentials being available, e.g. the archive:restore destination might be an empty directory, and the archive might not have database credentials.

We should support the Drush 8 archive:restore options related to the database credentials:
```
 --db-prefix                               An optional table prefix to use during restore.                                            
 --db-su                                   Account to use when creating the new database. Optional.                                   
 --db-su-pw                                Password for the "db-su" account. Optional.                                                
 --db-url=<mysql://root:pass@host/db>      A Drupal 6 style database URL indicating the target for database restore. If not provided, 
                                           the archived settings.php is used.                                                         
```
I don't think we necessarily need to support `--db-prefix`, but could, for convenience.

src/Commands/core/ArchiveRestoreCommands.php
weitzman: Agreed. In general, lets prefer a basic implementation for these commands, as opposed to full featured.

src/Commands/core/ArchiveDumpCommands.php
jabxjab: `eval()` has been removed

src/Commands/core/ArchiveDumpCommands.php
jabxjab: No, there is no overhead here. `validateSensitiveData()` analyzes files matching`sites/[subdir]/settings.php` pattern only

src/Commands/core/ArchiveRestoreCommands.php
jabxjab: Updated archive:restore command to limit the support for a local site only

jabxjab:
CI is failing due to `testInsecurePhpPackage`/`testNoInsecureProductionPhpPackage` tests.

weitzman:
I just fixed that test. Feel free to rebase/update your branch.

src/Commands/core/ArchiveRestoreCommands.php
```diff
@@ -0,0 +1,528 @@
...
+                        '!port' => $databaseSpec['port'] ?: dt('n/a'),
+                        '!prefix' => $databaseSpec['prefix'] ?: dt('n/a'),
+                    ]
+                )
+            )
+        ) {
+            throw new UserAbortException();
+        }
+
+        if (!$sql->drop($sql->listTablesQuoted())) {
```
greg-1-anderson, line 520: Also, if the database does not exist, we should attempt to create it.

greg-1-anderson:
I didn't think that the archive:restore tests should overwrite the sut. I started to fix that, but ran out of time.

greg-1-anderson:
archive:restore should do a `composer install` after it unpacks the files.

tests/functional/ArchiveTest.php
```diff
@@ -0,0 +1,427 @@
...
+            'archive:restore',
+            [],
+            array_merge($this->archiveRestoreOptions, [
+                'code' => null,
+                'code-source-path' => Path::join($this->extractPath, 'code'),
+                'destination-path' => $this->restorePath,
+            ])
+        );
+        $this->assertTrue(is_file(Path::join($this->restorePath, 'sut', $testFileName)));
+        $this->installComposerDependencies();
```
greg-1-anderson, line 203: The tests shouldn't have to call `composer install`; the `archive:restore` command should do it.

tests/functional/ArchiveTest.php
```diff
@@ -0,0 +1,427 @@
...
+        );
+        $this->assertRestoredSiteStatus();
+    }
+
+    /**
+     * Asserts the status of restored site.
+     */
+    private function assertRestoredSiteStatus(): void
+    {
+        $this->setupSettingsPhp();
```
greg-1-anderson, line 365: The tests shouldn't have to write the db credentials to settings.php; `archive:restore` should do it.

src/Commands/core/ArchiveRestoreCommands.php
```diff
@@ -102,6 +107,9 @@ public function restore(
         array $options = [
             'destination-path' => null,
             'overwrite' => false,
+            'docroot-dir' => self::DOCROOT_DIR,
```
greg-1-anderson, line 110: We shouldn't need `docroot-dir`. Drupal Finder can determine where the docroot is by examining the composer-installer paths in the composer.json file. Maybe the existing code is already using Drupal Finder?

src/Commands/core/ArchiveRestoreCommands.php
```diff
@@ -0,0 +1,611 @@
...
+            } catch (Throwable $t) {
+                $this->logger()->warning('Could not bootstrap Drupal site at destination to determine file path');
+            }
+        }
+
+        // Find the Drupal root for the archived code, and assume sites/default/files.
+        $composerRoot = $this->getDestinationPath();
+        $drupalFinder = new DrupalFinder();
+        if ($drupalFinder->locateRoot($composerRoot)) {
+            return Path::join($drupalFinder->getDrupalRoot(), 'sites/default/files');
```
greg-1-anderson, line 394: Here's an existing example where we instantiate a DrupalFinder instance to find the Drupal Root.

src/Commands/core/ArchiveRestoreCommands.php
jabxjab: @greg-1-anderson Agree, will replace with DrupalFinder-based implementation

src/Command/ServiceCommandlist.php
```diff
@@ -17,6 +17,6 @@ public function addCommandReference($command): void
 
     public function getCommandList(): array
     {
-        return $this->commandList;
+        return array_filter($this->commandList);
```
greg-1-anderson, line 20: Do you know why it became necessary to add this? Would it be possible to prevent empty items from being added to the list, rather than remove them every time the accessor is called?

greg-1-anderson:
@weitzman I think this PR is in pretty good shape now. Could you please review again?

src/Command/ServiceCommandlist.php
jabxjab: @greg-1-anderson Spent some time investigating it and did not found a real cause of this. What I found is that the dependency references for `config.export.commands` and `config.export.commands` turining into `NULL` once added which causes fatal error on Drupal10 CI pipeline but not on Drupal9-related ones.

rwagner00:
@weitzman Feedback addressed, let me know if there's any other issues. Thanks!

greg-1-anderson:
I added `@validate-php-extension Phar`; I think that makes this complete.

tests/functional/ArchiveTest.php
weitzman: I asked to remove `composer install` from restore. IMO its out of scope. Either the command should archive the vendor dir, or it shouldnt. Happy to discuss.

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,605 @@
...
+     * @option db Archive database SQL dump.
+     * @option destination The full path and filename in which the archive should be stored. If omitted, it will be saved to the drush-backups directory.
+     * @option overwrite Overwrite destination file if exists.
+     * @option description Describe the archive contents.
+     * @option tags Add tags to the archive manifest. Delimit several by commas.
+     * @option overwrite Do not fail if the destination file exists; overwrite it instead. Default is --no-overwrite.
+     * @option generator The generator name to store in the MANIFEST.yml file. The default is "Drush archive-dump".
+     * @option generatorversion The generator version number to store in the MANIFEST file. The default is Drush version.
+     * @option exclude-code-paths Comma-separated list of paths to exclude from the code archive.
+     *
```
weitzman, line 88: Please add a few @usage examples. Especially --destination. Same for restore command

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -0,0 +1,605 @@
...
+ */
+class ArchiveDumpCommands extends DrushCommands
+{
+    /**
+     * @var \Symfony\Component\Filesystem\Filesystem
+     */
+    private Filesystem $filesystem;
+
+    /**
+     * @var string
```
weitzman, line 36: These class property doxygen are duplicative and can be removed.

src/Commands/core/ArchiveRestoreCommands.php
```diff
@@ -0,0 +1,718 @@
...
+     *   1) code ("code" directory);
+     *   2) database dump file ("database/database.sql" file);
+     *   3) Drupal files ("files" directory).
+     * @param string|null $site
+     *   Destination site alias. Defaults to @self.
+     * @param array $options
+     *
+     * @throws \Exception
+     */
+    public function restore(
```
weitzman, line 102: Lets add a validate hook which checks `!elf::programExists('rsync)` and throws an exception if untrue.

tests/functional/ArchiveTest.php
greg-1-anderson: @weitzman: The comment above is old. @jabxjab's original implementation did not do a `composer install` in `archive:restore`. I believed that `archive:restore` should produce a functional site, so I told him to move the composer install to the command itself, and he did so. After that, you provided your review comment that you preferred it if `archive:restore` did not call `composer install`, so @rwagner00 reverted back to the old tests, which made GitHub display my old comment again.

jabxjab:
@weitzman feedback is addressed, thanks!

weitzman (approved):
This looks ready to me. Thanks for all the hard work. I'll leave the merge honors to @greg-1-anderson.

greg-1-anderson:
Thanks everyone!

PR #5149, opened 2022-05-19
Pass $app_root to drupal_rebuild().

weitzman:


weitzman:
We decided not to solve this in Drush https://github.com/drush-ops/drush/pull/4713

Issue #205, opened 2013-10-21
When installing/uninstalling modules, only prompt for confirmation about dependent modules


iant-ee:
Currently, when (un)installing (d7: enabling) a module you will be prompted to confirm if you want to install the module and its dependencies:

```
$ drush en admin_views
The following extensions will be enabled: admin_views, views, views_bulk_operations
Do you really want to continue? (y/n): y
admin_views was enabled successfully.
views was enabled successfully.
views_bulk_operations was enabled successfully.
```

You get a similar prompt even if a module doesn't have any unmet dependencies:

```
$ drush en views
The following extensions will be enabled: views
Do you really want to continue? (y/n): y
views was enabled successfully.
```

This leads to people always passing the -y switch to save time, and accidentally installing dependencies that they weren't planning to.

If the user has explicitly asked for a module to be enabled, then they are unlikely to immediately change their mind, so there seems little benefit asking them for confirmation. I would therefore propose that confirmation is only requested for dependencies, as the user might not have been expecting them. The above examples would then become:

```
$ drush en admin_views
The following extensions are dependencies and will also be enabled: views, views_bulk_operations
Do you really want to continue? (y/n): y
admin_views was enabled successfully.
views was enabled successfully.
views_bulk_operations was enabled successfully.
```

and

```
$ drush en views
views was enabled successfully.
```

I don't think the prompt "The following extensions are dependencies and will also be enabled" is well phrased, so I'm open to suggestions for that.

If I wrote a patch to do this, would it be accepted?

PR #5152, opened 2022-05-23
Make `pm:install` and `pm:uninstall` respect `--simulate` option

Kingdutch:
In Drush 8 it was possible to pass `--no` to `pm:uninstall` to let it output only what modules would be uninstalled without actually uninstalling any modules. This was removed in Drush 9 after discussion in #205 (PR #1057 was declined as a result).

However, we had some automation that used this to see what would be uninstalled so that this could be reported for user verification, even in the case only a single module would be uninstalled. This is now broken and that command uninstalls the module when it's the only one being uninstalled.

To preserve the new desired behaviour but still easily allow automation to determine what changes would be made by running `pm:install` or `pm:uninstall` this PR adds a `--dry-run` option that will not actually perform any of the changes.

An alternative implementation would be to provide entirely separate commands, but that feels like overkill.

### DIFF
--- a/src/Drupal/Commands/pm/PmCommands.php
+++ b/src/Drupal/Commands/pm/PmCommands.php
@@ -67,11 +67,13 @@ class PmCommands extends DrushCommands
      *
      * @command pm:install
      * @param $modules A comma delimited list of modules.
-     * @option dry-run Display what modules would be installed but don't install them.
      * @aliases in, install, pm-install, en, pm-enable, pm:enable
      * @bootstrap root
+     *
+     * @usage drush pm:install --simulate content_moderation
+     *    Display what modules would be installed but don't install them.
      */
-    public function install(array $modules, array $options = ['dry-run' => false]): void
+    public function install(array $modules): void
     {
         $modules = StringUtils::csvToArray($modules);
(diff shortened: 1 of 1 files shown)


weitzman:
Seems a good proposal to me. 

1. Lets unconditionally do our printing via $this->logger->debug(). We dont usually use writeln()
2. Drush already has a --simulate global option, so lets check for that via `if (\Drush\Drush::simulate())`

Kingdutch:
@weitzman 
> Lets do our printing via $this->logger->debug(). We dont usually use writeln()

I mostly copied the existing code here. Do you want to change the output of the command itself to use `debug()` instead of `writeln` too? I'd like to keep the output between `--simulate` and the actual execution the same, otherwise automation may still be difficult/unpredictable.

Kingdutch:
I've made changes 2 and 3 but have omitted 1. Changing to `logger()->debug()` would make the output hidden unless the `--debug` flag is passed as well, which would cause all sorts of output that's not useful to users who want to preview the results of their actions, it would also make automation a lot more difficult as unneeded output would have to be filtered out. I also don't see any other log levels that would make sense over normal output to `stdout`.

PR #5155, opened 2022-05-29
Bootstrap levels are `int` in DrupalBootLevels

mglaman:
When using PHPStan, there is an error with the following code:

Code:
```
#[CLI\Bootstrap(level: DrupalBootLevels::NONE)]
```

Error:
```
Parameter $level of attribute class Drush\Attributes\Bootstrap constructor expects string, int given. 
```

All constants are integers. When passing a stringified zero (`'0`) the attribute fails. So it's documented as a string but actually wants an integer.


weitzman:
thanks

PR #5158, opened 2022-06-15
The field:delete command deletes fields from all bundles instead of only the requested bundles.

### LINKED ISSUE #5157: The field:delete command deletes fields from all bundles instead of only the requested bundles.
**Describe the bug**
The field:delete command deletes fields from all bundles instead of only the requested bundles.

**To Reproduce**
1. Create a field and add it two bundles
2. Delete the field from one of those bundles using `drush field:delete`

**Expected behavior**
The field is deleted from the requested bundle.

**Actual behavior**
The field is deleted from both bundles.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 11.x
| Drupal version? | 9.x
| PHP version     | 8.x
| OS?             | Mac

DieterHolvoet:
Fixes #5157.


DieterHolvoet:
Going to merge this since test failures are unrelated.

PR #5162, opened 2022-06-17
--force option added to 'updatedb' command

f1mishutka:
Small change but really helpful for us. Would like to see it merged as soon as possible. Thank you!

## Motivation

"Do you wish to run the specified pending updates?" question is ok when running `drush updb` from shell.

But we use this command in CI-automation (with Jenkins). We want to have a way to apply database updates automatically in some dev environments. But if there are some updates Jenkins pipeline execution just freezes. With last line in log like this:
```
17:21:26 Do you wish to run the specified pending updates? (yes/no) [yes]:
```
So we really want to have `--force` option to be able to skip this question under such non-interactive shells.

Thank you!

## P.S.:

I intentionally left "Requirements check reports errors. Do you wish to continue?" question unaffected by this new option. Because this requirements question is much more serious and let it better stack instead of applying some incompatible updates silently even in dev environment.

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -29,12 +29,13 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
      * @command updatedb
      * @option cache-clear Clear caches upon completion.
      * @option post-updates Run post updates after hook_update_n and entity updates.
+     * @option force Do not ask 'Do you wish to run the specified pending updates?' question before running updates.
      * @bootstrap full
      * @topics docs:deploy
      * @kernel update
      * @aliases updb
      */
-    public function updatedb($options = ['cache-clear' => true, 'post-updates' => true]): int
+    public function updatedb($options = ['cache-clear' => true, 'post-updates' => true, 'force' => false]): int
     {
         $this->cache_clear = $options['cache-clear'];
         require_once DRUPAL_ROOT . '/core/includes/install.inc';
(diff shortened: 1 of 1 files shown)


f1mishutka:
> Please use existing --no-interaction global flag. Thats common to all Symfony Console apps. You can see all global flags with `drush topic core:global-options`

Thank you for suggestion!

`--no-interaction` will suppress all other questions. We need to suppress only this one.

weitzman:
I see. If you want to ensure that your requirements are met, run `drush rq` before running `drush updatedb` ... Leaving this PR closed

Chi-teck:
@f1mishutka  The other option is providing answers through stdin.
`echo 'yes' | drush updb`

f1mishutka:
> @f1mishutka The other option is providing answers through stdin.
> 
> ```
> echo 'yes' | drush updb
> ```

Thanks for suggestion!

Unfortunately this is not an option at all. You don't know what firts question will be. And you definitely don't want to answer "yes" blindly to something unexpected ("Do you want delete everything?" for example).

My point with this PR was to supress one exact question in a consistent predictable manner. All options given instead are about ignoring all the questions, not just this one.

Chi-teck:
Right, in case of positive answers `echo 'yes'`, `-y`, `-n` work same way here.
I think, it would make more sense to have "No" as default value for "Requirements check" question.

Same question was once asked in Drush Slack channel.
https://drupal.slack.com/archives/C62H9CWQM/p1599541992024900
Unfortunately, `drush rq` does not return non-zero exit code in case of erros. So you'll have to parse its output.

deviantintegral:
We just ran into this as well on a project. It was pretty confusing to see errors in the logs yet CI continuing to go on. I came up with the same potential solution, and found this PR looking for a `--force` flag.

Adding a call first to `core:requirements` does work in our specific case, because loading the module handler causes an exception to be thrown due to the missing module. However, it also feels like we shouldn't rely on its exit code, given that when things are working you get a `0` exit code even when requirements fail.

I think most users would expect `--no-interaction` or `--yes` to mean "don't confirm on potentially destructive actions", not "keep going even if there's major errors".

@weitzman is there anything that could be done to reopen or refile this PR in a way that would be merged? What if we be specific, and add a `--ignore-requirements-check` or even a `--skip-requirements-check` for those who for some reason may want to continue anyways?

Issue #5166, opened 2022-06-22
site:install doesn't escape database name when creating/dropping


MakerTim:
**Describe the bug**
When I run site:install with a database containing characters like `-` it breaks creating the database

**To Reproduce**
set DB_NAME to something with a `-` in the name
run site:install
"[warning] Failed to drop or create the database. Do it yourself before installing. ERROR 1064 (42000) at line 1: You have an error in your SQL syntax..."
example:
> DB_NAME=example-tim drush site:install 
also works with "--db-url=mysql://root:pass@localhost:port/example-tim

**Expected behavior**
dropped existing database & created the database with the `-` in the name

**Actual behavior**
[warning] Failed to drop or create the database. Do it yourself before installing. ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-tim' at line 1

**Workaround**
not having `-` in the name

### System Configuration
| Q               | A
| --------------- | ---
Drush version : 11.0.9 
Drupal version   : 9.3.12                                   
PHP 7.4.27 (cli) (built: Jan 26 2022 18:08:44) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.27, Copyright (c), by Zend Technologies
    with Xdebug v3.1.4, Copyright (c) 2002-2022, by Derick Rethans
OS: Linux

**Additional information**
PR incoming :)
-- see a lot of potential for SQL injection... 
not a real thread, but it's there!

weitzman:
> see a lot of potential for SQL injection

Drush users have CLI access so they can block users, get email addresses, edit/drop the DB, etc. SQL injection is a feature not a bug. Will look at the PR.

ericras:
Same issue as #4203

weitzman:
Lets discuss in #4204

PR #5167, opened 2022-06-22
site:install Fix: database containing certain characters breaks creating of database

MakerTim:
see https://github.com/drush-ops/drush/issues/5166 for context


weitzman:
This looks right but I'm a bit concerned about backward compat. Lets merge this when we branch for drush 12.

PR #5170, opened 2022-06-22
Use compatible method to create Response

andypost:
Response is created via undocumented method even in 3.4 https://symfony.com/doc/3.4/components/http_foundation.html#response


greg-1-anderson:
Rebased in #5361

PR #5172, opened 2022-06-25
Add url option to the core:route command

Chi-teck:
## Motivation
`drush route --path` is one of may favorite Drush commands. However, it feels a bit boring to manually extract path from URL.

## Proposed solution
Add `url` option. So that the URL copied from the browser address bar can be pasted directly to the command line.

Example:
```
drush route --url=https://example.com/node/1
```

### DIFF
--- a/src/Drupal/Commands/core/DrupalCommands.php
+++ b/src/Drupal/Commands/core/DrupalCommands.php
@@ -151,7 +151,7 @@ class DrupalCommands extends DrushCommands
      * @option path An internal path or URL.
      * @version 10.5
      */
-    public function route($options = ['name' => self::REQ, 'path' => self::REQ,'format' => 'yaml'])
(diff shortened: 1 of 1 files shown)


Chi-teck:
Aside issue. This line seems redundant.
https://github.com/drush-ops/drush/blob/11.0.9/src/Drupal/Commands/core/DrupalCommands.php#L163

weitzman:
This is a bit confusing because drush has a global option --uri and that means something similar/overlapping. Maybe just enhance the processing of --path so it can deal with a full url.

Chi-teck:
Well, yes. Actually when Drupal is installed into a sub-direcotry both `--uri` and `--url` options would be needed get to correct result. I've updated the PR to enhance processing of the `--path` option. The only concern I have is that the `path` is no longer the right name for that option. I have no idea how to name option that accepts both URL and path.

weitzman:
Thanks. IMO option name can remain as is.

PR #5174, opened 2022-06-30
Remove timeout for runserver command

jurgenhaas:
Running the server should never time out.

Would be great to backport that to Drush 10 as well.


weitzman:
also merged to 10.x

PR #5180, opened 2022-07-11
Allow to set the URL for the security advisories composer url via env variable

Cyberschorsch:
The URL which is used in the drush pm:security command is hard-coded and links to the drupal-security-advisories repo ( https://github.com/drupal-composer/drupal-security-advisories ). I really would like to have an easy option to use a different url if needed.

This PR sets the url for the source of the security check command to the environment variable DRUSH_SECURITY_ADVISORIES_URL if set, otherwise use the default url.

### DIFF
--- a/src/Commands/pm/SecurityUpdateCommands.php
+++ b/src/Commands/pm/SecurityUpdateCommands.php
@@ -102,7 +102,7 @@ class SecurityUpdateCommands extends DrushCommands
     protected function fetchAdvisoryComposerJson()
     {
         $client = new Client(['handler' => $this->getStack()]);
-        $security_advisories_composer_url = getenv('DRUSH_SECURITY_ADVISORIES_URL') ? getenv('DRUSH_SECURITY_ADVISORIES_URL') : 'https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/9.x/composer.json';
+        $security_advisories_composer_url = getenv('DRUSH_SECURITY_ADVISORIES_URL') ?? 'https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/9.x/composer.json';
         $response = $client->get($security_advisories_composer_url);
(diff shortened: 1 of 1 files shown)


weitzman:
Seems reasonable. 

- Please add a comment about this env variable not being a supported API.
- Pkease use the null coalesce php operator to save some characters,

Cyberschorsch:
@weitzman thank your for the fast feedback, really appreciate it. I made the changes you suggested.

weitzman:
I see that the tests are failing. Hopefully its a quick fix.

Cyberschorsch:
Since getenv() returns FALSE instead of NULL if the env variable is not set or empty we propably have to stick with the ternary operator.

src/Commands/pm/SecurityUpdateCommands.php
```diff
@@ -95,14 +95,20 @@ public function suggestComposerCommand($updates): void
...
+     *
      * @return mixed
      *
      * @throws \Exception
      */
     protected function fetchAdvisoryComposerJson()
     {
         $client = new Client(['handler' => $this->getStack()]);
-        $response = $client->get('https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/9.x/composer.json');
+        $security_advisories_composer_url = getenv('DRUSH_SECURITY_ADVISORIES_URL') ?? 'https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/9.x/composer.json';
```
weitzman:
```suggestion
        $security_advisories_composer_url = getenv('DRUSH_SECURITY_ADVISORIES_URL') ?: 'https://raw.githubusercontent.com/drupal-composer/drupal-security-advisories/9.x/composer.json';
```

weitzman (commented):
yeah, i was thinking of ternary. here is a suggested change

Issue #5184, opened 2022-07-18
The table of available updates seems wrong for post updates


Pasqualle:
**Describe the bug**
This is my output of drush updb (same with drush updbst):
```
 -------------- ------------------------------ --------------- ---------------------------------------
  Module         Update ID                      Type            Description
 -------------- ------------------------------ --------------- ---------------------------------------
  bupo_teasers   9000                           hook_update_n   9000 - Create bupo_teasers field.
  Update node_reference paragraphs.sers_field   post-update
 -------------- ------------------------------ --------------- ---------------------------------------

``` 
The post update description seems to be displayed in the wrong column.

**To Reproduce**
This is my post update hook declaration (in bupo_teasers.post_update.php) :
```
/**
 * Update node_reference paragraphs.
 */
function bupo_teasers_post_update_6_test_migrate_teasers_field(&$sandbox) {
```

**Expected behavior**
The post-update should be correctly listed. Update description in the last column.

**Actual behavior**
see the description

**Workaround**
update.php shows the list of update descriptions correctly.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.6.2
| Drupal version? | 9.3.18
| PHP version     | 8.0.19
| OS?             | Ubuntu (with WSL2)

**Additional information**
[UpdateDBCommands.php#L609](https://github.com/drush-ops/drush/blob/10.x/src/Commands/core/UpdateDBCommands.php#L609)
Trying to hard code the description like
`'description' => 'X',`
then the output is correct, the description X is displayed in the last column.

Pasqualle:
The source of the problem is windows style line ends used in the post_update file.
Switching to linux line ends solves this issue (without needing to change anything in drush).

https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Update%21UpdateRegistry.php/function/UpdateRegistry%3A%3AgetPendingUpdateInformation/9.3.x
This getPendingUpdateInformation() does not remove `\r` from the doc comment, and results in the broken table display.

Pasqualle:
I think this bug can be closed as developer not following coding standards.
https://www.drupal.org/docs/develop/standards/coding-standards

> Files should be formatted with \n as the line ending (Unix line endings), not \r\n (Windows line endings).

PR #5185, opened 2022-07-18
Fix db update status for post update hooks

Pasqualle:
Solution for issue: https://github.com/drush-ops/drush/issues/5184


Pasqualle:
Still broken with multi-line comments, with windows style line ends. See https://github.com/drush-ops/drush/issues/5184

PR #5193, opened 2022-07-27
Sort migration message

RoSk0:
Default sort order is based on source IDs hash which is not useful.


weitzman:
So what column is used for sorting in this PR?

RoSk0:
Hi @weitzman ,

I'm actually not sure now. I don't have that environment to reproduce and confusingly docs are not really helpful.

From memory we had a D7=>D9 node migration, so there was three source ID columns (`nid`, `vid`,`lang`) and with default output sorted by the source ID hash , messages related to a single source  node was randomly spread in the output. With that patch, messages related to the same source node was nicely grouped together in the output.

Noticed that the patch  missed to sort the output when `idlist` is provided. Will add that now.

PR #5200, opened 2022-08-08
Set a default account to make sure the correct timezone is set

### LINKED ISSUE #1166: Set default timezone
A number of my colleagues work on OSX, whose PHP implementation doesn't appear to set date.timezone. It should be pretty harmless to add `date_default_timezone_set()` somewhere early in our bootstrap to avoid lots of warnings like:

```
date(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone.
```

I've begun sometimes adding `date_default_timezone_set()` in hook_drush_init() in some of my contrib projects. This is pretty trivial, so I'm going to go ahead and add this unless there are objections. Any suggestions as to where?

DieterHolvoet:
Fixes #1166.

### DIFF
--- a/src/Boot/DrupalBoot8.php
+++ b/src/Boot/DrupalBoot8.php
@@ -248,6 +248,9 @@ class DrupalBoot8 extends DrupalBoot implements AutoloaderAwareInterface
 
         parent::bootstrapDrupalFull($manager);
         $this->addDrupalModuleDrushCommands($manager);
+
+        // Set a default account to make sure the correct timezone is set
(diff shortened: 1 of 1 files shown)


weitzman:
Seems reasonable to me. Lets merge in a couple days if there is no negative feedback.

DieterHolvoet:
Let's do this? 😃

PR #5203, opened 2022-08-12
Update queue:run to run garbage collection

marcaddeo:
This updates the queue:run command to run garbage collection against the
queue if it supports it. If we don't do this, queue items expirations
never actually expire until the next cron run.


weitzman:
Why is waiting for the next cron a bad idea? I'm not sure queue running implies garbage collection (i.e. it could be a separate command).

marcaddeo:
@weitzman I'm using a `DelayedRequeueException` to have my queue item reprocessed in 5 minutes, but we only run cron every 10 minutes. 

And then I'm running `queue:run <queue name>` every minute to ensure my queued items are quickly processed.

I realize I could just run cron more frequently, or instead of `queue:run`, but this just seemed like strange behavior to me when running `queue:run` quickly and not having my delayed requeued items process.

Core does queue garbage collection in `system_cron` before any queue workers are run, so it seemed slightly implied that it would be run for a queue runner to me.

I could see it being a separate queue command as well if you think that's the better route.

weitzman:
Thoughts? @juampynr @davereid @damiankloip and any other queue folks?

weitzman:
This looks useful and benign so in it goes. Sorry for the big delay.

juampynr:
Sorry, super late review but still: this looks great!

PR #5204, opened 2022-08-13
Make config.storage.sync injection optional and log any issues

### LINKED ISSUE #5195: Missing services are quietly ignored
**Describe the bug**

I am debugging a new project and they didn't send a settings.php and the one I put together accidentally ommitted `$settings["config_sync_directory"]`. drush simply omits the config import and export commands and there is no error message anywhere. I only realized when I gave up and went to `/admin/config/development/configuration` in the UI.

**To Reproduce**
What did you do?

Oh dear, what have I done, yeah, that's what I am asking myself in the long lonely nights. But in this case see above.

**Expected behavior**

Maybe something in `drush status`?

**Actual behavior**
What happened instead?

drush config:import is just missing

**Workaround**

Not really.

DieterHolvoet:
Fixes #5195.

### DIFF
--- a/src/Drupal/Commands/config/ConfigImportCommands.php
+++ b/src/Drupal/Commands/config/ConfigImportCommands.php
@@ -81,6 +81,11 @@ class ConfigImportCommands extends DrushCommands
         return $this->configStorageSync;
     }
 
+    public function setConfigStorageSync(StorageInterface|null $syncStorage): void
+    {
+        $this->configStorageSync = $syncStorage;
+    }
+
     public function getConfigCache(): CacheBackendInterface
     {
         return $this->configCache;
@@ -164,7 +169,6 @@ class ConfigImportCommands extends DrushCommands
     public function __construct(
         ConfigManagerInterface $configManager,
         StorageInterface $configStorage,
-        StorageInterface $configStorageSync,
         CacheBackendInterface $configCache,
         ModuleHandlerInterface $moduleHandler,
         // Omit type hint as it changed in https://www.drupal.org/project/drupal/issues/3161983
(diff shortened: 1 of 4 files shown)


weitzman:
@bircher @greg any thoughts? See linked issue. @greg-1-anderson

bircher:
I am ok with this approach. But I think we need to throw the exception also when trying to use the uninitialized service. Because otherwise it will be difficult to understand why the command doesn't work.

greg-1-anderson:
I don't have time to try it myself right now. The things that are unclear to me from the linked issue:

1. What happens in the Drupal admin UI when the config sync directory is undefined? Do any import / export functions work?
2. What happens in Drush if you run config import or export and specify a directory rather than relying on the config sync configuration, when the later is not defined?

Having an indicator in `drush status` seems like it would be very helpful to have as well.

src/Drupal/Commands/config/ConfigImportCommands.php
```diff
@@ -81,6 +81,11 @@ public function getConfigStorageSync(): StorageInterface
         return $this->configStorageSync;
     }
 
+    public function setConfigStorageSync(StorageInterface|null $syncStorage): void
```
greg-1-anderson: From the failed tests: this is not valid syntax in PHP 7.4.

DieterHolvoet:
> I am ok with this approach. But I think we need to throw the exception also when trying to use the uninitialized service. Because otherwise it will be difficult to understand why the command doesn't work.

@bircher do you mean adding a check in `getConfigStorageSync()`?

bircher:
Ok so I checked more closely again. And we should refactor this a lot. I don't know what part of this is considered "public api of drush" But I don't think it is a good idea to extend the import and export command classes.
Our code should work fine even when not checking for the property to be set because the code works with the path to the sync directory and checks if the path is the same as the one in the settings before accessing the sync storage property. So my concern from the previous comment is mitigated. And since I don't think these methods should be API it is fine to do without the check since php will already complain by itself.

bircher:
so tldr: looks good to me

PR #5209, opened 2022-08-18
Fix fatal error on Drupal 10 w/ symfony/console:6.2+

larowlan:
On Drupal 10 (with symfony/console 6.2.x) I'm getting this fatal:

```
PHP Fatal error:  Declaration of Drush\Style\DrushStyle::choice(string $question, array $choices, mixed $default = null): mixed must be compatible with Symfony\Component\Console\Style\SymfonyStyle::choice(string $question, array $choices, mixed $default = null, bool $multiSelect = false): mixed 
```

This is to fix that signature difference

### DIFF
--- a/src-symfony-compatibility/v6/Style/DrushStyle.php
+++ b/src-symfony-compatibility/v6/Style/DrushStyle.php
@@ -27,7 +27,7 @@ class DrushStyle extends SymfonyStyle
     {
         // Display the choices without their keys.
         $choices_values = array_values($choices);
-        $return = parent::choice($question, $choices_values, $default);
+        $return = parent::choice($question, $choices_values, $default, $multiSelect);
 
         return array_search($return, $choices);
     }
--- a/src-symfony-compatibility/v6/Style/DrushStyle.php
+++ b/src-symfony-compatibility/v6/Style/DrushStyle.php
@@ -23,7 +23,7 @@ class DrushStyle extends SymfonyStyle
         return parent::confirm($question, $default);
     }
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson (approved):

src-symfony-compatibility/v6/Style/DrushStyle.php
```diff
@@ -23,7 +23,7 @@ public function confirm(string $question, bool $default = true): bool
         return parent::confirm($question, $default);
     }
 
-    public function choice(string $question, array $choices, mixed $default = null): mixed
+    public function choice(string $question, array $choices, mixed $default = null, bool $multiSelect = false): mixed
```
weitzman, line 26: We need to pass along $multiSelect in the parent::choice call a few lines down.

weitzman (changes requested):

src-symfony-compatibility/v6/Style/DrushStyle.php
larowlan: :facepalm: fixed - to clarify this is only when we get to symfony/console 6.2

weitzman: its safe to merge this now right? its compatible with symfony 4?

larowlan: I think you have a different DrushStyle for symfony4 in the src-symfony-compatibility/v4 directory.

In terms of will it break on symfony/console: 6.1 - doesn't seem like it https://3v4l.org/ZXXWg

Issue #4935, opened 2021-12-16
webmozart/path-util is abandoned


danepowell:
**Describe the bug**
Installing Drush via Composer produces a warning:
> Package webmozart/path-util is abandoned, you should avoid using it. Use symfony/filesystem instead.

I know this is just a warning and end users can ignore it for now, but it's still a risk (bad things happen to abandoned packages) and the package needs to be removed/replaced.

**To Reproduce**
`composer require drush/drush`

**Expected behavior**
No warnings

**Actual behavior**
> Package webmozart/path-util is abandoned, you should avoid using it. Use symfony/filesystem instead.

**Workaround**
Ignore the warning

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 11.x

weitzman:
Oy. We use that everywhere!

greg-1-anderson:
Sad panda. It's a small library; I guess we could just copy it into Drush, or maybe fork it into Consolidation and adopt ownership.

We could review symfony/filesystem and see if it really does contain all of the methods in path-util; last time I checked, it didn't, but that was a while ago.

romulasry:
> Sad panda. It's a small library; I guess we could just copy it into Drush, or maybe fork it into Consolidation and adopt ownership.
> 
> We could review symfony/filesystem and see if it really does contain all of the methods in path-util; last time I checked, it didn't, but that was a while ago.

Any way to contact them?

romulasry:
I found this at their support page https://symfony.com/support

https://github.com/symfony/symfony/discussions

danepowell:
We faced this problem in https://github.com/typhonius/acquia-php-sdk-v2/pull/219 and found it easy enough to migrate to symfony/filesystem, all of the functions were the same. The only problem is that it breaks compatibility with Symfony 4.

Incompatibility with Symfony 4 is the biggest problem here and looks like it would affect Drush as well. I think Symfony should backport this to Symfony 4 since it's supposed to be supported for a few years yet.

danepowell:
I opened an upstream feature request with Symfony to backport Path functionality to Symfony 4: https://github.com/symfony/symfony/issues/44897

weitzman:
Thanks @danepowell

konfuzed:
[...]

PR #5227, opened 2022-09-08
webmozart/path-util is abandoned

Pasqualle:
#4935

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
     "symfony/var-dumper": "^4.0 || ^5.0 || ^6.0",
     "symfony/yaml": "^4.0 || ^5.0 || ^6.0",
     "webflo/drupal-finder": "^1.2",
-    "webmozart/path-util": "^2.1.0"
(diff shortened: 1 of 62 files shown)


weitzman:
I think we should use symfony/filesystem since we dont want to maintain our own PathUtil. See #4935

Pasqualle:
But that can't be used with Symfony 4, which means can't be used with Drupal 9, right?
Will there be a new drush for D10 only, or drush will use the abandoned webmozart/path-util until support for D9 is dropped?

weitzman:
Ah right. At the moment, we are not planning a D10 only Drush release. It will probably happen during D10's lifetime, but not for 10.0. So, I'd say that this PR is not much better than the current situation. Just because a package gets abandoned does not mean all its downstreams have to now maintain it. Maybe we can coax @danepowell to maintain a fork of path-util 👅

Pasqualle:
Fixed in https://github.com/drush-ops/drush/pull/5264

PR #5230, opened 2022-09-11
Update theme:enable to theme:install

### LINKED ISSUE #5229: theme:enable should be theme:install
**Describe the bug**
In Drupal 8.x and above, there is no longer a way to "enable/disable" themes. Instead we "install/uninstall" them.

However, Drush still has a theme:enable command, while theme:uninstall is fine.

This is confusing (enable is the opposite of uninstall?) and inconsistent with the Drupal UI.

**To Reproduce**
List commands in Drush and you'll see that theme:enable turns a module on, and theme:uninstall turns it off (which also involves adding/removing configuration, database tables, and/or content).

**Expected behavior**
The theme:enable command should be named theme:install for consistency with the Drupal UI and with the theme:uninstall command.

**Additional information**
#4588.

gitressa:
Fixes #5229.


weitzman:
LGTM. Any concerns, @greg-1-anderson?

PR #5254, opened 2022-09-27
Fixes #5253: Drush user-cancel command blocks the user account instead of deletion.

### LINKED ISSUE #5253: Drush user-cancel command blocks the user account instead of deletion
**Describe the bug**
Drush user-cancel command blocks an user account instead of deletion. For blocking users there is user-block command.

**To Reproduce**
Run `drush user-cancel username -y`

**Expected behavior**
The account is deleted, content is reassigned to the Anonymous user.

**Actual behavior**
The account is blocked.

**Workaround**
The only way to do it manually from Drupal UI.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 11.x/1
| Drupal version? | 9.x
| PHP version     | N/A
| OS?             | N/A

xopoc14:
Fixes #5253

### DIFF
--- a/src/Drupal/Commands/core/UserCommands.php
+++ b/src/Drupal/Commands/core/UserCommands.php
@@ -283,7 +283,7 @@ class UserCommands extends DrushCommands
                 $this->logger()->warning(dt('All content created by !name will be deleted.', ['!name' => $account->getAccountName()]));
             }
             if ($this->io()->confirm('Cancel user account?: ')) {
-                $method = $options['delete-content'] ? 'user_cancel_delete' : 'user_cancel_block';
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks. This needs a test to check reassignment, and is a backward compat change so would need to go into next major release.

xopoc14:
> Thanks. This needs a test to check reassignment, and is a backward compat change so would need to go into next major release.

Sure, I am busy at the moment, will do it as I have time

weitzman:
Would you mind editing the command description and options/examples so that they are clear on what happens to the user, and what happens to the content.

weitzman:
This got resolved with a new --reassign-content option

PR #5258, opened 2022-10-04
Bugfix: Drush should typehint interfaces rather than concrete implementations

dpi:
To allow service decoration
This is currently crashing out Drush when a module with a service decoration for `date.formatter` is enabled.


dpi:
Its not clear to me whether the CI failures are caused by this PR.

Seems unlikely.

weitzman:
Failures are unrelated.

dpi:
Thanks @weitzman !

PR #5262, opened 2022-10-07
Add pronunciation to README

Vancelot11:


weitzman:
Thanks. We used to have this in README but now its on our pretty web site. Also, the pronunciation in the PR differs from the recommended one. See https://www.drush.org/latest/#how-do-i-pronounce-drush

PR #5263, opened 2022-10-10
Add an example about deleting taxonomy terms.

ylp-ineris:


weitzman:
Thanks. I think we have sufficient examples already for this command. We dont aim to be comprehensive.

PR #5264, opened 2022-10-10
Issue #4935: Replace webmozart/path-util with symfony/filesystem

ossobuffo:
`symfony/filesystem` as of v5.4 is pretty much a drop-in replacement for `webmozart/path-util`, at least for all the drush use-cases.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,12 +45,12 @@
     "league/container": "^3.4 || ^4",
     "psy/psysh": "~0.11",
     "symfony/event-dispatcher": "^4.0 || ^5.0 || ^6.0",
+    "symfony/filesystem": "^5.4",
     "symfony/finder": "^4.0 || ^5 || ^6",
     "symfony/polyfill-php80": "^1.23",
     "symfony/var-dumper": "^4.0 || ^5.0 || ^6.0",
     "symfony/yaml": "^4.0 || ^5.0 || ^6.0",
-    "webflo/drupal-finder": "^1.2",
-    "webmozart/path-util": "^2.1.0"
+    "webflo/drupal-finder": "^1.2"
   },
   "require-dev": {
     "composer/installers": "^1.7",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "44a4a9422530d4f23e065c0245690160",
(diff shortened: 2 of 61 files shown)


weitzman:
Thanks for getting this going. Looks like [consolidation/site-alias](https://github.com/consolidation/site-alias) has a hidden dependency on webmozart. Lets fix that too. We can get that merged and released quickly.

Also note that you may see one security related test failure in drush thats unrelated to this PR.

ossobuffo:
@weitzman  See https://github.com/consolidation/site-alias/pull/55

effulgentsia:
This broke compatibility with both the Drupal 9 and Drupal 10 versions of `drupal/core-dev`, because the former requires `symfony/filesystem: ^4.4` and the latter requires `symfony/filesystem: ^6.1`. See also https://github.com/drupal/core-dev/pull/3.

ossobuffo:
I would imagine we could fix this by changing drush's composer.json filesystem reference to this:
```
"symfony/filesystem": "^4.4 || ^5.4 || ^6.1",
```
but this would require more testing.

weitzman:
Ouch. That approach won’t work for 4 because path functionality is not in file system v4. Looks like we gotta revert this.

Pasqualle:
It would be nice to allow symfony/filesystem version 6.

Can we just remove symfony/filesystem dependency for drush, as it is already required by consolidation/site-alias (by both consolidation/site-alias versions: ^3.1.6 and ^4).

```
$ composer why symfony/filesystem
chi-teck/drupal-code-generator 2.6.1    requires symfony/filesystem (^4.4 || ^5.1 || ^6)
consolidation/robo             3.0.10   requires symfony/filesystem (^4.4.9 || ^5 || ^6)
consolidation/self-update      2.0.5    requires symfony/filesystem (^2.5 || ^3 || ^4 || ^5 || ^6)
consolidation/site-alias       4.0.0    requires symfony/filesystem (^5.4 || ^6)
drush/drush                    11.x-dev requires symfony/filesystem (^4.4 || ^5.4)

$ composer why consolidation/site-alias
consolidation/site-process 5.1.1    requires consolidation/site-alias (^3 || ^4)
drush/drush                11.x-dev requires consolidation/site-alias (^3.1.6 || ^4)
```

https://packagist.org/packages/consolidation/site-alias#3.1.6
https://packagist.org/packages/consolidation/site-alias#4.0.0

weitzman:
It doesn’t make sense to me to to remove the dependency. We use it extensively. Happy to expand its constraint.

Pasqualle:
note: symfony/filesystem ^6.1 added in #5292

PR #5278, opened 2022-10-15
Symfony Path with symfony/filesystem 4.x

greg-1-anderson:
We'll try the tests on Symfony 4.x once, then we'll loosen up to allow 4 or 5.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
     "league/container": "^3.4 || ^4",
     "psy/psysh": "~0.11",
     "symfony/event-dispatcher": "^4.0 || ^5.0 || ^6.0",
-    "symfony/filesystem": "^4.4",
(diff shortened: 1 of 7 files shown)


greg-1-anderson:
[2200c9a](https://github.com/drush-ops/drush/pull/5278/commits/2200c9a7fce759f5143500fb667acbf0da26ef5c) has all tests green running on Symfony 4.x.

greg-1-anderson:
Implications for global Drush:

- Now we have symfony/filesystem ^5.4 in our vendor, which is incompatible with drupal/core-dev
- If you've added drupal/core-dev, you must be using Composer to manage your site
- If you're using Composer to manage your site, then you should also include Drush via Composer

So I think we're good here.

weitzman:
This looks great. Thanks @greg-1-anderson.

PR #5285, opened 2022-10-19
Remove symfony/filesystem dependency

Pasqualle:
https://github.com/drush-ops/drush/pull/5264#issuecomment-1284600799


weitzman:
Our own code used it so it’s a direct dependency of drush.

PR #5287, opened 2022-10-21
Allow resetting the status of multiple migrations

DieterHolvoet:
This PR adds the `all` and `tag` options to the `migrate:reset-status` command, together with the ability to pass a comma separated list of migration ids as argument.

### DIFF
--- a/src/Drupal/Commands/core/MigrateRunnerCommands.php
+++ b/src/Drupal/Commands/core/MigrateRunnerCommands.php
@@ -609,9 +609,7 @@ class MigrateRunnerCommands extends DrushCommands
             foreach ($migrations as $migrationId => $migration) {
                 /** @var MigrationInterface $migration */
                 $status = $migration->getStatus();
-                if ($status == MigrationInterface::STATUS_IDLE) {
-                    $this->logger()->warning(dt('Migration @id is already Idle', ['@id' => $migrationId]));
-                } else {
+                if ($status !== MigrationInterface::STATUS_IDLE) {
                     $migration->setStatus(MigrationInterface::STATUS_IDLE);
                     $this->logger()->success(dt('Migration @id reset to Idle', ['@id' => $migrationId]));
                 }
--- a/src/Drupal/Commands/core/MigrateRunnerCommands.php
+++ b/src/Drupal/Commands/core/MigrateRunnerCommands.php
@@ -570,29 +570,52 @@ class MigrateRunnerCommands extends DrushCommands
      *
      * @command migrate:reset-status
      *
-     * @param string $migrationId
-     *   The ID of migration to reset.
+     * @param string|null $migrationIds
+     *   Comma-separated list of migration IDs.
      *
      * @aliases mrs,migrate-reset-status
      *
      * @topics docs:migrate
      *
+     * @option all Process all migrations.
+     * @option tag A comma-separated list of migration tags to rollback
+     *
+     * @usage migrate:reset-status --all
+     *   Reset the status of all migrations
+     * @usage migrate:reset-status --tag=user,main_content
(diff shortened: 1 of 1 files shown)


DieterHolvoet:
One thing I wasn't sure about is whether we should log _Migration x is already Idle_ if passing the `all` or `tag` options, or whether we should only log migrations that were actually reset.

weitzman:
> One thing I wasn't sure about is whether we should log Migration x is already Idle if passing the all or tag options, or whether we should only log migrations that were actually reset.

Yes lets omit those log messages. Less noise.

DieterHolvoet:
Changed and rebased.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -570,29 +570,50 @@ public function stop(string $migrationId): void
...
-            $migration->setStatus(MigrationInterface::STATUS_IDLE);
-            $this->logger()->success(dt('Migration @id reset to Idle', ['@id' => $migrationId]));
+        $tags = $options['tag'];
+        $all = $options['all'];
+
+        if (!$all && !$migrationIds && !$tags) {
+            throw new \Exception(dt('You must specify --all, --tag, or one or more migration names separated by commas'));
+        }
+
+        if (!$list = $this->getMigrationList($migrationIds, $options['tag'])) {
```
claudiu-cristea, line 604: We've already assign `$options['tag']` to `$tags`. Why not use that var?

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -570,29 +570,50 @@ public function stop(string $migrationId): void
...
+            $this->logger()->error(dt('No migrations found.'));
+        }
+
+        foreach ($list as $migrations) {
+            foreach ($migrations as $migrationId => $migration) {
+                /** @var MigrationInterface $migration */
+                $status = $migration->getStatus();
+                if ($status !== MigrationInterface::STATUS_IDLE) {
+                    $migration->setStatus(MigrationInterface::STATUS_IDLE);
+                    $this->logger()->success(dt('Migration @id reset to Idle', ['@id' => $migrationId]));
```
claudiu-cristea, line 614: Maybe `Migration @id status reset to idle` (added the "status" word)

claudiu-cristea (changes requested):
This looks good but we need to enrich the tests to cover also the `--all` and `--tag` cases.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -570,29 +570,50 @@ public function stop(string $migrationId): void
...
-        /** @var MigrationInterface $migration */
-        $migration = $this->migrationPluginManager->createInstance($migrationId);
-        $status = $migration->getStatus();
-        if ($status == MigrationInterface::STATUS_IDLE) {
-            $this->logger()->warning(dt('Migration @id is already Idle', ['@id' => $migrationId]));
-        } else {
-            $migration->setStatus(MigrationInterface::STATUS_IDLE);
-            $this->logger()->success(dt('Migration @id reset to Idle', ['@id' => $migrationId]));
+        $tags = $options['tag'];
+        $all = $options['all'];
```
claudiu-cristea, line 598: What happens if some are using:

* `--all` with `--tag` and some migration names?
* but `--all` with `--tag`?
* or `--all` with a list of migrations?

Maybe we should remove `--all` and NO MIGRATIONS == ALL? Then:

* no arguments, no options: _All migrations_
* only `--tag`: _All migrations filtered by provided tags_
* migration IDs + `--tag`: should return error (or filter the list on tags?)

PR #5293, opened 2022-10-24
Warn if modules passed to pm-uninstall are not enabled

DieterHolvoet:
Now it prints _'Successfully uninstalled: \<moduleName\>'_ which is confusing IMO.

### DIFF
--- a/src/Drupal/Commands/pm/PmCommands.php
+++ b/src/Drupal/Commands/pm/PmCommands.php
@@ -170,17 +170,17 @@ class PmCommands extends DrushCommands
     {
         $modules = StringUtils::csvToArray($modules);
 
-        $enabled_modules = array_filter($modules, function ($module) {
+        $installed_modules = array_filter($modules, function ($module) {
             return $this->getModuleHandler()->moduleExists($module);
         });
-        if ($enabled_modules === []) {
-            throw new \Exception(dt('The following module(s) are not enabled: !list. No modules to uninstall.', ['!list' => implode(', ', $modules)]));
+        if ($installed_modules === []) {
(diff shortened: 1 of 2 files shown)


src/Drupal/Commands/pm/PmCommands.php
```diff
@@ -169,7 +169,18 @@ public function validateEnableModules(CommandData $commandData): void
     public function uninstall(array $modules): void
     {
         $modules = StringUtils::csvToArray($modules);
-        $list = $this->addUninstallDependencies($modules);
+
+        $enabled_modules = array_filter($modules, function ($module) {
+            return $this->getModuleHandler()->moduleExists($module);
+        });
+        if ($enabled_modules === []) {
+            throw new \Exception(dt('The following module(s) are not enabled: !list. No modules to uninstall.', ['!list' => implode(', ', $modules)]));
```
weitzman: I think the right language now is not enabled/disabled but installed/uninstalled. We just did a PR to change the from:enable to pm:install

weitzman (changes requested):
LGTM. Minor change requested.

src/Drupal/Commands/pm/PmCommands.php
DieterHolvoet: Changed

weitzman:
I'm wondering this this will break CI or scripts for anyone. I think it will, with the new Exception. We should make a new minor release, and not a patch release.

weitzman (approved):
Approved. Lets merge before we release a new minor. That way this doesn't accidentally go out in a patch release.

Issue #5283, opened 2022-10-19
Make Drush core command classes 'final'


weitzman:
Indicate that command classes should not be extended by marking them with the final keyword. Followup to #5260

greg-1-anderson:
Was also considering making this a requirement in the next major version of the annotated command library.

DieterHolvoet:
I can understand why we whould make all built-in Drush commands final but I don't really see why this should be a requirement for absolutely all commands, including custom/contrib. Especially if this PR is merged: https://github.com/consolidation/annotated-command/pull/272

greg-1-anderson:
Yeah, I agree that making command classes break if they were not final would be too inconvenient for contrib modules with Drupal commands. Maybe in the next major release of annotated commands I will be a bit more strict, but I won't require `final`.

weitzman:
Done

PR #5295, opened 2022-11-02
Do not allow commands in parent classes

greg-1-anderson:
Set flag in the Annotated Command commandFactory, if it exists, to ignore commands in parent classes.

This prohibition was added in annotated-command v4.5.7, but it will be reverted to maintain backwards compatibility. See https://github.com/consolidation/annotated-command/pull/277. Once this revert happens, Drush should continue to disallow extending command classes, to avoid side effects in modules.

### DIFF
--- a/src/Runtime/DependencyInjection.php
+++ b/src/Runtime/DependencyInjection.php
@@ -166,6 +166,11 @@ class DependencyInjection
         $factory = $container->get('commandFactory');
         $factory->setIncludeAllPublicMethods(false);
         $factory->setIgnoreCommandsInTraits(true);
+        // Temporary: once we have a stable release of Annotated Command with
+        // this flag, we can call it invariantly.
+        if (method_exists($factory, 'setIgnoreCommandsInParentClasses')) {
+            $factory->setIgnoreCommandsInParentClasses(true);
+        }
         $factory->addCommandInfoAlterer(new DrushCommandInfoAlterer());
 
         $commandProcessor = $container->get('commandProcessor');


greg-1-anderson:
I merged the Annotated Command PRs into the latest dev releases (4.x-dev and 5.x dev). We should merge this into Drush, then do a stable release of Drush, then do a stable release of Annotated Command.

No rush, as long as we don't forget.

DieterHolvoet:
Does that mean that the backwards compatibility break doesn't apply for Drush?

greg-1-anderson:
Backwards compatibility is only enforced for defined behavior. If you use a feature that's not supposed to be a feature because it's part of undefined behavior, then continuing use of that feature is not guaranteed. Command classes that need to have variable behavior based on environment should use _hasa_ rather than _isa_ (i.e. use dependency injection) in the implementation. Inheritance wasn't considered; it was neither specifically supported nor explicitly prohibited. For the library, we should be permissive. I'm undecided about whether to default this to `true` in the 5.x version of the library, but I'm inclined to let it go. It's safe to do this in the library because we don't provide any command classes. If people want to design their command classes to use inheritance to extend the behavior, that's on them.

Drush, on the other hand, has a lot of command classes. None of them were designed to be extended. If we allow folks to do this, internal changes to Drush could break folks using this undefined behavior. I think it's much safer to prohibit the extension of command classes in Drush by disabling the undefined behavior that we cannot support.

If you have a strong need for this feature to keep working in the context of Drush, appeal to @weitzman, he has final say. All we need to do in that instance is close this PR as will-not-fix. I recommend that we merge it instead.

greg-1-anderson:
I middle ground could be to add *final* to all of our command classes, and continue to let modules extend their command classes. I don't really like this idea either, but it's one level safer than just punting entirely.

DieterHolvoet:
Thanks for your explanation, makes sense! I'm not opposed to merging this, I was just curious about your reasoning.

weitzman:
> A middle ground could be to add final to all of our command classes, and continue to let modules extend their command classes.

I like that. I'm inclined to not break anyone's install. Folks can watch https://github.com/drush-ops/drush/issues/5283 for progress.

greg-1-anderson:
Annotated Command library 4.6.1 released; closing here.

PR #5298, opened 2022-11-07
Add --severity-min parameter to watchdog commands

jonathan1055:
Initial version of new `--severity-min` parameter.
Some tests failed locally, but I think that was an unrelated problem. 

This feature came out of a [discussion on slack](https://app.slack.com/client/T06GX3JTS/C62H9CWQM/thread/C62H9CWQM-1667491861.622799)

### DIFF
--- a/src/Drupal/Commands/core/WatchdogCommands.php
+++ b/src/Drupal/Commands/core/WatchdogCommands.php
@@ -57,7 +57,7 @@ class WatchdogCommands extends DrushCommands
      */
     public function show($substring = '', $options = ['format' => 'table', 'count' => 10, 'severity' => self::REQ, 'severity-min' => self::REQ, 'type' => self::REQ, 'extended' => false])
     {
-        $where = $this->where($options['type'], $options['severity'], $options['severity-min'], $substring);
+        $where = $this->where($options['type'], $options['severity'], $substring, 'AND', $options['severity-min']);
         $query = Database::getConnection()->select('watchdog', 'w')
             ->range(0, $options['count'])
             ->fields('w')
@@ -121,20 +121,20 @@ class WatchdogCommands extends DrushCommands
      * @usage  drush watchdog:tail
      *   Continuously tail watchdog messages.
      * @usage drush watchdog:tail "cron run successful"
-     *   Continously tail watchdog messages, filtering on the string <info>cron run successful</info>.
+     *   Continuously tail watchdog messages, filtering on the string <info>cron run successful</info>.
      * @usage drush watchdog:tail --severity=Notice
-     *   Continously tail watchdog messages, filtering severity of notice.
+     *   Continuously tail watchdog messages, filtering severity of notice.
      * @usage drush watchdog:tail --severity-min=Warning
-     *   Continously tail watchdog messages, filtering for a severity of warning or higher.
+     *   Continuously tail watchdog messages, filtering for a severity of warning or higher.
      * @usage drush watchdog:tail --type=php
-     *   Continously tail watchdog messages, filtering on type equals php.
+     *   Continuously tail watchdog messages, filtering on type equals php.
      * @aliases wd-tail,wt,watchdog-tail
      * @validate-module-enabled dblog
      * @version 10.6
      */
     public function tail(OutputInterface $output, $substring = '', $options = ['severity' => self::REQ, 'severity-min' => self::REQ, 'type' => self::REQ, 'extended' => false]): void
     {
-        $where = $this->where($options['type'], $options['severity'], $options['severity-min'], $substring);
+        $where = $this->where($options['type'], $options['severity'], $substring, 'AND', $options['severity-min']);
         if (empty($where['where'])) {
             $where = [
               'where' => 'wid > :wid',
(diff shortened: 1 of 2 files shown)


jonathan1055:
No test coverage yet obviously. Will add when the approach is agreed.

jonathan1055:
At @weitzman's request I re-worked this to be a `--severity-min` parameter. This can be used in `wd-show` and `wd-tail`. It is not (yet) used in the interactive `wd-list` as that would require more work to change the user input (but could be done if required). I have also not implemented it in `wd-del` as I did not think it necessary. But again it could be added there if required. The two parameters `--severity` and `--severity-min` cannot both be set on the same call, because the presence of one or the other one is how we tell which filtering to do.

jonathan1055:
The test failure is only on the D10 test build. This looks to be unrelated to my change, but I'm not really sure.
````
TTY mode requires /dev/tty to be read/writable.
````

File: src/Drupal/Commands/core/WatchdogCommands.php [hunk 1]
```diff
@@ -275,14 +281,16 @@ public function showOne($id, $options = ['format' => 'yaml']): PropertyList
...
+     *   Int or String for the minimum severity to return.
      * @param $filter
      *   String. Value to filter watchdog messages by.
      * @param $criteria
      *   ('AND', 'OR'). Criteria for the WHERE snippet.
      * @return
      *   An array with structure ('where' => string, 'args' => array())
      */
-    protected function where($type = null, $severity = null, $filter = null, $criteria = 'AND'): array
+    protected function where($type = null, $severity = null, $severity_min = null, $filter = null, $criteria = 'AND'): array
```
weitzman: Lets move severity_min to the end just in case someone is calling this method from custom code.

weitzman (changes requested):
LGTM. One minor adjustment needed. 

I think you canadd a few lines to WatchdogTest and get decent test coverage for the new option. Thanks.

src/Drupal/Commands/core/WatchdogCommands.php
jonathan1055: Yes, done.

jonathan1055:
The additional part of the watchdog test passes on mysql and postgres but fails on sqlite. The `wd-delete` using text filter did not delete the row.
https://app.circleci.com/pipelines/github/drush-ops/drush/4307/workflows/62071fed-aeb5-466a-b5f9-88e27532c403/jobs/25911

jonathan1055:
I could remove that `wd-delete` and the tests might pass on sqlite

tests/integration/WatchdogTest.php
```diff
@@ -37,6 +37,39 @@ public function testWatchdog()
         $output = $this->getOutput();
         $this->assertGreaterThanOrEqual($message_chars, substr_count($output, $char));
 
+        // Test deleting a watchdog message by filtering on text.
```
weitzman: Keeping this commented out seems fine to me, for now. Would it make sense to move down to "// Tests message deletion" section?

jonathan1055: ha ha, I was deleting it there, early, for convenience when developing the `wd-show` assertions, to reduce the screen output by removing that long message.  I wanted to keep the original notice message.

jonathan1055: I'll tidy up the tests, and have all the deletes in one place

jonathan1055: Decided to expand the delete testing and split the function into two.

jonathan1055: The debug output could be removed if we don't want that. But I thought it might illuminate something about why the tests fails on some builds.

jonathan1055:
In the mysql deletion test, after removing all messsges and adding the new, we have
````
>>>>> 1
---- -------------- ------- ---------- ------------------ 
  ID   Date           Type    Severity   Message           
 ---- -------------- ------- ---------- ------------------ 
  5    09/Nov 14:41   drush   Critical   *** Exterminate!  
  4    09/Nov 14:41   drush   Error      Obliterate        
  3    09/Nov 14:41   drush   Warning    Eliminate         
  2    09/Nov 14:41   drush   Notice     Delete            
  1    09/Nov 14:41   other   Info       Scrub             
 ---- -------------- ------- ---------- ------------------
````
But in the postgres and sqlite tests the ids continue from the previous test (even though those messages have been deleted). We see
````
>>>>> 1
---- -------------- ------- ---------- ------------------ 
  ID   Date           Type    Severity   Message           
 ---- -------------- ------- ---------- ------------------ 
  10   09/Nov 14:41   drush   Critical   *** Exterminate!  
  9    09/Nov 14:41   drush   Error      Obliterate        
  8    09/Nov 14:41   drush   Warning    Eliminate         
  7    09/Nov 14:41   drush   Notice     Delete            
  6    09/Nov 14:41   other   Info       Scrub             
 ---- -------------- ------- ---------- ------------------
````
So I think testing 'delete by id' is not going to be worth the effort.

jonathan1055:
The test build has not been triggered. Maybe it was because I did a force-push? I thought I had done that before and it was all OK.

jonathan1055:
The new integration tests pass in all builds. The failing section (same as before, rw for dev/tty) is the functional test in D10, which runs phpunit 9.5.26, whereas the others are all 9.5.25, I don't know if that is significant.

weitzman:
Test failure is unrelated. Thank you!

jonathan1055:
Thanks. That's my first committed contribution to drush :-)

PR #5307, opened 2022-11-10
Require latest annotated-command

weitzman:
Bump so that commands written with new-style option params are functional on all sites.


greg-1-anderson (approved):
Good idea

PR #5310, opened 2022-11-11
Remove unused param and only set tty when we can

weitzman:
Fixes a Drupal 10 test failure "TTY mode requires /dev/tty to be read/writable.". Example https://app.circleci.com/pipelines/github/drush-ops/drush/4320/workflows/70530dae-ef77-43cd-9fe9-d8f9713a71fa/jobs/25996/tests#failed-test-0

### DIFF
--- a/src/Commands/core/SshCommands.php
+++ b/src/Commands/core/SshCommands.php
@@ -51,7 +51,6 @@ class SshCommands extends DrushCommands implements SiteAliasManagerAwareInterfac
 
         $process = $this->processManager()->siteProcess($alias, $code);
         if (Tty::isTtySupported()) {
-            // @todo is this is still needed?
             $process->setTty($options['tty']);
         }
         // The transport handles the chdir during processArgs().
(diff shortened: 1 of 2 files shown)


src/Commands/core/SshCommands.php
```diff
@@ -49,7 +50,10 @@ public function ssh(array $code, $options = ['cd' => self::REQ, 'tty' => false])
         }
 
         $process = $this->processManager()->siteProcess($alias, $code);
-        $process->setTty($options['tty']);
+        if (Tty::isTtySupported()) {
```
greg-1-anderson, line 53: We still need to allow the user to select whether a TTY is desired or not. The other end of the connection might not support TTYs. I think that the `tty` CLI option should be preserved, and the `$process->setTty($options['tty']);` should merely be wrapped in `if (Tty::isTtySupported()) {`

greg-1-anderson (changes requested):
Please retain tty flag

src/Commands/core/SshCommands.php
weitzman: Ah right. I pushed a commit which does this. 

The $options for ssh method did not need to declare 'tty' key since thats done by `@optionset_proc_build`

PR #5311, opened 2022-11-11
Add drush command to clear the locale status

rolfvandekrol:
Updating modules does not automatically clear the locale status. This causes problems in our automatic deployments, because that means the `locale:update` command will sometimes try to use `.po` files for older versions of modules. 

We use `drupal-composer/drupal-l10n` to include the correct `.po` files in our build and don't allow downloading the `.po` files inside the production container. 

When we clear the locale status before running `locale:update`, it always uses the correct files.

We could potentially add this to our own scripts, but we think this is also useful for other Drush users.

### DIFF
--- a/src/Drupal/Commands/core/LocaleCommands.php
+++ b/src/Drupal/Commands/core/LocaleCommands.php
@@ -81,6 +81,18 @@ class LocaleCommands extends DrushCommands
     }
 
     /**
+     * Clears the translation status.
+     *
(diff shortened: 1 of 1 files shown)


Sutharsan (approved):
Code looks good. I understand the use case and agree that it is beneficial for others. I have objection to add this command.

Sutharsan:
@weitzman I'm fine with adding this to 11.x.

PR #5312, opened 2022-11-12
Document how to specify drush.services.yml files in package sub-modules

### LINKED ISSUE #5297: How to declare multiple drush.services.yml files in one composer.json
**Existing document**

https://www.drush.org/latest/commands/#specifying-the-services-file

**What are you attempting to do**

In some cases, a single project with a single `composer.json` file may have multiple sub-modules, each providing their own Drush commands and `drush.services.yml` files. How should this be handled in the project's root `composer.json` file?

Two scenarios where this happens:

1. Modules that have sub-modules.
2. Distributions that include modules.

Specifically, I am curious how to solve this for (2) in the farmOS distribution, where we have two different modules that provide Drush commands:

- https://github.com/farmOS/farmOS/blob/2.x/modules/core/migrate/drush.services.yml
- https://github.com/farmOS/farmOS/blob/2.x/modules/core/update/drush.services.yml

And here is the distro's `composer.json`: https://github.com/farmOS/farmOS/blob/2.x/composer.json

**In what way is the existing documentation unclear or incomplete**

It is unclear how to handle the above situation.

**What should the documentation say instead?**

TBD (and happy to open a PR once I understand!)

mstenta:
Fixes #5297

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -81,6 +81,8 @@ In this example, the file `drush-11-99.services.yml` loads commandfile classes t
 
 It is also possible to use [version ranges](https://getcomposer.org/doc/articles/versions.md#version-range) to exactly specify which version of Drush the services file should be used with (e.g. `"drush.services.yml": ">=11 <11.99"`).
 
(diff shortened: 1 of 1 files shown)


weitzman:
Thank you.

PR #5325, opened 2022-11-29
Use new process plugin name in generator help

weitzman:


greg-1-anderson (approved):
I believe you, but I don't have any specific knowledge about this new plugin.

weitzman:
This went in elsewhere

PR #5326, opened 2022-11-29
Update role examples to use single quotes

### LINKED ISSUE #5324: user:create and user:password don't recognize password starting with ! character
**Describe the bug**
The bug appears when creating a user with user:create, but also replacing the password with user:password does not work if the password starts with a characteristic dollar ('$'). For example, the password '$foobar12345', a normally acceptable password, does not work. Conversely, "£foobar12345" or ".foobar12345" or apparently any other first character, seems to work. 
Perhaps Drush tries to interpret the parameter as a variable? 

**To Reproduce**
drush user:create your.user --mail="your.user@email.com" --password="$foobar12345" 

**Expected behavior**
The password should be correctly registered.

**Actual behavior**
The password is not assigned or changed.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 11.2.1
| Drupal version? | 9.x
| PHP version     | 8.x
| OS?             | Linux

gitressa:
Fixes: #5324.

## Problem

The `user:password` example uses double quotes:

```
$ drush user:password -h
Set the password for the user account with the specified name.

Examples:
 drush user:password someuser "correct horse battery staple"
```

... but you cannot use double quotes to create passwords containing special character (for example `f!oobar12345`) but have to use single quotes:

> Strings in single quotes are not expanded by bash.

Otherwise you get an error:

```
$ drush user:password your.user "!foobar12345"
bash: !foobar12345: event not found
```

From https://superuser.com/questions/133780/in-bash-how-do-i-escape-an-exclamation-mark

## Solution

Use single quotes.

To stay consistent, I updated the other instances in examples to use single quotes as well. Also, I noticed that a few examples were not correct, so I updated them as well.

### DIFF
--- a/src/Drupal/Commands/core/UserCommands.php
+++ b/src/Drupal/Commands/core/UserCommands.php
@@ -168,7 +168,7 @@ class UserCommands extends DrushCommands
      * @option $uid A comma delimited list of user ids to lookup (an alternative to names).
      * @option $mail A comma delimited list of emails to lookup (an alternative to names).
      * @aliases urol,user-add-role
-     * @usage drush user-add-role "editor" user3
+     * @usage drush user-add-role 'editor' user3
      *   Add the editor role to user3
      */
     public function addRole(string $role, string $names = '', $options = ['uid' => self::REQ, 'mail' => self::REQ]): void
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks!

gitressa:
Perfect, thanks for a fast response!

PR #5327, opened 2022-12-02
Invoke hook_requirements_alter when hook_requirements is invoked

gaelg:
Should be self-explanatory. See https://www.drupal.org/project/drupal/issues/309040 and https://www.drupal.org/project/drupal/issues/3324995


weitzman:
i notice that updatedb command also checks requirements by calling drupal core's update_check_requirements(). That one doesnt call an alter hook. is that a core bug?

claudiu-cristea:
@weitzman, yes, see https://www.drupal.org/project/drupal/issues/3324995

PR #5328, opened 2022-12-05
Allow to specify/override ssh default-command

ceesgeene:
To be able to use `drush ssh` without arguments on hosts that don´t have bash installed, this PR allows to set the default command via config.


greg-1-anderson (commented):
Didn't notice this was assigned to me. Looks useful. It needs a rebase onto the 12.x branch at this point, though.

Issue #5332, opened 2022-12-13
[error] Deprecated function: explode(): Passing null to parameter #2 on drush status


guix77:
**Describe the bug**

```
 [error]  Message: /Deprecated function/: explode(): Passing null to parameter #2 ($string) of 
type string is deprecated in 
/Drush\Commands\core\StatusCommands::pathAliases()/ (line /231/ of 
//var/www/html/vendor/drush/drush/src/Commands/core/StatusCommands.php/). 
Drush\Commands\core\StatusCommands::pathAliases(Array, Object, Object) (Line: 
152)
Drush\Commands\core\StatusCommands->getPropertyList(Array) (Line: 82)
Drush\Commands\core\StatusCommands->status('', Array)
call_user_func_array(Array, Array) (Line: 257)
Consolidation\AnnotatedCommand\CommandProcessor->runCommandCallback(Array, 
Object) (Line: 212)
Consolidation\AnnotatedCommand\CommandProcessor->validateRunAndAlter(Array, 
Array, Object) (Line: 176)
Consolidation\AnnotatedCommand\CommandProcessor->process(Object, Array, 
Array, Object) (Line: 390)
Consolidation\AnnotatedCommand\AnnotatedCommand->execute(Object, Object) 
(Line: 255)
Symfony\Component\Console\Command\Command->run(Object, Object) (Line: 1039)
Symfony\Component\Console\Application->doRunCommand(Object, Object, Object) 
(Line: 275)
Symfony\Component\Console\Application->doRun(Object, Object) (Line: 149)
Symfony\Component\Console\Application->run(Object, Object) (Line: 124)
Drush\Runtime\Runtime->doRun(Array, Object) (Line: 51)
Drush\Runtime\Runtime->run(Array) (Line: 77)
require('/var/www/html/vendor/drush/drush/drush.php') (Line: 18)
drush_main() (Line: 143)
require('phar:///usr/local/bin/drush/bin/drush.php') (Line: 14)
```

**To Reproduce**

`drush status`

**Expected behavior**

No error

**Actual behavior**

Error

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 11.3.2
| Drupal version? | 9.4.9
| PHP version     | 8.1.13
| OS?             | Docker4Drupal => drupal-php container version 8.1-dev-4.38.5

weitzman:
Fixed in #5333

PR #5333, opened 2022-12-13
#5332 fix drush status

guix77:


weitzman:
thanks. I pushed a commit to use our helper instead.

guix77:
Thanks !

PR #5337, opened 2022-12-13
Relay update warnings to the user.

### LINKED ISSUE #5335: Inform the user when database updates have warnings
It is possible that a database update from a module cannot be executed. For example if the module's schema version is from a previous version of Drupal, or when a module uses the update hook number `_8000` which is reserved for the major D7 > D8 update.

In this case `update_get_update_list()` will return a warning for the offending module, rather than the usual `start` key that indicates the first update that can be executed.

In the Drupal core `update.php` page these warnings are shown to the end user before they start the update procedure:

image

We should probably do the same.

On a side note, when a module update has a warning, the missing "start" array key throws a warning:

> [warning] Undefined array key "start" UpdateDBCommands.php:456

Show the warnings to the user. The best solution would be to show them as part of the `drush update:status` command. The output of this command is shown before the user is asked for confirmation to execute the updates.

pfrenssen:
Fixes #5335

**How to test:**

Add the following line to `core/modules/node/node.install`:

```
function node_update_8000() {}
```

Run the `drush updatedb:status` command. A warning should be shown:

```
$ drush updatedb:status
 [warning] node: <em>node</em> module cannot be updated. It contains an update numbered as 8000 which is reserved for the earliest installation of a module in Drupal 8.x, before any updates. In order to update <em>node</em> module, you will need to install a version of the module with valid updates.
 [success] No database updates required.
```

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -104,7 +104,12 @@ class UpdateDBCommands extends DrushCommands implements SiteAliasManagerAwareInt
     {
         require_once DRUSH_DRUPAL_CORE . '/includes/install.inc';
         drupal_load_updates();
-        list($pending, $start) = $this->getUpdatedbStatus($options);
+        list($pending, $start, $warnings) = $this->getUpdatedbStatus($options);
+
+        // Output any warnings.
+        foreach ($warnings as $module => $warning) {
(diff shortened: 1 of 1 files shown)


weitzman (approved):
LGTM. Up to you if you want to make a test for this.

pfrenssen:
Thanks! This is an edge case which will only affect a minority of users, so I think a test does not bring much value.

It only happens if a developer makes a mistake in the naming of an update hook. Still a good idea to show the warnings in this case :)

Issue #5009, opened 2022-01-11
drush_backend_batch_process() and &$context


ultimike:
I have a batch that runs fine via a form - with the $context being populated and `$context['results']` available to my "finished" callback as expected.

I've created a custom Drush command for this batch, but when I run it with `drush_backend_batch_process();`, the `$results` array isn't populated as I expect in my "finished" callback. Everything else is working fine - the custom Drush command runs the batch with the expected results, other than the `$results` array being provided to the "finished" callback. 

Is this a known limitation of `drush_backend_batch_process();` or am I just missing something?

thanks,  
-mike

halisonfernandes:
Same issue here. I'm on 10.6.1 Drush version.

What I've noticed is that after the first interaction, it's loosing the `$context['results']`.

weitzman:
Sounds like a bug to me. Not expected.

omahm:
Getting the same here on 10.6.2. It's not passing the context between operations and the finish callback.

weitzman:
FYI Drush 10 is no longer supported https://www.drush.org/latest/install/#drupal-compatibility. @ultimike do you recall what version of Drush you were using?

omahm:
Thanks @weitzman, I upgraded to 11.0.9 and the issue still persists.

ultimike:
@weitzman When I posted this issue, I was running 11.0.5.

I have since updated to 11.0.9 and the problem persists 🙁

-mike

weitzman:
I have added test coverage which shows that $results is populated. $results is the 2nd param in a finished callback. See https://github.com/drush-ops/drush/pull/5144

Not sure why others are seeing differently. Can you share your batch code? FWIW I looked at https://git.drupalcode.org/project/examples/-/blob/3.x/modules/batch_example/batch_example.module to see how batches are implemented.

ultimike:
@weitzman Here's the relevant (I believe) code (some names have been changed to protect the innocent):

```
    // Create a Batch API item for each user.
    foreach ($users as $uid => $user) {
      $operations[] = ['my_module_batch_operation', [$uid]];
    }
    $batch = [
      'operations' => $operations,
      'finished' => 'my_module_finished',
      'file' => $this->extensionListModule->getPath('my_module') . '/my_module.batch.inc',
    ];
    batch_set($batch);
    if ($drush) {
      // For some reason, running via drush leaves the $results array in
      // my_module_finished() empty. See
[...]

PR #5354, opened 2023-01-10
Remove DrushBatchContext magic wrapper for batch .

halisonfernandes:
It should fix this issue #5009 

Tests are passing.

**Summary:**
The DrushBatchContext wrapper is causing issues to keep the batch $context values. It's mainly used for logging, but it can be done in the operations level instead of overriding the $context type.
The variable `$context['error_message']` is only being used by the drush batch processes and we can use `Drush::logger()->error();` as a replacement.
After this change, we can use the `$task_message` variable to retrieve the `$context['message']` value and log it to the user.

I tested the `updb` in one of my sites and this is the output:
```
 ----------------------- ------------------------------- --------------- ------------------------------------------------------------------------------ 
  Module                  Update ID                       Type            Description                                                                   
 ----------------------- ------------------------------- --------------- ------------------------------------------------------------------------------ 
  lightning_media_image   9001                            hook_update_n   9001 - Downloads the Cropper JavaScript library if needed.                    
  block_content           entity_changed_constraint       post-update     Clear the entity type cache.                                                  
  ctools                  remove_entitybundleconstraint   post-update     Invalidate the service container to force EntityBundleConstriant is Removed.  
  webform                 ckeditor                        post-update     Move from custom CKEditor to hidden 'webform_default' text format.            
 ----------------------- ------------------------------- --------------- ------------------------------------------------------------------------------ 


 Do you wish to run the specified pending updates? (yes/no) [yes]:
 > 

>  [notice] Update started: lightning_media_image_update_9001
>  [notice] Update completed: lightning_media_image_update_9001
>  [notice] Update started: block_content_post_update_entity_changed_constraint
>  [notice] Update completed: block_content_post_update_entity_changed_constraint
>  [notice] Update started: ctools_post_update_remove_entitybundleconstraint
>  [notice] Update completed: ctools_post_update_remove_entitybundleconstraint
>  [notice] Update started: webform_post_update_ckeditor
>  [notice] Update completed: webform_post_update_ckeditor
 [success] Finished performing updates.
```


weitzman:
This is ancient code and I'm a bit hesitant to remove it as I dont grok it well. It went in 13 years ago at https://www.drupal.org/node/1186480. Alas @jonhattan hasn't been active here for years.

Thanks for the deep dive. I've merged this into 12.x, since the tests suggest its safe. I'm inclined to leave 11.x alone.

jonhattan:
That code was an improvement introduced in the Drupal 6/7 + PHP 5.x epoch. A lot has changed since, and the regression may by anywhere, or perhaps the code had a bug that didn't manifest until recent.

FWIW The goal was to log messages generated within the operations  (ie. `Indexing item i of n`) rather than between operations. I guess this is now lost - not the worst thing.

halisonfernandes:
Thanks for merging.

@jonhattan the main difference is that now it's not possible to log messages in the middle of the operation using the `$context['message']` object, but the message will still be logged at the end of it. The recommendation would be to use the logger object or simply change the text to be `Indexed i item of n` instead of `Indexing item i of n` as it will be logged at the end of the operation only.

Having a quick look at the `search_api` module, it's doing it already:
```
      // Display progress message.
      if ($indexed > 0) {
        $context['message'] = static::formatPlural($context['results']['indexed'], 'Successfully indexed 1 item on @index.', 'Successfully indexed @count items on @index.', ['@index' => $index->label()]);
      }
```

PR #5358, opened 2023-01-19
Separate using git from isDecorated.

charginghawk:
This will make the output consistent between interactive and non-interactive sessions.

On Acquia servers, `drush cex --diff --no` outputs a single-line `Only in [directory]` when a config file is fully added or deleted, where `drush cex --diff` will show the full diff, ie the contents, of the file.


weitzman:
Thanks. Sorry for the delay.

Issue #4685, opened 2021-03-10
Stop using Drupal container for contrib command discovery and instantiation


weitzman:
Use Factory method ...

greg-1-anderson:
The general idea here is that we find command classes by discovery, like we used to do (and still do for site-wide commands). We could also consider discovering command classes via the autoloader, as Robo plugins do.

Once we have a list of command classes, we instantiate them via a static `create` factory method (that must exist? or only if it exists?). The create method is passed a DI container, and it can use that to fetch and inject its dependencies into the commandfile class.

League/container supports chained DI containers. Symfony DI did not; I'm not sure if it does now, but I suspect we might be able to chain from a League DI container to a Symfony (Drupal) DI container. If not, we might need to pass two DI containers to the create method.

weitzman:
Looks like neither the Symfony container nor the Drupal container (a close fork of Symfony) support delegate() so we'll need to pass two containers I think.

greg-1-anderson:
[Symfony containers](https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/DependencyInjection/Container.php) implement ContainerInterface, which is the requirement for deletage() in league / container, so I think that maybe we could delegate to the Drupal container from the Drush container.

We should only do this if we change our bootstrap process to do the delegation; if we don't do that, then we should pass two containers.

Also, just because we can pass just one container doesn't mean we should. I'm undecided on this point.

weitzman:
Could we do this in a backward compat way? Seems like we could mimic service instantiation so they continue to work?

greg-1-anderson:
Hm, yeah, that might work.

weitzman:
A bit more discussion about motivation for this (or lack thereof) https://drupal.slack.com/archives/C62H9CWQM/p1633443431221800?thread_ts=1633433650.221500&cid=C62H9CWQM

joachim-n:
It would be really nice if Drush looked for commands in all installed packages, not just modules.

There's a comment in DrupalKernelTrait::discoverServiceProviders() about this.

I think it's doable by instantiating Composer and getting a list of packages from it, but is the command list cached at all? Getting a list of packages out of Composer every single time is going to be a bit slow.

Seldaek:
If you need a list of packages, please use https://getcomposer.org/doc/07-runtime.md#installed-versions - there's also `getInstalledPackagesByType` which could be useful for this perhaps if the packages have defined types.

[...]

PR #5371, opened 2023-01-30
Upgrade to drupal-code-generator v3

Chi-teck:
Drush currently uses DCG 2 to generate code.

### Remaining tasks
- [x] Update `drush generate` command to use PHP attributes instead of annotations
- [x] Update `drush:command-file` generator
- [x] Figure out a way to register third-party generators
- [x] Update Drush tests
- [ ] Implement completion for generator names

The big question is how to deliver this change. Here are a couple complications.
1. DCG 3 is not API compatible with DCG 2. 
2. DCG 3 requires PHP 8.1 and Drupal 10.

Basically there are two options.
1. Wait for a new major Drush release that would only support Drupal 10.
2. Make `drush generate` compatible with both DCG versions.


Chi-teck:
This kind of critical, because Drupal 10 has been already released and code generated by DCG 2 is not fully compatible with it. In order to support multiple Drupal versions DCG would have to generate code with deprecations .

weitzman:
I dont feel strongly about it, but I think 1. is the best way forward. We could branch for Drush 12 now, with a target release date just before Drupalcon Pittsburgh (early June). Drush 12 would be Drupal 10 minimum. Here is the todo list for Drush 12 https://github.com/drush-ops/drush/milestone/7

weitzman:
_Update: ignore below. I had merged composer changes into 11.x by accident._

I changed the base of this PR to 12.x It says it has no merge conflicts but I dont see how composer.lock could not conflict. Please reroll if needed when you get a chance.

Chi-teck:
> Figure out a way to register third-party generators

Modules can subscribe to generator info event and register their generators. I am not sure if this is enough to resolve this task.

src/Commands/generate/GenerateCommands.php
```diff
@@ -20,40 +25,42 @@ class GenerateCommands extends DrushCommands implements AutoloaderAwareInterface
...
     {
+        // @todo Figure out a way to inject the container.
+        $container = \Drupal::getContainer();
+
+        // @todo Implement discovery for third-party generators.
+        $application = Application::create($container);
+        $application->setAutoExit(FALSE);
+
+        $application->addCommands([
+            new MigrationGenerator(),
```
weitzman: Any interest in moving Migration generator in DCG? If not thats , OK.

weitzman (commented):
How does a global generator subscribe to the event? Here is what we say about authoring global generators today https://www.drush.org/11.x/generators/#global-generators

src/Commands/generate/GenerateCommands.php
Chi-teck: DCG already have a generator for MigrationSource plugins. We could add a simple migration generator under Yaml namespace.

Chi-teck:
> How does a global generator subscribe to the event? Here is what we say about authoring global generators today https://www.drush.org/11.x/generators/#global-generators

Those need to be registered in the generate command before `$application->run();`

src/Commands/generate/GenerateCommands.php
Chi-teck: https://github.com/Chi-teck/drupal-code-generator/commit/6347e6ffea446194e37a8fb2db9067d29b0bd676

weitzman:
I updated the dcf generator a bit in https://github.com/drush-ops/drush/pull/5442. Still needs port to DCG3.

weitzman:
For global generators, is there any reason we can't keep using the code in `discoverPsr4Generators` - https://github.com/drush-ops/drush/pull/5371/files#diff-14a8581695da2d03094b84103528a480af8e2891e0590de222aa9b04a8d4d7f2L137-L144. That can be called right from the command method.

Chi-teck:
> I updated the dcf generator a bit in #5442. Still needs port to DCG3.

The dcf command has been updated to DCG3. Needs review.

Chi-teck:
> For global generators, is there any reason we can't keep using the code in `discoverPsr4Generators` - https://github.com/drush-ops/drush/pull/5371/files#diff-14a8581695da2d03094b84103528a480af8e2891e0590de222aa9b04a8d4d7f2L137-L144. That can be called right from the command method.

The link seems broken. Anyway we can use any discovery method as long as it can ignore DCG 2 generators. I would prefer something less magic, something that does not rely on reflection API and loading generator classes. That would allow to avoid difficulties in future major updates of DCG and symfony/console.

weitzman:
The diff wont show because it is inside the file thats deleted in the PR.

IMO we should merge this without major changes to generator discovery in drush (both global and module). Then we work on those in #4685

Chi-teck:
I've removed `discoverDrushGenerators`. I don't think it's worth implementing a discovery method for just a couple of generators. Also replaced `adjustDCGGenerators` with event subscriber.

Issue #5375, opened 2023-02-07
user:cancel doesn't support user_cancel_reassign method by default


stratulat-vladislav:
**Describe the bug**
Drush has **user:block** method which blocks the user(s). It also has **user:cancel** method which also blocks the user(s) if **--delete-content** option is not specified. I believe it should cancel account and reassign the content instead.

**To Reproduce**
`drush user:cancel username`

**Expected behavior**
>  [notice] Deleted user: username <username@example.com >.
>  [notice] Message: Account /username/ has been deleted.

**Actual behavior**
>  [notice] Blocked user: username <username@example.com>.
>  [notice] Message: Account /username/ has been disabled.

**Workaround**
No

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x
| Drupal version? | 9.x
| PHP version     | 8.x
| OS?             | Linux

stratulat-vladislav:
I have created 10.x PR here https://github.com/drush-ops/drush/pull/5376

stratulat-vladislav:
New PR for 11.x version here https://github.com/drush-ops/drush/pull/5380

stratulat-vladislav:
New PRs for 9.x https://github.com/drush-ops/drush/pull/5381 and 9.5.x https://github.com/drush-ops/drush/pull/5382 versions

stratulat-vladislav:
Closing this since Drush 11 was updated while older versions are no longer supported.

PR #5376, opened 2023-02-07
#5375 user:cancel to support user_cancel_reassign method by default

stratulat-vladislav:


weitzman:
I think this command description should be "Block user account and optionally delete or reassign its content". Lets add a new boolean option for reassign. Its description should say that content will belong to the Anonymous user.

stratulat-vladislav:
Why do we have to mention "Block" on **cancel** command while we have user:block command for this task?

stratulat-vladislav:
The documentation says: **Cancel the user account with the name username and anonymize all content created by that user.**, but it doesn't do that.
https://www.drush.org/latest/commands/user_cancel/

weitzman:
Because thats how it has historically worked and I'm not inclined to surprise folks with a behavior change.

stratulat-vladislav:
> I think this command description should be "Block user account and optionally delete or reassign its content". Lets add a new boolean option for reassign. Its description should say that content will belong to the Anonymous user.

Hi @weitzman. I have pushed further changes. Can you please have a look?

src/Drupal/Commands/core/UserCommands.php
```diff
@@ -121,7 +121,11 @@ public function renderRolesCell($key, $cellData, FormatterOptions $options)
      * @param string $names A comma delimited list of user names.
      * @aliases ublk,user-block
      * @usage drush user:block user3
-     *   Block the users whose name is user3
+     *   Block the user whose name is <info>user3</info>
```
stratulat-vladislav, line 124: Hi @weitzman. I have updated help text and usage examples as you did for 11.x branch.

weitzman:
Unfortunately, the 10.x branch is EOL as well. Tests have broken it seems so it wouldnt be prudent to merge this. Thanks for working on older versions of Drush.

PR #5379, opened 2023-02-14
Fix parameter name in ConfigCommands::__constructor()

Chi-teck:
This currently triggers the following notice on PHP 8.2.
> Deprecated function: Creation of dynamic property Drush\Drupal\Commands\config\ConfigCommands::$configStorage is deprecated in Drush\Drupal\Commands\config\ConfigCommands->__construct()

### DIFF
--- a/src/Drupal/Commands/config/ConfigCommands.php
+++ b/src/Drupal/Commands/config/ConfigCommands.php
@@ -84,7 +84,7 @@ class ConfigCommands extends DrushCommands implements StdinAwareInterface, SiteA
         if (isset($this->configStorageExport)) {
             return $this->configStorageExport;
         }
-        return $this->configStorage;
+        return $this->configStorageExport;
     }
 
     public function setImportTransformer(ImportStorageTransformer $importStorageTransformer): void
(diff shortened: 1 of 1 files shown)


Chi-teck:
Well, I am not actually certain if the provided fix is correct. Seems like `configStorage` and `configStorageExport` are used independently.
Here is the PR where this change was introduced #4201.

weitzman:
@bircher Thoughts?

bircher:
No this is not correct. A more correct fix would be to explicitly define `protected $configStorage`. The export storage is set via an optional setter (for BC with D8.7), and the one from the constructor is the active storage. If we don't need to support Drupal 8.7 then we could use both in the constructor and clean up all of the config command classes to be more consistent

weitzman:
Thanks. We dropped Drupal 8 compat with Drush 11 https://www.drush.org/latest/install/#drupal-compatibility. I'll close this and hopefully someone will start a new PR with cleanup as suggested.

andypost:
++ to commit only https://github.com/drush-ops/drush/pull/5379/commits/5a4a8c17f4015f2c340c04fdc9814675b4d8c2d3

PR #5380, opened 2023-02-14

stratulat-vladislav:


weitzman:
@stratulat-vladislav  do my help updates look good to you?

stratulat-vladislav:
@weitzman, absolutely!

src/Drupal/Commands/core/UserCommands.php
```diff
@@ -123,7 +123,11 @@ public function renderRolesCell($key, $cellData, FormatterOptions $options)
      * @option $mail A comma delimited list of emails to lookup (an alternative to names).
      * @aliases ublk,user-block
      * @usage drush user:block user3
-     *   Block the users whose name is user3
+     *   Block the user whose name is user3
+     * @usage drush user:cancel user3 --delete-content
```
stratulat-vladislav, line 127: Is this the right place to reference `user:cancel` method?

weitzman: Its a bit unusual, but I think it helps more than it hurts. Thats how discovery works. You are looking in one place and you discover you should be looking elsewhere.

PR #5381, opened 2023-02-14

stratulat-vladislav:


weitzman:
[Drush 9 is EOL](https://www.drush.org/latest/install/#drupal-compatibility). Drush 8 is supported but not recommended. At this point, its urgent fixes only. Lets close this and focus on 11.x

Issue #3782, opened 2018-11-13
Cron runs even if the site is in maintenance mode


defr:
**Describe the bug**

When running cron through the core supported way (cron.php in D7, /cron/{token} in D8), a check is in place to prevent running cron while the site is in maintenance mode and potentially going important schema changes (running database updates for example)

There's no such prevention when running through ```drush cron``` , which means that a cron might trigger at just the wrong moment and breaks the site.

**To Reproduce**

* ```drush state-set system.maintenance-mode 1```
* ```drush cron```

**Expected behavior**

Message stating that the site is in maintenance mode and cron run was aborted

**Actual behavior**

```drush cron``` returns a "Cron run successfully" (in the best case)

**Workaround**

Manually disabling cron before deployment, and re-enabling them afterwards is a viable workaround, but that's a lot less convenient.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 9.5.2
| Drupal version? | 8.6
| PHP version     | 7.2
| OS?             | Linux

weitzman:
Thats out of scope IMO. If you don't want Drush to run during maint mode, don't call it during maint mode. That is, adjust your cron script to check maint and not call cron as needed.

greg-1-anderson:
We could maybe put a policy file in the examples directory for folks to copy.

weitzman:
Having said that, I'm open to discussion on this.

defr:
After some more testing, if you enable Automated Cron in core, it'll launch for anonymous users for receive the maintenance page too, so putting an explicit state check outside of drush might be the right call ; a mention of this in docs/cron.md would be nice.

Sidenote: The not so easy to automate case is in Drush 8, where there might be a single ```drush @sites cron --quite``` call for all the sites in a multi site setup. With that alias being gone in drush 9, checking the maintenance mode status in each cron job isn't that much of an overhead.

weitzman:
Please open a docs change in a PR if desired.

weitzman:
Drupal core is planning to implement this https://www.drupal.org/project/drupal/issues/3318964

Until then, I've added new Drush commands to facilitate this in #5383. You can now do `drush maint:status && drush cron`. This will fail and not run cron if maint mode is enabled.

PR #5383, opened 2023-02-15
Add maint mode commands

weitzman:
Helps address https://github.com/drush-ops/drush/issues/3782 and https://drupal.slack.com/archives/C45SW3FLM/p1675287662331809 (Cron jobs that run while maint mode is on can break deployments, especially field add/edit during config import)

Example usages:
- `drush maint:status && drush purge`
- `drush maint:status && drush queue:run scheduled_transition_job`

### DIFF
--- a/docs/cron.md
+++ b/docs/cron.md
@@ -1,14 +1,14 @@
 Running Drupal cron tasks from Drush
 ====================================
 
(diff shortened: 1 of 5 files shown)


greg-1-anderson (approved):
I like it.

PR #5402, opened 2023-02-27
Attributes - DrupalCommands

weitzman:


src/Drupal/Commands/core/DrupalCommands.php
```diff
@@ -57,39 +42,32 @@ public function __construct(CronInterface $cron, ModuleHandlerInterface $moduleH
...
+    #[CLI\Usage(name: 'drush core:requirements', description: 'Show all status lines from the Status Report admin page.')]
+    #[CLI\Usage(name: 'drush core:requirements --severity=2', description: 'Show only the red lines from the Status Report admin page.')]
+    #[CLI\FieldLabels(labels: [
+        'title' => 'Title',
+        'severity' => 'Severity',
+        'sid' => 'SID',
+        'description' => 'Description',
+        'value' => 'Summary',
+    ])]
+    #[CLI\DefaultFields(fields: ['severity', 'value'])]
```
greg-1-anderson: Was there a reason you took out `title` from the default fields? Another note: in most cases, it's probably better to use `DefaultTableFields` instead of `DefaultFields` now. The former applies only to "human readable" formats, whereas the later applies to all formats. If DefaultFields is omitted (recommended), then formats such as json will default to showing all available information.

weitzman: Fixed. Now using DefaultTableFields everywhere.

PR #5408, opened 2023-02-27
Attributes - QueueCommands

weitzman:
- Adds a HookCustom attribute to register a tag implementation like `@validate-queue`. [Example usage](https://github.com/drush-ops/drush/pull/5408/files#diff-1900bf031c5b584de69f3be1f8f258797b6e815cbc99a0e1f81fcbdfdd5071f7R61). If we dont have this then folks will have to create Attribute classes for seldom used tags. If HookCustom class looks good, I will move it to annotated-command project.
- Usual conversion stuff


src/Attributes/HookCustom.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Drush\Attributes;
+
+use Attribute;
+use Consolidation\AnnotatedCommand\Parser\CommandInfo;
+
+#[Attribute(Attribute::TARGET_METHOD)]
+class HookCustom
```
greg-1-anderson: I would recommend renaming this to HookSelector.

src/Attributes/HookCustom.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    /**
+     * @param $name
+     *  The hook name which is being implemented.
```
greg-1-anderson: Maybe:

> The hook target which is being selected. If a hook has a target that begins with `@`, the
> remainder of the target is the name of the hook selector that must be present for that hook to run.

src/Attributes/HookCustom.php
```diff
@@ -0,0 +1,28 @@
...
+    public function __construct(
+        public string $name,
+        public array $arguments = [],
+    ) {
+    }
+
+    public static function handle(\ReflectionAttribute $attribute, CommandInfo $commandInfo)
+    {
+        $instance = $attribute->newInstance();
+        $commandInfo->addAnnotation($instance->name, implode(' ', $instance->arguments));
```
greg-1-anderson: The old annotation implementation had no choice but to use an untyped top-level annotation as its selector. Now, though, we have a typed attribute, a Hook Selector, so we could and probably should be more discerning. We could add a list of selectors to the commandInfo, and set data there instead of directly in the annotations.

Alternately, we could go the other way, and instead name this `CLI\Annotation`, and make it a general-purpose way to set generic annotations on a CLI attribute. We currently don't have any other use for this facility than a hook selector, but we might someday.

src/Drupal/Commands/core/QueueCommands.php
```diff
@@ -48,15 +52,13 @@ public function getQueueService(): QueueFactory
...
-     * @option time-limit The maximum number of seconds allowed to run the queue.
-     * @option items-limit The maximum number of items allowed to run the queue.
-     * @option lease-time The maximum number of seconds that an item remains claimed.
      */
+    #[CLI\Command(name: 'queue:run', aliases: ['queue-run'])]
+    #[CLI\Argument(name: 'name', description: 'The name of the queue to run, as defined in either hook_queue_info or hook_cron_queue_info.')]
+    #[CLI\Option(name: 'time-limit', description: 'The maximum number of seconds allowed to run the queue.')]
+    #[CLI\Option(name: 'items-limit', description: 'The maximum number of items allowed to run the queue.')]
+    #[CLI\Option(name: 'lease-time', description: 'The maximum number of seconds that an item remains claimed.')]
+    #[CLI\HookCustom(name: self::VALIDATE_QUEUE, arguments: ['name'])]
```
greg-1-anderson: Should we just make the parameters to this attribute be `name` and `argument` rather than `name` and `arguments`? All that we ever do with the arguments is implode with ` `; it seems more straightforward in this case to let the user define the string they want.

src/Attributes/HookCustom.php
greg-1-anderson: These are optional suggestions. The first is only possible in the annotations library itself, as we would have to also change the way hook selection worked. That's probably not a good idea, as it would prevent interoperation between the attribute implementation and the annotation implementation of commands. That may or may not be an issue, but with commands and hooks being defined in multiple modules, it likely would become an issue.

I don't know that the second suggestion is good either; I'm more inclined towards the specific name, unless we specifically think someone might misuse it to set generic annotation values. If we did, then maybe the more generic name might be better after all.

src/Drupal/Commands/core/QueueCommands.php
weitzman: I am just trying to be as native as possible to the command author. The fact that we implode later is a (slightly embarassing) implementation detail.

src/Attributes/HookCustom.php
weitzman: I like the first proposal here. Lets do it at some future date. IMO annotations should be deprecated by this library and removed at some far future date.

src/Drupal/Commands/core/QueueCommands.php
```diff
@@ -48,15 +52,13 @@ public function getQueueService(): QueueFactory
...
-     * @option time-limit The maximum number of seconds allowed to run the queue.
-     * @option items-limit The maximum number of items allowed to run the queue.
-     * @option lease-time The maximum number of seconds that an item remains claimed.
      */
+    #[CLI\Command(name: 'queue:run', aliases: ['queue-run'])]
+    #[CLI\Argument(name: 'name', description: 'The name of the queue to run, as defined in either hook_queue_info or hook_cron_queue_info.')]
+    #[CLI\Option(name: 'time-limit', description: 'The maximum number of seconds allowed to run the queue.')]
+    #[CLI\Option(name: 'items-limit', description: 'The maximum number of items allowed to run the queue.')]
+    #[CLI\Option(name: 'lease-time', description: 'The maximum number of seconds that an item remains claimed.')]
+    #[CLI\HookSelector(name: self::VALIDATE_QUEUE, arguments: ['name'])]
```
greg-1-anderson: Or, revisiting my previous comment on this subject, instead of `arguments: ['name']`, how about `value: 'name'` (and remove the corresponding implode)?

weitzman: Done. See https://github.com/consolidation/annotated-command/pull/290

src/Drupal/Commands/core/QueueCommands.php
weitzman: Done

PR #5409, opened 2023-02-28
Attributes - LocaleCommands

Sutharsan:
Annotation of all locale commands replace by attributes. Results tested manually with `drush list` and `--help `.

I've found two command documentation fixes one on the `locale:import` and one on the `locale:export` command. @weitzman if you are strict on this and don't want to include it in the PR, I will make another.


src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -11,19 +12,20 @@
 use Drupal\Core\State\StateInterface;
 use Drupal\language\Entity\ConfigurableLanguage;
 use Drupal\locale\PoDatabaseReader;
+use Drush\Attributes as CLI;
 use Drush\Commands\DrushCommands;
 use Drush\Exceptions\CommandFailedException;
 use Drush\Utils\StringUtils;
 
 class LocaleCommands extends DrushCommands
```
weitzman:
```suggestion
final class LocaleCommands extends DrushCommands
```

src/Drupal/Commands/core/LocaleCommands.php
```diff
@@ -58,11 +60,9 @@ public function __construct(LanguageManagerInterface $languageManager, ConfigFac
 
     /**
      * Checks for available translation updates.
-     *
-     * @command locale:check
-     * @aliases locale-check
-     * @validate-module-enabled locale
      */
+    #[CLI\Command(name: 'locale:check', aliases: ['locale-check'])]
```
weitzman: As of today, we are creating constants for command names. See https://github.com/drush-ops/drush/pull/5408/files#diff-1900bf031c5b584de69f3be1f8f258797b6e815cbc99a0e1f81fcbdfdd5071f7R21-R24 for an example. In addition to the commandfile, you can use the constants in tests, and site owners can use them in scripts.

weitzman (changes requested):
Great work! No problem with additional improvements sneaking in. Here are a couple minor change requests.

The Postgres failures are unrelated to this PR.

src/Drupal/Commands/core/LocaleCommands.php
weitzman: As per #5283

PR #5431, opened 2023-03-05
Attributes - MigrateRunnerCommands

claudiu-cristea:


claudiu-cristea:
The PostgreSQL failure occurs also in `12.x`

src/Drupal/Migrate/ValidateMigrationId.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Drush\Drupal\Migrate;
+
+use Attribute;
+use Consolidation\AnnotatedCommand\Parser\CommandInfo;
+
+#[Attribute(Attribute::TARGET_METHOD)]
+class ValidateMigrationId
```
weitzman: You can now extend NoArgumentBase to save a little code. See src/Attributes/OptionsetProcBuild.php for an example

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -57,64 +45,53 @@ class MigrateRunnerCommands extends DrushCommands
...
+     * Lists all migrations with current status.
      *
      * @throws \Exception
      *   If --names-only is used with --field having other value than 'id'.
      */
+    #[CLI\Command(name: 'migrate:status', aliases: ['ms', 'migrate-status'])]
+    #[CLI\Argument(name: 'migrationIds', description: 'Restrict to a comma-separated list of migrations. Optional.')]
+    #[CLI\Option(name: 'tag', description: 'A comma-separated list of migration tags to list. If only <info>--tag</info> is provided, all tagged migrations will be listed, grouped by tags.')]
+    #[CLI\Option(name: 'names-only', description: '[Deprecated, use --field=id instead] Only return names, not all the details (faster).')]
+    #[CLI\Help(description: 'List all migrations with current status.')]
```
weitzman: You can omit this since annotated-command still uses the method description.

weitzman (approved):
Nice work. Just 2 minor comments.

src/Drupal/Commands/core/MigrateRunnerCommands.php
claudiu-cristea: OK, I did. But I don't necessarily agree. I think, now, attributes should be the norm. I would rather remove the method description.

weitzman: Yeah, I think that both approaches are valid. I think keeping the description in Doxygen is slightly more readable and @greg-1-anderson agrees since he made this work in https://github.com/consolidation/annotated-command/pull/281.

weitzman (approved):
LGTM. Lets merge once all tests besides Postgres are passing.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -57,64 +45,52 @@ class MigrateRunnerCommands extends DrushCommands
...
+    #[CLI\FieldLabels(labels: [
+        'id' => 'Migration ID',
+        'status' => 'Status',
+        'total' => 'Total',
+        'imported' => 'Imported',
+        'needing_update' => 'Needing update',
+        'unprocessed' => 'Unprocessed',
+        'last_imported' => 'Last Imported',
+    ])]
+    #[CLI\DefaultFields(fields: [
```
weitzman, line 84:
```suggestion
    #[CLI\DefaultTableFields(fields: [
```
We are switching to DefaultTableFields everywhere. DefaultTableFields only changes the table view whereas other output formats will show all the fields by default.

src/Drupal/Commands/core/MigrateRunnerCommands.php
```diff
@@ -599,47 +527,32 @@ public function resetStatus(string $migrationId): void
...
+    #[CLI\ValidateModulesEnabled(modules: ['migrate'])]
+    #[Migrate\ValidateMigrationId()]
+    #[CLI\FieldLabels(labels: [
+        'level' => 'Level',
+        'source_ids' => 'Source ID(s)',
+        'destination_ids' => 'Destination ID(s)',
+        'message' => 'Message',
+        'hash' => 'Source IDs hash',
+    ])]
+    #[CLI\DefaultFields(fields: [
```
weitzman, line 548:
```suggestion
    #[CLI\DefaultTableFields(fields: [
```

weitzman (approved):

weitzman:
@greg-1-anderson What we are seeing here is that when a command uses DefaultFields attribute, then $options['fields'] is properly populated. When we use DefaultTableFields, $options['fields'] = ''. is this expected? Here is an example command declaration https://github.com/drush-ops/drush/pull/5431/files#diff-56e104b35aa4fe4d54637127e16c583fbd6125688fe31c09d0a040065930faf6R63-R94

weitzman:
Back to DefaultFields until that change is in AC

PR #5439, opened 2023-03-06
Update rector.php to use RectorConfig

DieterHolvoet:


DieterHolvoet:
@weitzman what did this part of the config do exactly?
```php
$services = $containerConfigurator->services();
$services->set(ParamTypeDeclarationRector::class);
$services->set(ReturnTypeDeclarationRector::class);
```

DieterHolvoet:
Why did you specify the `src` folder for the ignored rules? Pretty sure that if you leave that out, the rule is ignored for the whole project. Isn't that what we want?

weitzman:
I didn't put much thought into this. Was just trying to refctor to add return types and method type hints. Feel free to change this any way you think is best.

PR #5441, opened 2023-03-06
Add SetList::PHP_81 & apply suggested changes

DieterHolvoet:


DieterHolvoet:
I reviewed the changes, they seem okay.

weitzman:
Nit so sure about this. Maybe if we remove whatever is doing the widespread cadst to (string) and the `final const`. Dont think those add much value.

DieterHolvoet:
The string casts do provide value: they prevent errors in places where non-string values might be passed to built in functions that gained type hints in the last PHP versions. But there might be better ways to prevent these issues than the string casts.

And since we're making command classes final in 12.x, why not continue the theme and make these constants final as well? It doesn't seem like these should be able to be overridden in child classes anyway. But I don't care as much about these, if you don't trust them I'll remove these changes.

weitzman:
> The string casts do provide value: they prevent errors in places where non-string values might be passed to built in functions that gained type hints in the last PHP versions. But there might be better ways to prevent these issues than the string casts.

IMO these errors are helpful. They show the dev that she has been sloppy in the past, and should fix code to be more strict.

I only intend to restrict extension for the command classes. Other classes can be extended if a dev really needs to do lets remove the rule thats adding `final`. Thank you.

weitzman:
Thank you.

PR #5443, opened 2023-03-06
Stop using options hook in favor of Attribute

weitzman:
I realize that we could deprecate the options hook from annotated command and instead have Attributes directly add options to $commandinfo

Note: since we need to provide a default value for the option, I had to make `$commandInfo->addOptionOrArgumentDescription()` public

### DIFF
--- a/src/Attributes/OptionsetGetEditor.php
+++ b/src/Attributes/OptionsetGetEditor.php
@@ -10,7 +10,7 @@ class OptionsetGetEditor
 {
     public static function handle(\ReflectionAttribute $attribute, CommandInfo $commandInfo)
     {
-        $commandInfo->addOptionOrArgumentDescription($commandInfo->options(), 'editor' ,'A string of bash which launches user\'s preferred text editor. Defaults to <info>${VISUAL-${EDITOR-vi}}</info>.', [],'');
-        $commandInfo->addOptionOrArgumentDescription($commandInfo->options(), 'bg' ,'Launch editor in background process.', [], false);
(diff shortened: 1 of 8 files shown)


src/Attributes/OptionsetGetEditor.php
```diff
@@ -3,9 +3,14 @@
...
+use Consolidation\AnnotatedCommand\Parser\CommandInfo;
 
 #[Attribute(Attribute::TARGET_METHOD)]
-class OptionsetGetEditor extends NoArgumentsBase
+class OptionsetGetEditor
 {
-    const NAME = 'optionset_get_editor';
+    public static function handle(\ReflectionAttribute $attribute, CommandInfo $commandInfo)
+    {
+        $commandInfo->addOptionOrArgumentDescription($commandInfo->options(), 'editor' ,'A string of bash which launches user\'s preferred text editor. Defaults to <info>${VISUAL-${EDITOR-vi}}</info>.', [],'');
```
greg-1-anderson: Looks good, but I don't think that we should make `addOptionOrArgumentDescription` public. That's an internal function, and it's just an implementation detail that calling it on an option that hasn't been defined happens to work; it's not actually part of the contract of that method. Instead, we should make a new method of CommandInfo for adding an option. It's implementation can just be to call `$this->addOptionsOrArgumentDescription($this->options(), ... the other parameters passed in)`. That way, we have a well-named method with a defined contract, and we can adjust its implementation as needed if it ever becomes necessary to change how `addOptionOrArgumentDescription` works.

PR #5450, opened 2023-03-08
php strict_types on commandfiles

weitzman:


DieterHolvoet:
@weitzman the field:delete related test failure is fixed, the current failures are about migrations.

DieterHolvoet (approved):

PR #5459, opened 2023-03-11
Add composer-runtime-api to the composer.lock file

Chi-teck:
The lock file is not up to date with the latest changes in the composer.json, which makes it harder to synchronize pull requests.


Chi-teck:
By the way, do we really need the lock file under Git? If so I propose adding `composer validate` command to some CI workflow.

weitzman:
Good idea to add composer validate

PR #5460, opened 2023-03-11
Use newer Boot level constants

weitzman:


greg-1-anderson (approved):
Cool (I guess tests should pass first 😄 )

PR #5462, opened 2023-03-13
Use Composer Runtime API to determine Drush version

weitzman:
- [Composer says it is safe to get own version](https://getcomposer.org/doc/07-runtime.md#knowing-a-package-s-own-installed-version)
- When doing development on Drush itself, you might see see the version below which amuses me and seems "right enough" since the major version is correct.

```
Drush version : 12.9999999.9999999.9999999-dev
```

### DIFF
--- a/drush.info
+++ b/drush.info
@@ -1,1 +1,0 @@
-drush_version=12.0.0-dev
--- a/src/Drush.php
+++ b/src/Drush.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Drush;
 
+use Composer\InstalledVersions;
 use Robo\Runner;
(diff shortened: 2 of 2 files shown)


greg-1-anderson:
Feels a bit odd, but it's documented, so I guess we can use it, and it's preferable to what we were doing before.

Could we preg_replace the `9`s so that dev just comes out 12.x-dev? I understand why Composer does what it does, but I'd rather not show that as a version to the user. Unless you think folks will pass the output from `drush --version` version compare.

weitzman:
Lets leave the version as is because it gives a sensible value for \Drush\Drush::getMinorVersion. Happy to discuss more.

PR #5467, opened 2023-03-14
Auto-enable pg_trm extension for postgres during DB creation.

weitzman:


phil-s:
This should also happen in `dropOrCreate()`.

This is to cover the scenario where the database is provisioned by some external process which doesn't know about Drupal-specific requirements and therefore doesn't create the `pg_trgm` extension.  In that situation, `drush site:install` calls `dropOrCreate()` which drops all the tables but does not create the extension, and consequently the process errors out a bit later.

As `CREATE EXTENSION IF NOT EXISTS pg_trgm` is a no-op for the cases where the extension *did* already exist, I don't see a reason not to do this.

phil-s:
And although it looks to me as if nothing would normally call `SqlPgsql::createdbSql()`, I think the command should be included there as well.

PR #5470, opened 2023-03-15
Test Drush 8  on Drupal 9

greg-1-anderson:


weitzman:
I won't object to wider testing, but I will object to adding a checkbox for this combo to the [compat table](https://www.drush.org/11.x/install/#drupal-compatibility).

greg-1-anderson:
I'm not sure why the failing tests are pulling in Symfony 6; that makes no sense to me. I don't think I'm going to take time to fix this.

PR #5479, opened 2023-03-16
Render markdown for the console (for topic commands)

weitzman:
https://github.com/phppkg/cli-markdown is working OK for me. Code blocks get a bit of highlighting (not full syntax highlighting) - its helpful. A few bugs:

- [ ] Numbered lists render as bullets `drush docs:generators`
- [ ] using my Iterm2, the headings don't render in blue for me like in the [project's README](https://github.com/phppkg/cli-markdown). Using Terminal they are blue, so I have a color conflict - no idea how to avoid that.
- [ ] Links without associated text show with double urls. See end of `drush docs:cron`

image


src/Commands/DrushCommands.php
```diff
@@ -80,12 +81,20 @@ protected function logger(): ?DrushLoggerManager
...
+        if ($extension == ".htm" || $extension == ".html") {
             $tmp_file = drush_tempnam(basename($file));
             file_put_contents($tmp_file, drush_html_to_text(file_get_contents($file)));
             $file = $tmp_file;
         }
 
+        if ($extension == 'md') {
+            $parser = new CliMarkdown();
+            $rendered = $parser->render(file_get_contents($file));
+            $this->output()->write($rendered);
```
Chi-teck: I would add `OutputInterface::OUTPUT_RAW` option to all write calls in this printer. So the symfony/console will not try to format the output itself.

Chi-teck:
The output looks great. I have concerns only regarding additional dependencies.

https://github.com/phppkg/cli-markdown is relatively new one. Just 6 stars, not under active development. And it depends on https://github.com/cebe/markdown which seems abandoned.

src/Commands/DrushCommands.php
weitzman: Good catch. Done.

Chi-teck:
I wonder, if it a good idea to have `drush readme <module>` command.

Chi-teck:
> https://github.com/phppkg/cli-markdown is relatively new one. Just 6 stars, not under active development.

Well, `phppkg/cli-markdown` is just one class (~350 lines). We can fork it if needed. The main work is done by `cebe/markdown`.

Chi-teck:
The output in Linux (Konsole 22).
Screenshot_20230317_001130

weitzman:
I like the `readme` command suggestion. IIRC, the old `drush dl` would automatically show you the readme after you downloaded a project.

https://github.com/cebe/markdown  is indeed stale but its popular and probably doesnt need to change. I'm inclined to merge this.

Chi-teck:
On the other hand it wouldn't be a BC break if Drush drops this functionality in case of any issues with those third-party projects.

weitzman:
The implementation here changed to prefer the [bat](https://github.com/sharkdp/bat) program for rendering when available.

weitzman:
The output doesnt look excellent, so passing on this for now. Feel free to keep discussing and we can reopen if needed.

PR #5481, opened 2023-03-17
More type hints

weitzman:


src/Boot/DrupalBoot8.php
```diff
@@ -23,20 +23,9 @@ class DrupalBoot8 extends DrupalBoot implements AutoloaderAwareInterface
...
-    /**
-     * @var DrupalKernelInterface
-     */
-    protected $kernel;
-
-    /**
-     * @var Request
-     */
-    protected $request;
+    protected ?LoggrInterface $drupalLoggerAdapter = null;
```
klausi, line 26: typo LoggrInterface - does that mean this code is not actually used anywhere?

PR #5492, opened 2023-03-20
Drush 8 installation note

greg-1-anderson:
Is it okay to put a note about Drush 8 installation on the Drush 12.x (and maybe 11.x) branches? Seems cumbersome to create an 8.x branch of the docs site at this point.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -55,7 +55,7 @@ Drupal Compatibility
   </tr>
   <tr>
     <td> Drush 8 </td>
-    <td> 5.4.5+ </td>
+    <td> 5.4.5+ †</td>
     <!-- Released Nov 2015 -->
     <td> Nov 2023 </td>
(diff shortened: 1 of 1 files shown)


weitzman:
We already have a note at top of https://www.drush.org/12.x/install/. Do we need 2? 

> See the [Drush 8](http://docs.drush.org/en/8.x) or [Drush 9](https://docs.drush.org/en/9.x) docs for installing prior versions of Drush."

What I would like to see is a banner (or similar) at top of https://docs.drush.org/en/8.x/ that points to the current Drush docs. We have a mention on the home page that points to Drush 10 (should be 11). Ideally we have a banner on every page docs.drush.org and not just homepage.

greg-1-anderson:
Oh wait those docs sites are still being rendered. I'll move this there.

Issue #5505, opened 2023-03-28
Drush does not list all annotated commands


Chi-teck:
**To Reproduce**
1. Install [field_definition_dumper](https://github.com/Chi-teck/field_definition_dumper) module.
2. Check that `drush field:dump:types` command works.
3. Runs `drush list` command. 

**Expected behavior**
`field:dump:*` commands to be listed

**Actual behavior**
`field:dump:*` are not listed

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.x
| Drupal version? | 10.x
| PHP version     | 8.x
| OS?             | Linux

**Additional information**
Seems like Drush sets the commands "hidden" by some reason.

weitzman:
Fixed by https://github.com/consolidation/annotated-command/pull/299

weitzman:
These commands look pretty sweet. If @Chi-teck and @DieterHolvoet think these should move to core Drush, I support that. They are pretty small. They do need some basic tests.

Chi-teck:
I don't mind moving these commands to Drush.

DieterHolvoet:
Sure, we could move these to Drush! I'd rename them though, to be in line with other commands like `pm:list`, `role:list`, `queue:list`, etc. How about `field:type:list`, `field:widget:list` and `field:formatter:list`?

weitzman:
What about naming them field:types, field:widgets, field:formatters?

DieterHolvoet:
Sure, that works as well. @Chi-teck would you be up for creating the PR?

PR #5517, opened 2023-03-30
Field definition commands

Chi-teck:
This will bring [Field Definition Dumper](https://github.com/Chi-teck/field_definition_dumper) module into Drush core.

Ref: #5505


Chi-teck:
I am not sure if `FieldDefinitionCommands` is a good name for these class. May be `FieldPluginCommands` or something else would work better.

src/Drupal/Commands/field/FieldDefinitionCommands.php
```diff
@@ -0,0 +1,154 @@
...
+use Drush\Attributes as CLI;
+use Drush\Commands\DrushCommands;
+
+/**
+ * Defines field definition Drush commands.
+ */
+final class FieldDefinitionCommands extends DrushCommands {
+
+    public function __construct(
+        private readonly FieldTypePluginManagerInterface $typePluginManager,
```
weitzman, line 23: Love the promoted properties. I used there on our Attribute classes - good idea to use in commandfiles.

src/Drupal/Commands/field/FieldDefinitionCommands.php
```diff
@@ -0,0 +1,154 @@
...
+use Consolidation\OutputFormatters\StructuredData\RowsOfFields;
+use Drupal\Core\Field\BaseFieldDefinition;
+use Drupal\Core\Field\FieldTypePluginManagerInterface;
+use Drupal\Core\Field\FormatterPluginManager;
+use Drupal\Core\Field\WidgetPluginManager;
+use Drush\Attributes as CLI;
+use Drush\Commands\DrushCommands;
+
+/**
+ * Defines field definition Drush commands.
```
weitzman: Lets omit useless comments. Our PHPCS doesnt require them AFAIK.

src/Drupal/Commands/field/FieldDefinitionCommands.php
```diff
@@ -0,0 +1,154 @@
...
+
+    public function __construct(
+        private readonly FieldTypePluginManagerInterface $typePluginManager,
+        private readonly WidgetPluginManager $widgetPluginManager,
+        private readonly FormatterPluginManager $formatterPluginManager,
+    ) {
+        parent::__construct();
+    }
+
+    #[CLI\Command(name: 'field:types')]
```
weitzman: Lets create a constant for each command name.

src/Drupal/Commands/field/FieldDefinitionCommands.php
```diff
@@ -0,0 +1,154 @@
...
+            'id' => 'ID',
+            'label' => 'Label',
+            'default_widget' => 'Default Widget',
+            'default_formatter' => 'Default Formatter',
+            'settings' => 'Settings',
+        ],
+    )]
+    #[CLI\Help(
+        description: 'Lists field types.',
+        synopsis: 'drush field:types [options]',
```
weitzman: Lets rely on Usages instead of this synopsis.

weitzman: Also all other core commands use Doxygen for description and synopsis which is a bit more readable than Help attribute in my opinion and @greg-1-anderson opinion. Its a taste thing so OK to leave as is.

src/Drupal/Commands/field/FieldDefinitionCommands.php
```diff
@@ -0,0 +1,154 @@
...
+            'provider' => 'Provider',
+        ],
+    )]
+    #[CLI\DefaultFields(fields: ['id', 'label', 'default_settings', 'field_types'])]
+    #[CLI\Help(
+        description: 'Lists field widgets.',
+        synopsis: 'drush field:widgets [options]',
+    )]
+    #[CLI\FilterDefaultField(field: 'id')]
+    public function widgets(array $options = ['format' => 'yaml', 'field-type' => NULL]): RowsOfFields {
```
weitzman:
```suggestion
    public function widgets(array $options = ['format' => 'yaml', 'field-type' => self::REQ]): RowsOfFields {
```

src/Drupal/Commands/field/FieldDefinitionCommands.php
```diff
@@ -0,0 +1,154 @@
...
+            'provider' => 'Provider',
+        ],
+    )]
+    #[CLI\DefaultFields(fields: ['id', 'label', 'default_settings', 'field_types'])]
+    #[CLI\Help(
+        description: 'Lists field formatters.',
+        synopsis: 'drush field:formatters [options]',
+    )]
+    #[CLI\FilterDefaultField(field: 'id')]
+    public function formatters(array $options = ['format' => 'yaml', 'field-type' => NULL]): RowsOfFields {
```
weitzman:
```suggestion
    public function formatters(array $options = ['format' => 'yaml', 'field-type' => self::REQ]): RowsOfFields {
```

src/Drupal/Commands/field/FieldDefinitionCommands.php
```diff
@@ -0,0 +1,154 @@
...
+        );
+    }
+
+    /**
+     * Encodes rows.
+     *
+     * Some output formats i.e. 'table' expect that each row to be a scalar
+     * value.
+     */
+    private static function encodeDefinitions(array $definitions, string $format): mixed {
```
weitzman: See https://github.com/drush-ops/drush/blob/d013acfdb3fa0e48282e19640b065d584f9cb12e/src/Commands/core/StatusCommands.php#L87
https://github.com/weitzman/drush/blob/d013acfdb3fa0e48282e19640b065d584f9cb12e/src/Commands/core/StatusCommands.php#L87 for an approach we use in a few places for formatting table cells.

weitzman (commented):
Thanks! Ideally we add some tests as well.

src/Drupal/Commands/field/FieldDefinitionCommands.php
weitzman: Or, do what https://github.com/drush-ops/drush/blob/d013acfdb3fa0e48282e19640b065d584f9cb12e/src/Commands/core/SiteCommands.php#L99
https://github.com/weitzman/drush/blob/d013acfdb3fa0e48282e19640b065d584f9cb12e/src/Commands/core/SiteCommands.php#L99 and others do and default to `format=yaml`. That handles nested data easily.

src/Drupal/Commands/field/FieldDefinitionCommands.php
Chi-teck: Replaced with Usage attribute.

src/Drupal/Commands/field/FieldDefinitionCommands.php
Chi-teck: That encoder is needed not only for table format but for some other formats that does not support nested structures. Seems like renderer function is not used for `string` format.

src/Drupal/Commands/field/FieldDefinitionCommands.php
Chi-teck: How should we name these constants?

weitzman:
```
const TYPES =  'field:types'
```

src/Drupal/Commands/field/FieldDefinitionCommands.php
weitzman: I think all 3 commands should default to yaml format and return `UnstructuredListData` like https://github.com/weitzman/drush/blob/d013acfdb3fa0e48282e19640b065d584f9cb12e/src/Commands/core/SiteCommands.php#L99 . Then we wont need encodedefinitions() AFAICT.

weitzman: I made a PR into this PR with my proposed changes https://github.com/Chi-teck/drush/pull/1

weitzman:
I added some tests. Feedback welcome.

Chi-teck:
Looks good to me.

weitzman:
Tests are green. Anything else needed here @DieterHolvoet? If not feel free to merge.

DieterHolvoet:
Looks good! Merged.

PR #5524, opened 2023-04-02
Store test results in subdirs. try to fix Insights

weitzman:


weitzman:
It didnt fix anything but merged anyway

PR #5531, opened 2023-04-11
Fix #4494. Remove --post-update option from UpdateDB

### LINKED ISSUE #4494: drush updb does not clear caches if there are no post updates
**Describe the bug**

It looks like drush updb was optimized to only run a cache clear between updates and post updates if there are post updates and were updates.

If there are only updates that doesn't seem to happen.

That is becoming a more serious issue due to the Update cache backend that is now in core, that ignores cache writes, and for example the cached config storage only writes to cache and doesn't delete.

**To Reproduce**
Check value of config with drush cget. Run an update that changes that config value. Check again with drush cget. 

**Expected behavior**
New value is visible.

**Actual behavior**
The old value is still there.

**Workaround**
drush cr

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x/9.x
| Drupal version? | 8.x
| PHP version     | 7.x,5.x

**Additional information**

weitzman:
Per https://github.com/drush-ops/drush/issues/4494#issuecomment-1503165344

### DIFF
--- a/src/Commands/core/DeployCommands.php
+++ b/src/Commands/core/DeployCommands.php
@@ -34,11 +34,8 @@ final class DeployCommands extends DrushCommands implements SiteAliasManagerAwar
         $manager = $this->processManager();
 
         $this->logger()->notice("Database updates start.");
-        $options = ['no-cache-clear' => true];
-        $process = $manager->drush($self, UpdateDBCommands::UPDATEDB, [], $options + $redispatchOptions);
+        $process = $manager->drush($self, UpdateDBCommands::UPDATEDB, [], $redispatchOptions);
         $process->mustRun($process->showRealtime());
-
-        $this->cacheRebuild($manager, $self, $redispatchOptions);
 
         $this->logger()->success("Config import start.");
         $process = $manager->drush($self, ConfigImportCommands::IMPORT, [], $redispatchOptions);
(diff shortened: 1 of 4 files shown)


weitzman:
deploy command does suppresses any cache clears within updatedb and does an unconditional clear after updates and before config import. If we remove --no-cache-clear we will do two in a row in any deploy where there are updates. how should we handle this @alexpott? See https://github.com/drush-ops/drush/blob/de82698455ac0a88c050555f9ef904cafccbf120/src/Commands/core/DeployCommands.php#L33-L47

alexpott:
I think we should remove the first `$this->cacheRebuild($manager, $self, $redispatchOptions); ` from deploy. Calling updatedb from drush should always result in a cache clear even if there are no updates.

weitzman:
I implemented the unified cache clear from updatedb, which happends regardless of whether there were pending updates/post-updates or not.

I'm still a little hesitant to get rid of the --no-cache-clear option. That went in with #308. How sympathetic are folks to that request? Maybe these folks should call updatedb:status and avoid clearing if there is nothing to do?

greg-1-anderson:
Skipping cache clears is definitely a "no seatbelts" option. Folks who use this are implicitly promising to clear the cache later. There are a lot of things you can do to a Drupal site that cause the cache to become dirty, so I don't think this option is too sharp. Being able to speed up sequential operations in CI is compelling, though, so it seems desirable to maintain that capability.

weitzman:
OK, --no-cache-clear is back in, and deploy command no longer uses it since updatedb clears by default.

Issue #5534, opened 2023-04-13
site-install fails for database user passwords containing special characters


wdouglascampbell:
**Describe the bug**
Installation fails when the database user password specified in the --db-url parameter of drush site-install contains either of the following characters: `'`, `\` .

**To Reproduce**
Use the following:

```
drush site-install standard \
  --sites-subdir="www.example.com" \
  --db-url="mysql://test_db_user:Aa~%60%21%40%23%24%25%5e%26%2a%28%29_-%2b%3d%7b%5b%7d%5d%7c%5c%3a%3b%22%27%3c%2c%3e.%20%3f%2fAa~%60%21%40%23%24%25%5e%26%2a%28%29_-%2b%3d%7b%5b%7d%5d%7c%5c%3a%3b%22%27%3c%2c%3e.%20%3f%2f@localhost/test_drupal_db" \
  --db-su="root" \
  --db-su-pw="password" \
  --account-name="admin" \
  --account-pass="password" \
  --site-name="Test Site" \
  --site-mail="postmaster@example.com" \
  --yes
```

**Expected behavior**
Installation would be successful and database user password would work.  Password used is ```Aa~`!@#$%^&*()_-+={[}]|\:;"'<,>. ?/Aa~`!@#$%^&*()_-+={[}]|\:;"'<,>. ?/```

**Actual behavior**
The following error occurs:

```
 [warning] Failed to drop or create the database. Do it yourself before installing. ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '<,>. ?/Aa~`!@#$%^&*()_-+={[}]|\:;"'<,>. ?/'; GRANT ALL PRIVILEGES ON test_dru...' at line 1
```

In this particular case it is getting tripped up on the first single quote.  However, I also discovered when I attempted to remove the single quotes that while things would proceed further they would end with the following error:

```
In install.core.inc line 969:

  Resolve all issues below to continue the installation. For help configuring your database server, see the <a href="https://www.drupal.org/docs/installing-drupal">installation handbook<
  /a>, or contact your hosting provider.<div class="item-list"><ul><li>Failed to connect to your database server. The server reports the following message: <em class="placeholder">SQLSTA
  TE[HY000] [1045] Access denied for user &#039;test_db_user&#039;@&#039;localhost&#039; (using password: YES)</em>.<ul><li>Is the database server running?</li><li>Does the database exis
[...]

PR #5535, opened 2023-04-13
Quote special characters in mysql user password -- Fix issue #5534

wdouglascampbell:
Quote all occurrences of single quotes, semi-colons and backslashes in database user password.

### DIFF
--- a/src/Sql/SqlMysql.php
+++ b/src/Sql/SqlMysql.php
@@ -109,7 +109,7 @@ EOT;
 
             // For MariaDB, ALTER USER was introduced in version 10.2. Support
             // for 10.1 ended in October 2020.
-            $sql[] = sprintf("ALTER USER %s IDENTIFIED BY '%s';", $user, $dbSpec['password']);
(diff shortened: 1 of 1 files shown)


weitzman:
In 12.x we recently fixed a similar issue with the DB name and special chars. See https://github.com/drush-ops/drush/pull/5167/files. So I think the right fix here is to honor the $quoted param. This change should go into 12.x and then 11.x if desired.

wdouglascampbell:
I had not realized that the password could be quoted that way.  I will take a look and then submit a different pull request for 12.x.  How do I ensure it will also be applied to 11.x though?

weitzman:
You open a PR for 12.x and when thats merged you open the equivalent for 11.x (unless I can cherry pick the commit cleanly - then I will just do that)

wdouglascampbell:
Okay.  I discovered that using backticks does NOT work if the password contains a backtick.  I also realized after further testing that somehow my pull request didn't do what I intended.  I have fixed it and will resubmit it.

PR #5542, opened 2023-04-22
Use own DrupalFinder which pulls path from Composer API

weitzman:
- Implements our own DrupalFinder
- Keeps using webflo/drupal-finder when we want to discover roots that are different from the bootstrapped site. That happens in ArchiveDump and in Preflight (when you use a site alias with a local drupal root)

### DIFF
--- a/src/Boot/BootstrapManager.php
+++ b/src/Boot/BootstrapManager.php
@@ -14,7 +14,7 @@ use Consolidation\SiteAlias\SiteAliasManagerAwareInterface;
 use Consolidation\SiteProcess\ProcessManagerAwareInterface;
 use Consolidation\AnnotatedCommand\Input\StdinAwareInterface;
 use Consolidation\AnnotatedCommand\AnnotationData;
-use Drush\DrupalFinder\DrupalFinder;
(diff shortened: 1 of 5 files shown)


weitzman:
One benefit is that we no longer dont rely on any `extra` in the project's composer.json

claudiu-cristea:
What happens for the projects not using composer? Or Drush doesn’t support such sites?

weitzman:
Not supported, for several years now - "Drush only supports one install method. It requires that your Drupal site be built with Composer and Drush be listed as a dependency." https://www.drush.org/latest/install/

greg-1-anderson:
Drush 12 removes support for non-Composer-managed sites. I think that the "one install method" referenced above was added in Drush 9, although non-Composer-managed sites still "happened to work", mostly in support of ISP configurations.

greg-1-anderson (approved):
I like it; it's nice and clean. Would be a good thing for the upstream to adopt too. I'm fine with using our own implementation until that happens.

greg-1-anderson:
I don't have any good theories about why this is breaking on Windows.

PR #5547, opened 2023-04-29
Follow the DI pattern used by Drupal Forms/Controllers/Blocks to instantiate Drush command classes

greg-1-anderson:
If a Drush command handler class implements a static create method, then use it to instantiate the instance, passing it a Drupal container that it may use for dependency injection.

See [Dependency Injection for a Form](https://www.drupal.org/docs/drupal-apis/services-and-dependency-injection/dependency-injection-for-a-form) in the drupal.org documentation.


src/Application.php
```diff
@@ -337,6 +359,10 @@ public function configureAndRegisterCommands(InputInterface $input, OutputInterf
         $runner->registerCommandClasses($this, $commandClasses);
     }
 
+    protected function hasStaticCreateFactory(string $class)
```
greg-1-anderson: Unused, removed.

src/Application.php
```diff
@@ -58,6 +58,9 @@ class Application extends SymfonyApplication implements LoggerAwareInterface, Co
     /** @var TildeExpansionHook */
     protected $tildeExpansionHook;
 
+    /** @var string[] */
+    protected $bootstrapCommandClasses = [];
```
weitzman: Lets please use an actual type hint as well. 
```suggestion
    protected array $bootstrapCommandClasses = [];
```

src/Application.php
```diff
@@ -181,6 +184,11 @@ public function selectUri($cwd)
         return $this->bootstrapManager()->selectUri($cwd);
     }
 
+    public function bootstrapCommandClasses()
```
weitzman:
```suggestion
    public function bootstrapCommandClasses(): array
```

weitzman (commented):
Lovely ... unless I'm mistaken, this will make available commands from uninstalled modules. Those commands are not likely to work, so I dont think it would be a win.

greg-1-anderson:
Commands could check to see if they are able to run in their static factory method; e.g. if they ask for a service they define from the container, and it comes back null, they could decline to run. If the static factory returns null instead of creating an object, then the command is not shown.

We could add code to automatically skip commands that are in disabled modules. In order to do this, we'd need to be able to reliably map from a path to a module. One issue, though, is that commands are more in _projects_ than modules, so there might be some ambiguity about which modules to check for.

greg-1-anderson:
I have a thought about the uninstalled module issue. We could check the path to the implementing class of each commandfile found during discovery. For any of these that are inside the `modules`, `themes` or `profiles` directory, we will require that the namespace end `Drush\Commands\foo`. If there is no foo, then the commandfile is ignored (perhaps with a `debug` log message). If there is a `foo`, then `module_exists` (or whatever the best "is installed" check is) can be called at bootstrap time, and if the specified module is not enabled, then the commandfile will again be ignored. By convention, we presume that folks will always pick `foo` from among the modules in the project the commandfile is packaged with. I don't know if it would be practical for us to enforce that.

As for the issue of Drush version compatibility, folks could simply add `conflict` directives in their module's composer.json file.

weitzman:
Sounds good. I do think we need to focus on avoiding instantiating classes from uninstalled modules. Thats a guarantee that Drupal provides and Drush has followed suit for many years. I see that `$container->getParameter('container.modules')` gets you the paths of enabled modules. Also, [Path::isBasePath()](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Filesystem/Path.php#L690-L721) will be useful.

Perhaps we could convert woot module to show that command are available or unavailable depending on woot's enabled status.

greg-1-anderson:
Yes, I think I can avoid instantiating the class if the module is not enabled; I will add a test to the woot module. Do you suppose that it is necessary to test that the class is not instantiated? We'd have to test a side effect, so maybe it's ok to just trust?

Thanks for the pointers for the path checks. I was also thinking that we might be able to easily validate that `foo` is or is not a module inside a given project by looking for the info.yml file.

weitzman:
I think it suffices to verify that woot command is not in the help list.  

`container->getParameter('container.modules')` has already done that info.yml check for us. Oh maybe you want to know if a class is in a module and its uninstalled. The info.yml check sounds good, or perhaps we could get the list of module discovery paths from the container. Hmm, a bit awkward to get. See \Drupal\Core\Extension\ExtensionDiscovery::scan

greg-1-anderson:
The reason I was thinking we might want to do the check for the info file (or some other check to confirm that `foo` is a module in the project) is some wise module might make a namespace `Drupal/myproject/Drush/Commands/system`, and then the module would run regardless of whether any modules in `myproject` were enabled, since the system module is enabled. Maybe we don't care if people go out of their way to cut their seatbelts.

Also, I note that Drupal does not rely on Composer autoloading for modules, it adds its own autoloader. This means that most modules don't have an autoload section, but modules that use the containerless Drupal command mechanism would have to. This is not terrible, but it we'd have to decide on conventions for this. It might be odd to have two autoloaders loading from the same path. Maybe I could add a bit more code and rely on the Drupal autoloader for module Drush commands. The existing code would still be used for Drush commands in Composer projects that are not modules.

greg-1-anderson:
This version discovers the commands in the modules, and relies on the Drupal autoloader to instantiate the classes. It is therefore not possible to trick Drush into checking some other module's install status. We just use some simple well-known paths / namespaces in the module. Only a few directories are scanned per module, so the performance characteristics are very similar to the PSR-4 discovery.

Folks will still use PSR-4 discovery to add Drush commands to Composer projects that are not Drupal modules. If this mechanism is used to add a command to a module, then it will run, even if the module is not installed. However, doing this is not possible unless the user adds an `autoload` entry to their module's composer.json file. Since modules don't usually have autoload entries, doing this on accident shouldn't ever happen, and if we don't document the possibility, it would be fairly difficult to do deliberately.

composer.json
```diff
@@ -57,7 +57,7 @@
     "cweagans/composer-patches": "~1.0",
     "drupal/core-recommended": "^10",
     "drupal/semver_example": "2.3.0",
-    "phpunit/phpunit": ">=9",
+    "phpunit/phpunit": "^9",
```
weitzman, line 60: What led to forbidding phpunit 10? Its only a dev dependency so not a big deal - just curious.

weitzman (approved):
Great stuff. I think its ready for merge ... Do you have any opinion on whether we can pass a delegate container thing to the create( ) method so that commandfiles can also inject Drush services? Or pass 2 containers? Thats all follow-up material, if at all.

composer.json
greg-1-anderson: Appveyor started running on PHPUnit 10 and broke. I figure we can fix this as follow-on work.

greg-1-anderson:
We could experiment with passing a delegate. I think that league container can delegate to Symfony's DI, but not visa-versa. We could always make our own simple delegate object.  Commandfiles can automatically acquire certain Drush services by implementing available `*AwareInterface`s. Giving full access to the Drush container seems like it would be useful / forward-thinking. I'll look at that briefly next.

PR #5548, opened 2023-05-01
Update the Drush commandfile generator to use create()

weitzman:
I had to change the new module discovery code because we arent discovering commandfiles in [MODULE]/src/Commands. This is what the generator uses, woot uses, etc. Are we changing this deliberately? Also there may be more places where this change is needed.


weitzman:
Pretty sure you made this change deliberately so we now generate accordingly.

PR #5550, opened 2023-05-01
Update docs for create() method

weitzman:
command alterers and generators not yet updated. feedback/suggestions welcome.

PR #5552, opened 2023-05-02
Move commands over to static create instead of using drush services

greg-1-anderson:
Starting with config commands, done previously.


greg-1-anderson:
Part of this work involves defining the strategy for making the Drush container available for commands. There are several options open to us:

1. We could add the Drupal container to the Drush container as a [delegate container](https://container.thephpleague.com/4.x/delegate-containers/).

The reason I do not like this alternative is that the typehint in the commandfile's `create()` method would be for league/container (Drush's container), not Symfony's container (Drupal's container). (Symfony does not support delegate containers, so we can't go the other direction.) This is dissatisfying, because even though the container _acts_ like Drupal's container, it _looks_ like Drush's container. It might also cause problems for commands that want to pass the Drupal container to some Drupal API; recovering the delegate is not supported, and typehinting would prevent passing Drush's container.

2. We could pass both containers to the `create()` method (two parameters instead of one).

This would work fine, and it is expressive, but it looks clunky.

3. We could make our own wrapper class that implements Psr\Container\ContainerInterface, and has `has` and `get` methods that call through to the Drupal and Drush containers. We could also provide methods to recover properly-typehinted references to the Drupal and Drush container.

This is expressive, and not clunky vis-a-vis the API. This is definitely an improvement over the previous two options, but I am still a little dissatisfied with it, because the signature of a Drush `create()` method would still become different than the signature of a Drupal Form `create()` method. It seems to me that it is advantageous to reduce the cognitive load on the user and use a pattern that looks exactly like a similar familiar pattern.

4. We could add the Drush container to the Drupal container, and allow the user to get the Drush container via the `get()` method. 

This is not expressive, but it is documentable and understandable. Another reservation I have is that we are, in a very small way, tainting the Drupal container. I wouldn't want Drupal code to rely on this (to detect if they are being called from a Drush command's `create()` method, for example), although that's overall probably pretty unlikely.

5. We can leave it as it is right now; commandfiles that need the Drush container can implement `ContainerAwareInterface` and have it injected.

The downside of this is that the Drush container is then not available in the `create()` method. The Drush container isn't really designed to be used generally by Drush command files. Drush expects command files to implement `FooAwareInterface` to have a given `Foo` class injected, and other Drush services (not exposed through the inflection system) should not be used.

claudiu-cristea:
@greg-1-anderson is there any issue or docs to understand why we're doing this?

greg-1-anderson:
We are deprecating `drush.services.yml` and instead following the pattern used by Drupal Forms. Moshe has started some documentation: https://github.com/drush-ops/drush/pull/5550

The problems caused by having Drush commands in the Drupal container are minor, and mostly cleared up by a `drush cr`; however, it's better to not have them there, and this pattern is more expressive and more readable. Also, this pattern allows Drush commands in Composer libraries (that are not Drupal modules) to participate in Drupal DI, which is not possible with `drush.services.yml`.

greg-1-anderson:
Looks like `#2` is probably the best option, since it is simplest and the most expressive.

andypost:
+1 to current approach - it looks more readable

src/Commands/core/BatchCommands.php
```diff
@@ -18,6 +19,7 @@ class BatchCommands extends DrushCommands
     #[CLI\Command(name: self::PROCESS, aliases: ['batch-process'])]
     #[CLI\Argument(name: 'batch_id', description: 'The batch id that will be processed.')]
     #[CLI\Help(hidden: true)]
+    #[CLI\Bootstrap(level: DrupalBootLevels::FULL)]
```
greg-1-anderson, line 22: Note that if a command from a `drush.services.yml` file previously did not participate in DI, we have two choices:

1. Give it a static create method that injects no services.
2. Set its Bootstrap level to FULL.

Both of these will have the same effect, as the inclusion of a static create function will force a full bootstrap. I opted for explicitly setting the bootstrap level, as that's simpler, more direct, and if we wanted, in the future we could have commands in the same file that had less than a full bootstrap. (A static create method forces the bootstrap level to full for all commands in the class.)

weitzman (approved):
Great work 💪 . Feel free to merge when you are ready.

greg-1-anderson:
When working on the field commands, I realized that some of the entries in the drush.services.yml are annotated with a character (`?`) that indicates that the service is optional. The `set` method is not called when the service does not exist in this case. I am going to have to go back and review all of these, because calling `get` on the container by default throws an exception for missing services.

I am not sure what is going on with the failing local sql sync test. A command hook is not working right. The odd thing is that the hook is being called, but the data usually passed to the hook (the command options) is missing. So far I have no theories about why this is happening.

Beyond those two issues, most of the commands have ported over pretty easily.

weitzman:
Would it help to only `get` after checking with `has` method? `$container->has('config.storage.export')`

greg-1-anderson:
I'm going to write a one-line wrapper so that `create` methods still look clean. Maybe `has` would be better than the Symfony-specific constant.

greg-1-anderson:
Maybe `has` and `get` right in the `create` method would be better than a wrapper that needed to be documented and learned. I'll compare.

andypost:
@greg-1-anderson if only `has()` and `get()` is exactly `psr/container` 
Ref https://github.com/php-fig/container/blob/master/src/ContainerInterface.php

greg-1-anderson:
I'm going to hold of on finishing and merging this PR until after #5553 is working. The existing Drush "Drupal Command" classes make a good set of test cases for the service adapter introduced in that PR.

greg-1-anderson:
Continued in #5562

PR #5553, opened 2023-05-04
Create a service adapter for drush.services.yml

greg-1-anderson:
Current status: admire, but do not use.

A while back, @weitzman had an idea that we could create a drush.services.yml service adapter that could read these files and instantiate them ourselves, without using a Symfony compiler pass at all. This PR is the beginnings of just such a service adapter.

Currently, the existing Drush Drupal Kernel trait is being used to find drush.services.yml. Because of the way the Drupal Kernel trait is tied to the Drupal service container, and because we are no longer storing Drush services in the service container, the result at this time is that the very next **one** Drush call after a `drush cr` (e.g. `drush list`) will see all of the module commands, and each Drush call after that will not see any module commands at all. Factoring the drush.services.yml discovery out of this trait should clear up this problem.

More work may be needed for command info alterers, generators and etc., but it seems that the adapter is working fairly well so far, so the technique seems promising. The result is that with this adapter, we will not need to remove support for drush.services.yml, and folks won't need to rewrite their module Drush commands to take advantage of the server-containerless DI of Drush 12. The new static create factory syntax will still be recommended and preferred, of course, since it is clearer and easier to read, but using it doesn't have to be mandatory with this adapter.


src/Runtime/DrushServiceFinder.php
```diff
@@ -0,0 +1,24 @@
...
+
+use Psr\Container\ContainerInterface;
+use League\Container\Container as DrushContainer;
+
+/**
+ * Find drush.services.yml files.
+ */
+class DrushServiceFinder
+{
+
```
greg-1-anderson: Committed by mistake, but might end up using eventually. Intended as the home for the drush.services.yml discovery methods, but I put that work off and just used the existing code in the Drush Drupal Kernel trait for the initial PoC.

src/Runtime/ServiceInstantiator.php
```diff
@@ -0,0 +1,204 @@
...
+ * Use Dependency Injection Container to instantiate services.
+ */
+class ServiceInstantiator
+{
+    protected ContainerInterface $drushServicesContainer;
+    protected array $tags = [];
+
+    public function __construct(protected ContainerInterface $container, protected ContainerInterface $drushContainer)
+    {
+        $this->drushServicesContainer = new DrushContainer();
```
greg-1-anderson: Services created in a drush.services.yml file are stored in this container so that they can be referenced as needed e.g. as the parameter to some other service or command also in a drush.services.yml file. Usually, only services in the same file can reference each other, as there is no other way to ensure the initialization order of Drush services.

greg-1-anderson: Maybe with a little extra work, we could use the features of league container to lazy-initialize the commandfiles. I'm not sure if it's necessary to allow cross-Drush-service-file references, though. Since Drush services can reference Drupal services, it would be preferable to just make a Drupal service for any shared service needs a Drush command might have.

src/Runtime/ServiceInstantiator.php
```diff
@@ -0,0 +1,204 @@
...
+            if (isset($serviceFileData['services'])) {
+                $this->instantiateServices($serviceFileData['services']);
+            }
+        }
+    }
+
+    /**
+     * Given a drush.services.yml file (parsed into an array),
+     * instantiate all of the services referenced therein, and
+     * cache them in our dynamic service container.
```
weitzman: "cache them" - the drush containers are rebuilt from scratch on every request, right? maybe we can say 'collect them' instead. if we are dumping containers to disk then we a similar cache rebuild issue that plagues the current module defined commands.

src/Runtime/ServiceInstantiator.php
weitzman: Is this a third container? I have no idea if there is a better way but its at least a thing that will raise an eyebrow.

1. Original Drush container
2. New container containing Drupal service data
3. Drupal container.

weitzman (commented):
Looking great. Thanks for pursuing my musing from long ago.

src/Runtime/ServiceInstantiator.php
greg-1-anderson: We could re-use the drupal container or the drush container or just put it in an array. The scope is only during instantiation; it is not needed thereafter.

src/Runtime/ServiceInstantiator.php
greg-1-anderson: Ah, yeah, there is no persistent cache here.

src/Runtime/DrushServiceFinder.php
andypost: Maybe `LegacyServiceFinder` instead?

greg-1-anderson: Good suggestion; this code vanishes if support for drush.services.yml is ever removed, and we are going to deprecate that even if it sticks around for b/c.

greg-1-anderson:
Down to one failing test. In this PR, the archive restore command fails:

> The config sync directory is not defined in $settings["config_sync_directory"]

Sure enough, it is not defined. If I switch back to the 12.x branch, though, this same test works, but `config_sync_directory` still is not defined in any settings.php file. I'm not sure why the working test works. How can you have a config sync directory without defining the path in settings.php?

src/Drupal/Commands/generate/ApplicationFactory.php
```diff
@@ -48,10 +48,8 @@ public function discover(): array
         $global_generators = $this->discoverPsr4Generators();
 
         $module_generators = [];
-        if ($this->container->has(DrushServiceModifier::DRUSH_GENERATOR_SERVICES)) {
-            $module_generators = $this->container->get(DrushServiceModifier::DRUSH_GENERATOR_SERVICES)->getCommandList(
-            );
-        }
+        $serviceManager = \Drush\Drush::service('service.manager');
+        $module_generators = $serviceManager->getGenerators();
```
greg-1-anderson, line 48: @Chi-teck, a question about `GeneratorInfoAlter`ers: do these simply operate on a list of Symfony Console Command classes? Can I just loop over them here (once they are instantiated), and call alter on each? Seems that should work for the `alterGenerators()` method in this class, but I am not completely sure if that's always true.

I am removing the DrushDrupalKernel and not using the Symfony DI compiler pass to find and instantiate Drush commands, generators and etc. any longer, which is why the alter step is going away & needs to be replaced.

src/Runtime/ServiceManager.php
```diff
@@ -0,0 +1,57 @@
...
+ *   - Symfony Console commands
+ *   - Command info alterers
+ *   - Generators
+ *
+ * Most services are discovered via the PSR-4 discovery mechanism. Legacy
+ * services are injected into this object by the bootstrap handler
+ * (DrushBoot8) using the LegacyServiceFinder and LegacyServiceInstantiator
+ * classes.
+ *
+ * TODO: That's the intention, anyway. For now, we just stuff gererators here.
```
weitzman:
```suggestion
 * TODO: That's the intention, anyway. For now, we just stuff generators here.
```

src/Drupal/Commands/generate/ApplicationFactory.php
greg-1-anderson: Actually I just did a quick ad-hoc test, and discovered that these are running. I guess this is a feature of the Symfony Application object; I had presumed it was a feature of the Symfony DI container.

src/Runtime/LegacyServiceFinder.php
```diff
@@ -0,0 +1,58 @@
...
+
+    protected $drushServiceYamls = [];
+
+    public function __construct(protected $moduleHandler, protected DrushConfig $drushConfig)
+    {
+    }
+
+    public function getDrushServiceFiles()
+    {
+        if (empty($drushServiceYamls)) {
```
andypost: Probably it should be `$this->drushServiceYamls`

src/Runtime/LegacyServiceInstantiator.php
```diff
@@ -0,0 +1,201 @@
...
+class LegacyServiceInstantiator
+{
+    protected array $drushServicesContainer = [];
+    protected array $tags = [];
+
+    public function __construct(protected ContainerInterface $container)
+    {
+    }
+
+    public function loadServiceFiles(array $serviceFiles)
```
weitzman, line 45: Lets add doxygen, param hints, return type hints, etc.

weitzman (approved):

greg-1-anderson:
I've still got a little doxygen and a couple TODOs to do, but this is pretty close to ready to merge now.

greg-1-anderson:
Done with the refactoring. I'm planning on leaving PSR-4 command info alterers as future work, so this PR is now down to just "naming things" and docblock work. I'll merge as soon as I'm done with that, unless I get more feedback.

src/Runtime/LegacyServiceInstantiator.php
```diff
@@ -0,0 +1,267 @@
...
+     * @param object Resolved object from DI container
+     */
+    protected function resolveFromContainer($container, string $arg)
+    {
+        [$required, $arg] = $this->isRequired($arg);
+
+        // Exit early if the container does not have the service
+        if (!$container->has($arg)) {
+            if ($required) {
+                throw new \Exception("Big badda boom! This should be the same thing that the Drupal / Symfony DI container throws.");
```
greg-1-anderson: Here's one more functional change I still need to make.

PR #5556, opened 2023-05-05
field-create: Make it easier for field types extending entity reference to add target type/bundles options (11.x)

DieterHolvoet:
The `field-create` command provides two options for entity reference fields, `target-type` and `target-bundle`. Currently, these options only appear for the `entity_reference` field type, not field types extending it like [entity_reference_revisions](https://www.drupal.org/project/entity_reference_revisions).

Instead of `entity_reference_revisions` having to duplicate all code related to those two options, I think it would be a good idea to make it possible to somehow inherit the `entity_reference` code.

This is how we would be able to implement those two options for `entity_reference_revisions` using the implementation in this MR:

```php
<?php

namespace Drupal\entity_reference_revisions\Commands;

use Drush\Drupal\Commands\field\FieldEntityReferenceHooks;
use Symfony\Component\Console\Input\InputInterface;

class EntityReferenceRevisionsHooks extends FieldEntityReferenceHooks {

    /**
     * @hook on-event field-create-field-storage
     */
    public function hookFieldStorage(array $values, InputInterface $input): array
    {
        if ($input->getOption('field-type') === 'entity_reference_revisions') {
            $values['settings']['target_type'] = $this->getTargetType($input);
        }

        return $values;
    }

    /**
     * @hook on-event field-create-field-config
     */
    public function hookFieldConfig(array $values, InputInterface $input): array
    {
        if ($input->getOption('field-type') === 'entity_reference_revisions') {
            $values['settings']['handler_settings']['target_bundles'] = $this->getTargetBundles($input);
        }

        return $values;
    }

}
```

### DIFF
--- a/src/Drupal/Commands/field/FieldCreateCommands.php
+++ b/src/Drupal/Commands/field/FieldCreateCommands.php
@@ -202,10 +202,6 @@ class FieldCreateCommands extends DrushCommands implements CustomEventAwareInter
             $this->ensureOption('is-translatable', [$this, 'askTranslatable'], false);
             $this->ensureOption('cardinality', [$this, 'askCardinality'], true);
 
-            if ($this->input->getOption('field-type') === 'entity_reference') {
-                $this->ensureOption('target-type', [$this, 'askReferencedEntityType'], true);
-            }
-
             $this->createFieldStorage();
         }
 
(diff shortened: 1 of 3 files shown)


DieterHolvoet:
@weitzman does this look like a good idea to you, given the debate on whether or not to make all command classes final? If not, do you have a suggestion for another solution?

weitzman:
That decision is more of a "by default". It's ok if some commands deviate. Happy to proceed here however you think is best.

weitzman:
Lets remember to target 12.x as well.

DieterHolvoet:
This is ready. There's already a test to make sure the target-type option is asked when creating an entity reference field, so I don't think this needs extra test coverage.

PR #5560, opened 2023-05-11
Handle bool option values in Test trait

weitzman:


jonathan1055:
Thanks for looking at this. For reference this is where I discovered the problem
https://drupal.slack.com/archives/D01204U43N0/p1683825657850099

PR #5562, opened 2023-05-11
Static create commands

greg-1-anderson:
Continuation of #5552, rebased on top of #5553


greg-1-anderson:
I wasn't sure why the Sql command tests were failing after the conversion, so I reverted that part. The generators could take greater advantage of DI now, but overall I think it would be good to merge here now, move on to other things, and maybe come back to those issues later, as follow-on work.

greg-1-anderson:
I just remembered I was going to use `has` and avoid using the Symfony-specific constant here.

weitzman (approved):
Looks great. Feel free to merge when ready.

PR #5564, opened 2023-05-14
More Drupal and Drush service dependency injection in CacheCommands

weitzman:
@greg-1-anderson 

1. Does this look as expected?
2. Fixed by getting from Bootstrapmanager <strike>Does AutoloadAwareInterface still work? $this->autoloader() is not working (independent of this PR I think) - https://github.com/drush-ops/drush/blob/2afc8426c43cc51e515d1dd2838f3a489b7c940c/src/Commands/core/CacheCommands.php#L241</strike>

If this PR looks good, I will go through Commands looking for similar opportunities for DI. After all these years I'm still slightly hostile toward DI for all the indirection it adds to a codebase but I know that ship has sailed.

### DIFF
--- a/src/Commands/core/CacheCommands.php
+++ b/src/Commands/core/CacheCommands.php
@@ -6,14 +6,11 @@ namespace Drush\Commands\core;
 
 use Consolidation\AnnotatedCommand\Hooks\HookManager;
 use Drupal\Core\Cache\CacheTagsInvalidatorInterface;
-use Drupal\Core\Routing\RouteBuilderInterface;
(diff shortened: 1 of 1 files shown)


src/Commands/core/CacheCommands.php
```diff
@@ -40,6 +37,35 @@ final class CacheCommands extends DrushCommands implements CustomEventAwareInter
     const SET = 'cache:set';
     const REBUILD = 'cache:rebuild';
 
+    public function __construct(
+        private ContainerInterface $container,
```
greg-1-anderson: Doesn't look like you need to save the container.

PR #5565, opened 2023-05-14
Backport static create factory commands for Drush 11.x

greg-1-anderson:
This backports static create factory command handlers to Drush 11.x, supporting commands, hooks and info alterers for PSR-4 discovered commands and module commands. Does not include the port of Drush command implementations to use static create factory methods, or any of the modernized classes or refactoring from the 12.x version.

Having this basic capability in Drush 11.x will make module maintainers more likely to adopt static create factory commands earlier, since this will allow them to work across two major versions of Drush.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@ api
 sut/*
 !sut/drush
 sut/drush/sites/*test.site.yml
+!sut/modules
+sut/modules/contrib
 /sandbox/
 .env
 # Test fixtures
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -63,6 +63,7 @@ install:
   # - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
 
 test_script:
+  - composer info phpunit/phpunit
   - vendor/bin/phpunit --colors=always --configuration tests --testsuite functional --debug
   - vendor/bin/phpunit --colors=always --configuration tests --testsuite integration --debug
   - vendor/bin/phpunit --colors=always --configuration tests --testsuite unit --debug
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
     "ext-dom": "*",
     "chi-teck/drupal-code-generator": "^2.4",
     "composer/semver": "^1.4 || ^3",
-    "consolidation/annotated-command": "^4.7.0",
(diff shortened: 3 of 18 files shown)


weitzman:
Code looks good.

1. What happens to a commandfile that has a static create() method AND a drush.services.yml?
2. What do we do about the drush commandfile generator. Seems like that should start generating with create()? 
3. Any changes to the DI and command authoring docs? 
4. Will there be any warnings about the drush,services.yml constraint in a module's composer.json?

greg-1-anderson:
1. I think we need to address this in both 12.x and 11.x branches.
2. Yeah I suppose we could generate static create method classes in 11.x; definitely should do so in 12.x.
3. Yeah, could backport some of the docs here.
4. drush.services.yml processing is unchanged (adapter not backported), so this should still behave the same way.

I'll do 3 here now; maybe 1 and 2 should be follow-on work, done in 12.x branch first.

weitzman:
Well, 4) makes the composer.json stanza unneeded for modules that choose to use create(). Want to make sure we dont error or warn in this case.

weitzman:
We should probably just remove the debug message about the composer.json stanza. That feature is going away anyway.

greg-1-anderson:
Ah, you're right, there's no reason to encourage folks to pin down the version selection in composer.json even in the 11.x branch. We can just take that out.

greg-1-anderson:
`1.` I just tested writing a command with a redundant static create method *and* a drush.services.yml file, and it works fine, presuming that both are well-formed. Note that commands with a static create method must be in the namespace `...\Drush\Commands`, whereas the recommended location for drush.services.yml was previously `...\Commands`, so overlap is unlikely, unless folks happened to move their commands away from that location and happened to choose the next one. Since it doesn't cause a problem even when done, and it would take extra code to catch the "error", I think we can just leave this alone.

`4.` Removed the warning in this PR. I think it's already gone in 12.x, but I'll confirm. (**UPDATE:** Warning still exists in 12.x branch, removed in https://github.com/drush-ops/drush/pull/5570)

Still in progress on 2 and 3; will work on them in the 12.x branch first, then move here.

greg-1-anderson:
12.x branch already used static create when creating Drush command files; ported that code here.

greg-1-anderson:
Docs updates also backported; that should take care of 1-4; let me know if you notice any other gaps.

weitzman (approved):
LGTM.

PR #5566, opened 2023-05-15
More Di in commandfiles.

weitzman:


weitzman:
So many test failures. Can we not use a create method if a command is `#[CLI\Bootstrap(level: DrupalBootLevels::ROOT)]`? That what site:install is.

greg-1-anderson:
I haven't looked at this in earnest, but I will mention that if you add a static create method, then that automatically upgrades the bootstrap level of the command to `FULL`, which was the requirement for module commands, and the core commands in the `Drush\Drupal` namespace. From your comment, I presume this is what is going on here. If you want to add a static create for commands with a lower bootstrap level, we'll have to discuss. The Drupal container isn't available at `DrupalBootLevels::ROOT`, so we might need a different method signature that passes only the Drush container to the constructor, and uses a set method to inject the Drupal container later.

weitzman:
That makes sense. I've removed the SiteInstallCommands changes. IMO it would be ideal if we threw an Exception if you try to use create() with a non-Full command. 

We can revisit further changes in the future.

PR #5567, opened 2023-05-16
Add support for unix socket file paths to db_url

r7l:
This PR adds support for unix socket file paths (unix:///tmp/mysql.sock) when using `$db_url`. 

There seems to be a bug in some versions of PHP that ignores settings for MySQL sockets from php.ini and uses the configuration set during compile time (`--with-mysql-sock=/var/run/mysqld/mysqld.sock`) instead. The only way to fix this on PHP side would be recompiling PHP. But this might not be possible for everyone.

Having this bug prevents a user from using `drush site:install` as this command does not recognize anything set in `settings.php` and uses either PHP defaults or `--db_url` which does not support unix sockets.

Currently with this PHP bug, using this command:

```
drush si --db-url=mysql://user:pass@localhost/database
```

Will always lead to this error:

```
Failed to drop or create the database. Do it yourself before installing. ERROR 2002 (HY000): Can't connect to local server through socket '/var/run/mysqld/mysqld.sock' (2)
```

This PR allows a user use the following command: 

```
drush si --db-url=mysql://user:pass@unix:///tmp/mysql.sock/database
```


greg-1-anderson:
I like the idea of being able to use a unix socket with `drush site:install`, but if I am not mistaken, the `--db-url` format introduced here is new: no prior code ever used db urls in this form, right? I checked to see if I could figure out how to use Unix sockets with Drupal 6, but I couldn't find anything. I don't know if this is because unix socket support was introduced in Drupal 7 or later, or simply because Drupal 6 is so old that it's hard to find info like this.

If there is a historic format, we should try to match it here. If there is no historic format for representing unix sockets in DB URLs, we should consider closely what format we should use here. Your example looks odd:
```
drush si --db-url=mysql://user:pass@unix:///tmp/mysql.sock/database
```
Was that a typo? RFC-1739 says that the host component ("unix") must be separated from the path via a `/`, which is not part of the path. So `unix://` appears to be an invalid URL.

Did you mean to write it like this?
```
drush si --db-url=mysql://user:pass@unix//tmp/mysql.sock/database
```
That would give a path of `/tmp/mysql.sock` and a database `/database`. I am further wondering if we can just presume that unix sockets are always absolute, so that someone could write `unix/tmp/mysql.sock/database` (avoiding the double slash), and it would still mean that the socket was `/tmp/mysql.sock`. One wouldn't ever want a relative socket path, right?

We could also consider using the query parameters:
```
drush si --db-url=mysql://user:pass@unix/database?unix_socket=/tmp/mysql.sock
```
These are just suggestions; if there is historical precedence that I didn't find, we should follow its example. My last example looks most readable and clear to me, though.

r7l:
@greg-1-anderson Thanks allot for reviewing the PR.

Unix sockets are supported by Drupal and Drush for some time. I can't say when it was implemented exactly in Drupal 6 or 7. So far, it was working for me out of the box. Especially after [my last PR to Drush](https://github.com/drush-ops/drush/pull/4180) for Unix sockets. Usually Drupal or Drush would support Unix sockets through settings.php. But it's different during setups when settings.php is ignored by Drush and configuration is taken either by PHP defaults or `--db-url` option.

Btw. after some digging i've found that [this PHP bug existed before](https://bugs.php.net/bug.php?id=50829).

I am using Unix sockets allot and when it comes to format, i've never noted a real standard. I've also checked the mentioned RFC and the examples there all show `bla://` format. There even is RFC1738 that goes into more detail. 

So some projects do this and others do that. So i was trying to use something i've been using on other PHP related projects. For example Symfony is using a [similar syntax at some points](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpClient/Internal/AmpClientState.php#L170) as well as the [phpredis project](https://github.com/phpredis/phpredis/blob/35a7cc094c6c264aa37738b074c4c54c4ca73b87/README.md?plain=1#L85).

If you feel the number of slashes should be limited, i don't mind and could add it. We just need to agree on a format. Personally i would prefer the `:` to stay as it might be cleaner to look for it when parsing since you can easily mix a "unix" within the directory structure or even database hostname to cause issues. So instead of your suggestion:

```
drush si --db-url=mysql://user:pass@unix//tmp/mysql.sock/database
```

I would suggest this:

```
drush si --db-url=mysql://user:pass@unix:/tmp/mysql.sock/database
```

This is also not an uncommon syntax and makes sure we actually refering to Unix sockets. [Nginx does it like this for example](https://gist.github.com/ubermuda/7327264).

When it comes to your last suggestion with the additional query parameters, it would also be possible but i guess the code needed to use this would be more complex. It would need to have add additional checks for the parameter and maybe more. For that reason i would prefer the example shown above.

greg-1-anderson:
Based on the Symfony example, I can see the merit behind the way you originally implemented it, although I do still think that having a URI with a protocol embedded as a path inside of another URI is unusual. Maybe that's why nginx only puts one slash after the colon, to emphasize that the colon is not the same sort of separator as the `://` of the URI schema.

The reason I brought up Drupal 6 is that Drupal 7 introduced the `$databases` array, so if unix sockets worked in Drupal 6, then there would have to be some db url format to represent it. If such a thing existed, I'd prefer to follow that schema. If it doesn't, then maybe following Nginx's lead (`unix:/tmp/mysql.sock/database`) would be the way I would lean. I don't have a strong preference, they all look a little odd to me.

@weitzman do you have any preference?

weitzman:
I'm inclined not to merge this and leave the patch here for folks to apply via cweagans/composer-patches or otherwise. Nobody has mentioned this topic in like 10 years. I searched the web. Going back 10-13 years, it came up a couple times and each time the fix was to adjust your php configuration. Maybe thats not possible anymore, but the demand for this simply too low to justify keeping it around. Further, we'd need to support sockets for postgres as well. 

I'll leave this open a bit for more discussion.

r7l:
@weitzman Sad to read this. Sockets are supported by Drush and Drupal already by `settings.php` even with custom socket location configuration within this file. This PR is not a new feature. Instead it just extends the install option of Drush which relies on PHP defaults that might not work perfectly all the time.

PR #5569, opened 2023-05-17
Add locale:import:all - imports multiple custom translation po files

andypost:
Supersedes #4251


src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,78 @@ public function exportValidate(CommandData $commandData): void
...
+    /**
+     * Imports multiple translations from the defined directory.
+     *
+     * @throws \Exception
+     */
+    #[CLI\Command(name: self::IMPORT_ALL, aliases: ['locale-import-all', 'locale:import:all'])]
+    #[CLI\Argument(name: 'directory', description: 'The path to directory with .po files to import.')]
+    #[CLI\Option(name: 'type', description: 'String types to include, defaults to <info>not-customized</info>. Recognized values: <info>not-customized</info>, <info>customized</info>')]
+    #[CLI\Option(name: 'override', description: 'Whether and how imported strings will override existing translations. Defaults to the Import behavior configured in the admin interface. Recognized values: <info>none</info>, <info>customized</info>, <info>not-customized</info>, <info>all</info>')]
+    #[CLI\Usage(name: 'drush locale:import-all /var/www/translations', description: 'Import all custom translations from the defined directory.')]
```
Sutharsan: The word 'custom' is not applicable here. 
I suggest to add that the directory is _not_ searched recursively.
I also suggest to add that the translation language is automatically determined based on the file name (some-name.langcode.po).

src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,78 @@ public function exportValidate(CommandData $commandData): void
...
+    #[CLI\Usage(name: 'drush locale:import-all --types=customized --override=all', description: 'Import all custom translations from the defined directory and override any existing translation.')]
```
Sutharsan: No directory is provided in the example. /var/www/translations/custom

src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,78 @@ public function exportValidate(CommandData $commandData): void
...
+    #[CLI\ValidateModulesEnabled(modules: ['locale'])]
+    public function importAll($directory, $options = ['type' => self::OPT, 'override' => self::OPT])
+    {
+        if (!is_dir($directory)) {
+            throw new \Exception('The destination directory does not exist.');
```
Sutharsan: 'destination' is not applicable here.

src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,78 @@ public function exportValidate(CommandData $commandData): void
...
+        }
+
+        // Look for .po files in defined directory
+        $poFiles = glob($directory . DIRECTORY_SEPARATOR . '*.po');
+        if (empty($poFiles)) {
+            throw new \Exception('Translation files not found in the destination directory.');
```
Sutharsan: 'destination' is not applicable here.

src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,78 @@ public function exportValidate(CommandData $commandData): void
...
+        foreach ($poFiles as $file) {
+            // Ensure we have the file intended for upload.
+            if (file_exists($file)) {
+                $poFile = (object) [
+                    'filename' => basename($file),
+                    'uri' => $file,
+                ];
+                $poFile = locale_translate_file_attach_properties($poFile, $translationOptions);
+                if ($poFile->langcode == LanguageInterface::LANGCODE_NOT_SPECIFIED) {
+                    $this->logger->warning(dt('Can not autodetect language of file @file', [
```
Sutharsan: Add an explanation of the required file name pattern. `some-name.langcode.po`

Sutharsan (changes requested):
Initial review of the code. I did not test the command yet.

<strike>Note: Performing the same import with locale:import is possible but requires bash scripting.</strike>

src/Commands/core/LocaleCommands.php
Sergey-Orlov: updated description for this example

src/Commands/core/LocaleCommands.php
Sergey-Orlov: fixed

src/Commands/core/LocaleCommands.php
Sergey-Orlov: updated

src/Commands/core/LocaleCommands.php
Sergey-Orlov: updated

src/Commands/core/LocaleCommands.php
Sergey-Orlov: Added details about pattern

weitzman:
@Sutharsan Is all feedback addressed?

src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,78 @@ public function exportValidate(CommandData $commandData): void
...
+        $translationOptions['customized'] = $this->convertCustomizedType($options['type']);
+        $override = $this->convertOverrideOption($options['override']);
+        if ($override) {
+            $translationOptions['overwrite_options'] = $override;
+        }
+        $langcodes_to_import = [];
+        $files = [];
+        foreach ($poFiles as $file) {
+            // Ensure we have the file intended for upload.
+            if (file_exists($file)) {
```
Sutharsan: The other if conditions in this foreach loop use `continue` to skip processing the current `$file`. I agree with using this code style (early return). I'd like to use it consistently and apply it to this condition too, and remove these nested if conditions. With this being done, the comment 'Ensure we have the file ...' can go away. It is pretty obvious what is going on here.

Sutharsan:
I have tested the code and all works as expected. 

@weitzman If you provide an array of command aliases, drush shows all aliases except the first. Is this intended behavior?  (I see the same behavior with existing commands)
Since this is a new command why should we provide aliases? No BC required, no short code.

The command code:  
```
#[CLI\Command(name: self::IMPORT_ALL, aliases: ['locale-import-all', 'locale:import:all'])]
```

The output of `drush list`:
```
locale:                                                                                                                        
  locale:check                                 Checks for available translation updates.                                       
  locale:clear-status                          Clears the translation status.                                                  
  locale:export                                Exports to a gettext translation file.                                          
  locale:import                                Imports to a gettext translation file.                                          
  locale:import-all (locale:import:all)        Imports multiple translation files from the defined directory.                  
  locale:update                                Imports the available translation updates.                                      
```

src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,78 @@ public function exportValidate(CommandData $commandData): void
...
+    #[CLI\Option(name: 'override', description: 'Whether and how imported strings will override existing translations. Defaults to the Import behavior configured in the admin interface. Recognized values: <info>none</info>, <info>customized</info>, <info>not-customized</info>, <info>all</info>')]
+    #[CLI\Usage(name: 'drush locale:import-all /var/www/translations', description: 'Import all translations from the defined directory (non-recursively). Supported filename patterns are: {project}-{version}.{langcode}.po, {prefix}.{langcode}.po or {langcode}.po.')]
+    #[CLI\Usage(name: 'drush locale:import-all /var/www/translations/custom --types=customized --override=all', description: 'Import all custom translations from the defined directory (non-recursively) and override any existing translation. Supported filename patterns are: {project}-{version}.{langcode}.po, {prefix}.{langcode}.po or {langcode}.po.')]
+    #[CLI\ValidateModulesEnabled(modules: ['locale'])]
+    public function importAll($directory, $options = ['type' => self::OPT, 'override' => self::OPT])
```
weitzman:
```suggestion
    public function importAll($directory, $options = ['type' => self::REQ, 'override' => self::REQ])
```
The meaning of REQ is that a value must be supplied when the option is used.

src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,78 @@ public function exportValidate(CommandData $commandData): void
...
     }
 
+    /**
+     * Imports multiple translation files from the defined directory.
+     *
+     * @throws \Exception
+     */
+    #[CLI\Command(name: self::IMPORT_ALL, aliases: ['locale-import-all', 'locale:import:all'])]
+    #[CLI\Argument(name: 'directory', description: 'The path to directory with translation files to import.')]
+    #[CLI\Option(name: 'type', description: 'String types to include, defaults to <info>not-customized</info>. Recognized values: <info>not-customized</info>, <info>customized</info>')]
```
weitzman: Use suggestionValues param as per https://github.com/drush-ops/drush/blob/166af7acb72ad256cbc8c33fba58f5981683cd5a/src/Commands/config/ConfigCommands.php#L146
https://github.com/weitzman/drush/blob/166af7acb72ad256cbc8c33fba58f5981683cd5a/src/Commands/config/ConfigCommands.php#L146. Helps with shell completion.

weitzman (commented):
Yeah, we should omit "dash" aliases for new commands.

weitzman:
> If you provide an array of command aliases, drush shows all aliases except the first. Is this intended behavior? 

Looks like a bug. I think we should show only the first alias. Otherwise the first column gets too wide.

weitzman:
[Actually ListCommands hides 'dash' aliases](https://github.com/drush-ops/drush/blob/c2b51cf5c0b537cc989513ccaa05e3e229104d4f/src/Commands/help/ListCommands.php#L132-L139).

src/Commands/core/LocaleCommands.php
Sergey-Orlov: fixed

src/Commands/core/LocaleCommands.php
Sergey-Orlov: done

src/Commands/core/LocaleCommands.php
Sergey-Orlov: done, for **override** option as well

weitzman (approved):
LGTM. @Sutharsan will merge when satisfied.

Sergey-Orlov:
rebased forked branch, so all checks should be fine now

src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,80 @@ public function exportValidate(CommandData $commandData): void
...
+        $override = $this->convertOverrideOption($options['override']);
+        if ($override) {
+            $translationOptions['overwrite_options'] = $override;
+        }
+        $langcodes_to_import = [];
+        $files = [];
+        foreach ($poFiles as $file) {
+            // Ensure we have the file intended for upload.
+            if (!file_exists($file)) {
+                $this->logger->warning(dt('Can read file @file.', ['@file' => $file]));
```
Sutharsan:
```suggestion
                $this->logger->warning(dt('Can not read file @file.', ['@file' => $file]));
```

src/Commands/core/LocaleCommands.php
```diff
@@ -208,6 +209,80 @@ public function exportValidate(CommandData $commandData): void
...
+            ];
+            $poFile = locale_translate_file_attach_properties($poFile, $translationOptions);
+            if ($poFile->langcode == LanguageInterface::LANGCODE_NOT_SPECIFIED) {
+                $this->logger->warning(dt('Can not autodetect language of file @file. Supported filename patterns are: {project}-{version}.{langcode}.po, {prefix}.{langcode}.po or {langcode}.po.', [
+                   '@file' => $file,
+                ]));
+                continue;
+            }
+            if (!$this->getLanguageManager()->getLanguage($poFile->langcode)) {
+                $this->logger->warning(dt('Language @language not exists for file @file', [
```
Sutharsan:
```suggestion
                $this->logger->warning(dt('Language @language does not exist for file @file', [
```

Sutharsan (changes requested):
I found two sentences that need some improvement.

Sutharsan:
Once this is fixed, I will merge the PR.

Sergey-Orlov:
> Once this is fixed, I will merge the PR.

Fixed and rebased again

Sutharsan (approved):

Sutharsan:
Thanks all for working on this.

PR #5571, opened 2023-05-20
Modernize site:install command

greg-1-anderson:
Annotation data is available from the provided command metadata; we do not need to recover it from the command object.


weitzman:
Lgtm.

PR #5581, opened 2023-05-21
Discover generators same as commands, without backward compat

weitzman:


src/Boot/DrupalBoot8.php
```diff
@@ -273,11 +274,6 @@ public function addDrupalModuleDrushCommands(BootstrapManager $manager): void
             $commandInfoAlterers = array_merge($commandInfoAlterers, $commandInfoAlterersInThisModule);
         }
 
-        // Look up the generators from the legacy service instantiator and inject
-        // them into the service manager. The `generate` command will get the
-        // generators from the service manager.
-        $this->serviceManager->injectGenerators($legacyServiceInstantiator->taggedServices('drush.generator.v3'));
-
```
weitzman, line 280: We have no attempt at backward compat since generator API changed in Drush 12 anyway.

src/Commands/generate/ApplicationFactory.php
```diff
@@ -43,13 +47,19 @@ public function create(): Application
...
-        $module_generators = $serviceManager->getGenerators();
+        $module_generator_classes = [];
+        foreach ($this->container->get('module_handler')->getModuleList() as $moduleId => $extension) {
+            $path = DRUPAL_ROOT . '/' . $extension->getPath() . '/src/Drush/';
+            $module_generator_classes = array_merge(
+                $module_generator_classes,
+                $this->serviceManager->discoverModuleGenerators([$path], "\\Drupal\\" . $moduleId . "\\Drush")
+            );
+        }
+        $module_generators = $this->serviceManager->instantiateServices($module_generator_classes, $this->drush_container, $this->container);
```
weitzman, line 57: Generator discovery (except sitewide) happens within generate command, and not during DrupalBoot8.

src/Runtime/ServiceManager.php
```diff
@@ -205,7 +205,7 @@ public function discoverPsr4Generators(): array
     {
         $classes = (new RelativeNamespaceDiscovery($this->autoloader))
             ->setRelativeNamespace('Drush\Generators')
-            ->setSearchPattern('/.*Generator\.php$/')
+            ->setSearchPattern('/.*DrushGenerator\.php$/')
```
weitzman: Rename PSR4 generators to require `DrushGenerator` not just `Generator`. Matches how PSR4 commands are named. So CustomDrushGenerator is new name of our testing generator.

src/Runtime/ServiceManager.php
```diff
@@ -231,8 +231,8 @@ public function discoverModuleCommands(array $directoryList, string $baseNamespa
             ->setIncludeFilesAtBase(true)
             ->setSearchDepth(1)
             ->ignoreNamespacePart('src')
-            ->setSearchLocations(['Commands', 'Hooks', 'Generators'])
-            ->setSearchPattern('#.*(Command|Hook|Generator)s?.php$#');
+            ->setSearchLocations(['Commands', 'Hooks'])
+            ->setSearchPattern('#.*(Command|Hook)s?.php$#');
```
weitzman, line 227: Dont discover generators here.

sut/modules/unish/woot/src/Drush/Generators/ExampleGenerator.php
```diff
@@ -22,12 +27,19 @@ class ExampleGenerator extends BaseGenerator
...
      */
-    protected ModuleHandlerInterface $moduleHandler;
+    public function __construct(
+        protected ModuleHandlerInterface $moduleHandler,
+    ) {
+        parent::__construct();
+    }
 
-    public function __construct(ModuleHandlerInterface $moduleHandler = null)
+    public static function create(ContainerInterface $container): self
```
weitzman, line 36: Illustrates create() method for injecting dependencies into a generator.

src/Commands/generate/ApplicationFactory.php
weitzman: Seems like sitewide should also move out of Bootstrap?

greg-1-anderson: Sitewide commands are discovered during preflight, but then they are saved until bootstrap so that they may be constructed via their static `create` method. We have to wait until bootstrap to construct them, because the Drupal container isn't available until then. We could use a setter method for the Drupal container, but it seems much better to follow the static `create` pattern consistently, and not force folks to do DI during their container setter method, which would look awkward.

Sitewite commands that do not have a static create factory are instantiated during preflight via `new` or `createEarly()`.

src/Runtime/ServiceManager.php
greg-1-anderson: It somehow escaped my attention that PSR4 commands must be named `*DrushCommands`, and not just `*Commands`. Since we already require PRS4 commands and generators to be in a "Drush" namespace, I think that it is unnecessary to require the word "Drush" in the filename and class as well. Folks already find that the need to put the word "Commands" in the file/classname is a gotcha; being inconsistent about this when commands in modules don't need it seems like a second-order gotcha that we would be better off avoiding. I would strongly recommend that we keep this as just `*Generator.php`, and loosen up PSR4 commands to also just be `*Commands.php` to keep it simpler and more straightforward. Existing PRS4 commands named `*DrushCommands.php` will keep working if we do this.

If for some reason you think that the double-Drush restriction is important for some reason, then I agree we should be consistent here.  Note also, though, that we allow either plural or signular for module commands and generators, but require plural for commands and require singular for generators when PSR-4 discovered; I'd recommend also allowing both singular and plural for PSR4 commands and generators, for consistency

greg-1-anderson (approved):
Looks great, except I think we should drop the double-Drush requirement in PSR4 command and generator discovery, and allow `*\Drush\Commands\*Commands.php` / `*\Drush\Genrators\*Generators.php` for both module and PSR4 discovery. Up to you, though, feel free to merge the way you think is best.

docs/dependency-injection.md
```diff
@@ -27,6 +27,7 @@ The code that is responsible for providing the dependencies a class needs is usu
 
 create() method
 ------------------
+:octicons-tag-24: 12.0+
```
greg-1-anderson: This is backported to Drush 11, but not yet released; maybe this should be 11.6+?

docs/dependency-injection.md
```diff
@@ -55,7 +56,7 @@ omit the second parameter to the `create()` method.
 
 createEarly() method
 ------------------
-
+:octicons-tag-24: 12.0+
```
greg-1-anderson, line 59: _Not intending to backport createEarly atm_

src/Runtime/ServiceManager.php
weitzman: I"m fine with shortening the filename requirement. I only made it because I was seeing `sut/modules/unish/woot/src/Drush/Generators/ExampleGenerator.php` get picked up by **both** psr4 discovery and module discovery. How to avoid this double discovery problem when the filenames are the same between each? I suspect this affects both commands and generators.

weitzman: I've backed out the DrushGenerators.php suffix requirement. I'm still seeing ExampleGenerator getting discovered in $module_generators (correct) and $global_generators (incorrect). But its harmless as Application just accepts one of them. 

Lets shorten the filename requirement for commands in a new issue.

greg-1-anderson: The double discovery won't happen for most modules, because module class files are autoloaded via a Drupal-provided autoloader, and therefore do not appear in the Composer PSR-4 autoloader list. This should only be an issue for our examples. I think it's good to test that the double discovery is at least innocuous, because it's possible someone might needlessly add a PRS-4 autoload section to a module's composer.json file.

greg-1-anderson (approved):
LGTM

PR #5582, opened 2023-05-22
Deprecate IO trait

greg-1-anderson:
Favor passing $io via parameter injection rather than inflecting it via the IO trait.

I originally thought that I would get rid of `$this->io()` everywhere, but there are a lot of places where this is cross-referenced e.g. via traits that are used in multiple commands, so it would be a bit lengthy / tedious to unwind it all.

We'll see how this MVP work to partially remove such access holds up in testing.

### DIFF
--- a/src/Commands/config/ConfigExportCommands.php
+++ b/src/Commands/config/ConfigExportCommands.php
@@ -16,6 +16,7 @@ use Drush\Exceptions\UserAbortException;
 use Symfony\Component\Console\Output\BufferedOutput;
 use Symfony\Component\Filesystem\Path;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
 final class ConfigExportCommands extends DrushCommands
 {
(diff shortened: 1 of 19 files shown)


greg-1-anderson:
Shall we do just this much, maybe?

weitzman:
I'm a bit meh on this one. My objection to this parameter injection is that its not clear who is passing $io and why it works to just add the right param (and type hint?) to your command method. A less magical way to get $io in my mind is to fetch it from drush_container during create(). Or I'm fine with the status quo. Happy to discuss more.

greg-1-anderson:
The status quo is okay. I admit that parameter injection is a little magical. (Yes, it needs the type hint to work.) I don't like storing $io / $logger etc. in fields of the class is that it makes things pretty rough if you want to make a sub-call to some method, and hope to capture the logs or output in a different object than the one cached in the current class instance. This isn't really an issue for Drush. It's a big pain in some instances for Robo tasks.

We can kick this can down the road another version or two (or forever, maybe).

PR #5583, opened 2023-05-22
Always call the site-local Drush, if possible

greg-1-anderson:
With this change, we avoid sending `--root` when calling remote Drush commands, and instead execute the site-local Drush directly, by running the Drush found in `vendor/bin`.

In Drush 12, if the alias includes a `root`, then we'll presume that the target is a site-local Drush.

This is a step towards removing support for global Drush in remote calls. In Drush 13 we'd probably be happier if Drush 12 was not including unnecessary `--root` options (which we allow in Drush 12 for bc with Drush 11). We wouldn't have to do this in Drush 12, we could just require folks to set the `drush-script` path in the alias file.


greg-1-anderson:
Without this PR:
```
$ drush @example.dev status -s
 [notice] Simulating: ssh -p 2222 -o "AddressFamily inet" isp.io 'drush status --uri=dev-example.isp.io --root=/code/web'
```

With this PR:
```
$ drush @example.dev status -s
 [notice] Simulating: ssh -p 2222 -o "AddressFamily inet" isp.io '/code/vendor/bin/drush status --uri=dev-example.isp.io'
```

greg-1-anderson:
Failures are just places where the expectations have changed, and the tests need to be updated.

greg-1-anderson:
We could also change the output of `drush sa`

Current:
```
$ drush sa @self
'@self':
  root: /path/to/project/web
  uri: default
```
Proposed:
```
'@self':
  paths:
    drush-script: /path/to/project/vendor/bin/drush
  uri: default
```
With this change, `root` and `uri` would remain the preferred way to describe a site; this has the advantage of being easier to read, and less work for folks in terms of updating their site aliases and learning new best practices in the no-global-Drush era.

greg-1-anderson:
The downside is that this makes it impossible for Drush 12 to call a global Drush on a remote system. Under the new model, you usually won't / shouldn't remote dispatch to a different site, which means that the remote site should always have a site-local Drush, so this limitation is probably not terrible (and might be considered a feature).

src/SiteAlias/ProcessManager.php
```diff
@@ -65,6 +69,10 @@ public function drushScript(SiteAliasInterface $siteAlias)
         // If the provided site alias is for a remote site / container et. al.,
         // then use the 'drush' in the $PATH.
         if ($this->hasTransport($siteAlias)) {
+            if ($siteAlias->hasRoot()) {
+                return Path::join($siteAlias->root(), '../vendor/bin/drush');
```
weitzman: Lets check https://getcomposer.org/doc/07-runtime.md#binary-bin-dir-path-in-binaries for bin-dir and then fall back to `$environment->getVendorPath() . '/bin/drush'`

IMO its reasonable to assume that a codebase layout is consistent across environments.

weitzman (approved):
Seems like a step forward to me.

src/SiteAlias/ProcessManager.php
greg-1-anderson: Yes, seems good; if that assumption happens to not be true, then folks can always fall back to `path.drush-script`.

greg-1-anderson: Hm, according to the docs, `$_composer_bin_dir`, `$_composer_autoload_path` et. al. are only set for .phar files, so they will never be of any use to Drush.

greg-1-anderson: `$environment` is only available during preflight, but the vendor path is exported to `drush.vendor-dir` in the Drush config, so I can get it from there.

greg-1-anderson: Bother, `drush.vendor-dir` is an absolute path. I can use `drush.base-dir` to make it relative again, though.

weitzman: > Hm, according to the docs, $_composer_bin_dir, $_composer_autoload_path et. al. are only set for .phar files, so they will never be of any use to Drush.

Pretty recently, Composer started writing a proxy script at vendor/bin/drush that calls out to vendor/drush/drush/drush, Same for all other php binaries. Thats what set global vars for bin path and autoload path. It works for non-phars. See https://getcomposer.org/doc/articles/vendor-binaries.md#finding-the-composer-bin-dir-from-a-binary

weitzman:
Is there any simplification we can do for local Drush calls? --root is obsolete there too, since the DrushDrupalFinder uses Composer API to find location of drupal/core.

src/SiteAlias/ProcessManager.php
greg-1-anderson: Oh, I see. I was just looking at the Drush project. When I looked at Drush included as a dependency of another project, then I see that `$_composer_bin_dir` et. al. are defined. I was just confused by wha the docs meant by a "binary"; I see now that the script in `vendor/bin` now falls into that category, rather than just being a symlink as it was in the past.

src/SiteAlias/ProcessManager.php
```diff
@@ -35,17 +35,21 @@ public function drush(SiteAliasInterface $siteAlias, string $command, array $arg
...
+
         // Fill in the root and URI from the site alias, if the caller
         // did not already provide them in $options.
         if ($siteAlias->has('uri')) {
             $options += [ 'uri' => $siteAlias->uri(), ];
         }
-        if ($siteAlias->hasRoot()) {
+        // Include the --root parameter only if calling a global Drush.
+        // This is deprecated in Drush 12, and will be removed in Drush 13.
+        if ($siteAlias->hasRoot() && ($drushScript == "drush")) {
```
greg-1-anderson, line 47: > Is there any simplification we can do for local Drush calls?

If your site alias for your local Drupal site has a `--root` entry, then `$drush_script` will be an absolute path (to something relative to the root, or using `$_composer_bin_dir`). If your site alias for your local Drupal site does not have a `--root` entry, then we assume it's a multisite of the same Drupal install, and use the absolute path to the current running instance of Drush. Either way, the `--root` option is not passed along in the `exec`. The only way you get a `--root` here is if your alias has a `root` option and also defines `paths.drush_script` to be just "drush", i.e. the Drush in the $PATH. This is the one way that you could get Drush 12 to call a global Drush (earlier I incorrectly said there would be no way to do that).

The other way you can get simply "drush" here is if you're making a remote call using an alias that does not have a `--root` entry, e.g. a Pantheon alias. This will also call the global Drush. On Pantheon, the `drush` in the $PATH is `vendor/bin/drush`, as there is only one Drupal site in the container. I don't know that any other ISPs use this pattern. Pantheon used to have roots that looked like `/srv/binding/AN_ID_THAT_CAN_CHANGE/code/web`, but now it's always just `/code/web`, so we could start writing `root` entries into Pantheon aliases. Most folks are going to use `terminus drush` rather than Pantheon aliases, though, so this is something of an edge case.

weitzman: I was more thinking that in a local drush call like `drush --root=foo version`, the `--root` has no impact or should have no impact because the Drupal root is determined via the location of `drupal/core`.

greg-1-anderson: This code usually won't include the `--root` parameter. It will only be included if you also set `paths.drush-script` to "drush", which most current site aliases won't do, because that's the default in previous versions of Drush. So `--root` should go away in most instances.

PR #5584, opened 2023-05-22
Remove drush's own service providers.

weitzman:
I think these lines are no longer needed /cc @greg-1-anderson


greg-1-anderson (approved):
Yes, these can go!

weitzman:
Is there a reason why the sql commandfiles stay in /Drupal dir?

greg-1-anderson:
Yes, I tried to port them, and the tests failed. I couldn't figure out why they were breaking, seemed unrelated. Thought I would do one commandfile at a time, and see which one actually caused the problem. Haven't had time to circle back yet.

Issue #5587, opened 2023-05-26
Provide `NULL` value support for config-set without having to use `--input-format=yaml`


wimleers:
**Is your feature request related to a problem? Please describe.**
Just like #4081/#4781 added support for the empty string and empty array, we also need to provide support for `~` (NULL).

Example: a Text Editor config entity (`editor.editor.*`)'s `image_upload.max_dimensions.width` accepts integers for imposing a width restriction, but also `NULL` (`null` or `~` in YAML) for indicating there is _no_ width restriction. See https://www.drupal.org/node/2644838.

There's a handful of other config schemas in Drupal core that have `nullable: true` set that allow this, and many more in contrib.

**Describe the solution you'd like**
Allow `drush cset media.settings iframe_domain NULL -y` to set `null` to not only have
```
        // Special handling for empty array.
        if ($data == '[]') {
            $data = [];
        }
```
but also
```
        // Special handling for NULL.
        if ($data == 'NULL') {
            $data = NULL;
        }
```
(PR: #5588 — handles any upper- or lowercasing of this special value.)

With those 3 added lines, it'd work as expected, instead of generating:
```
iframe_domain: 'NULL'
```
it would generate:
```
iframe_domain: null
```

**Describe alternatives you've considered**
Work-arounds that work today:
- `drush cset media.settings iframe_domain null --input-format=yaml -y`
- `drush cset media.settings iframe_domain '~' --input-format=yaml -y`

**Additional context**
N/A

weitzman:
Done

PR #5588, opened 2023-05-26
cset - allow NULL as value

wimleers:
PR for #5587.

### DIFF
--- a/src/Commands/config/ConfigCommands.php
+++ b/src/Commands/config/ConfigCommands.php
@@ -171,9 +171,9 @@ final class ConfigCommands extends DrushCommands implements StdinAwareInterface,
             $data = [];
         }
 
-        // Special handling for empty array.
+        // Special handling for null.
         if (strtolower($data) == 'null') {
-            $data = NULL;
+            $data = null;
         }
 
         if ($options['input-format'] === 'yaml') {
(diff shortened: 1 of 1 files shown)


src/Commands/config/ConfigCommands.php
```diff
@@ -170,6 +171,11 @@ public function set($config_name, $key, $value, $options = ['input-format' => 's
             $data = [];
         }
 
+        // Special handling for empty array.
```
wimleers: 🙈 
```suggestion
        // Special handling for null.
```

src/Commands/config/ConfigCommands.php
```diff
@@ -170,6 +171,11 @@ public function set($config_name, $key, $value, $options = ['input-format' => 's
...
+        if (strtolower($data) == 'null') {
+            $data = NULL;
```
wimleers: Apparently different coding style than Drupal! We need this:
```suggestion
            $data = null;
```

weitzman:
Welcome to Drush, @wimleers!

T2L:
this change produces

`Error: ]  TypeError: strtolower(): Argument #1 ($string) must be of type string, array given in strtolower() (line 175 of /vendor/drush/drush/src/Commands/config/ConfigCommands.php`

when using something like `./vendor/bin/drush config:set config_ignore.settings ignored_config_entities '[]' --yes` with PHP8.1+

Issue #5572, opened 2023-05-20
Logger mesages are no longer retrievable in contrib phpunit tests


jonathan1055:
**Describe the bug**
A recent change in Drush 12 has side-effects for contrib modules who provide commands and test them in phpunit.

**To Reproduce**
The tests fail on DrupalCI and I can replicate the test failure locally on MacOS php8.1 with latest Drush 12 dev
https://www.drupal.org/project/scheduler/issues/3359998#comment-15050025 has more details

**Expected behavior**
Log messages created via a modules logger using `$this->logger->notice()` used to be availbale during phpunit test. They could be retrieved in the test using `$messages = $this->getErrorOutput()`. But this behavior stopped in the latest Drush 12 and the messages are not retrievable. Phpunit tests that were checking for message fail where they previously passed.

I have isolated the problem and it was created in [this commit](https://github.com/drush-ops/drush/pull/5553/files) from #5553 
Rebasing Drush 12 dev locally back to the [immediately preceding commit](https://github.com/drush-ops/drush/pull/5560/files) from #5560 and the tests pass

I'm not saying there is necessarily anything wrong with that commit. Maybe getting the log messages needs to be done differently now? 

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.x latest dev
| Drupal version? | 10.1 and 9.5
| PHP version     | any
| OS?             | Mac and DrupalCI

**Additional information**
The messages _are_ still written to the terminal when the command is run. It is only in phpunit tests that they are not in the message array.

greg-1-anderson:
That commit is a rather drastic one, in that it stops merging the drush.service.yml files into the Drupal kernel, and instead uses a bespoke adaptor class to instantiate command files. Overall, though, the behavior of Drush should be about the same as it was before, and off the top of my head, can't think of a reason why that logger behavior would be different.

Do you think that you could submit a failing test in Drush core (e.g. maybe in the woot module) that demonstrates the problem?

jonathan1055:
> Do you think that you could submit a failing test in Drush core (e.g. maybe in the woot module) that demonstrates the problem?

Yes sure will. I was wondering the best way to show you succinctly how the behaviour changed. I am totally unfamiliar `woot`. Do you mean add MR for https://github.com/drush-ops/drush/tree/12.x/sut/modules/unish/woot with example code and test?

greg-1-anderson:
Yes; that is a test fixture module, it should be a good place to demonstrate the behavior.

[...]

PR #5601, opened 2023-05-30
Replicate log message test failure

jonathan1055:
Issue #5572 

This first commit adds a test which fails at the latest dev of 12.x
It passes before the mr #5553 was committed.

### DIFF
--- a/sut/modules/unish/woot/drush.services.yml
+++ b/sut/modules/unish/woot/drush.services.yml
@@ -1,11 +1,16 @@
 services:
+  logger.channel.woot:
+    class: Drupal\Core\Logger\LoggerChannel
+    factory: logger.factory:get
+    arguments: ['woot']
   woot.manager:
     class: Drupal\woot\WootManager
-    arguments: ['@current_user']
+    arguments: ['@current_user', '@logger.channel.woot']
     tags:
       -  { name: drush.command }
   woot.command:
     class: Drupal\woot\Commands\WootCommands
+    arguments: ['@woot.manager']
     tags:
       -  { name: drush.command }
   greet.command:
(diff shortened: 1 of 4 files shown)


sut/modules/unish/woot/drush.services.yml
```diff
@@ -1,11 +1,16 @@
 services:
+  logger.channel.woot:
+    class: Drupal\Core\Logger\LoggerChannel
+    factory: logger.factory:get
```
greg-1-anderson, line 4: `drush.services.yml` is deprecated in Drush 12. A legacy service instantiator is provided for backwards compatibility purposes;  it is functional enough to satisfy the needs of most Drush commands, but it is not a full-service DI container. It does not support the `factory` element, for example.

I would suggest trying to set this up using the `create` method. See `WootStaticFactoryCommands` as an example.

greg-1-anderson:
See LegacyServiceInstantiator for the code that emulates the Symfony DI container yml processing mechanism. Simple PRs that increase the compatibility would be welcome, but we don't want the adapter to get too complicated. Ideally, folks will port their command files instead.

jonathan1055:
Thanks for the info. I have two questions though
(a) If I had not raised this issue, how would I know that `drush.services.yml` is deprecated in drush12? Maybe I missed some announcement? Usually when deprecated things are used we get messages. I just got a test failure, which was fortunate.
(b) Where would I find info on how to replace the functionality currently in `drush.services.yml`? I looked in `WootStaticFactoryCommands` for the example you gave, but it's not clear how I would go about making the changes

weitzman:
a) We will be sure to put this into the Drush 12.0.0 release notes. I think a debug message noting use of the deprecated API would be useful. That would go somewhere in https://github.com/drush-ops/drush/blob/12.x/src/Runtime/LegacyServiceInstantiator.php
b) https://www.drush.org/12.x/commands/. Add a create() method and constructor to the commandfile. Remove drush.services.yml

jonathan1055:
Thank you, that's really helpful.

PR #5602, opened 2023-05-31
Fix Drush 12 placed above headers

gitressa:


greg-1-anderson:
Thanks

gitressa:
Thanks for a fast reply and maintaining a wonderful Drupal tool.

PR #5605, opened 2023-05-31
Skip legacy generators

### LINKED ISSUE #5600: drush.services.yml is picked even when requirements are not met
**Describe the bug**
Specifying a `drush.services.yml` file in the `composer.json` of a Drupal module with a restriction that is not fulfilled (such as `^999`) loads the classes regardless.

**To Reproduce**
Change the restriction in a composer.json file of a module to `^999` and verify the commands defined in the `drush.services.yml` are still loaded.

**Expected behavior**
The `drush.services.yml` is ignored

**Actual behavior**
The commands in the `drush.services.yml` file are loaded

**Workaround**
Specify an non-existing yml.
```
    "extra": {
        "drush": {
            "services": {
                "non-existent.services.yml": ">11.99",
                "drush11.services.yml": "<11.99"
            }
        }
    }
```

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.0.0-rc3
| Drupal version? | 10.1.0-beta1
| PHP version     | 8.1
| OS?             | Linux

**Additional information**
I'm trying to provide a backwards compatible way to add v3 generators to the `extra_field` Drupal module
https://git.drupalcode.org/project/extra_field/-/blob/8.x-2.x/drush.services.yml
and I'm looking for a way to ignore the v1 and v2 generators. I think I'm going to use the workaround so far but I found the behavior odd.

It is probably not an easy fix since many Drupal modules declare the services file with something like `"drush.services.yml": "^10"` but that won't match drush 11 but currently it works for drush 11 because one of the files is always picked.

Chi-teck:
Fixes #5600


Chi-teck:
Checked this with extra_field module. Its DCG 2 generators don't produce fatal errors any more.

weitzman:
Oh I didn't realize that we had a problem in drush 12 as well. This looks good to me.

greg-1-anderson:
This will also prevent any hooks or commands bundled inside generator classes from being loaded. There's no way to fix that, since allowing the class file to load would cause it to break. Might be worth mentioning somewhere in the documentation.

Any idea what's up with the archive dump command? Didn't look into it, but on the surface looks like it might be unrelated, so I thought I'd rerun the most recent passing test to see if there are any regressions: https://app.circleci.com/pipelines/github/drush-ops/drush/5522/workflows/7179e927-4573-45e5-bb60-185dcaa29082

**UPDATE:** Yep, previously-passing test now failing in archive dump test, same as this PR.

**UPDATE 2:** Tests passed locally on Composer 2.5.1; upgrading to Composer 2.5.7 caused the failure here. Will fix in a separate PR.

greg-1-anderson:
Addressed issue with updated Composer in https://github.com/drush-ops/drush/pull/5606

greg-1-anderson (approved):

Chi-teck:
> This will also prevent any hooks or commands bundled inside generator classes from being loaded. 

Yes. There is no need for Drush 12 to execute hooks from DCG 2 generators.

PR #5609, opened 2023-06-02
field-create: Make it easier for field types extending entity reference to add target type/bundles options (12.x)

DieterHolvoet:
Same as https://github.com/drush-ops/drush/pull/5556, but for 12.x.

### DIFF
--- a/src/Commands/field/FieldEntityReferenceHooks.php
+++ b/src/Commands/field/FieldEntityReferenceHooks.php
@@ -1,0 +1,132 @@
+<?php
+
+namespace Drush\Commands\field;
+
+use Drupal\Core\Entity\EntityTypeBundleInfoInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drush\Commands\DrushCommands;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class FieldEntityReferenceHooks extends DrushCommands
+{
+    use EntityTypeBundleValidationTrait;
+
+    public function __construct(
+        protected EntityTypeManagerInterface $entityTypeManager,
+        protected EntityTypeBundleInfoInterface $entityTypeBundleInfo,
+    ) {
+    }
+
+    public static function create(ContainerInterface $container): static
+    {
+        return new static(
+            $container->get('entity_type.manager'),
+            $container->get('entity_type.bundle.info'),
+        );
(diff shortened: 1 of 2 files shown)


src/Commands/field/FieldEntityReferenceHooks.php
```diff
@@ -0,0 +1,132 @@
...
+namespace Drush\Commands\field;
+
+use Drupal\Core\Entity\EntityTypeBundleInfoInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drush\Commands\DrushCommands;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+class FieldEntityReferenceHooks extends DrushCommands
```
weitzman, line 14:
```suggestion
final class FieldEntityReferenceHooks extends DrushCommands
```
Lets add final unless we expect folks to extend this.

src/Commands/field/FieldEntityReferenceHooks.php
```diff
@@ -0,0 +1,132 @@
...
+    public static function create(ContainerInterface $container): static
+    {
+        return new static(
+            $container->get('entity_type.manager'),
+            $container->get('entity_type.bundle.info'),
+        );
+    }
+
+    /**
+     * @hook on-event field-create-field-storage
```
weitzman: these should be attributes

src/Commands/field/FieldEntityReferenceHooks.php
DieterHolvoet: We do expect people to extend this, that's the whole point of this PR. This was explained in https://github.com/drush-ops/drush/pull/5556.

src/Commands/field/FieldEntityReferenceHooks.php
DieterHolvoet: Do you know which attribute exactly? I can't seem to find an `OnEvent` attribute.

weitzman: https://github.com/drush-ops/drush/blob/0ddeed7a3f2ff5708135b7c9963194bb51fbfd4a/src/Drupal/Commands/sql/SanitizeCommentsCommands.php#L56
https://github.com/weitzman/drush/blob/0ddeed7a3f2ff5708135b7c9963194bb51fbfd4a/src/Drupal/Commands/sql/SanitizeCommentsCommands.php#L56

weitzman (approved):

PR #5614, opened 2023-06-05
Add drush sset examples for new Twig debugging / caching settings

DieterHolvoet:
As discussed in https://drupal.slack.com/archives/C62H9CWQM/p1684012867841549.

### DIFF
--- a/src/Commands/core/StateCommands.php
+++ b/src/Commands/core/StateCommands.php
@@ -62,9 +62,9 @@ final class StateCommands extends DrushCommands implements StdinAwareInterface
     #[CLI\Usage(name: 'drush sset system.maintenance_mode 1 --input-format=integer', description: 'Put site into Maintenance mode.')]
     #[CLI\Usage(name: 'drush state:set system.cron_last 1406682882 --input-format=integer', description: 'Sets a timestamp for last cron run.')]
     #[CLI\Usage(name: 'php -r "print json_encode(array(\'drupal\', \'simpletest\'));"  | drush state-set --input-format=json foo.name -', description: 'Set a key to a complex value (e.g. array)')]
-    #[CLI\Usage(name: 'drush state:set twig_debug true', description: 'Enable the Twig debug mode (since Drupal 10.1)')]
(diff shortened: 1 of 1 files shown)


DieterHolvoet:
A couple questions:
- Should I create an MR for 11.x as well?
- Do we want a separate command to change these settings, or are we going to stick with `state:set`?
- Should we get rid of the `drush sset system.maintenance_mode` example since there's a separate command now? Should I do that in this MR as well?

weitzman:
Does lowercase `true` as value work right? I only see handling for uppercase.

I think this MR can proceed as is. No need to remove maint example IMO.

DieterHolvoet:
> Does lowercase true as value work right?

No it doesn't, I changed it.

DieterHolvoet:
I do believe separate commands are a better idea. Now, if you want to toggle Twig debugging / caching, you have to run three Drush commands and you have to remember the three internal state keys. People who aren't aware that these values are kept in state - which is an implementation detail if you ask me, end users shouldn't have to care - will not be able to discover how to toggle these values. If we added commands, running `drush` would list them and people would immediately know what to do. Here's what I propose:
```
# Enable _Twig development mode_ (including _Twig debug mode_ & _Disable Twig cache_)
drush twig:dev 1
# Enable _Twig development mode_ and _Twig debug mode_ (not including _Disable Twig cache_)
drush twig:debug 1
# Enable _Twig development mode_ and _Disable Twig cache_ (not including _Twig debug mode_)
drush twig:cache 0
# Do not cache markup (disables render cache, dynamic page cache and page cache)
drush cache:markup 0
```

weitzman:
Yes, separate commands is better for those themers who need them. But that needs to balance against the clutter in the command list for those that dont. There are far more who dont. So, its a balance. I'm mulling it. We can (and have) add a ton of commands to core that are great but only apply to a small audience.

DieterHolvoet:
Okay, fair point! In that case I think this is good to go?

gitressa:
@DieterHolvoet Maybe you could be interested in #6267?

PR #5629, opened 2023-06-06
Sanity check service data in legacy service instantiator

greg-1-anderson:
Validate that service data looks reasonable before instantiating any services. Skip any service file that we do not have confidence in.

### DIFF
--- a/src/Runtime/LegacyServiceInstantiator.php
+++ b/src/Runtime/LegacyServiceInstantiator.php
@@ -99,8 +99,7 @@ class LegacyServiceInstantiator
      */
     protected function allServicesHaveClassElement(string $serviceFile, array $services)
     {
-        foreach ($services as $service => $data)
-        {
+        foreach ($services as $service => $data) {
             if (!isset($data['class'])) {
                 $this->logger->warning(dt('Service @service does not have a class element; skipping @file', ['@service' => $service, '@file' => $serviceFile]));
                 return false;
--- a/src/Runtime/LegacyServiceInstantiator.php
+++ b/src/Runtime/LegacyServiceInstantiator.php
@@ -11,6 +11,7 @@ use Symfony\Component\Console\Application;
 use Composer\Autoload\ClassLoader;
 use Drush\Command\DrushCommandInfoAlterer;
 use Psr\Container\ContainerInterface;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\Yaml\Yaml;
 use Robo\Robo;
 use Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException;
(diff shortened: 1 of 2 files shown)


src/Runtime/LegacyServiceInstantiator.php
```diff
@@ -55,12 +57,58 @@ public function loadServiceFiles(array $serviceFiles)
...
         }
     }
 
+    /**
+     * Validate service data before using it.
+     *
+     * @param string $serviceFile Path to service file being checked
+     * @param array $serviceFileData Parsed data from drush.services.yml
+     */
+    protected function isValidServiceData(string $serviceFile, array $serviceFileData)
```
weitzman:
```suggestion
    protected function isValidServiceData(string $serviceFile, array $serviceFileData): bool
```

src/Runtime/LegacyServiceInstantiator.php
```diff
@@ -55,12 +57,58 @@ public function loadServiceFiles(array $serviceFiles)
...
+    {
+        // If there are no services, then silently skip this service file.
+        if (!isset($serviceFileData['services'])) {
+            return false;
+        }
+
+        // We don't support auto-wiring
+        if (!empty($serviceFileData['services']['_defaults']['autowire'])) {
+            $this->logger->warning(dt('Autoloading not supported; skipping @file', ['@file' => $serviceFile]));
```
weitzman:
```suggestion
            $this->logger->warning(dt('Autowire not supported; skipping @file', ['@file' => $serviceFile]));
```

src/Runtime/LegacyServiceInstantiator.php
```diff
@@ -55,12 +57,58 @@ public function loadServiceFiles(array $serviceFiles)
...
+        return true;
+    }
+
+    /**
+     * Check all elements for required "class" elements.
+     *
+     * @param string $serviceFile Path to service file being checked
+     * @param array $services List of data from 'services' element from drush.services.yml
+     */
+    protected function allServicesHaveClassElement(string $serviceFile, array $services)
```
weitzman:
```suggestion
    protected function allServicesHaveClassElement(string $serviceFile, array $services): bool
```

src/Runtime/LegacyServiceInstantiator.php
```diff
@@ -55,12 +57,58 @@ public function loadServiceFiles(array $serviceFiles)
...
+    protected function isValidServiceData(string $serviceFile, array $serviceFileData): bool
+    {
+        // If there are no services, then silently skip this service file.
+        if (!isset($serviceFileData['services'])) {
+            return false;
+        }
+
+        // We don't support auto-wiring
+        if (!empty($serviceFileData['services']['_defaults']['autowire'])) {
+            $this->logger->warning(dt('Autowire not supported; skipping @file', ['@file' => $serviceFile]));
```
weitzman: I think we should downgrade these to info(). They are typically going to be from commandfiles that you dont care about. Its not like user is trying to run a command in the file.

PR #5641, opened 2023-06-08
TableEmptyMessage attribute

greg-1-anderson:
If this looks okay, then we can merge https://github.com/consolidation/annotated-command/pull/303 and https://github.com/consolidation/output-formatters/pull/108 and make releases.

Feel fee to comment on those PRs if you have any comments on naming things, or any other detail.


src/Attributes/TableEmptyMessage.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drush\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_METHOD)]
+class TableEmptyMessage extends \Consolidation\AnnotatedCommand\Attributes\TableEmptyMessage
```
greg-1-anderson, line 10: It seems really unfortunate that we need to duplicate all of the attributes in the \Consolidation\AnnotatedCommand\Attributes namespace here.

src/Commands/pm/PmCommands.php
```diff
@@ -235,6 +235,7 @@ public function validateUninstall(CommandData $commandData): void
     ])]
     #[CLI\DefaultTableFields(fields: ['package', 'display_name', 'status', 'version'])]
     #[CLI\FilterDefaultField(field: 'display_name')]
+    #[CLI\TableEmptyMessage(message: 'There are no items to list')]
```
greg-1-anderson, line 238: Run `drush pm:list --filter=package=Core` to see a list of core modules; run `drush pm:list --filter=package=Foo` to see the empty message.

src/Attributes/TableEmptyMessage.php
weitzman: I've accepted it in favor of having one namespace to autocomplete attribute names in an IDE. Custom code is welcome to use AC attributes directly as https://github.com/drush-ops/drush/blob/12.x/tests/fixtures/lib/Drush/Commands/ExampleAttributesDrushCommands.php#L80-L95 does.

weitzman (approved):
LGTM. Thanks.

weitzman:
Looks like https://github.com/consolidation/annotated-command/pull/303 needs to get merged still @greg-1-anderson

Issue #5643, opened 2023-06-09
SqlPgsql::dbExists checking exit code is not enough


Sweetchuck:
## Problem

[\Drush\Sql\SqlPgsql::dbExists()](https://github.com/drush-ops/drush/blob/11.x/src/Sql/SqlPgsql.php#L92-L92)
Always returns TRUE with PostgreSQL-15.

Exit code of the command `psql --no-psqlrc -t -c "SELECT 1 AS result FROM pg_database WHERE datname='foo'"` shows that the query ran successfully.
It doesn't care about the number of records.

When the database **IS** exists, then the stdOutput is `      1\n\n`
When the database **IS NOT** exists, then the stdOutput is `\n`. It is not an empty string, so it is converted to `TRUE`.

## Suggested solution

```php
return $process->isSuccessful()
    && trim($process->getOutput()) === '1';
```

_Originally posted by @Sweetchuck in https://github.com/drush-ops/drush/issues/5545#issuecomment-1581484525_

PR #5645, opened 2023-06-09
Issue #5643 - 11.x - Fix return value of the \Drush\Sql\SqlPgsql::dbExists()

Sweetchuck:
Issue #5643 - Fix return value of the \Drush\Sql\SqlPgsql::dbExists()


src/Sql/SqlPgsql.php
```diff
@@ -101,7 +101,9 @@ public function dbExists(): bool
         $process = Drush::shell($sql_no_db->connect() . ' -t -c ' . $query, null, $this->getEnv());
         $process->setSimulated(false);
         $process->run();
-        return $process->isSuccessful();
+
+        return $process->isSuccessful()
+            && trim($process->getOutput() === '1');
```
weitzman:
```suggestion
            && trim($process->getOutput()) === '1';
```

PR #5661, opened 2023-06-15
Return new user in user:create. Fixes #5660

### LINKED ISSUE #5660: Unable to capture user:create output from cy.exec()
**Describe the bug**
In Cypress, I'm using the following code to capture the output of user:create:

```
cy.exec(command, {failOnNonZeroExit: false}).then( (result) => {
  output = result.stdout
  console.log("execSync: " + output)
  return cy.wrap(output)
})

```
The command works but I receive no result when this is passed:
`drush  user:create  "Jayda Kozey" --mail="jayda.kozey@ethereal.email" --password="ManbQBMZ5ERXPWbQ2Y"`

However, the same code (which is [standard](https://docs.cypress.io/api/commands/exec#Yields-Icon-namequestion-circle)) is able to capture other output from Drush commands via cy.exec(), such as:

drush user:info --mail="jayda.kozey@ethereal.email" --format=json

"Output" is:

```
{
    "23": {
        "uid": "23",
        "name": "Jayda Kozey",
        "mail": "jayda.kozey@ethereal.email",
        "roles": [
            "authenticated"
        ],
        "user_status": "1"
    }
}
```

**To Reproduce**
In Cypress (or perhaps any other Node shell), run the cy.exec() and test the output.

**Expected behavior**
Should receive
`[success] Created a new user with uid 57`

Then I can match against "/Created a new user with uid ([0-9]+)/" to obtain the UID.

**Actual behavior**
"Output" above is ''.

**Workaround**
Once the account is created, use drush user:info with the account email and --format=json. Search for the UID in the json object (see below).

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 11.6
| Drupal version? | 10.0.9
| PHP version     | 8.1.3
| OS?             | macOs

**Additional information**
It would be great to correct this. However, another possible solution is to allow --format=json with user:create so that the UID is returned as part of the same structure as user:info, including all roles, etc. after all hooks have fired. (Or before the hooks, that's not the important bit.) 

Something like:

drush user:create "Jayda Kozey" --mail="jayda.kozey@ethereal.email" --password --format=json

"Output:" 

```
{
    "23": {
        "uid": "23",
        "name": "Jayda Kozey",
        "mail": "jayda.kozey@ethereal.email",
        "roles": [
            "authenticated",
            "custom_user_role_from_rules_or_similar"
        ],
        "user_status": "1"
    }
}
```

weitzman:


aangelinsf:
Thank you, will give it a whirl and report back if there are any issues.

Issue #5663, opened 2023-06-15
Enable drush archive-dump --extra (drush 8.x)


AndrewGearhart:
**Is your feature request related to a problem? Please describe.**
Drush archive-dump does not allow for sql flags such as --no-tablespaces which makes it fail without separate configuration of the database server. 

**Describe the solution you'd like**
It would be great if the same --extra flag that is available to the sql-dump command would work with drush archive-dump (drush ard). This allows for that flag to be used when necessary.

**Describe alternatives you've considered**
This can be solved with a user .my.cnf, however, you are then setting that setting for all mysqldump commands. You can also solve this by setting the process permission which is considered by the mysql maintainers to be a security issue.

**Additional context**
By adding this flag as a command option, it can use the same underlying functionality as drush sql-dump to create the mysqldump command and allow people to successfully use drush ard as they continue to maintain their Drupal 7 sites on the way to Drupal 10!

weitzman:
Drush 8 is getting security/significant fixes only. I suggest using drush configuration for sql-dump command.

PR #5664, opened 2023-06-15
fixes drush-ops/drush#5663 

AndrewGearhart:
adds --extra parameter to archive-dump to pass flag arguments to sql-dump command (fixing the ability to use drush ard with a normal user that doesn't have the processlist permission and who cannot/won't use the --no-tablespaces flag as a default setting)


weitzman:
Folks are welcome to use composer-patches to bring this into their builds. Drush 8 is not taking any new features. 

PRs to Drush 12 are still welcome of course.

PR #5666, opened 2023-06-17
Remove drush.bat

weitzman:


weitzman:
File is still in use by windows tests.

Issue #5109, opened 2022-03-31
Create command equivalent to Drupal console's site:mode


malcomio:
We are currently removing Drupal Console from our project dependencies.
One useful feature that it provides is the [site:mode command](https://drupalconsole.com/docs/en/commands/site-mode), which changes system performance configuration to "dev" or "prod" mode.

Ideally Drush would provide a similar command.

Here's the [source for the Console command](https://github.com/hechoendrupal/drupal-console/blob/90053d30f52427edb4e4941a9063acb65b5a2c1e/src/Command/Site/ModeCommand.php) - setting dev mode does the following:

```
 Configuration name: system.performance
 ------------------------- ---------------- ---------------- 
  Configuration key         Original Value   Override Value  
 ------------------------- ---------------- ---------------- 
  cache.page.use_internal                    false           
  css.preprocess            true             false           
  css.gzip                  true             false           
  js.preprocess             true             false           
  js.gzip                   true             false           
  response.gzip                              false           
 ------------------------- ---------------- ---------------- 

 Configuration name: views.settings
 -------------------------------- ---------------- ---------------- 
  Configuration key                Original Value   Override Value  
 -------------------------------- ---------------- ---------------- 
  ui.show.sql_query.enabled        false            true            
  ui.show.performance_statistics   false            true            
 -------------------------------- ---------------- ---------------- 

 Configuration name: system.logging
 ------------------- ---------------- ---------------- 
  Configuration key   Original Value   Override Value  
 ------------------- ---------------- ---------------- 
  error_level         verbose          all             
 ------------------- ---------------- ---------------- 

 New services settings
 ------------------------------------------ ------------- ------- 
  Service                                    Parameter     Value  
 ------------------------------------------ ------------- ------- 
  http.response.debug_cacheability_headers                 true   
  twig.config                                auto_reload   true   
  twig.config                                cache         false  
  twig.config                                debug         true   
 ------------------------------------------ ------------- ------- 

 // cache:rebuild
```

gitressa:
This would be a great feature to have in Drush.

weitzman:
The needs here are varied enough that I dont want to maintain this in core Drush. Users who want this have a few options:

1. Make a custom script that does exactly what you need
2. Make a Drush command and share it with the world. It can be a module or a standalone composer package.

gitressa:
@malcomio Maybe you could be interested in #6267?

PR #5668, opened 2023-06-20
Twig debug

Chi-teck:
This adds Drush command to enable/disable Twig debug mode.

This was previously requested as part of #5109. However, the proposed `site:mode` command is too wide, indeed. And it may need a lot of options to fit everyone's needs.

Unlike most of `site:mode` settings Twig debug mode cannot be set through settings.php or `drush cset`. It requires changing Twig parameters in `development.services.yml` file and rebuilding container. However, since Drupal 10.1.0 there is [a way](https://www.drupal.org/node/3359728) to enable it through Drupal states. So that the `drush twig:debug` command can be really simple.

Relevant issue on drupal.org
https://www.drupal.org/project/drupal/issues/3363676


src/Commands/core/TwigCommands.php
```diff
@@ -117,4 +122,32 @@ public function twigCompile(): void
             $this->logger()->success(dt('Compiled twig template !path', ['!path' => $relative]));
         }
     }
+
+    /**
+     * Enables Twig debug and disables caching Twig tempaltes.
```
weitzman:
```suggestion
     * Enables Twig debug and disables caching Twig templates.
```

src/Commands/core/TwigCommands.php
```diff
@@ -117,4 +122,32 @@ public function twigCompile(): void
...
+     *
+     * @see \Drupal\system\Form\DevelopmentSettingsForm::submitForm()
+     */
+    #[CLI\Command(name: self::DEBUG, aliases: ['twig-debug'])]
+    #[CLI\Argument(name: 'mode', description: 'Debug mode (on/off).')]
```
weitzman:
```suggestion
    #[CLI\Argument(name: 'mode', description: 'Debug mode. Recognized values: <info>on</info>, <info>off</info>.', suggestedValues: ['on', 'off'])]
```

weitzman (changes requested):

Chi-teck:
Fixed version number comparison.

gitressa:
Thanks! I tried enabling this feature, but can't see any changes ... where should I see the effect of this command?

```
$ drush twig:debug on
 [OK] Enabled twig debug.
```

Chi-teck:
That's just a CLI interface for admin/config/development/settings.

Apparently, it's no longer working because of https://www.drupal.org/node/3437162

weitzman:
Ugh. I wish Drupal would make an API for stuff like this. Then callers like Drush would be isolated from implementation shanges. Raw key-value is not an API.

PR #5677, opened 2023-06-27
Bring back Drupal kernel terminate event

weitzman:


src/Boot/DrupalBoot8.php
```diff
@@ -31,6 +31,7 @@ class DrupalBoot8 extends DrupalBoot
 
     public function __construct(protected ServiceManager $serviceManager, protected $autoloader)
     {
+        register_shutdown_function([$this, 'terminate']);
```
greg-1-anderson: Would it be better to leave this in BaseBoot and call parent::__construct here? If we're never going to have more than one kind of bootstrap class any more, then maybe we should instead simplify further. That might be a Drush 12 thing, though.

PR #5678, opened 2023-06-28
Add argument/option autocompletion to field commands

DieterHolvoet:
I added autocompletion to the `entityType`/`bundle` arguments of the field commands. Not sure if we have a way to test these kind of things yet.


weitzman:
Lgtm. Feel free to merge when satisfied.

Issue #5686, opened 2023-07-05
Drush installation is not detected for other site alias when having multiple local Drupal installations with aliases


ordermind:
**Describe the bug**
I have multiple local Drupal installations with site aliases. When I run a drush command for another site, I get the error message "[preflight] The alias @<alias> references a Drupal site that does not contain its own copy of Drush. Please add Drush to this site to use it."

**To Reproduce**
1. Setup two Drupal installations on the same server and install drush on both of them.
2. Define site aliases for the two installations
3. cd into the first installation and then run for example `drush @<other-site-alias> status`
4. You will now get the message "[preflight] The alias @<other-site-alias> references a Drupal site that does not contain its own copy of Drush. Please add Drush to this site to use it."

**Expected behavior**
The message should not be displayed as Drush is installed on both sites.

**Actual behavior**
The message is displayed, indicating that Drush needs to be installed on the other site even though it is already installed.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.x (latest dev release)
| Drupal version? | 10.0.9
| PHP version     | 8.1
| OS?             | Linux

**Additional information**
I dug around in the code and found some weird return types/values which seem to be causing this issue so I created a potential fix that I will include with this issue.

weitzman:
Lets discuss on the PR

PR #5687, opened 2023-07-05
#5686 Fix issue "Drush installation is not detected for other site al…

ordermind:
…ias when having multiple local Drupal installations with aliases"

### DIFF
--- a/src/Preflight/RedispatchToSiteLocal.php
+++ b/src/Preflight/RedispatchToSiteLocal.php
@@ -23,27 +23,26 @@ class RedispatchToSiteLocal
      * @param string $vendor The path to the vendor directory
      * @param PreflightLog $preflightLog A basic logger.
      *
-     * @return bool
-     *   True if redispatch occurred, and was returned successfully.
+     * @return array [$preflightDidRedispatch, $exitStatus]
      */
-    public static function redispatchIfSiteLocalDrush(array $argv, string $root, string $vendor, PreflightLog $preflightLog): bool
+    public static function redispatchIfSiteLocalDrush(array $argv, string $root, string $vendor, PreflightLog $preflightLog): array
     {
 
         // Try to find the site-local Drush. If there is none, we are done.
         $siteLocalDrush = static::findSiteLocalDrush($root);
         if (!$siteLocalDrush) {
-            return false;
+            return [false, 0];
         }
 
         // If the site-local Drush is us, then we do not need to redispatch.
         if (Path::isBasePath($vendor, $siteLocalDrush)) {
-            return false;
+            return [false, 0];
         }
 
         // Do another special check to detect symlinked Drush folder similar
         // to what the SUT sets up for Drush functional tests.
         if (dirname($vendor) === dirname($siteLocalDrush)) {
-            return false;
+            return [false, 0];
         }
 
         // Redispatch!
@@ -59,11 +58,7 @@ class RedispatchToSiteLocal
         $command .= ' ' . implode(' ', $args);
         passthru($command, $status);
 
-        // The $status variable is the exit status of the process, which is an
-        // integer. This method is supposed to return true when successful, so
-        // we need to convert it to a boolean accordingly.
-
-        return $status ? false : true;
+        return [true, $status];
     }
 
     /**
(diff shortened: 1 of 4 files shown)


ordermind:
fixes #5686

src/Runtime/Runtime.php
```diff
@@ -55,11 +55,11 @@ public function run($argv): int
     protected function doRun($argv, $output): int
     {
         // Do the preflight steps
-        $status = $this->preflight->preflight($argv);
+        $preflightSuccess = $this->preflight->preflight($argv);
```
greg-1-anderson, line 58: This code should remain as it was.

src/Preflight/RedispatchToSiteLocal.php
```diff
@@ -58,7 +58,12 @@ function ($item) {
         );
         $command .= ' ' . implode(' ', $args);
         passthru($command, $status);
-        return $status;
+
+        // The $status variable is the exit status of the process, which is an
+        // integer. This method is supposed to return true when successful, so
+        // we need to convert it to a boolean accordingly.
+
+        return $status ? false : true;
```
greg-1-anderson, line 66: This code should remain as it was.

src/Runtime/Runtime.php
greg-1-anderson: `false` does not indicate failure, so I would rename this something like `$preflightDidRedispatch`

src/Runtime/Runtime.php
```diff
@@ -55,11 +55,11 @@ public function run($argv): int
...
 
         // If preflight signals that we are done, then exit early.
-        if ($status) {
-            return $status;
+        if ($preflightSuccess) {
```
greg-1-anderson: Here is the real bug. The original code was [if ($status !== false)](https://github.com/drush-ops/drush/commit/1a7704edfe99a7a223f65154a823318e2ca09499#diff-fcb54a74a331c6112bd5cf5e32f08a4d0a39f42c16833b899b4fac0339d453c7R219). That is, Preflight would return `bool false` if `doRun` should continue, and an integer status code if a redispatch happened. This sort of coding convention was common in the PHP 5 era, but is no longer considered de rigor in PHP 8, and in fact is a deprecated practice.

src/Preflight/RedispatchToSiteLocal.php
greg-1-anderson: This function should always return true if it gets to this point. The comment in the header that says "and was returned successfully" was in error. See my other comment for a description of how the code used to work.

I think that the best fix would be to change this function to use `return [false, 0];` wherever it currently uses `return false`, and change this line to `return [true, $status];`. Then, `doRun` must record both return values, call them `$preflightDidRedispatch` and `$status` or something similar. It should then return `$status` if `$preflightDidRedispatch` is true.

greg-1-anderson (changes requested):
Thanks for looking at this. The original bug was caused by optimizing away a type check (`!== false`). I suggest a better fix would be to change the return tupe of the preflight function. Since preflight is an internal function, I don't think we need to consider this a breaking change.

src/Runtime/Runtime.php
ordermind: Thanks for the feedback, this should be fixed now.

src/Runtime/Runtime.php
ordermind: Thanks for the feedback, this should be fixed now.

src/Preflight/RedispatchToSiteLocal.php
ordermind: Thanks for the feedback, this should be fixed now.

greg-1-anderson (approved):
LGTM, Thank you for working on this problem.

@weitzman Does this resolution look OK to you?

weitzman:
Yes, LGTM. The one suggesion I have is to replace the various new `0` with `\Drush\Commands\DrushCommands::EXIT_SUCCESS`

greg-1-anderson (changes requested):
Use \Drush\Commands\DrushCommands::EXIT_SUCCESS as suggested by @weitzman

weitzman (approved):

PR #5693, opened 2023-07-07
Add FieldTextHooks

### LINKED ISSUE #5692: field:create: Add support for allowed text formats
It doesn't look like https://www.drupal.org/project/allowed_formats/issues/3254013 is ever going to be committed. Which is fine, because allowed text formats were added to Drupal core: https://www.drupal.org/project/drupal/issues/784672

Add a `FieldTextHooks` class with hooks adding an `allowed-formats` option to the `field:create` command. We'll be able to reuse the majority of the code from this MR: https://git.drupalcode.org/project/allowed_formats/-/merge_requests/2/diffs.

Since it was only added in Drupal 10.1, we should add a check to make sure the functionality is already present. Maybe somehow check the `field.field_settings.text` config schema for an `allowed_formats` key.

DieterHolvoet:
Fixes #5692.

### DIFF
--- a/tests/functional/FieldTest.php
+++ b/tests/functional/FieldTest.php
@@ -65,7 +65,7 @@ class FieldTest extends CommandUnishTestCase
         // Allowed formats
         $this->drush('field:create', ['unish_article', 'alpha'], ['field-name' => 'field_test4', 'field-label' => 'Text', 'field-type' => 'string', 'allowed-formats' => 'minimal'], null, null, self::EXIT_ERROR);
         $this->assertStringContainsString('The "--allowed-formats" option does not exist.', $this->getSimplifiedErrorOutput());
-        $this->drush('field:create', ['unish_article', 'alpha'], ['field-name' => 'field_test4', 'field-label' => 'Text', 'field-type' => 'text_long', 'allowed-formats' => 'baz'], null, null, self::EXIT_ERROR);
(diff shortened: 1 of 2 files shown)


DieterHolvoet:
@weitzman the test fails because this feature needs Drupal 10.1 to work. What should I do here? Increase the core version for the whole repo or just leave out the tests for now?

tests/functional/FieldTest.php
```diff
@@ -59,6 +59,14 @@ public function testFieldCreate()
         $this->assertStringContainsString('Success', $this->getErrorOutputRaw());
         $this->drush('field:create', ['unish_article', 'beta'], ['existing-field-name' => 'field_test3', 'field-label' => 'Body', 'field-widget' => 'text_textarea_with_summary'], null, null, self::EXIT_ERROR);
         $this->assertStringContainsString('Field with name \'field_test3\' already exists on bundle \'beta\'', $this->getErrorOutputRaw());
+
```
weitzman:
```suggestion
if (version_compare(\Drupal::VERSION, '10.1.0') < 0) {
    $this->markTestSkipped('Allowed formats available since Drupal 10.1.0');
}
```

weitzman (approved):

tests/functional/FieldTest.php
DieterHolvoet: Okay, cool. Will these tests currently ever run by doing this? Is Drush being tested using multiple Drupal versions, including Drupal 10.1?

weitzman:
At this time, the new testing lines will only get run in the test-sqlite-highest-functional job.

PR #5700, opened 2023-07-10
Defensive code in options hook for field:create

weitzman:
See https://github.com/drush-ops/drush/blob/958d93ecb11d850f9c9c03ee69e6f1599ecac14d/src/Commands/core/MkCommands.php#L358
https://github.com/weitzman/drush/blob/958d93ecb11d850f9c9c03ee69e6f1599ecac14d/src/Commands/core/MkCommands.php#L358


DieterHolvoet:
~Shouldn't we look into why this is throwing an exception? The fact that that option is missing smells like a bug to me.~ Missed your first comment. Fair enough, let's fix it this way.

DieterHolvoet (approved):

PR #5709, opened 2023-07-19
PsySh quick entity access

chx:


docs/repl.md
```diff
@@ -1 +1,3 @@
 The [php:cli command](commands/php_cli.md) is interactive PHP REPL with your bootstrapped site (remote or local). It’s a Drupal code playground. You can do quick code experimentation, grab some data, or run Drush commands. This can also help with debugging certain issues. See [this blog post](http://blog.damiankloip.net/2015/drush-php) for an introduction. Run `help` for a list of commands. 
+
+Entity classes are available without their namespace. For example, Node::load() works instead of Drupal\Node\entity\Noad::load().
```
weitzman, line 3:
```suggestion
Entity classes are available without their namespace. For example, Node::load() works instead of Drupal\Node\entity\Node::load().
```

weitzman (commented):
Could you also please add an Example to the cli command's help?

Issue #5712, opened 2023-07-21
Fail Drush Deploy if Database Updates Produce a Diff Against Tracked Configuration


adamzimmermann:
> If an update hook modifies configuration, then old configuration is imported, the changes made by the update hook are forever lost.
- https://www.drupal.org/project/drupal/issues/3110362

```
if (update hook changes config values to values different from config files) {
  abort deploy process
}
```

**Describe the solution you'd like**
```
$existing_diff = drush config:status
drush updatedb
$new_diff = drush config:status
if (diff_exists($existing_diff, $new_diff)) {
  fail the command?
}
```

**Describe alternatives you've considered**
We have talked about several, but nothing panned out as a serious viable alternative.

**Additional context**

Slack Discussion:

> It's not OK for updatedb to change config to values that differ from the deployment's saved config". Alas this complicates things badly.
- https://drupal.slack.com/archives/C45342CDD/p1689778091247189

weitzman:
Lets talk in PR

adamzimmermann:
Works for me. I'm unsure what the next steps there are though. I'll look to you for guidance.

PR #5713, opened 2023-07-21
Fail Drush Deploy if Database Updates Produce a Diff Against Tracked Configuration (#5389)

adamzimmermann:
## Motivation

- https://github.com/drush-ops/drush/issues/5712

## Known Issues

- I targeted this against 11.x as the site I'm working on daily won't let me get to 12.x due to issues I can't resolve at the moment. I'm happy to port this to 12.x once we agree on the general approach. This just helps me test it more easily.
- Adding some better logging/visibility into the differences it detected would be great. I'm happy to add that too once we have an agreement on the general approach/messaging we want.
- I'm unsure if the `return` is the preferred way to abort a process here, but figured someone with more knowledge could quickly point me in the right direction.

## Current State

I tested this locally and it does work though. I added an update hook that simply did this and it failed as expected. ✅ 
```
/**
 * Force a failure in the drush deploy command.
 */
function example_update_8004() {
  \Drupal::service('module_installer')->uninstall(['some_module']);
}
```

Screenshot 2023-07-21 at 10 34 19 AM

### DIFF
--- a/src/Commands/core/DeployCommands.php
+++ b/src/Commands/core/DeployCommands.php
@@ -36,7 +36,7 @@ class DeployCommands extends DrushCommands implements SiteAliasManagerAwareInter
 
         // Prepare custom options for checking configuration state.
         $configStatusOptions = $redispatchOptions;
-        $configStatusOptions['format'] = 'php';
+        $configStatusOptions['format'] = 'json';
         // Record the current state of configuration.
         $process = $manager->drush($self, 'config:status', [], $configStatusOptions);
         $process->mustRun();
--- a/src/Commands/core/DeployCommands.php
+++ b/src/Commands/core/DeployCommands.php
@@ -52,7 +52,7 @@ class DeployCommands extends DrushCommands implements SiteAliasManagerAwareInter
         // Record the current state of configuration.
         $process = $manager->drush($self, 'config:status', [], $configStatusOptions);
         $process->mustRun();
-        $newConfigDiff = unserialize($process->getOutput());
+        $newConfigDiff = $process->getOutputAsJson();
 
         // Check for new changes to active configuration that would be lost.
         if ($originalConfigDiff !== $newConfigDiff) {
--- a/src/Commands/core/DeployCommands.php
+++ b/src/Commands/core/DeployCommands.php
@@ -40,7 +40,7 @@ class DeployCommands extends DrushCommands implements SiteAliasManagerAwareInter
         // Record the current state of configuration.
         $process = $manager->drush($self, 'config:status', [], $configStatusOptions);
         $process->mustRun();
-        $originalConfigDiff = unserialize($process->getOutput());
+        $originalConfigDiff = $process->getOutputAsJson();
 
         $this->logger()->notice("Database updates start.");
         $options = ['no-cache-clear' => true];
--- a/src/Commands/core/DeployCommands.php
+++ b/src/Commands/core/DeployCommands.php
@@ -5,6 +5,7 @@ namespace Drush\Commands\core;
 use Consolidation\SiteAlias\SiteAlias;
 use Consolidation\SiteAlias\SiteAliasManagerAwareTrait;
 use Consolidation\SiteProcess\ProcessManager;
+use Drush\Drupal\Commands\config\ConfigCommands;
 use Drush\Commands\DrushCommands;
 use Drush\Drush;
 use Drush\SiteAlias\SiteAliasManagerAwareInterface;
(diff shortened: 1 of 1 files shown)


src/Commands/core/DeployCommands.php
```diff
@@ -33,13 +34,32 @@ public function deploy(): void
         $redispatchOptions = Drush::redispatchOptions();
         $manager = $this->processManager();
 
+        // Prepare custom options for checking configuration state.
+        $configStatusOptions = $redispatchOptions;
+        $configStatusOptions['format'] = 'php';
```
weitzman:
```suggestion
        $configStatusOptions['format'] = 'json';
```

src/Commands/core/DeployCommands.php
```diff
@@ -33,13 +34,32 @@ public function deploy(): void
...
+        // Record the current state of configuration.
+        $process = $manager->drush($self, 'config:status', [], $configStatusOptions);
+        $process->mustRun();
+        $originalConfigDiff = unserialize($process->getOutput());
```
weitzman:
```suggestion
        $originalConfigDiff = $process->getOutputAsJson();
```

src/Commands/core/DeployCommands.php
```diff
@@ -33,13 +34,32 @@ public function deploy(): void
...
         $options = ['no-cache-clear' => true];
         $process = $manager->drush($self, 'updatedb', [], $options + $redispatchOptions);
         $process->mustRun($process->showRealtime());
 
         $this->cacheRebuild($manager, $self, $redispatchOptions);
 
+        // Record the current state of configuration.
+        $process = $manager->drush($self, 'config:status', [], $configStatusOptions);
+        $process->mustRun();
+        $newConfigDiff = unserialize($process->getOutput());
```
weitzman:
```suggestion
        $newConfigDiff = $process->getOutputAsJson();
```

src/Commands/core/DeployCommands.php
```diff
@@ -33,13 +34,32 @@ public function deploy(): void
...
+
+        // Check for new changes to active configuration that would be lost.
+        if ($originalConfigDiff !== $newConfigDiff) {
+            $this->io()->error('Update hooks altered config that is about to be reverted during config import. Aborting.');
+            return;
```
weitzman:
```suggestion
            throw new \RuntimeException('Update hooks altered config that is about to be reverted during config import. Aborting.');
```

weitzman (commented):
Nice work! Also needs to go into 12.x first, and test coverage for should be expanded for deploy - that can be a followup.

src/Commands/core/DeployCommands.php
apotek: Curious what the motivation is to use JSON here given that this is all PHP tooling and a php native format will process more efficiently, 

JSON makes sense if you want to serialize between languages but I don't see the reason for it here.

weitzman: > php native format will process more efficiently,

Do you have a source for that claim? json_encode()/json_decode() are C functions and should process just as fast as serialize() and unserialize(). the motivation is that we already have a getOutputAsJson() function so using JSON saves some code (i.e. the unserialize() call).

weitzman:
Could I get permission to push to your branch? Or you can accept my suggestions from the UI. This looks great to me. Lets address those Known Issues and then we can merge this.

src/Commands/core/DeployCommands.php
adamzimmermann: Don't we want this if we are doing your other suggestion too? I see this is resolved though. 🤔

adamzimmermann:
I won't be on my computer much today, but believe I have the permissions adjusted. I'm happy to help with all of the needed adjustments next week, but I won't be around much this weekend. Hopefully I've unblocked things though. 🤞🏼

src/Commands/core/DeployCommands.php
adamzimmermann: Did you intend to resolve this?

src/Commands/core/DeployCommands.php
apotek: Here's an actual test result for versions of PHP from 5.4 to 8.27: https://3v4l.org/qFKvT.  It's (very crudely averaged) about 15-25% faster. The requirements to implement a parser for various data formats can lead to vastly different time complexities in the parsing routine (does the format allow for nesting, etc etc).

Thank you for prompting me to re-examine my previous understanding of the matter. Because while my comment came from a pure performance consideration (as well as the desire to push against the make-everything-JSON impulse at the moment), there are other, perhaps more important things to consider, and one of those is safety:  While `unserialize` does perform better, there is a drawback with using unserialize with regards to execution safety of the encoded data. I'm not sure how relevant that is in this case given that the output should not contain any user input. But it is a factor in favor of using JSON if there is even a slight chance that the input could be manipulated.

If we were trying to unserialize true objects, PHP's native unserializer would be the obvious winner as `json_decode` cannot restore PHP class types (nor magic methods and protected/private properties). But that's not what we are doing here. 

If we were exporting data that needed to be portable, JSON (or XML) would be the obvious choice rather than php serialized data. 

I guess it doesn't really matter. Raw speed would tilt in favor of `unserialize`. But if the drush community just generally prefers JSON for data portability or safety reasons, that's a good enough argument for me.

In the old days (pre php 8) JSON was not compiled into PHP by default so I would have made a stronger argument against using it here. But that's no longer a factor.

markdorison:
Do we think this can be the default behavior? I would love it to be, but will this change be disruptive to some?

weitzman:
I think it should only be disruptive when there is a genuine problem. Thus I think it runs always during a deploy. We can add a flag later if a need arises.

adamzimmermann:
Thank you all for chiming in and pushing this forward. 🙏🏼 

I'm hoping to find some time later today to target this to the 12.x branch and improve the logging.

Anything else that I should add to this before we consider is ready for a more final review?

I might need some help with test coverage too, but I'll take a swing at it.

apotek:
> Thank you all for chiming in and pushing this forward. 🙏🏼
> 
> I'm hoping to find some time later today to target this to the 12.x branch and improve the logging.
> 
> Anything else that I should add to this before we consider is ready for a more final review?
> 
> I might need some help with test coverage too, but I'll take a swing at it.

Thank you for your great work on this @adamzimmermann . The language (output) around this will be critical to a variety of users understanding why their deployment was blocked, and it strikes me that designing the user feedback will end up being a tangle in and of itself :)

adamzimmermann:
Closing in favor of https://github.com/drush-ops/drush/pull/5718.

PR #5714, opened 2023-07-23
Run batch process when (un)installing modules via config:import

claudiu-cristea:
## Problem

Say a module install process requires some batch operations, like adding some initial content or like `locale` module, which is downloading translations. When such a module is installed with `drush en some_module` the batch process is running because of these lines from `PmCommands::install()`:

```php
if (batch_get()) {
    drush_backend_batch_process();
}
```

However, not only `drush en some_module` installs a module. There is also `drush config:import`, which installs/uninstalls module based on changes detected in `core.extension` config. In such cases, the batch process doesn't run.

## Proposal

On `processExtensions` config import sync step, if there's a batch waiting to be triggered, call `drush_backend_batch_process()`.


claudiu-cristea:
I've added test coverage to a functional but than decided to move the whole class to integration. All green ;)

vever001:
Hello,

After updating to drush 12 I stumbled on what looks like a race condition with language config collections which lead me to this issue. I am getting an error on a multilingual site during `config:import`:
> [warning] Trying to access array offset on value of type null ConfigImporter.php:353
> > TypeError: array_diff(): Argument #2 must be of type array, null given in web/core/lib/Drupal/Core/Config/ConfigImporter.php on line 353 #0 web/core/lib/Drupal/Core/Config/ConfigImporter.php(353): array_diff()
> ...

At the beginning of `config:import`, there are several config collections (`language.*`) and `locale_system_update` gets called because several modules need to be enabled. This can cause some new config collections to be added if so, leads to the error above.
In my case it seems to happen when devel gets enabled via `config:import`.
I have an `mk` language in drupal. Before `locale_system_update` gets called, there is no corresponding `language.mk` config collection (in the `config` table) . But there is one after it.

This may not be easy to reproduce from a clean setup... but I'll try to do it when I have some time.
Conceptually the changes introduced in this PR make sense to me, but it seems to cause side effects, especially with locale.
Have you noticed any similar issue? Where would it be best to report this?

Thank you

claudiu-cristea:
@vever001 I haven't noticed but, indeed, this looks like a possible edge case. We probably need a new PR with a test that replicates the scenario to start with

vever001:
Thanks @claudiu-cristea, I opened https://github.com/drush-ops/drush/issues/5838, and will try to provide more info there.

PR #5717, opened 2023-07-24
add loadRevision functionality

chx:
next step in the entity work is to add Paragraph::loadRevision. We can just steal the code from EntityBase::load(), wrap it in eval and boom.

### DIFF
--- a/src/Commands/core/CliCommands.php
+++ b/src/Commands/core/CliCommands.php
@@ -308,7 +308,14 @@ final class CliCommands extends DrushCommands
         foreach ($this->entityTypeManager->getDefinitions() as $definition) {
             $class = $definition->getClass();
             $parts = explode('\\', $class);
-            class_alias($class, array_pop($parts));
+            eval(sprintf('class %s extends \%s {
+                public static function loadRevision($id) {
+                    $entity_type_repository = \Drupal::service("entity_type.repository");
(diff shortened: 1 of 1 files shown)


src/Commands/core/CliCommands.php
```diff
@@ -308,7 +308,14 @@ public function makeEntitiesAvailableWithShortClassNames(): void
         foreach ($this->entityTypeManager->getDefinitions() as $definition) {
             $class = $definition->getClass();
             $parts = explode('\\', $class);
-            class_alias($class, array_pop($parts));
+            eval(sprintf('class %s extends \%s {
```
weitzman, line 313: Needs code comment about what the heck we are doing. Also, can you think of a more elegant way than eval?

weitzman (approved):

heddn:
I have some final entity classes. this blows up on that.

chx:
You drank too much of the upstream kool aid to use final classes.

heddn:
I always use final classes for custom entities. And in the case of the example contrib config entities, I don't want to have to worry about BC with some strange use case extending my entity classes. Easier to just mark final.

chx:
/me shrugs

roll it back if you want, wouldn't be the first case this private and final idiocy ruins the party

I even wrote a pleading letter to php internals to give an out for readonly classes but they of course wouldn't listen

yay for theory to win over practice

PR #5729, opened 2023-08-15
Handle final entity class in php:cli

heddn:
Follow-up to https://github.com/drush-ops/drush/pull/5717

### DIFF
--- a/src/Commands/core/CliCommands.php
+++ b/src/Commands/core/CliCommands.php
@@ -308,7 +308,7 @@ final class CliCommands extends DrushCommands
     {
         foreach ($this->entityTypeManager->getDefinitions() as $definition) {
             $class = $definition->getClass();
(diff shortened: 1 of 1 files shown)


weitzman:
LGTM. Awaiting more comments then will proceed.

src/Commands/core/CliCommands.php
```diff
@@ -308,6 +308,10 @@ public function makeEntitiesAvailableWithShortClassNames(): void
     {
         foreach ($this->entityTypeManager->getDefinitions() as $definition) {
             $class = $definition->getClass();
+            $reflectionClass = new ReflectionClass($class);
```
heddn:
```suggestion
            $reflectionClass = new \ReflectionClass($class);
```
Should have manually test this earlier. Did now and found out a leading slash is needed.

Issue #5673, opened 2023-06-23
example.drush.yml should have drush.php.minimum-version


franfran:
**Describe the bug**
This is not a bug, but an enhancement.
In drush 11.x/12.x, the function confirmPhpVersion() will throw error if drush.php.minimum-version is not defined in drush.yml.  Since the key minimum-version is a requirement during Preflight, it will be better to include this in the example.drush.yml.

**To Reproduce**
- with PHP8, drush 11.x/12.x
- install drush
- cp examples/example.drush.yml ../drush.yml
- run any drush command, e.g. drush version

**Expected behavior**
Should be able to run without error.

**Actual behavior**
`Drush\Preflight\PreflightVerify::confirmPhpVersion(): Argument #1 ($minimumPhpVersion) must be of type string, null given`

**Workaround**
Add the key drush.php.minimum-version to drush.yml

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.x
| Drupal version? | 11
| PHP version     | 8.x
| OS?             | Mac/Linux/Windows

**Additional information**
In our build script, we have a step to copy the example.drush.yml right after installation.  There was no problem until we start upgrading to Drupal 11, which requires PHP8 and then the argument type checking is now in-place.

greg-1-anderson:
Providing a default value in the example file is a reasonable ask. The type error is also a bug that should be addressed.

AronNovak:
It also broke our deployment pipeline, so I created #5743 to address the type error.

PR #5743, opened 2023-08-31
Fix fatal error when no PHP version is specified in drush.yml

AronNovak:
#5673

### DIFF
--- a/src/Preflight/PreflightVerify.php
+++ b/src/Preflight/PreflightVerify.php
@@ -38,8 +38,11 @@ class PreflightVerify
      * @param string $minimumPhpVersion
      *   The minimum allowable php version
      */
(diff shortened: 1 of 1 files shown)


src/Preflight/PreflightVerify.php
```diff
@@ -38,8 +38,11 @@ public function verify(Environment $environment): void
      * @param string $minimumPhpVersion
      *   The minimum allowable php version
      */
-    public function confirmPhpVersion(string $minimumPhpVersion): void
+    public function confirmPhpVersion(string|null $minimumPhpVersion): void
     {
+        if (empty($minimumPhpVersion)) {
+            return null;
```
AronNovak:
```suggestion
            return;
```

Issue #5744, opened 2023-08-31
Drush SQL Sync fails when --source server has old drush.


jonpugh:
**Describe the bug**

We have an old Aegir server with ~10 sites, drupal 8.9.3 and drush8 installed locally. 

We have an updated codebase, drupal 9.5, drush 11, with remote aliases to the old sites.

Remote alias `drush @legacy.site sqlc` works fine. Remote `drush @legacy.site sql-dump` works as expected.

However, `drush sql-sync @legacy @new` fails with the error below.

I traced it down. The SQL dump works fine, it's the syncing back of the file that fails.

I opened up the code, and it looks like support for older drush sources was removed: https://github.com/drush-ops/drush/commit/14d55983120db638889de0ed56f8f386cdddc5ce#

So the dump works, but the output isn't JSON, so this little line of code fails:

```
  $json = $process->getOutputAsJson();
                $source_dump_path = $json['path'];
```

**To Reproduce**

Call `drush sql:sync @source @destination` where `@source` is an older site with drush8 and `@destination` is a site with drush 11 or higher.

**Expected behavior**
Successful sync.

**Actual behavior**

```
$ drush sql-sync @legacy.education @production.education
You will destroy data in productioneducat and replace with data from x.edu/educationx.

 Do you really want to continue? (yes/no) [yes]:
 > yes

 [notice] Starting to dump database on source.

In ProcessBase.php line 155:

  Output is empty.

Exception trace:
  at /var/aegir/projects/production/education/vendor/consolidation/site-process/src/ProcessBase.php:155
 Consolidation\SiteProcess\ProcessBase->getOutputAsJson() at /var/aegir/projects/production/education/vendor/drush/drush/src/Commands/sql/SqlSyncCommands.php:152
 Drush\Commands\sql\SqlSyncCommands->dump() at /var/aegir/projects/production/education/vendor/drush/drush/src/Commands/sql/SqlSyncCommands.php:59
 Drush\Commands\sql\SqlSyncCommands->sqlsync() at n/a:n/a
 call_user_func_array() at /var/aegir/projects/production/education/vendor/consolidation/annotated-command/src/CommandProcessor.php:276
 Consolidation\AnnotatedCommand\CommandProcessor->runCommandCallback() at /var/aegir/projects/production/education/vendor/consolidation/annotated-command/src/CommandProcessor.php:212
[...]

PR #5746, opened 2023-08-31
In sql:sync command, catch exception for missing JSON and set $source_dump_path from original options.

jonpugh:
*This is a second attempt at fixing #5744.  A better approach than #5745.*

Drush `sql:sync` currently fails without an indication why when source drush is lower than v9.6.

This is because before drush v9.6, `sql-dump` could not return JSON data. The actual dumping works fine. It's the part in SQL Sync where it takes the next step that fails.

This PR catches the exception triggered during the `sql:sync` command when the output of `sql:dump` is not in JSON.

Note that at this point, the external drush process has succeeded because of `$process->mustRun()`. We know that the sql dump took place. In fact, the only thing the code uses in the JSON output is the result file, which we already have in the options.

I know maintainability is important here, which is why I thought this technique might be acceptable.  This adds a catch to an uncaught exception instead of keeping old code lying around.

Thoughts?

-----

Context: Without this patch, using `drush sql:sync` *from* a site with drush 9.5 or earlier fails with no explanation, meaning using `drush sql:sync` for syncing from drupal 7 or 8 to modern doesn't work.  

The other commands, `drush sql:dump` and `drush sql:client` work fine when the target is drupal 7 or 8.

### DIFF
--- a/src/Commands/sql/SqlSyncCommands.php
+++ b/src/Commands/sql/SqlSyncCommands.php
@@ -141,8 +141,13 @@ final class SqlSyncCommands extends DrushCommands implements SiteAliasManagerAwa
             if ($this->getConfig()->simulate()) {
                 $source_dump_path = '/simulated/path/to/dump.tgz';
             } else {
-                $json = $process->getOutputAsJson();
-                $source_dump_path = $json['path'];
+                try {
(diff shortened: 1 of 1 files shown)


weitzman:
Interested folks are welcome to use this as a patch.

PR #5747, opened 2023-09-01
Check requirements when installing modules via cim

claudiu-cristea:
Checking module install requirements is performed when enabling a module via `pm:install` but modules can also be enabled via `config:import`.

We don't show the prompt in such cases but just abort the whole import as not installing modules might severely break the remaining imports

### DIFF
--- a/src/Commands/ModuleInstallRequirementsTrait.php
+++ b/src/Commands/ModuleInstallRequirementsTrait.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Drush\Commands;
+
+trait ModuleInstallRequirementsTrait
+{
+    /**
+     * Returns a list of modules not meeting the install requirements.
+     *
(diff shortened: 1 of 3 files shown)


claudiu-cristea:
IMHO, it's also a mistake that `pm:install` allows to install a module that has unmet requirements

weitzman:
Where does drupal core handle requirements checking of modules enabled during config import?

claudiu-cristea:
> Where does drupal core handle requirements checking of modules enabled during config import?

I don't think they do. But I think it's just missing

weitzman:
OK, then I think a core MR would be more appropriate than this PR. I'm not inclined to maintain this in Drush. This adds complication for drush deploy since its another way that updates can run and then config import can fail, thus leaving your DB in an unpleasant state.

PR #5755, opened 2023-09-08
Handle Drupal's web root being set to a symlink.

### LINKED ISSUE #5754: Drush 12 doesn't work with Drupal being in a symlink'd directory
**Describe the bug**
Drush fails to work when Drupal core's install path is a symlink. There's a check for an altered root being different to the root found another way in the preflight code, the first case uses `realpath()` and so resolves symlinks but the altered root does not causing the 2 methods to get different directories, even they're the same once symlinks are resolved.

**To Reproduce**

- We have `"www/core": ["type:drupal-core"],` in the `installer-paths` section
- Make the `www` a symlink to the folder Drupal's `core` folder is inside
- Run drush 12.x

**Expected behavior**
Drush to work.

**Actual behavior**
Drush fails to run with `[preflight] The alias references a Drupal site that does not contain its own copy of Drush. Please add Drush to this site to use it.`

**Workaround**
None

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.1.3.0
| Drupal version? | 10.1.2
| PHP version     | 8.1.22
| OS?             | Linux

MikeAnchor:
A pull request to fix #5754

### DIFF
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -322,8 +322,13 @@ class Preflight
                 return [$preflightDidRedispatch, $exitStatus];
             }
 
-            // Handle altered root containing a symlink.
-            $alteredRoot = FsUtils::realpath($alteredRoot);
+            // Handle altered root containing a symlink. Path::canonicalize()
+            // is called, because DrushDrupalFinder::getDrupalRoot() also calls
+            // it. realpath() will return a path with OS specific directory
+            // separators, whereas Path::canonicalize standardises on /. This
+            // means on Windows, the path returned by realpath() will have \
+            // but we need / to compare against $root.
+            $alteredRoot = Path::canonicalize(FsUtils::realpath($alteredRoot));
 
             // If the Drupal site changed, and the alternate site does not
             // contain its own copy of Drush, then we cannot continue.
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -13,6 +13,7 @@ use Drush\Config\ConfigLocator;
 use Drush\Config\DrushConfig;
 use Drush\Config\Environment;
 use Drush\SiteAlias\SiteAliasFileLoader;
+use Drush\Utils\FsUtils;
 use RuntimeException;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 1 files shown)


src/Preflight/Preflight.php
```diff
@@ -321,6 +322,9 @@ public function preflight($argv): array
                 return [$preflightDidRedispatch, $exitStatus];
             }
 
+            // Handle altered root containing a symlink.
+            $alteredRoot = FsUtils::realpath($alteredRoot);
```
greg-1-anderson: We tend to be fairly conservative about symlink usage, since they are fraught. However, we have the comparison immediately after this that insists that $alteredRoot must be the same as $root to continue, and $root is the root that we found from our shared `vendor`, so maybe this particular instance of symlink handling is okay. Up to @weitzman

greg-1-anderson (changes requested):
This change breaks Drush on Windows. The path compare below will need to be normalized in some way to fix this.

Alternately, perhaps the comparison could become `if (($alteredRoot != $root) && (FsUtils::realpath($alteredRoot) != $root)) {`, or something like that.

MikeAnchor:
Trying with a call to `Path::canonicalize()` due to https://www.php.net/manual/en/function.realpath.php#example-2286 - the `$alteredRoot` will have Windows directory separators but `$root` would have UNIX style which is why it was breaking on Windows.

greg-1-anderson (approved):
Seems fine to me if @weitzman is also 👍

greg-1-anderson (changes requested):
I missed the fact that this would in some configurations allow you to run a Drush that had a different vendor dir than the Drupal site. That should not be allowed.

Discussing whether this PR should be allowed at all. If it is, we'd have to change the comparison to check that the vendor dirs match, not just the root.

weitzman:
Are you using a site alias? If so please show the full debug log and the alias details.

MikeAnchor:
@weitzman nope, no alias. Can't run any drush commands at all, even just invoking `drush` gives the preflight error

weitzman:
This code is gone in 13.x so perfects the issue is fixed there. Lets leave 12.x alone IMO.

PR #5766, opened 2023-09-25
Replace usage of in_array() in MigrateExecutable::handleMissingSourceRows

### LINKED ISSUE #5765: Replace usage of in_array() in MigrateExecutable::handleMissingSourceRows
**Describe the bug**
Usage of `in_array()` in `MigrateExecutable::handleMissingSourceRows()` is proving to be very inefficient for migrations with a very large amount of rows.

**To Reproduce**
Run any migration ID with a very large amount of rows (eg 10,000+).
While the actual migration has a progress bar and lets you know when its finished, the logic in `handleMissingSourceRows()` will have the process seem like its frozen for an indeterminate amount of time.

**Actual behavior**
Running a migration ID with many rows (in my case over 300,000 for `upgrade_d7_file_private`) would take roughly 20-30 minutes for the actual migration, but would hang on `MigrateExecutable::handleMissingSourceRows()` for multiple hours before having to manually stop the process.

Using `in_array()` can be very inefficient as [it needs to compare all array values until it finds a match](https://stackoverflow.com/a/13483548) not to mention the current logic is trying to find an an array within an array of arrays.

**Workaround**
Instead of using `in_array()` the `$allSourceIdValues` property should be keyed with a unique ID in order to utilize `isset()`

Having a dedicated method to build the key off the source ID values can allow it to be used when writing to the `$allSourceIdValues` property in `MigrateExecutable::onPrepareRow()` and reading it within `handleMissingSourceRows()`.

Making this change to the example above with 300k rows, brought this post-migration logic to finish within a few minutes instead of multiple hours.

mdolnik:
PR to fix #5765

### DIFF
--- a/src/Drupal/Migrate/MigrateExecutable.php
+++ b/src/Drupal/Migrate/MigrateExecutable.php
@@ -214,6 +214,22 @@ class MigrateExecutable extends MigrateExecutableBase
     }
 
     /**
+     * The array key to use to represent Source ID values.
+     *
+     * Helps speed up array searched by serializing the value for a key.
+     *
+     * @param array $source_id
+     *   The Source ID array.
+     *
+     * @return string
+     *   The array key to represent Source ID values.
+     */
+    protected function getSourceIdKey(array $source_id): string
+    {
+      return implode('-', $source_id);
+    }
+
+    /**
      * Handles missing source rows after import.
      *
      * Detect if, before importing, the destination contains rows that are no
@@ -242,7 +258,8 @@ class MigrateExecutable extends MigrateExecutableBase
         $destinationIds = [];
         while ($idMap->valid()) {
             $mapSourceId = $idMap->currentSource();
-            if (!in_array($mapSourceId, $this->allSourceIdValues)) {
+            $sourceIdKey = $this->getSourceIdKey($mapSourceId);
+            if (!isset($this->allSourceIdValues[$sourceIdKey])) {
                 $destinationIds[] = $idMap->currentDestination();
             }
             $idMap->next();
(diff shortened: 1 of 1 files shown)


src/Drupal/Migrate/MigrateExecutable.php
```diff
@@ -213,6 +213,22 @@ public function onPreImport(MigrateImportEvent $event): void
...
+     *
+     * @param array $source_id
+     *   The Source ID array.
+     *
+     * @return string
+     *   The array key to represent Source ID values.
+     */
+    protected function getSourceIdKey(array $source_id): string
+    {
+        return implode('-', $source_id);
```
claudiu-cristea, line 229: I wonder how safe is this

mdolnik: I don't know the full extent of what is possible to appear in the `$source_id` array values, but from my investigation it is mostly entity IDs, revision IDs, and simple strings. Of which, most cases involve one or two values.

I am open to alternate ways of serializing the array values into a unique key.

debart-nathan: Hi.
I run into this issue while doing a diagnostic on what was slowing the import even after it was marked as finished.

And want to put my grain of salts.

The case i can see where the proposed serialisation can cause problem is very specific.
[A_B, C] has the same key than [A,B_C].

you could use hash function but you would still run into the [Pigeon hole principle](https://en.m.wikipedia.org/wiki/Pigeonhole_principle).

the only way to avoid that would be to let the user decide by a cli argument what separator to use, including the option to turn the feature off and use the old system.

PR #5767, opened 2023-09-26
Better way to detect core-recommended

prudloff-insite:
This is a followup to https://github.com/drush-ops/drush/pull/5684.
`$hasCoreRecommended` is not an associative array.


weitzman:
pm:security command is being removed in #5775.

PR #5769, opened 2023-09-27
add role name parameter - fixes #5768 for 11.x

### LINKED ISSUE #5768: role:list arguments do not match documentation
**Describe the bug**
According to the description of `role:list`:

> If a role name is provided as an argument, then all of the permissions of that role will be listed

**To Reproduce**
drush role:list authenticated

**Expected behavior**
Only the relevant role should be displayed.

**Actual behavior**
the following error is reported:

```
  No arguments expected for "role:list" command, got "authenticated".  
```

**Workaround**
None

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.x and 11.x
| Drupal version? | 10.x

malcomio:
fixes #5768 for 11.x


weitzman:
A backport of #5770 would be preferable over this PR

malcomio:
see #5772

PR #5775, opened 2023-10-06
Remove pm:security in favor of composer audit

weitzman:


deviantintegral:
This broke existing CI checks we had calling this command. Removing a whole command feels like a b/c break to me. Could this be changed to warn that it will be removed in drush 13?

weitzman:
I'm comfortable with removing a command in a minor, as long as there is mention in the release notes. I suggest pinning the drush dependency if you want to prevent this in the future.

PR #5785, opened 2023-10-18
Fixes #5761 Get the relative path to the project root as the descript…

### LINKED ISSUE #5761: Drush site alias commands stop working with version 12
**Describe the bug**
Drush site alias commands that worked with Drush 11 are now not working with Drush 12.

**To Reproduce**
Upgraded drush install to Drush 12

**Expected behavior**
Executing drush @self.live status would return a status listing for the site defined for the alias @self.live

**Actual behavior**
Command returned an exit code 127 (command not found)

**Workaround**

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.2.0.0
| Drupal version? | 10.0.8
| PHP version     | 8.1
| OS?             | Mac Local, Linux on remote server

We have drush-installer installed in the global path on the remote server.

**Additional information**
Running command with Drush 11 installed passing verbose flags reveals that the shell command executed by Drush is:

`ssh -t -o PasswordAuthentication=no <username>@<ip_address> 'drush status -vvv --uri=https://my.website.com --root=/home/forge/my.website.com/public` <== executes drush successfully on remote server

Running command with Drush 12 installed passing verbose flags reveals that shell command executed by Drush is:

`ssh -t -o PasswordAuthentication=no <username>@<ip_address> '/home/forge/bin/drush status -vvv --uri=https://my.website.com` <== fails to execute on remote server

The site alias file did not change, only the version of drush installed.

Drush alias file (`./drush/sites/self.site.yml`):

```
live:
  host: <ip_address>
  user: <username>
  root: /home/forge/my.website.com/public
  uri: https://my.website.com
stage:
  host: <ip_address>
  user: <username>
  root: /home/forge/staging.my.website.com/public
  uri: https://staging.my.website.com

```

Two observations:

1. `--root=/home/forge/my.website.com/public` is not being passed into the drush command
2. The path to the drush installation is not picking up the location of the project specific drush install (probably because of `--root` not being passed)

joelpittet:
Get the relative path to the project root as the docblock says not the drush base path.

This works for me but I'm not sure if the runtime.project is the correct config to use or if there is a better way?  It seems like the drush.base-dir is not the correct config though and it's broken for my remote alias commands.

### DIFF
--- a/src/SiteAlias/ProcessManager.php
+++ b/src/SiteAlias/ProcessManager.php
@@ -96,9 +96,9 @@ class ProcessManager extends ConsolidationProcessManager
     protected function relativePathToVendorBinDrush()
     {
         $absoluteVendorBin = $_composer_bin_dir ?? Path::join($this->getConfig()->get('drush.vendor-dir'), 'bin');
-        $basePath = $this->getConfig()->get('drush.base-dir');
+        $projectRoot = $this->getConfig()->get('runtime.project');
 
-        $relativeVendorBin = Path::makeRelative($absoluteVendorBin, $basePath);
+        $relativeVendorBin = Path::makeRelative($absoluteVendorBin, $projectRoot);
 
         return Path::join($relativeVendorBin, 'drush');
     }


joelpittet:
Fixes #5761

weitzman:
Hoping @greg-1-anderson can take a look.

greg-1-anderson (approved):
I'd have to check to see how `drush.base-dir` is calculated if you want a full comparison, but r`untime.project` is taken from Composer, and is literally what is wanted here, so I think this is a much better config variable to use here. And the tests pass, so ship it.

PR #5788, opened 2023-10-19
Avoid rebuilding Drupal container at end of updatedb

weitzman:
I'm not 100% sure this a good idea but I think it is.

### DIFF
--- a/src/Commands/core/UpdateDBCommands.php
+++ b/src/Commands/core/UpdateDBCommands.php
@@ -11,13 +11,16 @@ use Consolidation\SiteAlias\SiteAliasManagerAwareTrait;
 use Drupal\Core\Database\Database;
 use Drupal\Core\Utility\Error;
 use Drush\Attributes as CLI;
+use Drush\Boot\BootstrapManager;
 use Drush\Boot\DrupalBootLevels;
 use Drush\Commands\DrushCommands;
 use Drush\Drupal\DrupalUtil;
 use Drush\Drush;
 use Drush\Exceptions\UserAbortException;
 use Drush\Log\SuccessInterface;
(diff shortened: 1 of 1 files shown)


weitzman:
Hmm. Tests failed and its unclear at first glance if the problem is with a test or with the code change. Dont think I will work on this further at the moment.

greg-1-anderson:
Overall I think it's a good idea to make Drush `cr` behave as closely to Drupal core's rebuild as possible; however, if we pass in the Kernel, then we should probably call `$kernel->invalidateContainer();`, [as drupal_rebuild does](https://git.drupalcode.org/project/drupal/-/blob/11.x/core/includes/utility.inc?ref_type=heads#L33). I'm not sure that our end behavior would then be any different, but I think I'd prefer it this way anyway.

greg-1-anderson:
I didn't see your comment about the failed tests. I imagine they'd probably start passing again if you called invalidateContainer(). I don't think this (the change in technique in this PR) is super important.

weitzman:
This PR only affects updatedb - no change to `cr`. 

Calling invalidateContainer would negate any performance benefit from the PR.

greg-1-anderson:
Hm, my brain started thinking about updatedb, but at some point accidentally wandered into `rebuild.php` land. Not sure how I ended up there.

In any event, [Drupal's update function](https://git.drupalcode.org/project/drupal/-/blob/11.x/core/modules/system/src/Controller/DbUpdateController.php?ref_type=heads#L362) [1] calls `drupal_flush_all_caches` without any parameters, so I am guessing we can't shortcut away the kernel rebuild here either.

[1] There are other places it is called as well.

weitzman:
Lets leave it as is.

PR #5796, opened 2023-10-26
Refs #5785. Fix path to remote drush

weitzman:
The Doxygen misled us in #5785. We want Drupal root and not project root as base path. Use DrupalFinder to get that.


joelpittet:
As long as I provide the `root` to the `drupal root`/`web root` this totally works.

PR #5808, opened 2023-11-10
--ignore-roles for sql:sanitize 12.x, based on PR #4725

Berdir:
See existing discussion in PR #4725.

This is just a rebase, wanted to make this available when I did the work anyway.


weitzman:
Sounds like this is still in Draft?

weitzman:
After more thought, this looks merge worthy to me. I'll merge this unless someone sees a problem with it in the next day or so.

It would be nice to have test coverage. See https://github.com/drush-ops/drush/blob/12.x/tests/functional/SqlSyncTest.php#L110-L150

Issue #5797, opened 2023-10-27
TypeError: ArrayObject::__construct(): Argument #1 ($array) must be of type array, bool given in ArrayObject->__construct()


alayham:
**Describe the bug**

I am updating a complex site from D9.5 to D10.2 dev. I can run the db updates manually, and the site works fine, but with `drush updb` I get the following error after listing the available updates.
 ```
 // Do you wish to run the specified pending updates?: yes.                                                             

>  [error]  TypeError: ArrayObject::__construct(): Argument #1 ($array) must be of type array, bool given in ArrayObject->__construct() (line 15 of /var/www/drupal/vendor/consolidation/output-formatters/src/StructuredData/AbstractListData.php) #0 /var/www/drupal/vendor/consolidation/output-formatters/src/StructuredData/AbstractListData.php(15): ArrayObject->__construct()
> #1 /var/www/drupal/vendor/consolidation/output-formatters/src/StructuredData/UnstructuredListData.php(21): Consolidation\OutputFormatters\StructuredData\AbstractListData->__construct()
> #2 /var/www/drupal/vendor/drush/drush/src/Commands/core/UpdateDBCommands.php(132): Consolidation\OutputFormatters\StructuredData\UnstructuredListData->__construct()
> #3 [internal function]: Drush\Commands\core\UpdateDBCommands->process()
> #4 /var/www/drupal/vendor/consolidation/annotated-command/src/CommandProcessor.php(276): call_user_func_array()
> #5 /var/www/drupal/vendor/consolidation/annotated-command/src/CommandProcessor.php(212): Consolidation\AnnotatedCommand\CommandProcessor->runCommandCallback()
> #6 /var/www/drupal/vendor/consolidation/annotated-command/src/CommandProcessor.php(175): Consolidation\AnnotatedCommand\CommandProcessor->validateRunAndAlter()
> #7 /var/www/drupal/vendor/consolidation/annotated-command/src/AnnotatedCommand.php(387): Consolidation\AnnotatedCommand\CommandProcessor->process()
> #8 /var/www/drupal/vendor/symfony/console/Command/Command.php(326): Consolidation\AnnotatedCommand\AnnotatedCommand->execute()
> #9 /var/www/drupal/vendor/symfony/console/Application.php(1081): Symfony\Component\Console\Command\Command->run()
> #10 /var/www/drupal/vendor/symfony/console/Application.php(320): Symfony\Component\Console\Application->doRunCommand()
[...]

PR #5809, opened 2023-11-10
Correct fix for batch update scenario

Berdir:
Fix for #5797

### DIFF
--- a/includes/batch.inc
+++ b/includes/batch.inc
@@ -113,8 +113,20 @@ function _drush_backend_batch_process($command = 'batch-process', $args = [], $o
         $batch['id'] = $batch_storage->getId();
       }
       catch (IntegrityConstraintViolationException) {
-        // Remove calls to nextId() once 10.0 is no longer supported.
-        $batch['id'] = \Drupal::database()->nextId();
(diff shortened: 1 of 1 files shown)


includes/batch.inc
```diff
@@ -113,8 +113,20 @@ function _drush_backend_batch_process($command = 'batch-process', $args = [], $o
         $batch['id'] = $batch_storage->getId();
       }
       catch (IntegrityConstraintViolationException) {
-        // Remove calls to nextId() once 10.0 is no longer supported.
-        $batch['id'] = \Drupal::database()->nextId();
+        // @todo this is here to support the update path to deprecate
+        // Connection::nextId(). Remove when Drupal 10.1 and lower is no longer
+        // supported.
```
weitzman, line 118: What exactly should be removed after 10.1 is unsupported. We dont need to catch the exception at all anymore?

Berdir: correct. the exception is thrown when the 10.1 => 10.2 update function hasn't been run yet, because the id field is not a serial/autoincrement yet. (the update function is incorrectly named system_update_10100(), it should have been system_update_10200()

PR #5812, opened 2023-11-14
Add prepending slash to user:login destination URL

gitressa:
`node/add/article` works in Drush 12, but not in Drush 8. `/node/add/article` works in both.


gitressa:
Strike that. The problem is that a user id or user name is required in Drush 8, for it to work.

Works: `drush user:login 1 /node/add/article`
Doesn't work: `drush user:login /node/add/article`

PR #5819, opened 2023-11-24
Add tip that `--destination` starting point is Drupal root 

gitressa:
It would be nice if the `drush archive:dump` starting point was from where the command was run ... which is normally here, for most projects using `drupal/recommended-project projects`:

```
$ tree -L 1
.
├── composer.json
├── composer.lock
├── vendor
└── web
```

I guess it's not possible? So perhaps adding the tip that `--destination` starting point is Drupal root would help people in this situation?

1. Run `drush archive:dump --destination=archive.tar.gz`
1. Don't see the file in the folder from where it was run
1. Look in "web" folder by chance and realize they need to run `drush archive:dump --destination=../archive.tar.gz`

### DIFF
--- a/src/Commands/core/ArchiveDumpCommands.php
+++ b/src/Commands/core/ArchiveDumpCommands.php
@@ -64,7 +64,7 @@ final class ArchiveDumpCommands extends DrushCommands
      */
     #[CLI\Command(name: self::DUMP, aliases: ['ard'])]
     #[CLI\ValidatePhpExtensions(extensions: ['Phar'])]
-    #[CLI\Option(name: 'destination', description: 'The full path and filename in which the archive should be stored, relative to Drupal root, normally "web" for drupal/recommended-project projects. If omitted, it will be saved to the drush-backups directory.')]
+    #[CLI\Option(name: 'destination', description: 'The full path and filename in which the archive should be stored, relative to Drupal root, normally "web" for drupal/recommended-project projects. If omitted, it will be saved to the /tmp directory.')]
     #[CLI\Option(name: 'overwrite', description: 'Overwrite destination file if exists.')]
     #[CLI\Option(name: 'code', description: 'Archive codebase.')]
     #[CLI\Option(name: 'exclude-code-paths', description: 'Comma-separated list of paths (or regular expressions matching paths) to exclude from the code archive.')]
--- a/src/Commands/core/ArchiveDumpCommands.php
+++ b/src/Commands/core/ArchiveDumpCommands.php
@@ -64,7 +64,7 @@ final class ArchiveDumpCommands extends DrushCommands
      */
     #[CLI\Command(name: self::DUMP, aliases: ['ard'])]
     #[CLI\ValidatePhpExtensions(extensions: ['Phar'])]
-    #[CLI\Option(name: 'destination', description: 'The full path and filename in which the archive should be stored, relative to Drupal root, normally "web" for drupal/recommended-project projects. If omitted, it will be saved to the /tmp directory.')]
+    #[CLI\Option(name: 'destination', description: 'The full path and filename in which the archive should be stored. Any relative path will be calculated from Drupal root (usually <info>web</web> for drupal/recommended-project projects). If omitted, it will be saved to the configured temp directory.')]
     #[CLI\Option(name: 'overwrite', description: 'Overwrite destination file if exists.')]
     #[CLI\Option(name: 'code', description: 'Archive codebase.')]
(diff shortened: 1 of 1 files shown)


weitzman:
Good idea. I think the usual way is to add to the description of the destination option: 'relative to Drupal root'.

gitressa:
Ah yes, I updated the patch:

```
#[CLI\Option(name: 'destination', description: 'The full path and filename in which the archive should be stored, relative to Drupal root, normally "web" for drupal/recommended-project projects. If omitted, it will be saved to the drush-backups directory.')]
```

Though, it seems like this isn't true. Should it be removed?

> [...] If omitted, it will be saved to the drush-backups directory.

weitzman:
Yes, lets please remove that. It was copied from another option I guess.

gitressa:
We could remove it, or maybe update it instead, since it is actually saved to the `/tmp` directory? (I updated the MR to `If omitted, it will be saved to the /tmp directory.`)

```
$ drush archive:dump
/tmp/archive.tar.gz
```

Also, would it be possible to show the path, like this? It would be a big help, but I am not sure where to tweak the code...

```
$ drush archive:dump --destination=archive.tar.gz
Saved in /web/archive.tar.gz
```

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -64,7 +64,7 @@ final class ArchiveDumpCommands extends DrushCommands
      */
     #[CLI\Command(name: self::DUMP, aliases: ['ard'])]
     #[CLI\ValidatePhpExtensions(extensions: ['Phar'])]
-    #[CLI\Option(name: 'destination', description: 'The full path and filename in which the archive should be stored. If omitted, it will be saved to the drush-backups directory.')]
+    #[CLI\Option(name: 'destination', description: 'The full path and filename in which the archive should be stored, relative to Drupal root, normally "web" for drupal/recommended-project projects. If omitted, it will be saved to the /tmp directory.')]
```
weitzman:
```suggestion
    #[CLI\Option(name: 'destination', description: 'The full path and filename in which the archive should be stored. Any relative path will be calculated from Drupal root (usually <info>web</web> for drupal/recommended-project projects). If omitted, it will be saved to the configured temp directory.')]
```

gitressa:
Thanks, the wording is good now. Though this bit needs tweaking: 

`<info>web</web>` should be `<info>web</info>`

Do you think improved feedback like `Saved in /web/archive.tar.gz` is possible?

weitzman:
It gives the path now but not the 'Saved in' prefix. Its not easy to add that prefix because the return value can be used by machine callers. This pattern is common in Drush. See the URLs returned by user:login and browse, for example.

gitressa:
Thanks for clarifying about the prefix, but it seems like the `/web/` part is not included:

```
$ drush archive:dump --destination=archive.tar.gz
archive.tar.gz
```

As a minimum, showing the actual path would be helpful, whereas the "Saved in" is not so important, but more a nice-to-have.

weitzman:
Not ideal, but this is reporting the --destination as you provided it. And now that destination is documented as being drupal root relative. If you dont provide a --destination, it reports `/tmp/...`

gitressa:
Yes, and after trying some things, and looking around, you realize that you need to step out of the `/web` folder adding `../` like this, to get the archive file to land where you want it:

```
drush archive:dump --destination=../archive.tar.gz
```

I just think it would be a nice service to include the "relocated document root" in the returned value, to save the user some time ... but maybe it's complicated to include the relocated document root?

Another argument for returning a more complete path, is that some users might not even realize that a new potentially large file has just been created, and it might get committed with Git, which would be unfortunate.

gitressa:
Perhaps a general tip, when creating files (not just `archive:dump`) could be added, that the project is using a relocated web folder? This could give a crucial hint on where to look.

```
$ drush archive:dump --destination=archive.tar.gz
Note: The project is using the relocated document root "web"
archive.tar.gz
```

weitzman:
I've implemented the full path as you originally suggested - https://github.com/drush-ops/drush/pull/5821

gitressa:
Perfect thanks, this will be a great help, and save much time.

PS. Remember to update `<info>web</web>`.

weitzman:
Thanks. Done. https://www.drush.org/12.x/commands/archive_dump/

gitressa:
Thanks, it outputs the path from DDEV, but the important part `/web` is included, giving a hint in that direction.
```
$ drush archive:dump --destination=archive.tar.gz
/var/www/html/web/archive.tar.gz
```
Also, this command may mostly be used on servers, where the path displayed is much more likely to be the actual path.

PR #5823, opened 2023-11-27
Use the new Laravel Prompts

weitzman:
#### Todo
- [ ] Get https://github.com/laravel/prompts/pull/100 merged, or otherwise work around the exit(). 

See [Laravel Prompts](https://laravel.com/docs/10.x/prompts). Please comment or reaction with your feedback. Things to note:

- Helpful hints can be provided in addition to question
- Nice presentation of errors (see Field Name in screencast)
- Required error nicely presented
- Confirmation questions nicely presented (see Translatable)
- Great UX for selecting one choice (see Field Type) or multiple choices (see Allowed Formats)

[asciicast](https://asciinema.org/a/623617)

config:get
======
- Demonstrates the `suggest` helper. This adds a search+scroll UX for picking a config object.

[asciicast](https://asciinema.org/a/0upS0k7SCOnyLF8lQ2ROuuar4)


DieterHolvoet:
Maybe change the implementation of $this->io()->choice() instead? Less changes, + this means that people can still override it.

jonhattan:
I think that showing more than 5 rows in select widgets (ej Entity Type) improves UX

weitzman:
image

One small issue I'm seeing is that the Prompts renders a nice `Cancelled` message and then calls `exit(1)` after a user hits CTRL-C at the prompt. This causes the Drush ShutdownHandler to log `[warning] Drush command terminated abnormally.`. Ideally we would not log that. In 2022 we [discussed removing this handling](https://drupal.slack.com/archives/C62H9CWQM/p1641305631265300?thread_ts=1641139242.236300&cid=C62H9CWQM) but decided to keep it for now. Any replacement would need to [keep these tests passing](https://github.com/drush-ops/drush/pull/4037/files). [I opened an issue in Prompts to discuss removing the exit()](https://github.com/laravel/prompts/pull/100).

I see that php:cli has a similar issue and it just registers success before launching the shell - https://github.com/drush-ops/drush/blob/24033d38d2ca3f91d7855899777ce642b93198e0/src/Commands/core/CliCommands.php#L99-L102. We could do that with prompts, I suppose.

ChristophWeber:
Wow, this is cool! Such as nice marriage of command line interface and interactivity. It will massively reduce cognitive load, especially when running lesser used drush commands, and for those of use who don't use drush daily.

src/Commands/field/EntityTypeBundleAskTrait.php
```diff
@@ -40,7 +41,8 @@ protected function askEntityType(): ?string
                 : $entityTypeDefinition->getLabel();
         }
 
-        if (!$answer = $this->io()->choice('Entity type', $choices)) {
+        if (!$answer = select('Entity type', $choices, required:  true)) {
+            // @todo No longer needed with required field?
```
DieterHolvoet: This is required in case the `-y` option is passed without the `bundle` or `entityType` arguments. Can be reproduced by running `drush fc -y`. Do Laravel prompts respect the `-y`/`--yes`/`--no-interaction` option?

DieterHolvoet:
I really think we should update `Drush\Style\DrushStyle` to use the Laravel prompt methods instead of changing this in every single Drush core command. If not, contrib and custom commands will still be using the Symfony console style prompts and the user experience will be inconsistent. We can still add new functionalities like hints as optional arguments to the `DrushStyle` methods.

src/Commands/field/EntityTypeBundleAskTrait.php
weitzman: Laravel prompts fall back to Symfony Console for non-interactive requests. The bug you mention suggests that we should automatically set non-interactive for -y if we arent already.

weitzman:
I will experiment with changing DrushStyle instead of changing all the callers.

src/Commands/ConfiguresPrompts.php
```diff
@@ -0,0 +1,177 @@
...
+use Laravel\Prompts\MultiSearchPrompt;
+use Laravel\Prompts\MultiSelectPrompt;
+use Laravel\Prompts\PasswordPrompt;
+use Laravel\Prompts\Prompt;
+use Laravel\Prompts\SearchPrompt;
+use Laravel\Prompts\SelectPrompt;
+use Laravel\Prompts\SuggestPrompt;
+use Laravel\Prompts\TextPrompt;
+use Symfony\Component\Console\Input\InputInterface;
+use Unish\UnishTestCase;
```
DieterHolvoet: This causes the following error for me:
```
Fatal error: Uncaught Error: Class "Unish\UnishTestCase" not found
```

src/Commands/field/EntityTypeBundleAskTrait.php
DieterHolvoet: You're right, `-y` still works. I don't know if this should be considered a bug. Do you mean this part
```
throw new \InvalidArgumentException(dt('The entityType argument is required.'));
```
shouldn't have to be there, and we should deal with that on a more general level?

weitzman: Yeah thats what I meant but I'm not sure what the right approach is. Happy to go with whatever you think.

src/Commands/ConfiguresPrompts.php
weitzman: Fixed

src/Commands/field/EntityTypeBundleAskTrait.php
DieterHolvoet: Let's keep the `InvalidArgumentException` around for now, changing that would be out of scope for this feature.

weitzman:
I've changed Drush Style as suggested. Seems like this will work well. Thanks for insisting on it :), @DieterHolvoet. I reverted the changes to field command. I can re-implement those if you are not up for doing it. The main thing I had done was to move validation into the validation parameter of the prompt. This gives a better error presentation when validation fails.

DieterHolvoet:
I had a look at the dependencies of laravel/prompts:
```
> composer show --tree laravel/prompts
laravel/prompts v0.1.13
├──ext-mbstring *
│  └──php >=7.1
├──illuminate/collections ^10.0|^11.0
│  ├──illuminate/conditionable ^10.0
│  │  └──php ^8.0.2
│  ├──illuminate/contracts ^10.0
│  │  ├──php ^8.1
│  │  ├──psr/container ^1.1.1|^2.0.1
│  │  │  └──php >=7.4.0
│  │  └──psr/simple-cache ^1.0|^2.0|^3.0
│  │     └──php >=8.0.0
│  ├──illuminate/macroable ^10.0
│  │  └──php ^8.1
│  └──php ^8.1
├──php ^8.1
└──symfony/console ^6.2|^7.0
   ├──php >=8.1
   ├──symfony/deprecation-contracts ^2.5|^3
   │  └──php >=8.1
   ├──symfony/polyfill-mbstring ~1.0
   │  └──php >=7.1
   ├──symfony/service-contracts ^2.5|^3
   │  ├──php >=7.2.5
   │  ├──psr/container ^1.1
   │  │  └──php >=7.4.0
   │  └──symfony/deprecation-contracts ^2.1|^3
   │     └──php >=8.1
   └──symfony/string ^5.4|^6.0
      ├──php >=8.1
      ├──symfony/polyfill-ctype ~1.8
      │  └──php >=7.1
      ├──symfony/polyfill-intl-grapheme ~1.0
      │  └──php >=7.1
      ├──symfony/polyfill-intl-normalizer ~1.0
      │  └──php >=7.1
      └──symfony/polyfill-mbstring ~1.0
         └──php >=7.1
```
There aren't really any significant new dependencies except the couple `illuminate` packages. My only concert is that at some point, the `symfony/console` version constraint might start to drift away from `drupal/core`'s `symfony/console` version constraint, but that probably shouldn't stop us from committing this.

DieterHolvoet:
@weitzman about the changes to field:create. I don’t have a lot of spare time ATM, so if you have time to re-apply them, feel free!

PR #5829, opened 2023-12-04
Fix #5667: Use CREATE USER IF NOT EXISTS, preserve existing GRANTS

### LINKED ISSUE #5667: [regression] drush 8.3.3 site-install: DROP USER drops all grants
**Describe the bug**

drush 8.3.3 introduced a change (#4261) that added a `DROP USER` command before recreating the DB user. Dropping the DB user implicitly removes all grants for the user. This is a problem if the same user is used for accessing DBs of multiple sites.

**To Reproduce**

```bash
# Install site A
drush site-install --db-url=mysql://user:pw@localhost/drupal_site_a --db-su=root --db-su-pw=super-secret minimal
# Site A works.

# Install site B
drush site-install --db-url=mysql://user:pw@localhost/drupal_site_b --db-su=root --db-su-pw=super-secret minimal
# Site A is broken because the user can’t access the first database anymore.
```

**Expected behavior**

Existing grants for the DB user should not be harmed by installing a new Drupal site.

**Actual behavior**

Existing grants are deleted when running `drush si`.

**Workaround**

Manage databases and grants outside of drush by not passing `--db-su` or downgrade to versions before 8.3.3

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.3.3
| Drupal version? | 7.96
| PHP version     | 7.4.33
| OS?             | Debian 12

**Additional information**

I was able to `git bisect` the problem to b559117576bea1f958cb5e8151af5161b3f6d3ce .

torotil:
This fixes #5667 by not deleting the DB  user on each site-install.

### DIFF
--- a/lib/Drush/Sql/Sqlmysql.php
+++ b/lib/Drush/Sql/Sqlmysql.php
@@ -94,8 +94,7 @@ EOT;
       //   We can't easily know what IP adderss or hostname would represent our server.
       $domain = ($this->db_spec['host'] == 'localhost') ? 'localhost' : '%';
       $user = sprintf("'%s'@'%s'", $this->db_spec['username'], $domain);
-      $sql[] = sprintf("DROP USER IF EXISTS %s;", $user);
-      $sql[] = sprintf("CREATE USER %s IDENTIFIED WITH mysql_native_password;", $user);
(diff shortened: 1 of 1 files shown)


torotil:
The build error doesn’t seem related to my change. :shrug:

PR #5830, opened 2023-12-04
Update commands.md

Flower7C3:


weitzman:
This is contrary to the documented usage of content tabs in our site geenrator - https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#usage. Do you have a specific reason for suggesting the change?

Issue #5836, opened 2023-12-12
Drush 11 uses mixed type (PHP 8.0.0)


BramDriesen:
**Describe the bug**
We noticed that the latest version of Drush 11 (11.6.0) is using the `mixed` type which is only available since PHP 8.0.0. Drush 11 however is claiming to support PHP 7.4+. 

**To Reproduce**
`drush cim --partial --source"some/directory/"`

```
[error]  TypeError: Argument 1 passed to Drush\Drupal\Commands\config\ConfigCommands::getDirectory() must be an instance of Drush\Drupal\Commands\config\mixed or null, string given, called in C:\home\site\wwwroot\data_web\website\vendor\drush\drush\src\Drupal\Commands\config\ConfigImportCommands.php on line 220 in Drush\Drupal\Commands\config\ConfigCommands::getDirectory() (line 393 of C:\home\site\wwwroot\data_web\website\vendor\drush\drush\src\Drupal\Commands\config\ConfigCommands.php) #0 C:\home\site\wwwroot\data_web\website\vendor\drush\drush\src\Drupal\Commands\config\ConfigImportCommands.php(220): Drush\Drupal\Commands\config\ConfigCommands::getDirectory()
#1 [internal function]: Drush\Drupal\Commands\config\ConfigImportCommands->import()
#2 C:\home\site\wwwroot\data_web\website\vendor\consolidation\annotated-command\src\CommandProcessor.php(276): call_user_func_array()
#3 C:\home\site\wwwroot\data_web\website\vendor\consolidation\annotated-command\src\CommandProcessor.php(212): Consolidation\AnnotatedCommand\CommandProcessor->runCommandCallback()
#4 C:\home\site\wwwroot\data_web\website\vendor\consolidation\annotated-command\src\CommandProcessor.php(176): Consolidation\AnnotatedCommand\CommandProcessor->validateRunAndAlter()
#5 C:\home\site\wwwroot\data_web\website\vendor\consolidation\annotated-command\src\AnnotatedCommand.php(391): Consolidation\AnnotatedCommand\CommandProcessor->process()
#6 C:\home\site\wwwroot\data_web\website\vendor\symfony\console\Command\Command.php(255): Consolidation\AnnotatedCommand\AnnotatedCommand->execute()
#7 C:\home\site\wwwroot\data_web\website\vendor\symfony\console\Application.php(1039): Symfony\Component\Console\Command\Command->run()
[...]

PR #5837, opened 2023-12-12
Drush 11 uses mixed type (PHP 8.0.0) #5836

BramDriesen:
https://github.com/drush-ops/drush/issues/5836


greg-1-anderson:
In theory, Drush 11 is EOL; however, some folks are likely still relying on Drush 11 if they are still stuck on PHP 7.4, so perhaps a post-EOL release might be warranted. I guess this wasn't caught because Drush 11 is only testing against PHP 8.0.

Not sure why the test is failing; it's something to do with a re-entrant problem &/or preflight not correctly identifying that the autoloader of the site being bootstrapped is the same as the Drush autoloader.

Anyway, going to merge this PR so that at least 11.x-dev, at a minimum, has the fix.

PR #5845, opened 2023-12-29
Log CLI errors using Drupal error handler

Chi-teck:
Currently PHP errors that happen in Drush commands are not logged to Drupal. When Drush commands are running in background through Cron or Supervisor there is no way to notice these errors. For that reason I have to prepend each custom Drush command with the following line.
```php
\set_error_handler('_drupal_error_handler');
```
The downside of this is that it'll not display a error when the command is running in foreground.

The simplest solution is to call the Drupal error handler manually.

### DIFF
--- a/src/Runtime/ErrorHandler.php
+++ b/src/Runtime/ErrorHandler.php
@@ -24,6 +24,7 @@ class ErrorHandler implements LoggerAwareInterface, HandlerInterface
 
     public function errorHandler($errno, $message, $filename, $line)
     {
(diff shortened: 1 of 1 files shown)


weitzman:
Drupal may not be bootstrapped yet. thsi doesnt look right to me. Also see the comment at top 

```
**
 * Log PHP errors to the Drush log. This is in effect until Drupal's error
 * handler takes over.
 */
```

weitzman:
Not planned.

PR #5856, opened 2024-01-26
Skip abstract entity type classes from being exposed with their short…

ceesgeene:
It is possible the entity type class is abstract. For example the [quiz module](https://www.drupal.org/project/quiz) uses this for its QuizResultAnswer and QuizQuestion.

When executing `drush php` on a site with the quiz module enabled the following exception is triggered:

```
Fatal error: Class QuizResultAnswer contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods (Drupal\quiz\QuizAnswerInterface::score, Drupal\quiz\QuizAnswerInterface::getResponse) in {...}/vendor/drush/drush/src/Commands/core/CliCommands.php(319) : eval()'d code on line 1
```

### DIFF
--- a/src/Commands/core/CliCommands.php
+++ b/src/Commands/core/CliCommands.php
@@ -314,7 +314,7 @@ final class CliCommands extends DrushCommands
             $parts = explode('\\', $class);
             $end = end($parts);
             // https://github.com/drush-ops/drush/pull/5729 and https://github.com/drush-ops/drush/issues/5730.
-            if ($reflectionClass->isFinal() || class_exists($end)) {
+            if ($reflectionClass->isAbstract() || $reflectionClass->isFinal() || class_exists($end)) {
                 continue;
             }
             // Make it possible to easily load revisions.


weitzman:
@chx any thoughts?

chx:
knee jerk reaction: 

Untitled

I will try to think on something. Reflection probably.

chx:
Ah sorry for not getting back to you, I was thinking this over just today on the plane: I would argue the opening statement is not correct.

> It is possible the entity type class is abstract. 

Possible? In the sense that core doesn't throw an error at you for doing it, sure it is possible but there can be any number of calls which will die. If you have an abstract base class for your entities do not mark it with `@ContentEntityType`. This is pure nonsense. Such an entity can not be `::create()` 'd loaded or anything. It's not an entity. And our annotations do not inherit so you do not even save work by doing this.

weitzman:
That makes sense. Quiiz is misbehaving here. Anyway, this PR is already merged and it added just a small condition so lets keep it. Thanks everyone.

PR #5857, opened 2024-01-26
Remove the use of `mt_rand`

bbeversdorf:
`mt_rand` does not generate cryptographically secure values (https://www.php.net/manual/en/function.mt-rand.php). This logic should be updated to follow best practices because this key should be random. Although the outcome is the same, because of the limited character set, it sets precedence to follow good security practices.


weitzman:
what does drupal 10 use? i'd like to call its routine, if possible.

PR #5858, opened 2024-01-27
Add --filter option to core:route to limit results.

mrbaileys:
At the moment, drush core:route either returns all routes, or you can get information about an individual route by using the --name or --path options. I would be useful to allow users to limit the results returned by `drush core:route` by specifiying --filter.

ex. `drush core:route --filter=views_ui` to get all views_ui.* routes.

### DIFF
--- a/src/Commands/core/DrupalCommands.php
+++ b/src/Commands/core/DrupalCommands.php
@@ -133,12 +133,14 @@ final class DrupalCommands extends DrushCommands
     #[CLI\Command(name: self::ROUTE, aliases: ['route'])]
     #[CLI\Usage(name: 'drush route', description: 'View all routes.')]
     #[CLI\Usage(name: 'drush route --name=update.status', description: 'View details about the <info>update.status</info> route.')]
+    #[CLI\Usage(name: 'drush route --filter=devel', description: 'Show only routes starting with devel.')]
     #[CLI\Usage(name: 'drush route --path=/user/1', description: 'View details about the <info>entity.user.canonical</info> route.')]
     #[CLI\Usage(name: 'drush route --url=https://example.com/node/1', description: 'View details about the <info>entity.node.canonical</info> route.')]
     #[CLI\Option(name: 'name', description: 'A route name.')]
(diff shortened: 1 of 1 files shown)


weitzman:
The --filter functionality that most commands have is provided automatically by https://github.com/consolidation/output-formatters. Instead of this PR, I'd love to get simple --filter working for all yaml commands. Any thoughts, @greg-1-anderson ?

greg-1-anderson:
Yes, I agree. The `--filter` option is not a direct feature of output-formatters, but is instead provided by [filter-via-dot-access-data](https://github.com/consolidation/filter-via-dot-access-data). Drush is [already configured to use this project](https://github.com/drush-ops/drush/blob/1f9e78ad1aace1efa26bbaaa2e30c628ff8f5bf6/src/Runtime/ServiceManager.php#L105).

To add filters to a command, you merely need to annotate the command with `@filter-default-field` or `@filter-output`. I [didn't find either of these in use in Drush](https://github.com/search?q=repo%3Adrush-ops%2Fdrush+filter-default-field&type=code) when I did a Github search, so I don't know if this feature is actually being used by any commands right now, or if there is an affordance to select it with an attribute instead of an annotation.

weitzman:
Both of those annotations are in use in their Attribute equivalent.

Looks like all thats needed here is to add the https://github.com/drush-ops/drush/blob/12.x/src/Attributes/FilterDefaultField.php attribute with a value of `name`. For extra credit, implement the https://github.com/drush-ops/drush/blob/12.x/src/Attributes/FieldLabels.php attribute and document the fields that describe a route. An example of a yaml output command that does this is [site:alias](https://github.com/drush-ops/drush/blob/12.x/src/Commands/core/SiteCommands.php#L95) 

Folks can learn more about filtering at https://www.drush.org/12.x/output-formats-filters/#output-filters

weitzman:
This command outputs slightly differently depending depending on whether a --path has been provided (a list versus a detail). The list view cant support --filter unless we put name and path keys in the output. Thats not backward compatible so I am leaning towards not fixing this issue. Use grep for now.

Issue #5862, opened 2024-02-08
DrushLoggerServiceProvider is not registered after a cache rebuild


mcdruid:
**Describe the bug**

There's an inconsistency with how logging behaves in drush on a fresh install vs. after a cache rebuild.

As far as I can tell, this is because after the cache rebuild the `DrushLoggerServiceProvider` doesn't register the drush logger with the container. I'm not sure why this is.

**To Reproduce**

Clean install:

```
$ drush -y si

$ drush ev "\Drupal::logger('Test')->error('Hello, this is an error from drush');"
 [error]  Hello, this is an error from drush 
```

Now rebuild caches and try again:

```
$ drush cr
 [success] Cache rebuild complete.
$ drush ev "\Drupal::logger('Test')->error('Hello, this is an error from drush');"
$
```

This is a hacky way to access / view the registered services, but hopefully proves the point:

before (drush has registered its logger):

```
$ drush ev "var_dump((fn() => \$this->loggers)->call(\Drupal::logger('foo')));"

array(1) {
  [0] =>
  array(2) {
    [0] =>
    class Drupal\dblog\Logger\DbLog#4641 (4) {
      protected $connection =>
      class Drupal\mysql\Driver\Database\mysql\Connection#874 (25) {

...snip...

    }
    [1] =>
    class Drush\Log\DrushLog#4724 (2) {
      protected Drupal\Core\Logger\LogMessageParserInterface $parser =>
      class Drupal\Core\Logger\LogMessageParser#4704 (0) {
      }
      protected ?Psr\Log\LoggerInterface $logger =>
      NULL
    }
  }
}
```

after (only dblog is registered):

```
$ drush ev "var_dump((fn() => \$this->loggers)->call(\Drupal::logger('foo')));"

array(1) {
  [0] =>
  array(1) {
    [0] =>
    class Drupal\dblog\Logger\DbLog#1056 (4) {
      protected $connection =>
      class Drupal\mysql\Driver\Database\mysql\Connection#874 (25) {

...snip...

    }
  }
}
```

**Expected behavior**

Drush continues to successfully register its logger after a cache rebuild, and therefore the behaviour of the cli is consistent.

**Actual behavior**

After a `drush cr` drush doesn't do its own logging as its logger is AWOL.

**Workaround**

I've not yet worked out if there's a way of restoring the drush logger once it's no longer being registered.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.4.3.0
[...]

PR #5863, opened 2024-02-08
DrushLoggerServiceProvider is not registered after cache rebuild

mcdruid:
Initial attempt to address #5862


weitzman:
Merged to 12.x, and cherry picked to 13.x. Thanks for digging into this.

PR #5883, opened 2024-03-02
Fix site install command broken in D11

### LINKED ISSUE #5882: Site install command is broken in D11
**Describe the bug**

Site install command is broken in D11, most probably because of https://www.drupal.org/project/drupal/issues/3256524.

In D11, the non-interactive installer array should not use the driver name, but the driver namespace as key; https://git.drupalcode.org/project/drupal/-/commit/6490db80696b8db659d45516d7678d19b6a02bcb#999cee0a2d033814d9a2432c81ded7dcf9543d11_133_133

**To Reproduce**

Run vendor/bin/drush site-install standard --db-url=$SIMPLETEST_DB -y, 

where $SIMPLETEST_DB = "sqlite://localhost/test_database?module=sqlite" 

latest D11 codebase

```
 You are about to:
 * Create a sites/default/settings.php file
 * CREATE the 'localhost/test_database?module=sqlite' database.

 // Do you want to continue?: yes.                                              

Notice: ] Starting Drupal installation. This takes a while.
Notice: ] Performed install task: install_select_language
Notice: ] Performed install task: install_select_profile
Notice: ] Performed install task: install_load_profile
Notice: ] Performed install task: install_verify_requirements
Error: ]  The submitted value sqlite in the Database type element is not allowed. 
Warning: ] Undefined array key "database" Connection.php:106

In install.core.inc line 967:
                                                                               
  The submitted value <em class="placeholder">sqlite</em> in the <em class="p  
  laceholder">Database type</em> element is not allowed.                       
                                                                               

Error: Process completed with exit code 1.

```
**Expected behavior**
Successful install

**Actual behavior**
See above

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.4
| Drupal version? | 11.x
| PHP version     | 8.3

mondrake:
Fixes #5882


mondrake:
We need to get https://www.drupal.org/project/drupal/issues/3425129 solved here, otherwise the driver Connection class cannot be loaded.

mondrake:
This is doing more than what the title says, in general it's about Drupal 11 compatibility.

The PR requires as **minimum stable** the next patch release of Drupal 10.2, that will incorporate the fix at https://www.drupal.org/project/drupal/issues/3425129 .

site-install with a `--db-url` option will work now with Drupal canonical DB URLs; found out the ones used here in Drush were quite out of the standard.

Work is needed to get Drush site install interacting with CLI questions, not sure worth doing here or in a follow up.

mondrake:
We would need `chi-teck/drupal-code-generator` to be able to generate a unish_article.info.yml file supporting Drupal core ^11 to pass the highest tests.

mondrake:
I think this is as good as it could go for the scope of this issue. Evidently there are other D11 compatibility issues that need to be addressed, but I'd say it'd better to have individual follow-ups.

tests/unit/MigrateRunnerTest.php
```diff
@@ -184,6 +182,11 @@ protected function getDatabaseConnection()
         return $connection;
     }
 
+    public static function webroot(): string
+    {
+        return Path::join(Path::canonicalize(dirname(__DIR__, 2)), 'sut');
+    }
+
```
weitzman, line 195: Why are we overrideing the webroot from the parent? Seems like it does same thing.

mondrake: The parent of this test is vanilla `TestCase`, so we won't have that there. Unless I'm missing something?

PR #5895, opened 2024-03-09
Update SiteInstallCommands for interactive install

### LINKED ISSUE #5894: Fix interactive site:install for Drush 13
**Describe the bug**
Follow up of #5882 

#5882 fixed support for the `--db-url` option of the site:install command to make it D11 compatible. Running site:install with no options now does fail.

**To Reproduce**
Run site:install with no options.

**Expected behavior**
Installed site

**Actual behavior**
Failure

**Workaround**
Use `--db-url` option.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 13.x
| Drupal version? | 11.x
| PHP version     | 8.x

**Additional information**
~~As a prerequisite we need [DatabaseDriver::load should have a non-service alternative for class loading](https://www.drupal.org/project/drupal/issues/3426789) committed to Drupal core.~~ Edit: this is no longer necessary for #5895.

mondrake:
Fixes #5894.


mondrake:
Here I needed to boot the installer kernel to get some services necessary to list db drivers, string translation, etc

Could use some help for a better solution.

The bright side of this is that the input required to the user is much more guided and provides the same information as the web installer.

mondrake:
OK this is reviewable now I think.

PR #5904, opened 2024-03-12
Fix aliases for "php:cli" command

webflo:


weitzman:
Thanks. If you could merge https://github.com/webflo/drupal-finder/pull/64 that would be great. Thanks!

weitzman:
Backported to 12.x as well.

PR #5915, opened 2024-03-16
Add --extra-dump option to archive:dump

liberatr:
The --extra-dump flag works fine with sql:dump. This adds the same option to archive:dump, which passes the option on to the sql command in turn.


src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -95,6 +97,7 @@ public function dump(array $options = [
         'generator' => InputOption::VALUE_REQUIRED,
         'generatorversion' => InputOption::VALUE_REQUIRED,
         'exclude-code-paths' => InputOption::VALUE_REQUIRED,
+        'extra-dump' => false,
```
weitzman: Value should be `self::REQ`. See sql:sync for an example.

weitzman:
Tests will pass if you merge in latest 13.x

Issue #5917, opened 2024-03-18
Drush commands throw TypeError: ucfirst() when $databases array is empty


Supreetam09:
**Describe the bug**
We recently upgraded from Drupal Core 10.1.8 to 10.2.3 and Drush from 11.6.0 to 12.4.3

After update `DocrootUpdate` task during ACSF deployment started failing:
```
>>> \drush8 -r '/mnt/tmp/drupal.01dev/acsf_temporary_repo/1709655175/docroot' status --fields drupal-version

TypeError: ucfirst(): Argument #1 ($string) must be of type string, null given in ucfirst() (line 122 of /mnt/tmp/drupal01dev/acsf_temporary_repo/1709655175/vendor/drush/drush/src/Sql/SqlBase.php).
```

Its not an issue with ACSF but with drush only. **ALL** basic drush commands failing with the same error.

**To Reproduce**
The same command was running fine in local but failing in ACSF. We figured out its only happening when `$databases` is set to empty array in `settings.php` and no overrides gets applied. If we make `$databases` empty - issue is reproducible in local too.
And yes - drush commands like `drush status` etc runs ok with empty databases.

In local, `$databases` is defined in `local.settings.php` - which does not get deployed - so at first the issue wasn't reproducible in local and we thought its some issue with ACSF only.
Also since its not yet deployed to cloud, and cloud database overrides settings isn't available during deployment/docrootUpdate task, hence `$databases` is empty `[]` - and thats how error in ACSF was thrown and identified.

**Note:** Setting `$databases` to empty array in Drupal Core 10.1.8 and Drush 11.6.0 - doesn't throw the error

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.4.3
| Drupal version? | 10.2.3
| PHP version     | 8.1.27
| OS?             | Mac

Supreetam09:
Created PR https://github.com/drush-ops/drush/pull/5918

weitzman:
Fixed via that PR

Issue #5777, opened 2023-10-09
sql:dump outputs deprecation warning with MariaDB 11


LordAro:
**Describe the bug**

`sql:dump` uses a hardcoded `mysqldump` command. Since mariadb v11 `mysqldump` now outputs a deprecation warning

https://github.com/drush-ops/drush/blob/cb2610d5aa007b80bae4126b4cfb5454cd14dfe3/src/Sql/SqlMysql.php#L163

**To Reproduce**
What did you do?

Run `drush sql:dump`

**Expected behavior**
What did you expect would happen?

`drush sql:dump` to work cleanly

**Actual behavior**
What happened instead?

```
mysqldump: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb-dump' instead
```
output to stderr

**Workaround**
Is there another way to do the desired action?

Not that I'm aware of. Maybe dump command should be configurable?

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 11.6 (but also git HEAD)
| Drupal version? | 9.5.11
| PHP version     | 8.1.23
| OS?             | Linux (but all)

**Additional information**

```
$ mariadb --version
mariadb from 11.1.2-MariaDB, client 15.2 for debian-linux-gnu (x86_64) using  EditLine wrapper
```

weitzman:
We could add a driver for mariadb. We just have to be able to tell when thats in use over mysql.

terryzwt:
`drush sqlc --extra=--skip-ssl` as a trick

PR #5918, opened 2024-03-18
Fix null param to ucfirst() fatal error when DB information is missing

Supreetam09:
Fixes https://github.com/drush-ops/drush/issues/5917


LordAro:
This undid the fix for #5777 ( #5787 )

PR #5869, opened 2024-02-19
Fix #5838: Run batch operations at the end of config import.

vever001:
See https://github.com/drush-ops/drush/issues/5838

PR #5921, opened 2024-03-19
Prevent batch-process hidden commands to throw a fatal error if the g…

DuaelFr:
…iven batch does not exist anymore.

`UnstructuredListData` has been introduced in #3750 and #3758 to display the return of `drush_batch_command()`.
As this function can return `FALSE`, we need to check the value before creating the return object to prevent a fatal error due to the mistyped argument (ArrayObject only allow the value to be an array or an object, not a boolean).

### DIFF
--- a/includes/batch.inc
+++ b/includes/batch.inc
@@ -152,7 +152,10 @@ function _drush_backend_batch_process($command = 'batch-process', $args = [], $o
       $process->setTimeout(null);
       // Suppress printing stdout since a JSON array is returned to us here.
       $process->run($process->showRealtime()->hideStdout());
-      $result = $process->getOutputAsJson();
+      try {
+        $result = $process->getOutputAsJson();
+      }
+      catch (\InvalidArgumentException) {}
       $finished = !$process->isSuccessful() || (isset($result['drush_batch_process_finished']) && $result['drush_batch_process_finished'] === TRUE);
     }
   }
--- a/src/Commands/core/BatchCommands.php
+++ b/src/Commands/core/BatchCommands.php
@@ -20,9 +20,9 @@ class BatchCommands extends DrushCommands
     #[CLI\Argument(name: 'batch_id', description: 'The batch id that will be processed.')]
     #[CLI\Help(hidden: true)]
     #[CLI\Bootstrap(level: DrupalBootLevels::FULL)]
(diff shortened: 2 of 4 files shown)


DuaelFr:
This prevents the fatal error but it's still triggering some exception:
```
In ProcessBase.php line 155:
                    
  Output is empty.  
```

DuaelFr:
Last commit wraps the `$process->getOutputAsJson()` call into a try catch block.

weitzman:
Can you say a bit more about when this happens? Is there an update hook that causes this?

vever001:
Interesting, we stumbled on a similar error with xdebug 3.3+ in https://github.com/drush-ops/drush/pull/5869

> Can you say a bit more about when this happens? Is there an update hook that causes this?

+1
Do you have xdebug "develop" mode enabled by any chance?
In any case, I have not checked in detail but maybe it could be a good idea to harden this/make it more resilient.

DuaelFr:
> Can you say a bit more about when this happens? Is there an update hook that causes this?

In my case, the issue happened first when I was trying to enable a module containing default content (from the default_content module). I traced it down to modules using a batch within a `hook_modules_installed()` implementation (like the `locale` module).

> Do you have xdebug "develop" mode enabled by any chance?

That's a good point! I'm still failing to reproduce this on a fresh Drupal + Drush install so I guess the size of the codebase matters to run into what seem to be a race condition.

> In any case, I have not checked in detail but maybe it could be a good idea to harden this/make it more resilient.

I agree. I'm not sure my approach is the best here. The thing is that knowing that the `drush_batch_command()` function is expected to return `FALSE` in some cases should prevent us of using its return as an array without checking first. What would be the best handling of this kind of edge case?

vever001:
I traced this in https://github.com/drush-ops/drush/pull/5869 and found that the batch record was not written in time in DB when it should be. This relates to transactions and https://www.drupal.org/project/drupal/issues/3405976 (see also upstream xdebug issue and follow up tickets). This is relatively easy to reproduce with xdebug "develop" mode enabled. Disabling it solves it for now, as it causes all kinds of crazy issues.

weitzman:
Thanks for chasing that down. I'm sure that was quite a debugging session.

PR #5925, opened 2024-03-21
Trust existing config to have a profile set correctly

alexpott:
This also and more importantly allow sites with no profile to be installed from config. This is now a reality after https://www.drupal.org/project/drupal/issues/1170362 has landed.

Note that for this to actually work we need to land https://www.drupal.org/project/drupal/issues/3432602 too.


weitzman:
LGTM. We use PSR12 so needs a small fix.

weitzman:
Also cherry picked to 12.x

PR #5927, opened 2024-03-23
A site local Drush no longer redispatches to another site local Drush. And remove archive:restore

weitzman:
You must call the right drush to begin with. For example, we would no longer allow the feature that was fixed in #5686 

One motivation for removing this is to remove the DrupalFinder dependency. A second motivation is general simplification of Preflight.

There is one usage of DrupalFinder in ArchiveRestoreCommands that needs to be redone so it does not use DrupalFinder. I would love some help with that.


greg-1-anderson (approved):
LGTM

PR #5934, opened 2024-03-25
Drush commands for workspaces - #5933

### LINKED ISSUE #5933: Drush commands for workspaces
t would be useful to provide Drush commands for workspaces actions.

For example, to automatically publish a workspace in a test environment.

A drush command to publish a workspace

**Additional context**
I'd created https://www.drupal.org/project/drupal/issues/3435823, but realised that it belongs in drush, rather than drupal core.

malcomio:
Aims to fix #5933


src/Commands/core/WorkspacesCommands.php
```diff
@@ -0,0 +1,63 @@
...
+    ];
+
+    // Does this workspace have any content to publish?
+    $diff = $workspace_publisher->getDifferringRevisionIdsOnSource();
+    if (empty($diff)) {
+      $this->io()->warning(dt('There are no changes that can be published from %source_label to %target_label.', $args));
+      return;
+    }
+
+    $workspace->publish();
```
malcomio: this fails with wse enabled, because `\Drupal\wse_preview\Negotiator\CookieWorkspaceNegotiator::unsetActiveWorkspace` tries to get the user session

src/Commands/core/WorkspacesCommands.php
```diff
@@ -0,0 +1,63 @@
...
+    );
+  }
+
+  /**
+   * Publish a workspace.
+   */
+  #[CLI\Command(name: 'workspaces:publish')]
+  #[CLI\Argument(name: 'id', description: 'The workspace to publish.')]
+  #[CLI\Usage(name: 'workspaces:publish stage', description: 'Publish the stage workspace')]
+  public function commandName($id) {
```
amateescu: We need to ensure that the module is enabled: `#[CLI\ValidateModulesEnabled(modules: ['workspaces'])]`

src/Commands/core/WorkspacesCommands.php
amateescu: I'll look into this asap :)

src/Commands/core/WorkspacesCommands.php
```diff
@@ -0,0 +1,63 @@
...
+      $container->get('entity_type.manager'),
+    );
+  }
+
+  /**
+   * Publish a workspace.
+   */
+  #[CLI\Command(name: 'workspaces:publish')]
+  #[CLI\Argument(name: 'id', description: 'The workspace to publish.')]
+  #[CLI\Usage(name: 'workspaces:publish stage', description: 'Publish the stage workspace')]
```
weitzman:
```suggestion
  #[CLI\Usage(name: 'workspaces:publish stage', description: 'Publish the stage workspace')]
  #[CLI\ValidateModulesEnabled(modules: ['workspaces'])]
```

weitzman:
This needs Drush test coverage. I think an integration test will fit well. As setup the command would create a workspace, make it active, create a node in the workspace, run the drush command, and ensure the node is now published.

src/Commands/core/WorkspacesCommands.php
weitzman: Has this been resolved in wse?

src/Commands/core/WorkspacesCommands.php
```diff
@@ -0,0 +1,64 @@
...
+    return new static(
+      $container->get('workspaces.operation_factory'),
+      $container->get('entity_type.manager'),
+    );
+  }
+
+  /**
+   * Publish a workspace.
+   */
+  #[CLI\Command(name: 'workspaces:publish')]
```
weitzman: We should use class constant instead of string for command name. See other drush core commands.

weitzman (commented):
Ideally we add an interact() method that lets you pick a workspace from a select if the argument is not provided. I guess show the 20 most recently active workspaces.

weitzman:
We can skip the interact() method for now.

I stubbed out a test but it needs more work and I cant work on this anymore. Hopefully someone else can.

weitzman:
Note that the 'stage' workspace is no longer auto-created by Drupal core so the test will have to create its own workspace.

weitzman:
I added a test. which workspaces should we show in a select list when none is provided? All published? Sorted by alphabetical or some kind of reverse chrono?

Issue #5942, opened 2024-03-27
Add option to update site UUID on config import


andycarlberg:
**Is your feature request related to a problem? Please describe.**
Drupal 8+ requires a site UUID to match for config import to be successful. This presents a lot of positives but can also be frustrating when working with multiple environments. We have seen some solutions to this, most revolving around getting the UUID defined in exported config and setting the target environment to use the same UUID. Right now, this requires a custom command or another tool or module on top of Drush to accomplish.

**Describe the solution you'd like**
Would it make sense to just make this an option on the `config:import` Drush command? Drush has all the tools to figure out what the exported UUID is and set it accordingly. 

I would not expect this be the default state for the command. It is probably also in conflict with certain other flags (like `--partial`).

**Describe alternatives you've considered**
Acquia BLT used to quietly handle this for us on a number of projects. Since it is EOL, I'm looking at alternatives. A pre-command-hook could be used to implement this feature (as I have done on projects where this is an issue). There's also a module that can be installed on Drupal to remove the requirement but this feels like overkill.

weitzman:
I'm not initially excited by handling this in core Drush, but lets discuss merits once someone makes a PR.

andycarlberg:
I went ahead and opened a PR. This worked on a project locally but it is just a first pass.

The site UUID has to be addressed at least once on most projects using more than one environment or server. It seems simple and valuable. This replaces adding a hook to every project for a commonly needed feature.

alexpott:
Why do you need this feature - why don't the site IDs match? Are you installing minimal and trying to import a new config set over the top? If so why aren't you using drush si --existing-config?

It would be great to have some more info on the exact use-case.

The only use-cases I have for something like this is when an install profile does not support installing from config. Which might well be the case with Acquia stuff but since that is all deprecated maybe you should consider switching to your own profile where you can add a config/sync directory and manage the config from there. See https://www.drupal.org/project/install_profile_generator for something that can do that for you.

Also in Drupal 10.3.x we're getting the ability to uninstall install profiles and they will be installable from config (once the one outstanding issues lands).

My 2 pennies are that everything is moving away from this need and so this should not be added to Drush.

andycarlberg:
I've been on leave for a bit and wasn't aware of that change coming in
10.3.x.

I don't want to get into the business reasons we don't have a profile for
this but, yes, usually we can't install from config.

If this problem is going away though, I guess it's not a necessary change.

PR #5946, opened 2024-03-29
#5942: Add option to override site uuid on config import.

andycarlberg:
This adds an option to the `config:import` command to set the site UUID automatically, pulling from the config sync directory by default.

This is my first contribution to Drush and I'm not very familiar with the internals. If I've handled something wrong here, happy to change it. This was really just a first pass at addressing a problem I've noted.


weitzman:
https://github.com/drush-ops/drush/issues/5942

Issue #5952, opened 2024-04-08
Database update step for `drush deploy` behaves differently than `drush updb --no-cache-clear`


elaman:
**Describe the bug**
We use `drush deploy` in our CI, however recently we had two deploy failures. The most recent one being `simple_sitemap` module [changing its services definition](https://git.drupalcode.org/project/simple_sitemap/-/compare/4.1.8...4.1.9?from_project_id=59644&page=2&straight=false#2bbb4789145a3391adaffb95cbb6d83d07939e50). In both cases database update step in `drush deploy` executed in a different way than `drush updb  --no-cache-clear` it is trying to create a shortcut for.

**To Reproduce**
- Install vanilla Drupal 10. Use `drush si` to install it.
- Install old version of simple_sitemap `composer require drupal/simple_sitemap:4.1.8` and enable it using `drush en simple_sitemap`.
- Update simple_sitemap to the latest version `composer require drupal/simple_sitemap:4.1.9`.
- Execute `drush deploy`.

**Expected behavior**
The command should've updated the database and imported the configs successfully.

**Actual behavior**
Immediately received an error:

> TypeError: Drupal\simple_sitemap\Queue\QueueWorker::__construct(): Argument #2 ($key_value) must be of type Drupal\Core\KeyValueStore\KeyValueFactoryInterface, Drupal\Core\State\State given, called in /var/www/html/web/core/lib/Drupal/Component/DependencyInjection/Container.php on line 259 in /var/www/html/web/modules/contrib/simple_sitemap/src/Queue/QueueWorker.php on line 153 #0 /var/www/html/web/core/lib/Drupal/Component/DependencyInjection/Container.php(259): Drupal\simple_sitemap\Queue\QueueWorker->__construct(Object(Drupal\simple_sitemap\Settings), Object(Drupal\Core\State\State), Object(Drupal\simple_sitemap\Queue\SimpleSitemapQueue), Object(Drupal\simple_sitemap\Logger), Object(Drupal\Core\Extension\ModuleHandler), Object(Drupal\Core\Entity\EntityTypeManager), Object(Drupal\Core\ProxyClass\Lock\DatabaseLockBackend))
> #1 /var/www/html/web/core/lib/Drupal/Component/DependencyInjection/Container.php(177): Drupal\Component\DependencyInjection\Container->createService(Array, 'simple_sitemap....')
[...]

PR #5953, opened 2024-04-08
#5952: Database update step for `drush deploy` behaves differently than `drush updb --no-cache-clear`

elaman:
Fixes the issue #5952

As a reminder, Drush's usage of UpdateKernel started in https://github.com/drush-ops/drush/pull/5953. This PR would again use it in the main request of _deploy_ command, in addition to deploy's subrequest to _updatedb_.


bircher:
I think that makes sense.

alexpott:
Can this run without boostraping and no kernel - the site the commands operate on is potentially having significant change. This command really should try to load as little from it as possible.

weitzman:
That sounds really useful and easy to accomplish. I'll push that change today.

weitzman:
Merged, and backported to 12.x

elaman:
@weitzman  thank you!

Issue #5960, opened 2024-04-19
Environment variables ignored if the value provided in the configuration file 


AlexSkrypnyk:
**Describe the bug**
Given `$HOME/.drush/drush.yml` file with the following content:
```
options:
  root: myroot
  uri: mysite.local
```
When I call

```
DRUSH_OPTIONS_URI=other.domain ./vendor/bin/drush st
```
Then I see the output
```
Drupal version : 10.1.4                                      
Site URI       : mysite.local
```

**Expected behavior**
Then I see the output
```
Drupal version : 10.1.4                                      
Site URI       : other.domain
```

**Workaround**
Is there another way to do the desired action?

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.5.11
| Drupal version? | 10.1
| PHP version     | 8.1
| OS?             | Mac

**Additional information**

The issue lies in Drush `ConfigLocator`. The constructor adds the user context after env context. `consolidation/log`'s `ConfigOverlay::findContext()` looks up the contexts in reversed order hence the `options.uri` defined in `/home/.drush/drush.yml` (user context) supersedes the `DRUSH_OPTIONS_URI` var (env context).

In `ConfigLocator::__contruct()`
```
$this->config->addPlaceholder(self::DRUSH_CONTEXT);
if (!empty($envPrefix)) {
    $envConfig = new EnvConfig($envPrefix);
    $this->config->addContext(self::ENV_CONTEXT, $envConfig);
}
$this->config->addPlaceholder(self::USER_CONTEXT);
```
If I move `USER_CONTEXT` above `ENV_CONTEXT` the `DRUSH_OPTIONS_URI` will take precedence as expected.

weitzman:
Fixed in #5961

AlexSkrypnyk:
This will be available in Drush 13. Not backported to Drush 12.

PR #5961, opened 2024-04-19
Issue #5960: Environment variables ignored if the value provided in the user configuration file.

sonnykt:
Github issue: https://github.com/drush-ops/drush/issues/5960

Current behaviour:
Screenshot 2024-04-19 at 2 30 58 PM

Fixed behaviour:
Screenshot 2024-04-19 at 2 35 56 PM


greg-1-anderson (approved):
LGTM

tests/unit/ConfigLocatorTest.php
```diff
@@ -91,6 +91,24 @@ function ($item) {
         $this->assertEquals($expected, $aliasPaths);
     }
 
+    /**
+     * Test an env var supersedes user's configuration file.
+     */
+    public function testEnvVar()
+    {
+        putenv('TEST_HOME=An env overridden setting');
```
weitzman: How about making sure that we get user value initially, and that it changes after the env var is set.

sonnykt: Isn't the existing test `testLoadAll()` already checking that?

sonnykt: Anyway, I've updated the test.

AlexSkrypnyk:
@weitzman 
Could you please let me know if there anything else that needs to be done here to help merging this. Thank

weitzman:
Merged ... Even though this is a bug fix, I'm hesitant to backport it as some sites may be relying on the buggy behavior.

PR #5963, opened 2024-04-24
Allow command classes to set their own logger channel

jurgenhaas:
Following [this thread on Drupal Slack](https://drupal.slack.com/archives/C62H9CWQM/p1713769812436529) the ServiceManager inflection overrides the logger channel of a Drush command, if that came with its own channel to e.g. log to the Drupal channel e.g. in watchdog.

With this PR, the service manager would only set its logger if the command had not defined their own before.


jonathan1055:
Is there a way that I could test this change in a phpunit test run in gitlab_templates pipeline? Happy to help investigate, but not sure how to do it.

jurgenhaas:
@jonathan1055 not sure TBH. I haven't ever php unit tested a module in a drush context. Not sure if that's a thing.

jonathan1055:
The contrib module provides drush commands, and I have a phpunit test that checks the commands work as expected. I have had to disable that test group for a while. It is still working correctly on DrupalCI phpunit test, but not in gitlab pipelines.

src/Commands/DrushCommands.php
```diff
@@ -65,6 +66,18 @@ protected function io(): DrushStyle
         return $this->io;
     }
 
+    /**
+     * Sets a logger, if none is available yet.
+     *
+     * @param LoggerInterface $logger
+     */
+    public function setLoggerIfEmpty(LoggerInterface $logger): void
```
greg-1-anderson: Is this used by anything other than the service manager? If not, I think it would be better to just move the condition to `inflect`, and omit the extra trivial function in the base class.

jurgenhaas: We can't do that, the logger variable is protected and can't be accessed from the service manager.

weitzman: what about making logger() public and using it here? maybe only on 13.x branch due to BC.

```
if (!$object->logger()) {
  $object->setLogger($container->get('logger'))
}
```

greg-1-anderson: You can tell whether the logger variable is set via https://github.com/drush-ops/drush/blob/95e0acf9ac0cb2a235e217c31d46b048f08a5536/src/Commands/DrushCommands.php#L84

src/Runtime/ServiceManager.php
```diff
@@ -452,7 +452,7 @@ public function inflect(DrushContainer $container, $object): void
             $object->setConfig($container->get('config'));
         }
         if ($object instanceof LoggerAwareInterface) {
```
greg-1-anderson: OK, I guess we can't check `$object->logger()` here as I suggested unless we make `logger()` public. Did not consider impact of that, I think that method is defined in another project.

src/Runtime/ServiceManager.php
```diff
@@ -452,7 +452,7 @@ public function inflect(DrushContainer $container, $object): void
...
-            $object->setLogger($container->get('logger'));
+            $object->setLoggerIfEmpty($container->get('logger'));
```
greg-1-anderson: `setLoggerIfEmpty()` is not a method of `LoggerAwareInterface`, so there is a mismatch between the `instanceof` above and the usage here.

src/Commands/DrushCommands.php
jurgenhaas: That method is protected, so we can't call it from outside, i.e. from the service manager.

src/Runtime/ServiceManager.php
weitzman: That method is in Drush so we can change it but only for Drush 13 due to BC. I'll open a new PR for that, and fixup this one for Drush 12.

weitzman: I went the other way around. The Drush 12.x PR is at #5980. This PR will be for 13.x

weitzman:
Now this is a small, helpful change for 13.x. Thanks everyone.

Issue #5971, opened 2024-04-26
Twig compile - provide searchpaths argument


drupalranger:
**Is your feature request related to a problem? Please describe.**
When running `drush twigc` in drush 12, it will compile templates for all modules and all themes that exist in filesystem, without taking application setup into account, and without checking if the modules/themes are actually installed or not. It would be nice to have a quick way to decide which directories should be processed.

**Describe the solution you'd like**
Leave default functionality as it is, but allow to override `$searchpaths` parameters, so we can decide which directories we want to compile. Existing example is already available in `twig:unused` command. 

**Describe alternatives you've considered**
Create custom drush command I guess. 

**Additional context**
I simply would like to compile templates only for some selected directories, not for all modules and themes that are available in the filesystem. I believe leaving default command behavior as it is + adding optional argument will be the quickest way to achieve this.

weitzman:
See comment in PR

PR #5972, opened 2024-04-26
[#5971] twig compile - add optional searchpaths argument

drupalranger:
#5971
Add optional $searchpats argument to the twig compile command


weitzman:
We already only scan enabled modules/themes. I dont think we want more customization in the core command. Feel free to make own command or drush script.

PR #5976, opened 2024-05-01
Fix text prompts

weitzman:


DieterHolvoet (approved):
Didn't test, but changes look good!

src-symfony-compatibility/v6/Style/DrushStyle.php
```diff
@@ -50,17 +51,18 @@ public function choice(string $question, array $choices, mixed $default = null,
...
-        string $default = '',
         bool|string $required = false,
         ?\Closure $validate = null,
         string $hint = ''
-    ): void
-    {
-        // @todo return type of parent has changed
-        // return (new TextPrompt($label, $placeholder, $default, $required, $validate, $hint))->prompt();
+    ): mixed {
+        assert($validator === null, 'The $validator parameter is non-functional. Use $validate instead.');
```
YesCT, line 64: Why change from validator to validate? Is this cause of a dependency upstream change?? (Sorry for this comment on a 2 year ago change, we have some code we don't run often, and just noticed this change, and we are wondering what kind of testing we should do. Maybe no testing needed and only search and replace. Hm.)

PR #5978, opened 2024-05-01
Added overriding of output verbosity level to show a backtrace on errors

joachim-n:


weitzman:
Unless I am missing something, this has been available for a few years with the `-vvv` option. This is provided by Symfony Console and nothing Drush specific. No PR needed.

joachim-n:
The -vvv option gives me a LOT of noisy debug stuff about Drush that isn't relevant. I just want to see backtrace when something I am working on crashes.

weitzman:
Thats a niche need IMO. Feel free to make a personal or contrib commandfile which adds this option. See https://github.com/drush-ops/drush/blob/13.x/src/Commands/core/DrupliconCommands.php as an example

joachim-n:
I don't see how it's niche. Anytime you're developing something that runs over Drush, you run the risk of getting exceptions, and you want to see their backtrace. You don't want 41 lines of Drush debugging that's not relevant being output with every command.

This includes things like developing migrations, for instance, which I would not call niche.

I'm not sure how I could do that with a custom command, since the change needs to be in the Application class.

weitzman:
I expected you to disagree on niche. Thats why I kindly suggested a way forward. And of course you protested without even trying that way. I'll decide whats niche on my projects, and you decide whats niche on yours.

joachim-n:
I don't think it's about 'my projects' vs 'yours'. 

Are you saying you don't often crash Drush, or that most of the times you crash Drush, you don't care about seeing the backtrace?

PR #5986, opened 2024-05-05
Add a progress bar to the queue:run command

DieterHolvoet:
Just an idea for a small improvement I had: the queue:run command could use a progress bar, especially in case of large queues.

### DIFF
--- a/src/Commands/core/QueueCommands.php
+++ b/src/Commands/core/QueueCommands.php
@@ -74,7 +74,10 @@ final class QueueCommands extends DrushCommands
             $queue->garbageCollection();
         }
 
-        $this->io()->progressStart($items_limit ?: $queue->numberOfItems());
+        $max = $items_limit ?: $queue->numberOfItems();
+        if ($max > 0) {
+            $this->io()->progressStart($max);
+        }
 
         while ((!$time_limit || $remaining > 0) && (!$items_limit || $count < $items_limit) && ($item = $queue->claimItem($lease_time))) {
             try {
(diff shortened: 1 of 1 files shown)


weitzman:
If the queue worker logs messages, are they still shown? If so, I think this LGTM

DieterHolvoet:
I just tested and it looks like Drupal logs are never logged to the console: not with the progress bar, not without.

DieterHolvoet:
One thing we might want to change - not sure though - is to only start a progress bar once the first queue item is processed. What do you think?

CleanShot 2024-05-06 at 21 19 56@2x

weitzman:
Thanks. I'm also curious about any drush log messages that come from the worker. Maybe thats too uncommon to care about. 

The screenshot should also show the total number of items, right? Otherwise progress is a bit hard to understand. Once that is done, we could arguably get rid of the log message that reports progress.

DieterHolvoet:
> Thanks. I'm also curious about any drush log messages that come from the worker. Maybe thats too uncommon to care about.

Here's what that looks like:
CleanShot 2024-05-06 at 21 40 02@2x

Multiple logs per queue item:
CleanShot 2024-05-06 at 21 40 44@2x

> The screenshot should also show the total number of items, right?

It does, except when there's 0 items to be processed, which is why I proposed to not even start the progress bar at all in that case.

weitzman:
Thats super ugly IMO. Once a progress bar starts repeating lines it loses its value IMO. Happy to discuss more.

DieterHolvoet:
I agree, but I'm not sure if there's anything straightforward we can do about this. Also haven't tested this yet with Laravel Prompts https://laravel.com/docs/11.x/prompts#progress

DieterHolvoet:
On the other hand, progress bars are also used by the `entity:delete` and `entity:save` commands, the same thing could happen there as well. And what would be the alternative? How could we display both a progress bar and a list of messages? I just experimented a bit with console sections, but I can't seem to get it to work for our use case: https://symfony.com/blog/new-in-symfony-4-1-advanced-console-output

weitzman:
Right, there isnt. a great alternative. And thats why this command hasnt been made a progress bar for years. The other two commands you mention were converted by eager contributors, not me. I merged their PRs because others liked it. I actually dislike progress bars on CLI personally.

PR #5997, opened 2024-05-10
Directly add recipe command from Drupal 11

weitzman:
This direct addition of recipe command has pluses and minuses. 

The upside is that it will automatically gain powers as Drupal core builds out the command. The downside is users will incorrectly file issues in Drush for it, and the command help is weaker than an Annotated command.

A recipe path should be absolute or relative to Drupal root. Hopefully Drupal will update its command help to note that.

### DIFF
--- a/src/Runtime/ServiceManager.php
+++ b/src/Runtime/ServiceManager.php
@@ -12,6 +12,7 @@ use Consolidation\Filter\Hooks\FilterHooks;
 use Consolidation\SiteAlias\SiteAliasManagerAwareInterface;
 use Consolidation\SiteProcess\ProcessManagerAwareInterface;
 use Drupal\Component\DependencyInjection\ContainerInterface as DrupalContainer;
+use Drupal\Core\Recipe\RecipeCommand;
 use DrupalCodeGenerator\Command\BaseGenerator;
 use Drush\Attributes\Bootstrap;
 use Drush\Boot\DrupalBootLevels;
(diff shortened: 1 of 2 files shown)


gitressa:
Thanks! Does this add support for commands like this?

`drush site:install recipes/starshot --db-url=mysql://user:pass@dbhost/db`

Originally asked in https://github.com/phenaproxima/starshot-prototype/issues/94#issuecomment-2132224833.

weitzman:
No, thats not supported yet.

gitressa:
Thanks for a fast response, I created #6011.

src/Runtime/ServiceManager.php
```diff
@@ -12,6 +12,7 @@
 use Consolidation\SiteAlias\SiteAliasManagerAwareInterface;
 use Consolidation\SiteProcess\ProcessManagerAwareInterface;
 use Drupal\Component\DependencyInjection\ContainerInterface as DrupalContainer;
+use Drupal\Core\Recipe\RecipeCommand;
```
alex-moreno, line 15: would be good to add some readme or doc somewhere on how to run this, however simple it may seem

weitzman: the recipe command is documented like any other command - `drush list` to see its name and then `drush recipe --help`. It also has a docs page - https://www.drush.org/13.x/commands/recipe/. 

do you have anything else in mind?

alex-moreno: ah, got it, thanks Moshe

PR #6005, opened 2024-05-14
Remove extra parenthesis

emersonreisdev:
The parenthesis can be seen here https://www.drush.org/12.4.3/site-aliases/#altering-aliases

image


greg-1-anderson:
Thank you.

PR #6022, opened 2024-06-04
Restore SqlMariaDB override

LordAro:
#5918 removed the call to `SqlMysql::make`. Not sure why.

Refixes #5777.

Additionally uses `dumpProgram()` in an additional place.

I'm not quite sure of how things get backported/forward ported or whatever. But it seems that #5918 was only applied to 12.x, not 13.x. Shout if I need to do something different.


weitzman:
Thanks. If you have time to make a PR for 13.x that would be great.

LordAro:
Currently there's no need, as 13.x still contains the original condition ( #5918 was never forward-ported ? )

weitzman:
OK, but you said you replaced another instance of mysqldump? That needs to go into 13.x?

PR #6023, opened 2024-06-04
Improve dumpProgram handling in mysql database driver, and fix for null param to ucfirst for 13.x

LordAro:
* Use dumpProgram in an additional place in SqlMysql
* Fix null param to ucfirst() fatal error when DB information is missing

Forward ported from #6022 & #5918 (except fixed to not require #6022), as requested.


LordAro:
I'm pretty sure that CI failure is nothing to do with me

PR #5992, opened 2024-05-08
Fixes #5991: Adds Convert Symlinks Option 

rwagner00:
Resolves: https://github.com/drush-ops/drush/issues/5991

This PR adds a `convert-symlink` option to the `archive-dump` command which will convert symlinks to standard/static files and directories in order to resolve a bug where symlinks external to the project would cause an error on dump.

PR #6024, opened 2024-06-04
Fixes #5991: Adds Convert Symlinks Option #5992

### LINKED ISSUE #5991: Symlinks to directories outside project break archive dump.
**Describe the bug**
When doing `archive-dump` for code, if a symlink in the code leads to a directory outside the project code being archived, an error will be thrown.

**To Reproduce**
1. Create a symlink inside a project whose linked content is outside the project
2. Run `drush archive:dump --code`

**Expected behavior**
An archive will be generated successfully.

**Actual behavior**
```
In ArchiveDumpCommands.php line 213:

  [UnexpectedValueException]
  Iterator RecursiveIteratorIterator returned a path "{{symlink-destination}}" that is not in
  the base directory "{{tmp-directory}}"
```

**Workaround**
Remove the symlink

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.x/11.x/12.x 
| Drupal version? | 10.x/9.x/8.x/7.x
| PHP version     | 8.x/7.x
| OS?             | Mac/Linux

**Additional information**
This has only been verified in some configurations, but likely exists in most or all of them.

rwagner00:
Resolves: https://github.com/drush-ops/drush/issues/5991

This PR adds a convert-symlink option to the archive-dump command which will convert symlinks to standard/static files and directories in order to resolve a bug where symlinks external to the project would cause an error on dump.

### DIFF
--- a/src/Commands/core/ArchiveDumpCommands.php
+++ b/src/Commands/core/ArchiveDumpCommands.php
@@ -268,9 +268,9 @@ final class ArchiveDumpCommands extends DrushCommands
         foreach ($iterator as $file) {
             if (
                 $file->isLink() && ($convert_symlinks === true || strpos(
-                                                                      $file->getPathName(),
-                                                                      $archivePath
-                                                                  ) == 0)
+                    $file->getPathName(),
+                    $archivePath
+                ) == 0)
             ) {
                 $target = readlink($file->getPathname());
 
@@ -280,9 +280,9 @@ final class ArchiveDumpCommands extends DrushCommands
                     file_put_contents($file->getPathname(), $content);
                 } elseif (
                     is_dir($target) && ($convert_symlinks === true || strpos(
-                                                                          $file->getPathName(),
-                                                                          $archivePath
-                                                                      ) == 0)
+                        $file->getPathName(),
+                        $archivePath
+                    ) == 0)
                 ) {
                     $path = $file->getPathname();
                     unlink($path);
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Note that original review was done in #5992, but that PR was against the 12.x branch.

tests/functional/ArchiveTest.php
```diff
@@ -116,4 +116,30 @@ public function testArchiveDumpCommand(): void
...
         );
     }
+
+    public function testArchiveDumpSymlinkSwapCommand(): void
+    {
+      // Sites that contain symlinks to files outside the project root cause
+      // critical errors. To test for this, we manually create a symlink
+      // and then run archive:dump. If it completes at all, the symlink fix
+      // from Issue #5991 is working.
+        $linktarget      = Path::join($this->getSandbox(), '../symlinktest.txt');
```
greg-1-anderson, line 126: We should not write files outside of the sandbox.

tests/functional/ArchiveTest.php
```diff
@@ -116,4 +116,30 @@ public function testArchiveDumpCommand(): void
...
+        $linkdestination = Path::join($this->getSandbox(), 'symlinkdest.txt');
```
greg-1-anderson, line 127: The symlink must be written inside the SUT in order for it to affect the test. Maybe run the test once without the fix to prove that the test fails. Like this, it should still pass. Use $this->webroot() or $this->getComposerRoot() to find paths inside the SUT.

src/Commands/core/ArchiveDumpCommands.php
```diff
@@ -237,6 +242,74 @@ private function createManifestFile(array $options): void
...
+     *  Whether to convert all symlinks.
+     * @param string $archivePath
+     *   The file path of the archive.
+     *
+     * @return void
+     */
+    public function convertSymlinks(
+        $convert_symlinks,
+        $archivePath
+    ): void {
```
greg-1-anderson, line 259: See https://github.com/drush-ops/drush/pull/5992/files#r1619382407, not resolved yet.

rwagner00: Parameter types added

greg-1-anderson:
Continued in #6030

PR #6026, opened 2024-06-07
Fixes #5999: Allow drush site:install to work with recipes.

### LINKED ISSUE #5999: Allow Drush site-install to install from recipes
Drupal can now be installed using recipes, which bypasses the need for install profiles.

Users are currently using the `drupal quick-start` and `drupal install` scripts to install from recipes, but that only allows for an sql-lite database.  A must more real-world example will be to allow this from Drush, like most folks do to install Drupal now.

Add an option like `drush site-install --recipe /path/to/recipe` which would follow the steps in the Drupal installer and install Drupal with no install profile, and apply the recipel

Allow for selecting no install profile.  We could then use the proposed `drush recipe` to apply recipes.

**Additional context**
N/A

greg-1-anderson:
POC was pretty straightforward. Lightly tested. Seems to work with core standard recipe. Have not tried with Starshot. Additional work may be needed. Needs tests.

### DIFF
--- a/src/Commands/core/SiteInstallCommands.php
+++ b/src/Commands/core/SiteInstallCommands.php
@@ -50,7 +50,7 @@ final class SiteInstallCommands extends DrushCommands
      * Install Drupal along with modules/themes/configuration/profile.
      */
     #[CLI\Command(name: self::INSTALL, aliases: ['si', 'sin', 'site-install'])]
-    #[CLI\Argument(name: 'profile', description: 'An install profile name, or a path to a directory containing a recipe. Defaults to <info>standard</info> unless an install profile is marked as a distribution. Use <info>minimal</info> for a bare minimum installation. Additional info for the install profile may also be provided with additional arguments. The key is in the form <info>[form name].[parameter name]</info>')]
+    #[CLI\Argument(name: 'recipeOrProfile', description: 'An install profile name, or a path to a directory containing a recipe. Defaults to <info>standard</info> unless an install profile is marked as a distribution. Use <info>minimal</info> for a bare minimum installation. Additional info for the install profile may also be provided with additional arguments. The key is in the form <info>[form name].[parameter name]</info>')]
     #[CLI\Option(name: 'db-url', description: 'A Drupal 10 style database URL. Required for initial install, not re-install. If omitted and required, Drush prompts for this item.')]
     #[CLI\Option(name: 'db-prefix', description: 'An optional table prefix to use for initial install.')]
     #[CLI\Option(name: 'db-su', description: 'Account to use when creating a new database. Must have Grant permission (mysql only). Optional.')]
@@ -73,9 +73,9 @@ final class SiteInstallCommands extends DrushCommands
     #[CLI\Usage(name: 'drush si core/recipes/standard', description: 'Install from the core Standard recipe.')]
     #[CLI\Bootstrap(level: DrupalBootLevels::ROOT)]
     #[CLI\Kernel(name: Kernels::INSTALLER)]
-    public function install(array $profile, $options = ['db-url' => self::REQ, 'db-prefix' => self::REQ, 'db-su' => self::REQ, 'db-su-pw' => self::REQ, 'account-name' => 'admin', 'account-mail' => 'admin@example.com', 'site-mail' => 'admin@example.com', 'account-pass' => self::REQ, 'locale' => 'en', 'site-name' => 'Drush Site-Install', 'site-pass' => self::REQ, 'sites-subdir' => self::REQ, 'config-dir' => self::REQ, 'existing-config' => false]): void
+    public function install(array $recipeOrProfile, $options = ['db-url' => self::REQ, 'db-prefix' => self::REQ, 'db-su' => self::REQ, 'db-su-pw' => self::REQ, 'account-name' => 'admin', 'account-mail' => 'admin@example.com', 'site-mail' => 'admin@example.com', 'account-pass' => self::REQ, 'locale' => 'en', 'site-name' => 'Drush Site-Install', 'site-pass' => self::REQ, 'sites-subdir' => self::REQ, 'config-dir' => self::REQ, 'existing-config' => false]): void
     {
-        $additional = $profile;
+        $additional = $recipeOrProfile;
         $recipeOrProfile = array_shift($additional) ?: '';
         $form_options = [];
         foreach ($additional as $arg) {
@@ -132,6 +132,9 @@ final class SiteInstallCommands extends DrushCommands
         ];
 
         if ($recipe) {
+            if (version_compare(\Drupal::VERSION, '10.3.0') < 0) {
+                throw new \Exception('Recipes are only supported on Drupal 10.3.0 and later.');
+            }
             $settings['parameters']['recipe'] = $recipe;
         }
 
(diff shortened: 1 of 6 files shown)


greg-1-anderson:
I did a little more ad-hoc testing, and also provided a functional test. Seems to work as one might expect.

greg-1-anderson:
Right now the code reverts to the profile-checking code if the provided path isn't a recipe. This leads to an odd error if the path is wrong, for example. Since profile names cannot contain /, perhaps we shoudl give a better error message when a path is used. Testing for Drupal >= 10.3.0 would also be a good idea.

rfay:
But `ddev drush si -y demo_umami --account-pass=admin` is hard-wired into my fingers. How am I going to learn something new???

src/Commands/core/SiteInstallCommands.php
```diff
@@ -50,7 +50,7 @@ public function __construct(
      * Install Drupal along with modules/themes/configuration/profile.
      */
     #[CLI\Command(name: self::INSTALL, aliases: ['si', 'sin', 'site-install'])]
-    #[CLI\Argument(name: 'profile', description: 'An install profile name. Defaults to <info>standard</info> unless an install profile is marked as a distribution. Use <info>minimal</info> for a bare minimum installation. Additional info for the install profile may also be provided with additional arguments. The key is in the form <info>[form name].[parameter name]</info>')]
+    #[CLI\Argument(name: 'profile', description: 'An install profile name, or a path to a directory containing a recipe. Defaults to <info>standard</info> unless an install profile is marked as a distribution. Use <info>minimal</info> for a bare minimum installation. Additional info for the install profile may also be provided with additional arguments. The key is in the form <info>[form name].[parameter name]</info>')]
```
weitzman: Lets rename this arg to recipeOrProfile, if we are keeping this multi-purpose argument.

weitzman (commented):
This looks great .

1.  I'm a little hesitant to re-use the argument for both profile and recipe. An alternative would be to add a new --recipe option. If we do that we would want a -r short option because those accept spaces before the value and thus are easier to tab complete with a path. This approach is more compatible with bash completion as well.
2. Does this PR work as expected these re-install scenarios? (Update: MW: recipes have no persistence at all so I guess these work fine)
    1. User does not specify a profileOrRecipe
    2. User specifies a recipe that is different from the installed site
 3. Once recipes support user input (see https://www.drupal.org/project/distributions_recipes/issues/3304895), are we well positioned to support it? We already accept input via trailing optional key=value arguments. I guess approach works for recipes as well.
 4. We should add that improved error message when the recipe (or profile) is not found

src/Commands/core/SiteInstallCommands.php
greg-1-anderson: I can go with either decision on whether profile and recipe are combined or separate, but I would lean towards keeping them combined for uniformity with the "drupal install" command in Drupal core. Agree about the rename.

greg-1-anderson:
I'm not sure why the failing code was running in [fdd05b0](https://github.com/drush-ops/drush/pull/6026/commits/fdd05b0a494ce731e681cb214bda2ad4ad0f2d8e), seems like a pre-existing edge case. Just marked the test skipped in https://github.com/drush-ops/drush/pull/6026/commits/cce1a826ee33ae49b91150aacc3441b14dc4f7a9.

I'd suggest that we merge this to 13.x-dev to get a little broader testing, and handle input for recipes as follow-on work, once it's supported in Drupal. LMK if there are any further changes you'd like to see here before we merge.

src/Commands/core/SiteInstallCommands.php
```diff
@@ -50,7 +50,7 @@ public function __construct(
      * Install Drupal along with modules/themes/configuration/profile.
      */
     #[CLI\Command(name: self::INSTALL, aliases: ['si', 'sin', 'site-install'])]
-    #[CLI\Argument(name: 'profile', description: 'An install profile name. Defaults to <info>standard</info> unless an install profile is marked as a distribution. Use <info>minimal</info> for a bare minimum installation. Additional info for the install profile may also be provided with additional arguments. The key is in the form <info>[form name].[parameter name]</info>')]
+    #[CLI\Argument(name: 'recipeOrProfile', description: 'An install profile name, or a path to a directory containing a recipe. Defaults to <info>standard</info> unless an install profile is marked as a distribution. Use <info>minimal</info> for a bare minimum installation. Additional info for the install profile may also be provided with additional arguments. The key is in the form <info>[form name].[parameter name]</info>')]
```
weitzman:
```suggestion
    #[CLI\Argument(name: 'recipeOrProfile', description: 'An install profile name, or a path to a directory containing a recipe. Defaults to <info>standard</info> unless an install profile is marked as a distribution. Use <info>minimal</info> for a bare minimum installation. Additional info for the install profile may also be provided with additional arguments. Use the format <info>[form name].[parameter name]</info>')]
```

src/Commands/core/SiteInstallCommands.php
```diff
@@ -70,12 +70,13 @@ public function __construct(
     #[CLI\Usage(name: 'drush si --account-pass=mom', description: 'Re-install with specified uid1 password.')]
     #[CLI\Usage(name: 'drush si --existing-config', description: 'Install based on the yml files stored in the config export/import directory.')]
     #[CLI\Usage(name: 'drush si standard install_configure_form.enable_update_status_emails=NULL', description: 'Disable email notification during install and later. If your server has no mail transfer agent, this gets rid of an error during install.')]
+    #[CLI\Usage(name: 'drush si core/recipes/standard', description: 'Install from the core Standard recipe.')]
```
weitzman, line 73: The path is relative to the CWD and not Drupal root, right? Lets state this like we do for other path related arguments.

tests/functional/SiteInstallTest.php
```diff
@@ -35,4 +40,70 @@ public function testSiteInstallPrefix()
...
+        $error_output = $this->getErrorOutput();
+        $this->assertStringContainsString('Recipes are only supported on Drupal 10.3.0 and later.', $error_output);
+    }
+
+    /**
+     * Test functionality of installing a site with a recipe.
+     */
+    public function testSiteInstallRecipe()
+    {
+        if (version_compare(\Drupal::VERSION, '10.3.0') < 0) {
```
weitzman: Lets use `$this->isDrupalGreaterThanOrEqualTo()`.

weitzman (approved):
Approved with minor feedback

src/Commands/core/SiteInstallCommands.php
greg-1-anderson: The path is currently relative to the Drupal root. Relative to CWD is more consistent to other Drush commands, though. I'll adjust the code and comment.

greg-1-anderson: Decided to search both cwd and Drupal root. Core recipes are in core/recipes, so Drupal-root-relative recipe paths are most convenient for installing those. If a module provided a recipe, then a Drupal-root-relative path would also be helpful. The Starshot prototype ships recipes in the "recipes" directory, located at the project root, so cwd-relative recipe paths are useful for anything that follows that convention.

We could consider refining this further in the future, e.g. maybe automatically searching DRUPAL_ROOT/core/recipes, or CWD/recipes when a simple name is provided. Note, however, that "standard" exists as both a recipe and a profile. The result of installing either of these looks about the same; I am not sure if there might be differences, or if shifting the default from the standard profile to the standard recipe would have any noticeable impact.

greg-1-anderson: If we did the above, and someone converted demo_umami to a recipe, then @rfay wouldn't need to learn to type a different `drush si` command. This was an unintended side benefit of the thought, not the primary motivator.

greg-1-anderson: @rfay The above was just a humorous reference to https://github.com/drush-ops/drush/pull/6026#issuecomment-2158343029; that exact command would work as written to install from a recipe, if demo_umami were reimplemented as a recipe. You might have to type "starshot" instead of "demo_umami", though, as things evolve. Who knows, someone might build "starshot_umami".

Issue #6021, opened 2024-06-03
drush cron is sometimes missing info log output "Cron run completed"


klausi:
**Describe the bug**
We run Drupal cron with drush to catch errors. When the sentence "Cron run completed" is no in the output then we know it is an error
```sh
#!/bin/bash

# Only print drush output when it does not contain "Cron run completed."
# Usage: drush-cron9.sh /var/www/jobiqo-decoupled/backend

cd "$1" || exit
DRUSH_OUTPUT=$(/usr/bin/drush cron -v 2>&1)
RETURN_CODE=$?
# Check if the success message has a translation.
# Use it as regex pattern for grep.
SUCCESS_MESSAGE=$(/usr/bin/drush ev 'print t("Cron run successful.") . "|Cron run completed.";')
# If the success message is not in the output then we print all the output.
# shellcheck disable=SC2086
if echo $DRUSH_OUTPUT | grep -vqE "$SUCCESS_MESSAGE"; then
  echo "$DRUSH_OUTPUT"
fi
exit $RETURN_CODE
```

This worked fine in Drush 11, but since the upgrade to Drush 12.5.2 the cron command does not always print all info log messages. Somehow Drush or the Drupal site can get into a state where all info log messages are not printed in Drush output. A `drush cr` fixes it and then the output is correct again with all info log messages.

A broken `drush cron -v` output look like this, note the missing info messages at the end:
```
 [info] Drush bootstrap phase: bootstrapDrupalRoot()
 [info] Change working directory to /var/www/example/site/backend/web
 [info] Initialized Drupal 10.1.8 root directory at /var/www/example/site/backend/web
 [info] Drush bootstrap phase: bootstrapDrupalSite()
 [info] Initialized Drupal site example.com at sites/default
 [info] Drush bootstrap phase: bootstrapDrupalConfiguration()
 [info] Drush bootstrap phase: bootstrapDrupalDatabase()
 [info] Successfully connected to the Drupal database.
 [info] Drush bootstrap phase: bootstrapDrupalFull()
 [info] Starting bootstrap to none
 [info] Drush bootstrap phase 0
 [info] Try to validate bootstrap phase 0
```

Correct would be:
```
 [info] Drush bootstrap phase: bootstrapDrupalRoot()
 [info] Change working directory to /var/www/example/site/backend/web
 [info] Initialized Drupal 10.1.8 root directory at /var/www/example/site/backend/web
 [info] Drush bootstrap phase: bootstrapDrupalSite()
[...]

PR #6028, opened 2024-06-10
Remove unused drupalLoggerAdapter code (12.x)

klausi:
While trying to track down #6021 I found some dead code in DrupalBoot8.

`drupalLoggerAdapter` is not used anywhere and there is a typo in the type hint `LoggrInterface`. Was introduced in #5481 by mistake.

Phpstan would have caught this - could make sense to set it up?


weitzman:
phpstan is running. see the current branch - 13.x. 12.x is in maint mode so not sure about this PR.

klausi:
Opened #6029 to do the same for 13.x.

PR #6040, opened 2024-06-17
Command to regenerate htaccess

labya:


weitzman:
Thanks. I hope someone will maintain this in Contrib. Or this could be added as an example for php:eval command. As a full  command this is too niche for core Drush IMO. A leaner command list helps keep Drush approachable and maintainable.

prudloff-insite:
I created a module that does this: https://www.drupal.org/project/drush_create_htaccess

PR #6046, opened 2024-06-21
Fixes #5999: Allow drush site:install to work with recipes. (#6026)

### LINKED ISSUE #5999: Allow Drush site-install to install from recipes

greg-1-anderson:
Backport of #6026


weitzman:
I'm not inclined to backport this feature as long as Drush 13 supports Drupal 10.

PR #6047, opened 2024-06-21
Fix path to files directory when it is a symlink

greg-1-anderson:
archive:dump does not do the right thing when sites/default/files is a symlink to a different directory.


greg-1-anderson:
Did ad-hoc testing, and this fixes archive:dump on Pantheon sites. In order to write a test for this, we'd need a SUT that was configured like Pantheon sites. I don't think I'd want the main Drush SUT to always have these additions. The alternative would be to add a secondary site fixture, and have some Unish code that copied it to the sandbox, and added Drush as a path repository. I don't know if there might in the future want to have tests that test other Drupal site configurations.

I can write that test, but I'm not sure if you want the complexity. Probably wouldn't be too many lines.

weitzman (approved):
IMO we can skip that test

PR #6057, opened 2024-07-09
Sanitize user names on sql-sanitize.

### LINKED ISSUE #4609: Sanitize user names on sql-sanitize
Currently it's possible for users to use their real names as a username and it isn't sanitized by `sql-sanitize`.

We could set it to a pattern in a similar way to how email addresses are currently sanitized.

I can provide my own post-command sql-sanitize to do this, I'm just wondering if there'd be interest in moving it into drush proper?

primsi:
Fixes #4609

### DIFF
--- a/src/Commands/sql/sanitize/SanitizeUserTableCommands.php
+++ b/src/Commands/sql/sanitize/SanitizeUserTableCommands.php
@@ -165,7 +165,8 @@ final class SanitizeUserTableCommands extends DrushCommands implements SanitizeP
      *   - the table name;
      *   - the column name.
      */
-    protected function getFieldTableDetails(string $entity_type_id, string $field_name): array {
+    protected function getFieldTableDetails(string $entity_type_id, string $field_name): array
+    {
         $storage = $this->entityTypeManager->getStorage($entity_type_id);
         if (!$storage instanceof SqlEntityStorageInterface) {
             $context = ['!entity_type_id' => $entity_type_id];
--- a/src/Commands/sql/sanitize/SanitizeUserTableCommands.php
+++ b/src/Commands/sql/sanitize/SanitizeUserTableCommands.php
@@ -8,7 +8,10 @@ use Drupal\Core\Database\Connection;
 use Drupal\Core\Database\Query\SelectInterface;
 use Consolidation\AnnotatedCommand\CommandData;
 use Consolidation\AnnotatedCommand\Hooks\HookManager;
+use Drupal\Core\Database\Query\Update;
+use Drupal\Core\Entity\EntityFieldManagerInterface;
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks! Ideally we add test coverage for this. We can probably add on tgo an existing test.

src/Commands/sql/sanitize/SanitizeUserTableCommands.php
```diff
@@ -80,6 +84,20 @@ public function sanitize($result, CommandData $commandData): void
...
 
+        // Sanitize username.
+        if ($this->isEnabled($options['sanitize-username'])) {
+            [$name_table, $name_column] = $this->getFieldTableDetails('user', 'name');
+            [$uid_table, $uid_column] = $this->getFieldTableDetails('user', 'uid');
+            assert($uid_table === $name_table);
+
+            // Updates usernames to the pattern user_%uid.
+            $query
+                ->condition($uid_column, 0, '>')
```
andriokha, line 95: That condition's already added further up.
```suggestion
```

src/Commands/sql/sanitize/SanitizeUserTableCommands.php
```diff
@@ -80,6 +84,20 @@ public function sanitize($result, CommandData $commandData): void
...
+                ->expression($name_column, "CONCAT('user_', $uid_column)");
```
andriokha, line 96: Looking at the email code, it appears we'd need a different format for different DB drivers I think?

src/Commands/sql/sanitize/SanitizeUserTableCommands.php
```diff
@@ -80,6 +84,20 @@ public function sanitize($result, CommandData $commandData): void
             $messages[] = dt('User emails sanitized.');
         }
 
+        // Sanitize username.
+        if ($this->isEnabled($options['sanitize-username'])) {
+            [$name_table, $name_column] = $this->getFieldTableDetails('user', 'name');
+            [$uid_table, $uid_column] = $this->getFieldTableDetails('user', 'uid');
+            assert($uid_table === $name_table);
```
andriokha, line 91: FWIW to my mind this clashes a little with the existing style where the table and column names have been hardcoded. Not sure if that means it's better to stick with the current style or update the others personally (:

If we _do_ keep it, I think we need to catch any exceptions, we don't want to explode the sanitize command as a whole.

andriokha (commented):
Thanks, hope it's ok I put down some thoughts.

MakerTim:
At SWIS we use [this patch](https://bitbucket.org/!api/2.0/snippets/swisnl/eq6544/8165ad309e565699e0b1932c50f81d347ce5047f/files/6057.patch) for 100% of our projects since `2024-10-07` without any issues

PR #6061, opened 2024-07-14
Autowire Drupal container params and plain values

claudiu-cristea:
Autowiring works now for services but not for Drupal container parameters or even plain values.

It should be possible to autowire Drupal container params and plain values in his way:

```php
public function __construct(
    #[Autowire(param: 'container.modules')]
    protected array $installedModules,
    #[Autowire(value: 'a plain string')]
    protected string $string,
)
{
    parent:: __construct();
}

### DIFF
--- a/src/Commands/AutowireTrait.php
+++ b/src/Commands/AutowireTrait.php
@@ -2,6 +2,9 @@
 
 namespace Drush\Commands;
 
+use Drupal\Component\DependencyInjection\ContainerInterface as DrupalContainer;
+use Drush\Drush;
+use League\Container\Container as DrushContainer;
 use Psr\Container\ContainerInterface;
 use Symfony\Component\DependencyInjection\Attribute\Autowire;
 use Symfony\Component\DependencyInjection\Exception\AutowiringFailedException;
@@ -16,7 +19,18 @@ use Symfony\Component\DependencyInjection\Exception\AutowiringFailedException;
  */
 trait AutowireTrait
 {
-  /**
+    /**
+     * Limit to service and param or plain value.
+     *
+     * @see \Symfony\Component\DependencyInjection\Attribute\Autowire::__construct
+     */
+    private const ACCEPTED_AUTOWIRE_ARGUMENTS = [
+        0 => 'value',
+        1 => 'service',
+        4 => 'param',
+    ];
+
+    /**
      * Instantiates a new instance of the implementing class using autowiring.
      *
      * @param ContainerInterface $container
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/Commands/AutowireTrait.php
Comment (weitzman) [inline src/Commands/AutowireTrait.php:1]: This file is based on https://github.com/drupal/drupal/blob/11.x/core/lib/Drupal/Core/DependencyInjection/AutowireTrait.php. Looks like this PR wants to expand its duties considerably. I'm not opposed to that, but I fear that this may get tricky to maintain as Symfony evolves.

File: src/Runtime/ServiceManager.php
Comment (weitzman) [inline src/Runtime/ServiceManager.php:1]: I'm hoping that @greg-1-anderson can take a look at changes in this file.

**Review by weitzman — COMMENTED** [review]
Thanks for the tests

**Comment by claudiu-cristea** [comment]
Last commit deserves better test coverage

**Comment by weitzman** [comment]
League container doesn't even support container parameters and string so AutowireTrait how about we cheat and go right to the delegated container for those. If you have a compound container, you can do $container->getContainer() to get the Drupal container, and then ->getParameter() to get the param.

**Comment by claudiu-cristea** [comment]
I think I have something but no idea how add a new League container service in tests, see AutowireArgumentsTest::testWithDrushContainer()

File: tests/fixtures/lib/Drush/Commands/AutowireTestCommands.php
Comment (weitzman) [inline tests/fixtures/lib/Drush/Commands/AutowireTestCommands.php:1]:
```
Could not instantiate Custom\Library\Drush\Commands\AutowireTestCommands: Cannot autowire service "woot.autowire_test": argument "$argListContainerService" of method "Custom\Library\Drush\Commands\AutowireTestCommands::_construct()", you should configure its value explicitly. [1.06 sec, 8.13 MB]
```
I am seeing an error when I run `drush st -vvv`. I'm not sure why this commandfile loaded when I just `ddev start`, and am not running tests.

File: src/Runtime/ServiceManager.php [hunk 1]
```diff
@@ -372,22 +372,60 @@ public function instantiateServices(array $bootstrapCommandClasses, DrushContain
...
+        $type = ltrim((string) $params[0]->getType(), '?');
+        if ($drupalContainer && is_a($type, DrupalContainer::class, true)) {
+            // The factory create() method explicitly expects a Drupal container as 1st argument.
+            $args[] = $drupalContainer;
+        } elseif (is_a($type, DrushContainer::class, true)) {
+            // The factory create() method explicitly expects a Drush container as 1st argument. Don't add a 2nd
+            // argument. If the Drupal container has been initialized, it's already a delegate.
+            return [$drushContainer];
+        } elseif ($drupalContainer && is_a($type, ContainerInterface::class, true)) {
+            // The factory create() method expects a container of any type as st argument and the Drupal container has
```
weitzman, line 398: what do we mean "has already been initialized"?

src/Runtime/ServiceManager.php
```diff
@@ -372,22 +372,60 @@ public function instantiateServices(array $bootstrapCommandClasses, DrushContain
...
+            $args[] = $drupalContainer;
+        }
+
+        // Add Drush container as 2nd argument if the method expects one.
+        if (isset($params[1])) {
+            $type = ltrim((string) $params[1]->getType(), '?');
+            if (is_a($type, ContainerInterface::class, true)) {
+                $args[] = $drushContainer;
+            }
+        }
```
weitzman, line 409: I dont like commands and AutowireTrait having multiple numbers of params and type hints in create(). How about if we ask commands going forward to always type hint to `Psr\Container\ContainerInterface`. That receives the compound container. AutowireTrait could pull the drupalContainer out and get container params as needed. We need to declare a service for the drupalContainer in the league container. [We already store a reference to the league container inside itself](https://github.com/drush-ops/drush/blob/5b1d3d8a7c518d2882d604bfcd14ee3902fc568f/src/Runtime/DependencyInjection.php#L69). So this would be a similar service.

weitzman:
Some Drupal core work in this area - https://www.drupal.org/project/drupal/issues/3558292

claudiu-cristea:
> Some Drupal core work in this area - https://www.drupal.org/project/drupal/issues/3558292

I know, but Drush is more complex because of 2 containers we try to accommodate in one shot

Issue #6067, opened 2024-07-19
Add entity:count command to display count of specific entity


tibezh:
**Is your feature request related to a problem? Please describe.**
In a specific case I need to know count of specific entities in Drupal website.

**Describe the solution you'd like**
I think it's nice to have feature, just to know count of entities by specific type, support filter by bundles.

**Describe alternatives you've considered**
We can use a `drush php:eval` with a PHP code, but it is not flexible.

tibezh:
PR with `entity:count` command was added.
Could somebody check, please?

tibezh:
PR is closed:

> Not enough interest for core drush.

=(

PR #6068, opened 2024-07-19
Add entity:count command (#6067).

tibezh:
Covers https://github.com/drush-ops/drush/issues/6067.


weitzman:
Not enough interest for core drush.

PR #6071, opened 2024-07-23
Fix PHPStan by using correct type

alexpott:


alexpott:
This was broken in #5471

alexpott:
Might need to be backported to 12.x too

PR #6076, opened 2024-07-31
Warning about use of hook_update_N()

damienmckenna:
Adds a note to make it more clear that using hook_update_N() can lead to problems.


weitzman:
Can we link to Drupal's docs instead. Its not really Drush's place to explain this. deploy/updatedb is just a runner.

weitzman:
Lets reopen if a link to drupal docs arises.

PR #6080, opened 2024-08-02
Fix return type of Drush\Commands\DrushCommands::logger()

DieterHolvoet:
Since Drush 13, I get the following error when running any Drush command:
```
AssertionError: assert(is_null($this->logger) || $this->logger instanceof DrushLoggerManager) in assert() (line 76 of /Users/dieterholvoet/Projects/sites-personal/whathappens/vendor/drush/drush/src/Commands/DrushCommands.php)
```

It's caused by `Drupal\ultimate_cron\Commands\UltimateCronCommands` of the Ultimate Cron module assigning a Drupal logger to `$this->logger`. Is there any reason the return type is `Drush\Log\DrushLoggerManager` and not `Psr\Log\LoggerInterface`?

### DIFF
--- a/src/Commands/DrushCommands.php
+++ b/src/Commands/DrushCommands.php
@@ -13,7 +13,6 @@ use Drush\Attributes as CLI;
 use Drush\Config\ConfigAwareTrait;
 use Drush\Drush;
 use Drush\Exec\ExecTrait;
-use Drush\Log\DrushLoggerManager;
 use Drush\SiteAlias\ProcessManager;
 use Drush\Style\DrushStyle;
 use GuzzleHttp\HandlerStack;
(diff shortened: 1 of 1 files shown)


weitzman:
Yes, there is. Drush's log filtering via -v and -vvv options wont work with the UltimateCron logger I dont know why Ultimate Cron is doing that replacement. I don't see why it is needed.

I added some more docs about the right way to do it, just in case it is needed. See #6083

DieterHolvoet:
Adding this in the constructor doesn't work:
```php
$this->logger()->add('drupal', $logger->get('ultimate_cron'));
```
since `$this->logger()` returns NULL at that point. Is this expected? Should I change it to this?
```php
\Drush\Drush::logger()->add('drupal', $logger->get('ultimate_cron'));
```

weitzman:
That seems reasonable to me.

PR #6082, opened 2024-08-02
Fix wrong keyvalue store being used for the deploy hook registry

### LINKED ISSUE #6081: Deploy hook registry is reset after update to Drush 13
**Describe the bug**
After updating from Drush 12 to Drush 13 and running `drush deploy`, the command attempted to execute all existing deploy hooks, hooks that have already been executed a long time ago.

**To Reproduce**
Update from Drush 12 to Drush 13 and run `drush deploy`.

**Expected behavior**
Don't re-run old deploy hooks.

**Actual behavior**
Old deploy hooks were re-run.

**Workaround**
No.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 13.0.0
| Drupal version? | 10.3.1
| PHP version     | 8.3.9
| OS?             | Mac

**Additional information**
/

DieterHolvoet:
Fixes #6081.

### DIFF
--- a/src/Commands/core/DeployHookCommands.php
+++ b/src/Commands/core/DeployHookCommands.php
@@ -59,7 +59,7 @@ final class DeployHookCommands extends DrushCommands
                 $this->root = $root;
                 $this->sitePath = $site_path;
                 $this->enabledExtensions = array_merge(array_keys($module_list), array_keys($theme_handler->listInfo()));
-                $this->keyValue = $key_value_factory->get('post_update');
+                $this->keyValue = $key_value_factory->get('deploy_hook');
                 $this->updateType = 'deploy';
             }
         };


weitzman:
LGTM, but what are the ramifications if a D10 site upgrades its Drush to 13, while staying on D10? Past deploy hooks will run?

DieterHolvoet:
Yes, that's what happened to me.

weitzman:
Sorry I was unclear. If this or goes in, it could negatively affect a site that is only upgrading their drush and not their Drupal.

DieterHolvoet:
No that’s what I meant, I only updated Drush, not Drupal. This is obviously a mistake, no? It used to be deploy_hook, then it was mistakingly(?) changed to post_update, which is a different kind of update hook.

weitzman:
Yes it's a mistake and you have the right fix. I'm just wondering if we should copy data into the new registry to avoid problems. Or something.

DuaelFr:
I can confirm the issue (introduced in https://github.com/drush-ops/drush/commit/f9186d6062aa33a77b5a217178b9c2c97eb16665) and the fix.
I believe it's a critical one as it can lead to data loss given the wide usage of deploy hooks.

The only problem I can see here is that this fix would land in a 13.0.x version and might cause issues to people that ran their deploy hooks with 13.0.0 for the first time. I believe we need an upgrade path for this specific use case.

bircher:
Yes I think this should be changed as soon as possible.
It will not affect people upgrading drush from 12 to 13+fix, only for people going from 13-fix to 13+fix, but that is for sure a much smaller group.

bircher:
I wrote more on slack, I will post it here too so that someone might find it.

We merge this and fix drush 13 and release a new version with it.
So drush 12 is ok, then some versions of drush 13 are broken, and then later versions of drush 13 are ok again.
* If you update from drush 12 to the fixed drush 13, all is good.
* if you update from drush 12 to a broken drush 13 then you risk running old hooks again, you notice and you don't run the deploy hooks and you update to the fixed drush 13 and you are fine.
* If you started with a a broken drush 13 or updated and didn't notice and you have run drush deploy with the broken drush 13 you can fix this in the following way:
update to the fixed drush 13 without adding any new deploy hook in that deployment, then run:
 `drush deploy:mark-complete`. 
Now the correct key value store knows about the hooks that were run with the wrong key value store. And you are good too.

PR #6092, opened 2024-08-16
Add #[CLI\OptionsetSql] to sanitize()

### LINKED ISSUE #6091: Allow  sql:sanitize on any database
Currently, there seems to be no possibility to sanitize a database that isn't defined as the default database.

Add the `--database` option to make it possible to select any database that is defined in `settings.php`.

Boegie:
Closes #6091


Boegie:
I highly expect this to need tests and/or fail the current test suites, but locally it at least the `--database` option works, showing the conceptual "niceness" of the current code base.

weitzman:
Not planned, but folks should feel free to patch their drush to use it.

PR #6107, opened 2024-09-09
entity:save supports a state transition

weitzman:


src/Commands/core/EntityCommands.php
```diff
@@ -101,28 +109,38 @@ public function doDelete(string $entity_type, array $ids): void
     #[CLI\Option(name: 'bundle', description: 'Restrict to the specified bundle. Ignored when ids is specified.')]
     #[CLI\Option(name: 'exclude', description: 'Exclude certain entities. Ignored when ids is specified.')]
     #[CLI\Option(name: 'chunks', description: 'Define how many entities will be loaded in the same step.')]
-    #[CLI\Option(name: 'publish', description: 'Publish entities as they are saved.')]
+    #[CLI\Option(name: 'publish', description: 'Publish entities as they are saved. ')]
```
opi, line 112: Why this extra space ? [/nitpick]

weitzman: bug.

PR #6119, opened 2024-09-25
Disable xdebug by default to improve performance

webflo:
Xdebug maybe enabled in the dev environment, but most of the time its not the goal to debug drush commands. 

Speeds up drush cr, cex  tremendously.

This feature was in the old Drush Launcher. The implementationn in this PR is simpler, due to Xdebug 3.

PHPStan and Composer also disable by default.

### DIFF
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -255,7 +255,7 @@ class Preflight
         // Now that we know the value, set debug flag.
         $this->logger()->setDebug($this->preflightArgs->get(PreflightArgs::DEBUG, false));
 
-        // Give hint if a developer might be trying to debug Drush
+        // Give hint if a developer might be trying to debug Drush.
         if (extension_loaded('xdebug')) {
             $this->logger()->log(strtr('If you are trying to perform step debugging of Drush, see !url', ['!url' => 'https://www.drush.org/latest/usage/#xdebug']));
         }
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -256,7 +256,7 @@ class Preflight
         $this->logger()->setDebug($this->preflightArgs->get(PreflightArgs::DEBUG, false));
 
         // Give hint if a developer might be trying to debug Drush
-        if (extension_loaded('xdebug') || empty(ini_get('xdebug.mode'))) {
+        if (extension_loaded('xdebug')) {
             $this->logger()->log(strtr('If you are trying to perform step debugging of Drush, see !url', ['!url' => 'https://www.drush.org/latest/usage/#xdebug']));
         }
 
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -257,7 +257,7 @@ class Preflight
 
         // Give hint if a developer might be trying to debug Drush.
         if (extension_loaded('xdebug')) {
-            $this->logger()->log(strtr('If you are trying to perform step debugging of Drush, see !url', ['!url' => 'https://www.drush.org/latest/usage/#xdebug']));
+            $this->logger()->log(strtr('Drush disables Xdebug by default. To override this, see !url', ['!url' => 'https://www.drush.org/latest/commands/#xdebug']));
         }
 
         // Do legacy initialization (load static includes, define old constants, etc.)
--- a/src/Preflight/Preflight.php
+++ b/src/Preflight/Preflight.php
@@ -255,6 +255,11 @@ class Preflight
         // Now that we know the value, set debug flag.
         $this->logger()->setDebug($this->preflightArgs->get(PreflightArgs::DEBUG, false));
 
+        // Give hint if a developer might be trying to debug Drush
+        if (extension_loaded('xdebug') || empty(ini_get('xdebug.mode'))) {
+            $this->logger()->log(strtr('If you are trying to perform step debugging of Drush, see !url', ['!url' => 'https://www.drush.org/latest/usage/#xdebug']));
+        }
+
         // Do legacy initialization (load static includes, define old constants, etc.)
         $this->init();
 
--- a/.ddev/config.yaml
+++ b/.ddev/config.yaml
@@ -27,4 +27,4 @@ web_environment:
 #  - UNISH_DB_URL=pgsql://db:db@db:5432?module=pgsql
   - DRUSH_OPTIONS_URI=$DDEV_PRIMARY_URL
   - EDITOR=nano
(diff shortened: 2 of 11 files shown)


weitzman:
A related change in DDEV is proposed at https://github.com/ddev/ddev/pull/6565

andriokha:
I don't know if I'm unusual in this, but just to provide a data point I do regularly (daily) use xdebug with various drush commands, especially `drush ev`. It's a bit of a hassle that I have to stop my IDE listening when eg. `drush cr`, but since xdebug 3 it's not a performance issue (if I stop listening) and it's quick and easy to do. But as long as it's configurable I don't see a problem at all, just thought I'd mention the usecase.

It might be nice to provide a command line option similar to [phpstan](https://phpstan.org/user-guide/command-line-usage#--xdebug) to make it easier to allow xdebug for users with docker/VMs where setting an env var can be a little more painful than just adding it to the start of the line?

weitzman:
I think you are paying a perf penalty if all you do is stop listening. Your XDEBUG_MODE=debug so drush is slowing down. Compare with Xdebug not even enabled if you want. You probably dont mind the slowdown, and I usually dont either. But I'm not a good use case because I actually want to debug Drush much of the time.

I dont think a Drush option makes sense because we'd have to parse it in Bash and then strip it. Instead we recommend either setting the env variable or calling `vendor/bin/drush.php` instead of `vendor/bin/php`. This is mentioned in the docs in this PR.

andriokha:
> I think you are paying a perf penalty if all you do is stop listening. Your XDEBUG_MODE=debug so drush is slowing down. 

Right you are: I had thought this was improved to the point of being negligible with xdebug 3, but on a larger site I use I can see an almost 30% increase in the time of a `drush cr` with xdebug on and not being listened to. Thanks!

> Instead we recommend either setting the env variable or calling vendor/bin/drush.php instead of vendor/bin/php. This is mentioned in the docs in this PR.

Thanks, and yeah there's totally a way to do it. For me who tends to use `ddev drush` I think I'll either end up making a custom `drush.php` ddev command, or using `ddev ssh`, which is not the end of the world, just less perfect than `--xdebug` would've been from my PoV (:

weitzman:
I can see why `ddev drush --xdebug` would be handy. Same for Lando et al. If someone wants to add that to the `/drush` shell script, we can evaluate its complexity. The rough flow would be

- if request contains --xdebug
    - Set env variable XDEBUG_ON
    - Strip --xdebug so that drush.php does not see it.

webflo:
https://github.com/drush-ops/drush/pull/6124/commits/a5864fc64d42a61f4c1c87b07e77afeb1929cd9b / https://github.com/drush-ops/drush/pull/6124 has the --xdebug flag.

I defined it as application global option. This means that it does not have to be removed, even if it has no function in the application itself.

weitzman:
Thanks. I pushed that commit to this PR and tweaked the docs a bit. All that is left now is some courage to merge or courage to decline. I'm mulling it.

src/Preflight/Preflight.php
```diff
@@ -255,6 +255,11 @@ public function preflight($argv): array
         // Now that we know the value, set debug flag.
         $this->logger()->setDebug($this->preflightArgs->get(PreflightArgs::DEBUG, false));
 
+        // Give hint if a developer might be trying to debug Drush.
+        if (extension_loaded('xdebug')) {
+            $this->logger()->log(strtr('If you are trying to perform step debugging of Drush, see !url', ['!url' => 'https://www.drush.org/latest/usage/#xdebug']));
```
andriokha: *If* we're going to keep `--xdebug`, does it make sense to mention it directly here? Maybe something like
> If you are trying to perform step debugging of Drush use the --xdebug option, or see !url for more options.

weitzman: I improved the wording and fixed the url. I opted not to mention --xdebug there because the dev needs to decide which approach suits her better. Thats best done by reading the docs.

alexpott:
This should be like composer and emit a message if xdebug is installed so users can learn about it.

alexpott:
Oh it seems like PHPStan and Composer no longer emit a message... oh well... this will catch people out. I guess TIL.

andriokha:
> This should be like composer and emit a message if xdebug is installed so users can learn about it.

I think that should be in place, are you not seeing it?

https://github.com/drush-ops/drush/blob/3d1ebc7d470b90d1718d236ebbaaa1def909cc0c/src/Preflight/Preflight.php#L258-L261

nicoguerinet:
"This issue introduces a breaking change with Drush execution.
 In some cases, if your bin/drush file doesn't have the correct execution rights, you have to run it with `php vendor/bin/drush`.
 Now, you need to use `php vendor/bin/drush.php` or `sh vendor/bin/drush`.
 It might be useful to document this change."

greg-1-anderson:
Proposing reverting this in #6146

mglaman:
We just upgraded to Drush 13 and this caused a breaking change as we were calling Drush with our `php` version. And now it crashes since it's a Bash script.

guerinetn:
> We just upgraded to Drush 13 and this caused a breaking change as we were calling Drush with our `php` version. And now it crashes since it's a Bash script.

You could change php by sh vendor/bin/drush or php vendor/bin/drush.php or simply remove the php before vendor/bin

mglaman:
It's still a breaking change that has consequences across our codebase and others that interact with it. Luckily we can use `~13.2.0` as a constraint.

weitzman:
I object to using "breaking changes" here, as it pertains to the semver guidelines. Semver only applies to supported usage of a tool (and even then, its mostly about API). Drush has documented for years that the way to specify which PHP should be run is to use the PATH feature of the OS. See the _Note_ at https://www.drush.org/13.x/install/.

If you prepend the drush call with /path/to/non-standard-php, you will only get that desired php for the first drush call, and not the subprocesses. Subprocesses are very common - see updatedb, sql:sync, config:pull, deploy, etc.

FWIW, the workaround to get the old behavior is super simple. You don't need to revert. You can call `drush.php` instead of `drush`.

RobinHoutevelts:
> > This should be like composer and emit a message if xdebug is installed so users can learn about it.
>
> I think that should be in place, are you not seeing it?

No, you do not see it. The `PreflightLog` only outputs when you run drush with `--debug`.

IMO It's not reasonable to expect someone to run `drush --debug` to figure this out. I never suspected `drush` to be overriding the `XDEBUG_MODE` variable. It was sheer luck that I figured this out.

PR #6121, opened 2024-09-27
PHPStan configuration update to fix ModuleHandler->loadInclude()

### LINKED ISSUE #6120: ModuleHandler->loadInclude() call is not being processed
As originally opened in https://github.com/mglaman/phpstan-drupal/issues/741, phpstan is not loading the includes when testing. 

We need to update the phpstan config to load the necessary rules from phpstan-drupal.

andyg5000:
closes #6120

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -5,8 +5,6 @@ parameters:
         - phpstan-bootstrap.php
     paths:
         - src
-    drupal:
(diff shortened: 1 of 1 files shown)


weitzman:
I had trouble pushing to the source branch so I started a new branch and PR at https://github.com/drush-ops/drush/pull/6122.

PR #6124, opened 2024-10-01
Disable xdebug by default to improve performance (--xdebug)

webflo:


webflo:
Fixed by https://github.com/drush-ops/drush/pull/6119

PR #6126, opened 2024-10-06
Update commands doc to make specific reference to the `AutowireTrait` requirement

bradjones1:
I wrote my own command (without using the generator) and it turns out that implementing `AutowireTrait` is required. Otherwise your commands are ignored silently and it's quite the WTF condition. Let's make this more explicit. As it's written I thought the reference was to autowiring in general, and not a requirement that I implement a specific trait.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -2,7 +2,7 @@
 
 !!! tip
 
-      1. Drush 13 expects commandfiles to use [Autowire](https://github.com/drush-ops/drush/blob/13.x/src/Commands/AutowireTrait.php) to inject Drupal and Drush dependencies. Prior versions used a [drush.services.yml file](https://www.drush.org/11.x/dependency-injection/#services-files) which is now deprecated and will be removed in Drush 13.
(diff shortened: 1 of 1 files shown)


docs/commands.md
```diff
@@ -2,7 +2,7 @@
 
 !!! tip
 
-      1. Drush 13 expects commandfiles to use [Autowire](https://github.com/drush-ops/drush/blob/13.x/src/Commands/AutowireTrait.php) to inject Drupal and Drush dependencies. Prior versions used a [drush.services.yml file](https://www.drush.org/11.x/dependency-injection/#services-files) which is now deprecated and will be removed in Drush 13.
+      1. Drush 13 expects commandfiles to use the [AutowireTrait](https://github.com/drush-ops/drush/blob/13.x/src/Commands/AutowireTrait.php) to inject Drupal and Drush dependencies. Prior versions used a [drush.services.yml file](https://www.drush.org/11.x/dependency-injection/#services-files) which is now deprecated and will be removed in Drush 13.
```
weitzman:
```suggestion
      1. Drush 13 expects commandfiles to use the [AutowireTrait](https://github.com/drush-ops/drush/blob/13.x/src/Commands/AutowireTrait.php) to inject Drupal and Drush dependencies. Prior versions used a [drush.services.yml file](https://www.drush.org/11.x/dependency-injection/#services-files) which is now deprecated and will be removed in Drush 14.
```

bradjones1:
Good catch on the removal version, didn't even look at that. Thanks for committing this!

PR #6128, opened 2024-10-07
Use the mariadb executable for SQL commands where possible

codebymikey:
On newer versions of MariaDB e.g. 11.4.3, the following warning is shown:
```
mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
```

### DIFF
--- a/src/Sql/SqlMariaDB.php
+++ b/src/Sql/SqlMariaDB.php
@@ -10,6 +10,14 @@ class SqlMariaDB extends SqlMysql
 {
     use ExecTrait;
 
+    public function command(): string
+    {
+        if (self::programExists('mariadb')) {
+            return 'mariadb';
+        }
+        return parent::command();
+    }
+
     public function dumpProgram(): string
     {
         if (self::programExists('mariadb-dump')) {


weitzman:
Thanks. Can we not count on `mariadb` existing? Mysql and postgres count on their program existing.

codebymikey:
I did think that, but did it this way to maintain as much backwards compatibility as possible just like the dump command does.

Based off the [MariaDB documentation](https://mariadb.com/kb/en/mariadb-command-line-client/):
> Prior to [MariaDB 10.5](https://mariadb.com/kb/en/what-is-mariadb-105/), the client used to be called `mysql`, and can still be accessed under this name, via a symlink in Linux, or an alternate binary in Windows.

Drupal 10 [currently requires](https://www.drupal.org/docs/getting-started/system-requirements/database-server-requirements) MariaDB 10.3.7+ and can confirm from a quick test in [docker lab](https://labs.play-with-docker.com/), the `mariadb` binary only started being included around 10.4 despite what the doc says:

```sh
~$ docker run -it --rm mariadb:10.3.7 /bin/sh -c 'which mariadb'
~$ docker run -it --rm mariadb:10.4 /bin/sh -c 'which mariadb'
/usr/bin/mariadb
```

So the executable check will need to be removed once support for MariaDB <10.5 is dropped.

andypost:
is it possible to backport to 12.x?

codebymikey:
Is it worth creating a separate PR for this, or couldn't the maintainers cherry-pick the commit over to 12.x?

PR #6135, opened 2024-10-14
Deprecate Annotated Commands in favor of native Symfony Console commands

weitzman:
### Status
- We will present this work at Drupalcon Vienna and hope to gain consensus for moving most of our commands into Drupal core or Drupal CMS. 
- This PR is backward compatible. Therefore we intend to merge it into the 13.x branch. Drush 14 may remove support for old style commands. That is a separate decision.

### Open issues
- [x] _Keeping the feature as is_. Keep dependency on `consolidation/filter-via-dot-access-data`, move that into output-formatters, or drop the feature. See [\Drush\Formatters\FormatterTrait::alterResult](https://github.com/drush-ops/drush/pull/6135/files#diff-5a0f3e0205efb6c5c7f49a5f009d333550d7a735a4a0e70f904b8dae0451109dR48-R50)
- [x] redispatch.hook [currently looks at AnnotationData ](https://github.com/drush-ops/drush/blob/cdcedee6808ed43124c635203b9248a07249ce78/src/Runtime/RedispatchHook.php#L51)in order to stop redispatches. It needs to be able to operate on Console commands.
- [x] Topics are not part of the Console data model. I'm thinking we append links to www.drush.org pages at bottom of  a command's help.
- [ ] _Update: we will do this when the PR is ready for merge_. Move FormatterTrait and related Attributes into consolidation/output-formatters.
- [x]  _Update Moshe hacked image:flush into Dex and it worked!_ Attempt to re-use the new style commands and listeners in a separate vanilla Console app, or use Dex. The goal is to verify how consumable these commands are.
- [x] _Update: we chose Attributes_. Decide between two methods of delivering OptionSets. [SqlDumpCommand](https://github.com/drush-ops/drush/pull/6135/files#diff-6b1aae27e0c9c77cf7d57ac0f70cc0a4265380426291fa0fa47766fabf0c164f) demonstrates both approaches:
    1. The _tableSelection_ optionset is added by adding #[OptionsetTableSelection] on the class (i.e. keep the status quo). This signals [OptionsetTableSelectionListener](https://github.com/drush-ops/drush/pull/6135/files#diff-fa8a26480deb403a73c729cbdc0a482d2f699c80cc30f55c476e275043db7450) to add the options. 
    2. The _sql_ optionset is added by [calling a static method from configure()](https://github.com/drush-ops/drush/pull/6135/files#diff-6b1aae27e0c9c77cf7d57ac0f70cc0a4265380426291fa0fa47766fabf0c164fR59). No Attribute is involved. This is less magical (good) but perhaps slightly harder to discover as a command author.
- [x] _Update: we chose Attributes_. Same as above, for validators. [ImageFlushCommand](https://github.com/drush-ops/drush/pull/6135/files#diff-3b7cd99de69343fefe9c81829f97ac88a82904e64b45cdda528de7974483518a) demonstrates the two approaches.
    1. The `#[ValidateModulesEnabled]` Attribute signals to [ValidateModulesEnabledListener](https://github.com/drush-ops/drush/pull/6135/files#diff-322d213e3e246519d83057f5807a8bdd9b7c30b1c4c900e19ce9bdec46e71b1d) to do its validation work
    2. [The top of execute() calls the static Validators::entityLoad()](https://github.com/drush-ops/drush/pull/6135/files#diff-3b7cd99de69343fefe9c81829f97ac88a82904e64b45cdda528de7974483518aR68)

### Description
It is our hope that Drush commands will be consumed by a future CLI in Drupal core. To make our commands more consumable, this PR changes Drush commands to directly extend Symfony's Command class. 

Converted commands are [sql:dump](https://github.com/drush-ops/drush/pull/6135/files#diff-6b1aae27e0c9c77cf7d57ac0f70cc0a4265380426291fa0fa47766fabf0c164f), [twig:unused](https://github.com/drush-ops/drush/pull/6135/files#diff-adb922e5dacf604fec2453710d6cd73c99d8ea00a026bf04718863037f6213ad) and [image:flush](https://github.com/drush-ops/drush/pull/6135/files#diff-3b7cd99de69343fefe9c81829f97ac88a82904e64b45cdda528de7974483518a). Some highlights:
1. Each command gets its own file. Attributes are placed on the class instead of the method.
1. Use Symfony's #[AsCommand] Attribute to declare command name /description/alias. The Drush #[Command] Attribute is deprecated.
1. Our commands extend Symfony's `Command` (not DrushCommands) 💪. 
1. Use configure() and interact() methods as per usual with Symfony commands. Drush's #[Argument] and #[Option] PHP Attributes are deprecated.
1. Drush and Drupal services may still be autowired. This is how you access the logger. Build own $io as needed.
1. Commands that wish to offer multiple _output formats_ (yes please!) should: 
    - Implement the new [Formatter Attribute](https://github.com/drush-ops/drush/blob/13.x/src/Attributes/Formatter.php).
    - `execute()` is largely boilerplate. By convention, do your work in a `doExecute()` method instead.
    - See [TwigUnusedCommand](https://github.com/drush-ops/drush/pull/6135/files#diff-adb922e5dacf604fec2453710d6cd73c99d8ea00a026bf04718863037f6213ad) or [SqlDumpCommand](https://github.com/drush-ops/drush/pull/6135/files#diff-6b1aae27e0c9c77cf7d57ac0f70cc0a4265380426291fa0fa47766fabf0c164f) as examples.

Console recently introduced invokable commands - https://github.com/symfony/symfony/pull/59340 ([blog post](https://symfony.com/blog/new-in-symfony-7-3-invokable-commands-and-input-attributes)). This PR fully supports them when used Symfony with 7.4+ (releases in Nov 2025). An example in Drush is [MyCatCommand](https://github.com/drush-ops/drush/pull/6135/files#diff-5bb079709d3d4a2ea36433495886abc90287495f5b58e7cb0cc60c149c8b9a65). This will become the recommended way to write commands after Nov 2025 when you only care about supporting Drupal 11+.

We have more plans for simplifying Drush's Preflight and removing the [consolidation/annotated-command](https://github.com/consolidation/annotated-command) and [consolidation/robo](https://github.com/consolidation/robo) dependencies - this is just a start!


src/Formatters/FormatterTrait.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Format the structured data as per user input and the command definition.
+     */
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $configurationData = $this->getConfigurationData($this);
+        $formatterOptions = new FormatterOptions($configurationData, $input->getOptions());
+        $formatterOptions->setInput($input);
+        $data = $this->doExecute($input, $output);
```
greg-1-anderson: I was thinking maybe doExecute should be named "fetch()"

sut/modules/unish/woot/src/Drush/EventListener/WootDefinitionListener.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+    }
+
+
+    public function __invoke(ConsoleDefinitionsEvent $event): void
+    {
+        foreach ($event->getApplication()->all() as $id => $command) {
+            if ($command->getName() === 'woot:altered') {
+                $command->setAliases(['woot-new-alias']);
+                $this->logger->debug(dt("Module 'woot' changed the alias of 'woot:altered' command into 'woot-new-alias' in " . __METHOD__ . '().'));
```
claudiu-cristea, line 27: I think test should be adapted

greg-1-anderson:
We had talked about moving FormatterTrait to consolidation/output-formatters. I would like to do that. I think this implies that  [src/Attributes/FieldLabels.php](https://github.com/drush-ops/drush/pull/6135/files#diff-77e85b11336db8c2b0c161f3b0762d5a908d5e228f51c43bf3af77c5dd5049af) and the other format-related attributes would also need to move over. We can maintain compatibility with Drush 13 if we make Drush\Attributes\FieldLables & c. extend the consolidation/output-formatters attribute.

Should I go ahead and do that, and update this PR to use the output-formatters copy?

src/Formatters/FormatterTrait.php
weitzman: I think doExecute() better communicates the wrapping we are doing.

sut/modules/unish/woot/src/Drush/EventListener/WootDefinitionListener.php
weitzman: Done. Thanks.

weitzman:
I'm undecided about moving FormatterTrait and the related Attributes. I do agree that both are coupled and probably should live together. When authoring a command, its helpful for all Attributes to be in same namespace so you can browse them in your IDE. If we move the Formatter attributes, we bifurcate them, or we thinly extend the ones provided by output-formatters. We do that thin extension today for those provided by annotated-command, but I always found that slightly awkward. Lets mull it.



_Update: the implementation now complete._ Maybe @greg-1-anderson has a suggestion about how to implement --filter in the new paradigm. See [DrushCommandAlterer](https://github.com/drush-ops/drush/blob/13.x/src/Command/DrushCommandInfoAlterer.php) and [FilterHooks](https://github.com/consolidation/filter-via-dot-access-data/blob/cb2eeba41f8e2e3c61698a5cf70ef048ff6c9d5b/src/Hooks/FilterHooks.php). In this PR I've added `#[CLI\FilterDefaultField(field: 'template')]` to twig:unused, and [\Drush\Formatters\FormatterTrait::addFormatterOptions] auto-adds the --filter option when needed. Lets discuss how to implement result altering. I'm hoping that command authors don't have to worry about. Maybe we add result altering to `\Drush\Formatters\FormatterTrait::execute` just before formatting? 

`--filter` is a seldom used feature of Drush and perhaps we should drop it.

src/Commands/core/ImageFlushCommand.php
```diff
@@ -0,0 +1,80 @@
...
+    description: 'Flush all derived images for a given style.',
+    aliases: ['if', 'image-flush']
+)]
+#[CLI\ValidateEntityLoad(entityType: 'image_style', argumentName: 'style_names')]
+#[CLI\ValidateModulesEnabled(modules: ['image'])]
+final class ImageFlushCommand extends Command
+{
+    use AutowireTrait;
+
+    const FLUSH = 'image:flush';
```
Chi-teck: I propose we always to declare visibility scope and type hint for constants. As it just a reference for the command can we reserve `NAME` name for referencing command names?
```php
public const string NAME = 'image:flush';
```

src/Commands/core/ImageFlushCommand.php
```diff
@@ -0,0 +1,80 @@
...
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        // Needed for non-interactive requests.
+        if ($input->getOption('all')) {
+            $input->setArgument('style_names', implode(',', array_keys(ImageStyle::loadMultiple())));
+        }
+
+        foreach (ImageStyle::loadMultiple(StringUtils::csvToArray($input->getArgument('style_names'))) as $style_name => $style) {
+            $style->flush();
+            $this->logger->success(dt('Image style !style_name flushed', ['!style_name' => $style_name]));
```
Chi-teck: What's the purpose of `dt` function here?  The PSR-3 loggers must be capable of string interpolation. Also success method is not a part of  PSR-3. So the proper typehint for it should be `LoggerInterface&SuccessInterface`.

weitzman: Thanks for bringing up the typehint. When I use LoggerInterface&SuccessInterface, AutoWireTrait fails to find a service. I don't want to hint to DrushLoggerManager since Drupal will not want to inject that if it ever takes over our commands. We could potentially add more smarts to our AutoWireTrait. See https://www.drupal.org/project/drupal/issues/3417735 and https://symfony.com/blog/new-in-symfony-5-4-dependencyinjection-improvements#autowire-union-and-intersection-types for background. Thoughts? 

I removed dt() from the log messages - good idea. I have seen a styling issue with this approach (illegible text on a red background) when used with error() but we can deal with that when it comes up.

src/Application.php
```diff
@@ -352,4 +352,14 @@ protected function addListeners($commandfileSearchpath): void
         $listenerClasses = $this->serviceManager->filterListeners($listenerClasses);
         $this->serviceManager->addListeners($listenerClasses, Drush::getContainer());
     }
+
+    // Remove a command. Initially used by WootDefinitionListener and its test.
+    public function remove(string $id): void
+    {
+        $rf = new \ReflectionProperty(\Symfony\Component\Console\Application::class, 'commands');
+        $rf->setAccessible(true);
```
greg-1-anderson: Not super excited to do this outside of tests. Is this for some sort of "alter" hook? If we want to do this, maybe we should, during preflight, add commands to some other list and provide a specific hook to alter that before adding to the application. That might be complicated, since today we add in two steps (commands from modules being added later).

Alternately, instead of removing a command, maybe we could change its name, remove its aliases and hide it?

I'm not sure why I think that setAccissible is worse than that.

src/Commands/core/ImageFlushCommand.php
greg-1-anderson: Can we maybe just get rid of the success log level now?

greg-1-anderson:
Yes, I was thinking that filter could happen in the format trait.

I think that consolidation/output-formatters should have the trait and format attributes, so that they are available when that library is used outside of Drush. I'll probably duplicate the code there regardless, even if Drush doesn't use it. Perhaps thin extensions would be a workable solution, where necessary.

src/Commands/core/ImageFlushCommand.php
Chi-teck: I think custom log levels is tricky thing. Though the PSR-3 is a bit lenient about this.

> Calling this method with a level not defined by this specification MUST throw a Psr\Log\InvalidArgumentException if the implementation does not know about the level. Users SHOULD NOT use a custom level without knowing for sure the current implementation supports it.


The alternative approach would by just passing special context to logger. So loggers that support this can handle this messages in a different way. For example, console logger could display it in green.
```php
$logger->info('Log message', ['success' => true]);
```

src/Commands/core/ImageFlushCommand.php
weitzman: The type hint requires PHP 8.3 and we are only at 8.2. Lets do that part whenever we bump min PHP.

weitzman:
OK, when this PR is ready to go in, lets move the FormatterTrait and format Attributes, and Drush may thinly extend the Attributes.

src/Commands/core/ImageFlushCommand.php
weitzman: > $logger->info('Log message', ['success' => true]);

That looks good to me. I think we have to use notice() or else it wont show up without -v. I would love to have a wrapper like $this->success() but we extend Command now and have nowhere to put that code.

weitzman: My [last commit](https://github.com/drush-ops/drush/pull/6135/commits/a194c59554249057f5150e8c91b794e10ab2adaa) explores using $this->io->success(). Screenshot below. What do you guys think?

image

Chi-teck: I think `$io` is more appropriate tool for such messages than logger.

Chi-teck:
The formatter thing looks complex and too magic. Will it save much time for command developers?

I think `FormatterTrait` should have just one simple getter for the formatter instance.

Something like this.
```php
private static function getOutputFormatter(): FormatterManager {
  return \Drupal::service(FormatterManager::class);
}
```

It seems quite easy to apply it manually.
```php
protected function configure(): void
{
    $this
       ->addArgument('some-arg', InputArgument::REQUIRED, 'Description');
    self::getOutputFormatter()->configureCommand($this);
}

public function execute(InputInterface $input, OutputInterface $output): int
{
  $data = [];
  self::getOutputFormatter()->write($output, $input->getOption('format'), new RowsOfFields($data));
  return self::SUCCESS;
}
```

I would actually inject the formatter through constructor rather than through trait.

Chi-teck:
Given that `Command::addOption()` is a public method. We could potentially add `format` option automatically to commands that indicate a need for it through some PHP attribute.

Also curious if it is possible to integrate the formatter with `DrushStyle`. For instance `SymfonyStyle` has methods like `table()`, `listing()`, etc.  Why can't `DrushStyle` have something like `writeFormatted()`?

src/Runtime/DependencyInjection.php
```diff
@@ -169,11 +163,16 @@ protected function addDrushServices(Container $container, ClassLoader $loader, D
...
 
-    protected function alterServicesForDrush($container, Application $application): void
+    protected function alterServicesForDrush($container, Application $application, InputInterface $input, OutputInterface $output): void
     {
         $paramInjection = $container->get('parameterInjection');
         $paramInjection->register('Symfony\Component\Console\Style\SymfonyStyle', new DrushStyleInjector());
 
+        Robo::addShared($container, 'io', DrushStyle::class)
+            ->addArguments([$input, $output]);
+        // @todo Does this alias interfere with the paramInjector above? Could we drop that feature?
```
greg-1-anderson: Parameter injection only happens when an Annotated Command is dispatched. We should probably keep this until Robo and Annotated Command are removed completely from Drush. That could happen in Drush 14, but waiting for Drush 15 would give folks time to convert old commands.

Is it necessary to put `io` in the DI container? Doing this makes it harder to re-use code, e.g. calling a command object, but capturing its output in a buffer instead of letting it go to stdout. Symfony enables this by passing $input and $output in as parameters, e.g. in the `exectue` method. Robo struggles with this due to the way the DI container is used, especially with respect to io and the logger. Parameter injection was a mechanism / attempt to go back to the Symfony model without breaking backwards compatibility. However, "optional parameters" are unusual / magical; folks don't know to look for them or use them, and as a result, the feature was seldom used.

weitzman: The only reason why I'm injecting io instead of having everyone do `$io = new DrushStyle($input, $output)` during interact() and execute() is that Drupal could inject its own Style and the Command would not require any change. I'm not sure if thats a good enough reason to risk the problems that you describe. I'm OK with omitting this from the container if you think its wise.

greg-1-anderson: I think it's better to avoid putting $output, $input, $io and $logger in the container. We're a ways off from removing all of those, but I'd resist adding more in the same category.

weitzman:
1. The output formatter system is opt-in. Commands who dont care can ignore it.
2. The doExecute and getFormatterOptions() are a bit of a dance, I agree. I'm open to suggestions. Note that your example didn't pass FormatterOptions which is where most of the complexity resides. Also note that it needs to calculated during configure (for automaticOptions - only valid formats are listed in the description and autocomplete), and execute (for the actual filtering and formatting)

I attempted to remove a bit of magic for commands that format  - see [SqlDumpCommand](https://github.com/drush-ops/drush/pull/6135/files#diff-6b1aae27e0c9c77cf7d57ac0f70cc0a4265380426291fa0fa47766fabf0c164f). The command no longer has to add PHP Attributes like #[DefaultTableFields]. Instead, the command directly implements a getFormatterOptions() method to specify its configurationData. This is discussed a bit more in _Open Issues_ above.

src/Commands/core/ImageFlushCommand.php
greg-1-anderson: Yes, we can construct a Style object from $input and $output, which are parameters to execute. The logger is a little more fraught; we don't want people making their own logger from $output, but calling a command directly with a separate logger is not convenient with the logger passed via the DI container. I therefore think that moving folks away from the logger and to a Style method is preferable for the "success" method. I'm less sure about the fate of the rest of the logger, but at least this much is a positive change.

src/Application.php
weitzman: > add commands to some other list and provide a specific hook to alter that before adding to the application

I think thats more more easily done once we use a container to store commands and maybe use a CommandLoader. Yes, this is just used by Woot today  - we can remove it if it offends the eyes.

greg-1-anderson:
I like the requirement to call some method in `configure()` to set up the formatter options. I am not very fond of using a static method for this, though. Perhaps calling a method of some injectable class would be better. Unsure, though.

greg-1-anderson:
I am also wondering if formatter options can somehow be set up by calling methods in `configure()`, just as args and options are set up, rather than having folks provide a `getFormatterOptions()` method, which is a different pattern. Maybe the FormatterTrait could provide something.

weitzman:
The key bits that only the command knows the values for stuff like DefaultTableFields, FieldLabels, etc. Everything else is machinery. 

> I am also wondering if formatter options can somehow be set up by calling methods in configure()

We could add methods if we extended Command, like AnnotatedCommand does. Laravel extends it too. I'm not sure how loathsome that would be to Drupal.

greg-1-anderson:
Maybe those methods could be in the FormatterTrait.

src/Attributes/FilterDefaultField.php
```diff
@@ -1,12 +1,27 @@
...
+     */
+    public function __construct(
+        public string $field
+    ) {
+    }
+
+    public static function handle(\ReflectionAttribute $attribute, CommandInfo $commandInfo)
+    {
+        $args = $attribute->getArguments();
+        $commandInfo->addAnnotation('filter-default-field', $args['field']);
```
greg-1-anderson: Here and elsewhere, `KEY` is defined in the class but not used in handle()

src/Commands/OptionSets.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Drush\Commands;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputOption;
+
+class OptionSets
+{
+    public static function sql(Command $command)
```
greg-1-anderson: Maybe the static method technique of holding option sets would sit better if it lived in some "sql" namespace, rather than putting all option set domains in the same class. (This would be more obvious if there were more than one here.)

src/Commands/core/TwigUnusedCommand.php
```diff
@@ -0,0 +1,110 @@
...
+    {
+        $this
+            ->addArgument('searchpaths', InputArgument::REQUIRED, 'A comma delimited list of paths to recursively search')
+            // Usages can't have a description with plain Console :(. Use setHelp() if desired as per  https://github.com/symfony/symfony/issues/45050
+            ->addUsage('twig:unused /var/www/mass.local/docroot/modules/custom,/var/www/mass.local/docroot/themes/custom')
+            ->setHelp('Immediately before running this command, web crawl your entire web site. Or use your Production PHPStorage dir for comparison.');
+        $this->addFormatterOptions(RowsOfFields::class);
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
```
greg-1-anderson: Why does this command have its own copy of execute?

src/Formatters/FormatterTrait.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    /**
+     * Adds options to the command definition based on data type. The --format description is dynamic.
+     *
+     * @param string $dataType
+     *   Usually the same as the return type of a doExecute() method.
+     */
+    public function addFormatterOptions(string $dataType): void
+    {
+        $inputOptions = $this->formatterManager->automaticOptions($this->getFormatterOptions(), $dataType);
```
greg-1-anderson: I think this can be refactored to avoid the intermediate instances and private field access. This can wait until moving this code into output-formatter.

src/Commands/OptionSets.php
weitzman: True. The benefit of putting multiple methods on a single OptionSet class is that the optionsets are more discoverable. The class analogous to our existing https://github.com/drush-ops/drush/blob/13.x/src/Commands/OptionsCommands.php

src/Attributes/FilterDefaultField.php
weitzman: Just pushed a fix with a couple tweaks. Depends on another PR in output-formatters https://github.com/consolidation/output-formatters/pull/113

src/Commands/core/TwigUnusedCommand.php
weitzman: I removed execute() from FormatterTrait and into the Command. See https://github.com/drush-ops/drush/pull/6135/commits/6216ad68e4f0a41013e879ecb206877abaf57be5. This design has a bit less magic and a bit more boilerplate for the command. I think its a decent tradeoff.

greg-1-anderson: If format commands have to copy boilerplate code, it seems that there's little need to factor out `doExecute`. I think I still favor putting execute in the trait, with `doExecute` renamed to something like `getOutputData()`.

I often prefer to Write Everything Twice rather than make all code perfectly DRY, but in this case I think it's useful to have the boilerplate code in the trait.

How about this:

Every command's `execute()` method calls a differently-named method in the formatter trait. Something like this:
```
    public function execute(InputInterface $input, OutputInterface $output): int
    {
         $rowsOfFields = $this->iCanNameThisMethodWhateverILike();
         $this->writeFormattedOutput($input, $output, $rowsOfFields);
         return self::SUCCESS;
    }
```
This eliminates hiding the implementation of `execute()` in the trait, while still providing the boilerplate code for commands to use.

I am still thinking that formatterOptions should be set up in `configure()`, for consistency with Symfony Console patterns (and also to give us a way to add the `--format` flag automatically), but `setInput()` would still be called by `writeFormattedOutput()`.

weitzman: [That proposal is very close to where the code is now.](https://github.com/drush-ops/drush/pull/6135/files#diff-adb922e5dacf604fec2453710d6cd73c99d8ea00a026bf04718863037f6213adR59-R65) FormatterTrait no longer care about the doExecute() method name or its return type hint. 

In the current code getFormatteroptions() is called during configure() and again during execute(). I think you are proposing that it get called once during configure() and stored in a property in FormatterTrait? That works for me.

I'm not sure in your proposal where filtering is happenning. In the link above, its in the call to `$this->alterResult($data, $input)`

weitzman: I made a commit which implements what I think you re suggesting - https://github.com/drush-ops/drush/pull/6135/commits/f923a801dd50d4aa17fd2390790336bc962785a3

greg-1-anderson: Yeah, just a minor adjustment, but I like it a lot better.

src/Attributes/ValidateEntityLoad.php
```diff
@@ -8,7 +8,9 @@
 use Consolidation\AnnotatedCommand\CommandData;
 use Consolidation\AnnotatedCommand\CommandError;
 use Drush\Utils\StringUtils;
+use JetBrains\PhpStorm\Deprecated;
 
+#[Deprecated(replacement: 'Copy \Drush\Commands\core\ImageFlushCommand::validateEntityLoad into command and call during execute()')]
```
greg-1-anderson: Could we maybe use a [command hook](https://symfony.com/doc/current/components/console/events.html#the-consoleevents-command-event) to find and dispatch validate hooks?

src/Attributes/FilterDefaultField.php
weitzman: I have experimented with that approach earlier in this PR and can't decide whats best. Currently the validation is done right in execute() and isn't not DRY. See [ImageFlushCommand](https://github.com/drush-ops/drush/pull/6135/files#diff-3b7cd99de69343fefe9c81829f97ac88a82904e64b45cdda528de7974483518aR65-R101). Compare with https://github.com/drush-ops/drush/blob/f923a801dd50d4aa17fd2390790336bc962785a3/src/Listeners/ValidateModulesEnabledListener.php and its associated Attribute. IMO the current approach is less magical and I'm slightly valuing that these days.

src/Attributes/ValidateEntityLoad.php
weitzman: I have experimented with that approach earlier in this PR and can't decide whats best. Currently the validation is done right in execute() and isn't not DRY. See [ImageFlushCommand](https://github.com/drush-ops/drush/pull/6135/files#diff-3b7cd99de69343fefe9c81829f97ac88a82904e64b45cdda528de7974483518aR65-R101). Compare with https://github.com/drush-ops/drush/blob/f923a801dd50d4aa17fd2390790336bc962785a3/src/Listeners/ValidateModulesEnabledListener.php and its associated Attribute. IMO the current approach is less magical and I'm slightly valuing that these days.

weitzman:
I posted a status update to the issue description here. It says

> This PR is on long term hold. Since Drush is already bundled with Drupal CMS, most newcomers to Drupal are already getting Drush's benefits. If a member of Drupal Leadership expresses intent to use these commands directly in core or CMS, I (moshe) will restart work on this.

andypost:
makes sense to link Ideas issue which probably is the next to decide https://www.drupal.org/project/ideas/issues/3159647

weitzman:
Symfony Console recently released 7.3 which includes https://github.com/symfony/symfony/pull/59340 ([blog post](https://symfony.com/blog/new-in-symfony-7-3-invokable-commands-and-input-attributes)). This PR should probably be adjusted accordingly. Thoughts are welcome.

src/Attributes/OptionsetTableSelection.php
```diff
@@ -18,7 +18,6 @@ public static function handle(\ReflectionAttribute $attribute, CommandInfo $comm
         $commandInfo->addOption('tables-key', 'A key in the $tables array.', [], DrushCommands::REQ);
         $commandInfo->addOption('skip-tables-list', 'A comma-separated list of tables to exclude completely.', [], DrushCommands::REQ);
         $commandInfo->addOption('structure-tables-list', 'A comma-separated list of tables to include for structure, but not data.', [], DrushCommands::REQ);
-        $commandInfo->addOption('skip-tables-list', 'A comma-separated list of tables to exclude completely.', [], DrushCommands::REQ);
```
weitzman, line 21: This line was a duplicate

Issue #6137, opened 2024-10-16
Regression: Drush 13.3 does not properly handle SIGQUIT signal


Chi-teck:
That apparently is caused by #6124.

Here is example of a simple long running Drush command

```php

#[AsCommand( 'example')]
final class ExampleCommand extends Command implements SignalableCommandInterface {

  /**
   * Signal flag.
   */
  private bool $stop = FALSE;


  /**
   * {@inheritdoc}
   */
  protected function execute(InputInterface $input, OutputInterface $output): int {
    for ($i = 1; $i <= 60; $i++, sleep(1)) {
      echo '.';
      if ($this->stop) {
        echo 'Signal received...', \PHP_EOL;
        sleep(5);
        echo 'Done!', \PHP_EOL;
        break;
      }
    }
    return 0;
  }

  /**
   * {@inheritdoc}
   */
  public function getSubscribedSignals(): array {
    return [\SIGTERM, \SIGQUIT, \SIGINT];
  }

  /**
   * {@inheritdoc}
   */
  public function handleSignal(int $signal): false|int {
    $this->stop = TRUE;
    return FALSE;
  }

}
```

Start the above command and send SIGQUIT (Ctrl  + \\) to it.

**Output Drush 13.2 (correct)**
```
$ ./vendor/bin/drush example
....^\.Signal received!
Done!
```

**Output Drush 13.3 (wrong)**
```
$ ./vendor/bin/drush example
....^\.Signal received!
Quit (core dumped)
$ Done!
```

Note that in the last example "Done!" appears when Drush was already terminated. If a command is managed through some process manager, i.e. Supervisor,  such behavior causes orphan processes.

Chi-teck:
The new way to launch Drush creates two additional shell processes besides the PHP. Also some people reported that Drush 13.3 is not working on Windows because of this change.

webflo:
@Chi-teck Could you try to run ``./vendor/drush/drush/drush.php`` directly? Just to make sure, this executable has the correct behaviour?

Chi-teck:
Yes, drush.php works correctly.

```
$ ./vendor/bin/drush example
.....^\.Signal received...
Quit (core dumped)
$ Done!

$ ./vendor/drush/drush/drush.php example
.....^\.Signal received...
Done!
```

webflo:
In the original PR I implemented it without exec, because the Composer Wrapper script also works this way. 

[...]

PR #6138, opened 2024-10-17
Fix xdebug handling for sub processes

webflo:
The xdebug session is dropped for drush subprocesses. Use export to make it persistent.

### DIFF
--- a/drush
+++ b/drush
@@ -16,7 +16,7 @@ parse_commandline()
 		_key="$1"
 		case "$_key" in
 			--xdebug)
-				DRUSH_ALLOW_XDEBUG=1
+				export DRUSH_ALLOW_XDEBUG=1
 				return
 				;;
 		esac


weitzman:
LGTM. I'll wait a day for feedback. Any chance you can chime in at https://github.com/drush-ops/drush/issues/6137?

claudiu-cristea:
Not sure it's related. Working on contrib with https://github.com/ddev/ddev-drupal-contrib. No special config just following the docs. Drush is 13.3.0. When I try to debug:

```bash
$ ddev drush --xdebug en webform
Xdebug: [Step Debug] Could not connect to debugging client. Tried: host.docker.internal:9003 (fallback through xdebug.client_host/xdebug.client_port).
```

and it doesn't stop at breakpoints

weitzman:
@claudiu-cristea - that looks like xdebug extension is not enabled in DDEV. Thats a prerequisite for using --xdebug.

PR #6140, opened 2024-10-17
Replace subshell with exec

webflo:
Related to https://github.com/drush-ops/drush/issues/6137

---

This replaces the sub-process with exec, so the signals are treated correctly. But it won't fix the issue if drush is invoked from ```./vendor/bin/drush``` (the composer wrapper script, creates a sub-process too).

### DIFF
--- a/drush
+++ b/drush
@@ -31,4 +31,4 @@ then
   export XDEBUG_MODE=off
 fi
 
-"$DRUSH_PHP" "$@"
+exec "$DRUSH_PHP" "$@"


weitzman:
LGTM. Lets see if anyone else has thoughts ... Please remove from draft when you are satisfied.

Chi-teck:
`./vendor/bin/drush` is a primary endpoint for Drush. Given it does not handle signals you will have to update your Drush launcher, CI scripts, crontabs, etc. Everything that calls Drush...

Chi-teck:
Also these shell wrappers increase the number of processes. Drush 13.3 for each command starts 3 processes. I have a project that runs ~30 Drush workers in background, which produces 90 processes. With this PR it'll be reduced to 60 processes I guess, but it is still not good.

Chi-teck:
Filed a ticket for Composer
https://github.com/composer/composer/issues/12164

webflo:
@weitzman Done, Composer adopted the same solution.

webflo:
@Chi-teck Could you do a test with ```composer self-update --snapshot```? This has the new exec code

Chi-teck:
Confirm. With this PR and updated composer no additional shell processes are created. Also signals are handled correctly.

PR #6146, opened 2024-10-25
Revert drush bash script; return to native php front controller

greg-1-anderson:
Drush 13.3.0 replaced the native php front controller `drush` script with a bash wrapper, similar to the "full service" script provided in Drush 8, although on a much smaller scale.  Starting with Drush 9, though, we decided that Drush was not going to be in the business of setting up the user's shell, and instead presumed that users would do this themselves, and then call the Drush php front controller directly.

There are two main behavior changes switching back to bash:

- An extra shell is created, meaning that the terminal process is no longer the parent process of Drush.
- The subprocess can cause problems with signal handling (see https://github.com/drush-ops/drush/issues/6137).
- Since the language of the entrypoint script has changed, it is no longer possible to run `php <flags> drush`

Making this change in a minor release breaks the uniformity of launching Drush that has been consistent since Drush 9. If we do want to ultimately go back to bash, perhaps even restoring some of the features of the Drush 8 script, I think it would be better to wait and do it in the next major version of Drush.

As a compromise, this PR proposes that we move the bash entrypoint script to `bin/drush`. This allows XDebug users, or anyone else who prefers the bash script to put `./vendor/drush/drush/bin` in their $PATH in front of `./vendor/bin`, which would automatically run the bash script instead of the php script whenever it is available. Long term, I'd prefer to maintain this new pattern for the bash script, to minimize the number of times that we change how Drush is launched (currently, only changed in Drush 2 and Drush 9, at the moment).


andypost:
Handling signals is the strong argument for dockerized usage

weitzman:
Composer fix is in for signal issue https://github.com/composer/composer/pull/12176. I think this PR is no longer needed.

andypost:
++ to close the issue as `EXEC` now used

PR #6151, opened 2024-10-30
Update deploy documentation to clarify that the Drupal API isn't completely unavailable

andriokha:
This was triggered by [a Slack conversation](https://drupal.slack.com/archives/C62H9CWQM/p1730311183659209) where someone was uncertain exactly what it meant when the page said the Drupal API was not allowed. I think this could be particularly confusing if you aren't already primed on writing `hook_update_N()` implementations.

Just a suggestion, not entirely sure it's better...

Oh and I also muddied the waters by adding in whether it's implemented by Drupal or not. A contrib author might not realize they can't rely on the deploy hook, so I thought we could maybe add that.

Thanks!

### DIFF
--- a/docs/deploycommand.md
+++ b/docs/deploycommand.md
@@ -13,15 +13,20 @@ drush deploy:hook
 ```
 
 ## Authoring update functions
-Below are the 3 types of update functions run by this command, in order. Choose the most appropriate for your need. 
+Below are the 3 types of update functions run by this command, in order. Choose the most appropriate for your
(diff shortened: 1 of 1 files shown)


andriokha:
Maybe better with a warning box?
image

weitzman:
I want to say less, not more. It isn't the responsibility of deploy docs to educate devs on how to write an update hook. Lets change the wording to 'Limited API' and have that link to https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_update_N/11.x

weitzman:
Actually, I just merged it. Thanks.

PR #6153, opened 2024-11-01
Relax auto-discoverable command file/class naming convention

claudiu-cristea:
From https://mglaman.dev/blog/auto-discovery-global-commands-drush

> Personal note: I find the file naming requirement a bit redundant since we already have a specific namespace, and the class is inspected to ensure it isn't abstract, an interface, and is a subclass of Drush\Commands\DrushCommands.

Agree, lets relax the file/class naming convention `*DrushCommands` -> `*Commands`


weitzman:
LGTM. Will merge if nobody objects.

heddn:
This caught me when I converted from drush.services.yml to only autowire. I had the class in the right folder but it still wasn't being discovered, much less autowired. The more simple of requirements, the better.

PR #6182, opened 2024-12-13
Test with PHP 8.4

greg-1-anderson:
I don't have any strong opinions on what version of PHP we test with, but I wanted to see this run once, at least, to see how it fares now that there are 8.4 releases for all of the Consolidation components.


greg-1-anderson:
The "highest" tests all pass, so my work here is done; we can close.

The "lowest" tests are failing. We could fix that by ensuring that "lowest" runs on a lower PHP version, but I don't care about that too much one way or another. Leaving open in case someone wants to fix, otherwise feel free to close.

PR #6185, opened 2025-01-01
Add name of database to drush error message on "table not found" message. 

AJRepo:
I was doing a multi-site migration. The drush warning messages didn't identify which database drush was checking. This made the diagnosis take longer. This patch modifies reports to include the database name and clarify grammar to make it clearer that the check drush was doing is looking for a specific table named key_value in that database.

### DIFF
--- a/src/Boot/DrupalBoot8.php
+++ b/src/Boot/DrupalBoot8.php
@@ -167,7 +167,7 @@ class DrupalBoot8 extends DrupalBoot
             return false;
         }
         if (!$connection->schema()->tableExists('key_value')) {
-            $this->logger->info('key_value table not found. Database may be empty.');
+	    $this->logger->info('Can connecct to database ' . $connection_options['database'] .' but the table key_value was not found. Database ' . $connection_options['database'] . ' may be empty.');
             return false;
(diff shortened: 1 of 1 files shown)


weitzman:
Too niche.

AJRepo:
> Too niche.

The point of better error messages is to deal with edge cases. In this case the "edge" case is a user new to this part of drush as this kind of error had not been seen before by this user, and was frustrated by your error messages that lacked critical diagnostic information.  

Dismissing complaints by "new" users as "niche?"   We get that you are experts in the product so you don't need detailed error messages. A software base which is an encouraging ecosystem is supportive of users seeing new (to them) messages by helping to create clear error messages so you can grow your user  base through non-frustrating, non-vague warnings. 

This patch: 

* Addresses a key issue that users new to this part of the functionality are helped by clear and descriptive error messages which includes in the report the name of the database being tested.

* Fixes vagueness in the wording of error messages. 

* doesn't change any base functionality.  

Would ask you re-view this closing and dismissal of this PR by a new user who finds your error message one that creates an unnecessary barrier to new user adoption.

Issue #6187, opened 2025-01-03
mysql connextions overwrite/ignore global settings


MakerTim:
**Describe the bug**
From 11.4.4-MariaDB and onwards SSL checking is enabled by default.
To disable this behavior/revert the old behavior, one of its ways is to specify it in a configuration file.
Something like: `/etc/mysql/my.cnf`
```
[client-mariadb]
disable-ssl-verify-server-cert

```
This works like a charm, but not for drush, because they use the argument `--defaults-file` what specifies a file to load configuration from but also disables all other “default” configuration files.

Could we switch and use `--defaults-extra-file` instead? This still allows connecting with credentials to the database by file and has the least side effects.
But still loads other default files in.
![manual for mariadb](https://github.com/user-attachments/assets/5cc4772c-de4a-45ac-be74-045c2f3206e2)

**To Reproduce**
create a configuration file `/etc/mysql/my.cnf` with the following contents
```
[client-mariadb]
disable-ssl-verify-server-cert

```
Connect to a database that has SSL disabled or an invalid certificate.

**Expected behavior**
A successfull connection to mysql, eg
```
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MySQL connection id is ...
```

**Actual behavior**
What happened instead?
Got the following error: `ERROR 2026 (HY000): TLS/SSL error: Certificate verification failure: The certificate is NOT trusted.`

**Workaround**
Adding an argument `--extra=--skip-ssl` to all sql-based commands does work as well, but is not desirable.
Also, for dump `--extra-dump=--skip-ssl` is needed.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 13.3.3.0
| Drupal version? | 10.x
| PHP version     | 8.x
| OS?             | Linux

**Additional information**
Kinda related, or I did find this issue while working on https://github.com/drush-ops/drush/issues/6183
All started when mysql-client updated on the server.

MakerTim:
![image](https://github.com/user-attachments/assets/d0d9ec1f-adea-45e3-b7c5-0381820cb8af)

weitzman:
See the link in the PR where we tried this and ultimately decided against.

hyrsky:
In case anyone else runs into this issue, a [`drush.yml`](https://www.drush.org/13.x/using-drush-configuration/) with

```yml
command:
  sql:
    query:
      options:
        extra: "--skip-ssl"
    cli:
      options:
        extra: "--skip-ssl"
[...]

PR #6188, opened 2025-01-03
Fix mysql default configuration files being ignored (#6187)

MakerTim:
See #6187 for context.

Instead of using --defaults-file we will use --defaults-extra-file
This preserves the other configuration files that may be required for a connection.

See 
* https://mariadb.com/kb/en/mariadb-command-line-client/#option-files
* https://dev.mysql.com/doc/refman/8.4/en/option-file-options.html
image


weitzman:
See #4334

Issue #6183, opened 2024-12-17
D7 installation fails with Drush 8.x: Incorrect table name 'mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead


csandanov:
**Describe the bug**
Drupal 7 installation fails when using latest stable mariadb-client (Alpine 3.21)

**To Reproduce**
- Clone D7 from https://github.com/drupal-composer/drupal-project.git
- Install via Drush `drush si`

**Expected behavior**
Installation completes without errors

**Actual behavior**
Error thrown

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 8.5.0
| Drupal version? | 7.x
| PHP version     | 8.x
| OS?             | Linux

**Additional information**

```shell
Executing: mysql --defaults-extra-file=/tmp/drush_fdoGcc --database=drupal --host=mariadb --silent  < /tmp/drush_jbNBcc > /dev/null
Notice: y: SHOW TABLES; [0.14 sec, 7.46 MB]                             [notice]
Executing: mysql --defaults-extra-file=/tmp/drush_hDkIad --database=drupal --host=mariadb --silent  < /tmp/drush_JLiEad
  mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
Notice: y: DROP TABLE `mysql: Deprecated program name. It will be       [notice]
removed in a future release, use '/usr/bin/mariadb' instead` [0.17
sec, 7.47 MB]
Executing: mysql --defaults-extra-file=/tmp/drush_fLmGfg --database=drupal --host=mariadb --silent  < /tmp/drush_gMdBfg
Error: to create database: mysql: Deprecated program name. It will be    [error]
removed in a future release, use &#039;/usr/bin/mariadb&#039;
instead--------------DROP TABLE `mysql: Deprecated program name. It
will be removed in a future release, use &#039;/usr/bin/mariadb&#039;
instead`--------------ERROR 1103 (42000) at line 1: Incorrect table
name &#039;mysql: Deprecated program name. It will be removed in a
future release, use &#039;/usr/bin/mariadb&#039; inste&#039; [0.2
sec, 7.47 MB]
Returned from hook drush_core_pre_site_install [0.2 sec, 7.47 MB]        [debug]
Notice: dispatch complete [0.2 sec, 7.42 MB]                            [notice]
  mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead
  --------------
[...]

PR #6189, opened 2025-01-03
Fix mysql via mariadb warns about being deprecated (#6183)

MakerTim:
See #6183 for context

In short, this PR will prevent the following message while being fully compatible with mysql as a dropin replacement

"mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead"
When the mysql-binary is removed, this code-change will still be fully working with mariadb.

### DIFF
--- a/src/Sql/SqlMysql.php
+++ b/src/Sql/SqlMysql.php
@@ -4,10 +4,14 @@ declare(strict_types=1);
 
 namespace Drush\Sql;
 
+use Drush\Drush;
+use Drush\Exec\ExecTrait;
 use PDO;
 
 class SqlMysql extends SqlBase
 {
+    use ExecTrait;
+
     protected string $version;
 
     public string $queryExtra = '-A';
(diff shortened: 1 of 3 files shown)


weitzman:
https://github.com/drush-ops/drush/issues/6183 is for Drupal 7 which requires Drush 8.x. This PR is for Drush 13. Are you sure this is needed in Drush 13?

MakerTim:
I understand that the issue has the starting direction of Drupal-7.
But the underlying issue is that mariadb-client throws this warning.

I can make a new issue if you like that more, but since both cases inside issue #6183 where SSL issues with a warning, which is fixed by this pr, This is to stop red-harrings and stop the warning from comming.

greg-1-anderson:
Seems like this will be needed for anyone using the most recent versions of the mariadb client. I'm not sure which version introduced the behavior described in the above-referenced issue.

MakerTim:
It seems introduced in 11.4.x, we encounter it in 11.4.4

src/Sql/SqlMysql.php
```diff
@@ -17,15 +21,19 @@ class SqlMysql extends SqlBase
      */
     public static function make(array $dbSpec, array $options)
     {
-        // First get a MySQL instance
-        $instance = new self($dbSpec, $options);
+        // If the mysql version reports that it is MariaDB, use MariaDB as client.
+        $process = Drush::shell('mysql --version');
+        $process->setSimulated(false);
+        $process->run();
+        if ((!$process->isSuccessful() || strpos($process->getOutput(), 'MariaDB') !== false) && self::programExists('mariadb')) {
```
weitzman: I can't figure out what the double parens do here. Did you mean to surround so that the `||` condition is more clearly evaluated? Also, lets keep using str_contains() and not strpos()

weitzman: Also needs code style fix

`PSR12.Classes.OpeningBraceSpace.Found at /home/wodby/drush/src/Sql/SqlMariaDB.php (8:1)`

MakerTim: `||` and `&&` ordering is not always clear to read and changes in every language, so my nature makes them explicit so when im changing languages there is no question if the order is correct

PR #6191, opened 2025-01-06
Add Blackfire.io to credits

DieterHolvoet:
Blackfire.io is granting us a free open source license in return of mentioning them in the credits.


DieterHolvoet:
Can we see a preview of what's that going to look like? The logo might need to be resized.

DieterHolvoet:
I resized the Blackfire.io logo to be as high as the PhpStorm logo.

PR #6197, opened 2025-01-06
Allow config.storage service to be decorated

### LINKED ISSUE #6196: ConfgCommands using an incorrect typehint
**Describe the bug**
\Drush\Commands\config\ConfigCommands typehints on `protected StorageCacheInterface $configStorage`  but really want is required there is a \Drupal\Core\Config\StorageInterface - this approach was probably taken because there are multiple services in core.services.yml that satisfy \Drupal\Core\Config\StorageInterface but only one that has \Drupal\Core\Config\StorageCacheInterface. But this means that config commands stop working if you decorate config.storage and do not implement StorageCacheInterface.

**To Reproduce**
See https://www.drupal.org/project/wse/issues/3497004

**Expected behavior**
config commands are expected to work when config.storage is decorated

**Actual behavior**
config commands do not work when config.storage is decorated

**Workaround**
WSE config can make it's decorated class implement a no-op StorageCacheInterface but the code in Drush is wrong.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 13.x
| Drupal version? | 11.x
| PHP version     | 8.x

**Additional information**

alexpott:
Fixes #6196


weitzman:
Thanks. I'm hoping @bircher can review this one.

weitzman:
The test failure is unrelated. Curious folks should read https://drupal.slack.com/archives/C62H9CWQM/p1736182163381409

bircher:
As I wrote on slack, I am :+1: for this. I would go even further and also remove the `hasImportTransformer` and `getImportTransformer` methods but I don't want to scope creep this issue unnecessarily if @alexpott and @weitzman prefer to keep it as is.

bircher (approved):

PR #6203, opened 2025-01-07
Use drupal 11.x branch in highest tests (was 11.0.x)

weitzman:


weitzman:
I pushed a change for us using minimal everywhere instead of allowing extension scan of tests dir. Still lots of failures so unsure if thats a good direction or not. Its probably slower.

claudiu-cristea:
Trying to fix in #6210

claudiu-cristea:
Reduced to only ImageTest failure in _highest._ No idea on how to fix it.

Hints: Seems that `StreamWrapperManager::$wrappers` doesn't get populated on the first run, so "public" stream wrapper is not discovered. If I run the test the 2nd time it passes. Running manually also passes

It might caused by https://www.drupal.org/node/3414687 ?

Issue #4899, opened 2021-11-10
drush deploy disregards php's memory_limit parameter


jcnventura:
**Describe the bug**
Deploying a site using `drush deploy` is showing me out-of-memory errors even though I'm running it with the following command line:
`php -d memory_limit=-1 vendor/bin/drush deploy`.

**To Reproduce**
Execute `php -d memory_limit=-1 vendor/bin/drush deploy`.

**Expected behavior**
Drush would not run out of memory, and finish deploying the site.

**Actual behavior**
Drush ran out of memory at PHP's configured memory limit.

**Workaround**
Instead of running `drush deploy`, run the following set of commands in sequence:
```
php -d memory_limit=-1 vendor/bin/drush updatedb --no-cache-clear
php -d memory_limit=-1 vendor/bin/drush cache:rebuild
php -d memory_limit=-1 vendor/bin/drush config:import
php -d memory_limit=-1 vendor/bin/drush cache:rebuild
php -d memory_limit=-1 vendor/bin/drush deploy:hook
```

However, this defeats the purpose of having the deploy command. Also it transfers the burden of what exactly the deploy command does back to my CI deploy script.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 10.3.6
| Drupal version? | 9.2.7
| PHP version     | 7.4.25
| OS?             | Linux

**Additional information**
I believe the problem here is that even though the drush deploy command is indeed running in an unlimited memory environment, drush is starting new scripts that do not inherit the memory_limit setting from the deploy command.

weitzman:
Correct. Thats by design. Drush no longer tries to manage your php environment (nor your $PATH). You have to do that before you call drush. The approach that we suggest is at https://github.com/drush-ops/drush/issues/3294#issuecomment-618962852. If you are running `deploy` on a remote machine, you can configure env variables like `PHPRC` and `PATH` in your alias. See `#env-vars` key at https://www.drush.org/latest/site-aliases/#additional-site-alias-options

Also, you should consider a real value for memory_limit instead of disabling it. See https://github.com/drush-ops/drush/issues/1947

PR #6204, opened 2025-01-07
Add queue:run-all command

### LINKED ISSUE #6062: Run queue tasks continuously
We use Drupal's [Queue](https://api.drupal.org/api/drupal/core%21core.api.php/group/queue/11.x) feature to handle background tasks. However, many of these tasks are time sensitive and need to be executed in the background as soon as possible.

Drush doesn't provide a way to run a specific queue as a daemon, it can only be invoked on a schedule.

I would really like a way to run Drupal queue as a long running task in the foreground. Then I can use systemd, docker, or k8s to keep it running.

Maybe a `queue:watch` command that is like:
https://github.com/drush-ops/drush/blob/151b70439cdacda5b95546ca853f271d0fe981cf/src/Commands/core/QueueCommands.php#L60-L113
but that is running in an infinite loop (with a `usleep()`?) until the time limit or item limit is reached? Basically, it should keep running even if there are zero items in the queue right now. 

It might be nice to also limit the number of iterations by the amount of memory (i.e. `memory_get_usage()`) which would allow the script to be kept alive for as long as possible.

* Running in a cron tab, but the most frequent you can set that to is once a minute.
* Using a loop in a shell script that runs `drush queue:run`, but then we are doing the work to bootstrap Drupal and connect to the database on every iteration.

Related:
* [Advanced Queue Runner](https://www.drupal.org/project/advancedqueue_runner)

**Additional context**
N/A

DieterHolvoet:
Adds a `queue:run-all` command that does the same as `queue:run`, but for all available queues instead of a specific one. It takes the same options, but without the argument. The limit options count across all queues instead of a specific queue.

Also adds a `--memory-limit` option that takes either a number that ends with g/m/k (the format of the `memory_limit` ini setting) or a percentage. The command will run until that amount of memory is consumed. This option is added to both `queue:run` and `queue:run-all`.

Also adds a `--daemon` option that keeps the command running until `items-limit`, `lease-limit` or `memory-limit` is reached - if defined. Else, it keeps going indefinitely. This option is added to both `queue:run` and `queue:run-all`.

Fixes #6062.

### DIFF
--- a/src/Commands/core/QueueCommands.php
+++ b/src/Commands/core/QueueCommands.php
@@ -61,28 +61,30 @@ final class QueueCommands extends DrushCommands
     #[CLI\Complete(method_name_or_callable: 'queueComplete')]
     public function run(string $name, $options = ['time-limit' => self::REQ, 'items-limit' => self::REQ, 'lease-time' => self::REQ]): void
     {
-        $time_limit = (int) $options['time-limit'];
-        $items_limit = (int) $options['items-limit'];
-        $start = microtime(true);
         $worker = $this->getWorkerManager()->createInstance($name);
         $info = $this->getWorkerManager()->getDefinition($name);
-        $end = time() + $time_limit;
         $queue = $this->getQueue($name);
-        $count = 0;
-        $remaining = $time_limit;
+
+        $time_limit = (int) $options['time-limit'];
+        $start = microtime(true);
+        $end = time() + $time_limit;
+        $time_remaining = $time_limit;
         $lease_time = $options['lease-time'] ?? $info['cron']['time'] ?? 30;
+        $items_count = 0;
 
         if ($queue instanceof QueueGarbageCollectionInterface) {
             $queue->garbageCollection();
         }
 
-        while ((!$time_limit || $remaining > 0) && (!$items_limit || $count < $items_limit) && ($item = $queue->claimItem($lease_time))) {
-            $this->doRun($queue, $worker, $name, $item, $count);
-            $remaining = $end - time();
+        while (!$this->hasReachedLimit($options, $items_count, $time_remaining) && ($item = $queue->claimItem($lease_time))) {
+            if ($this->processItem($queue, $worker, $name, $item)) {
+                $items_count++;
+            }
+            $time_remaining = $end - time();
         }
 
         $elapsed = microtime(true) - $start;
-        $this->logger()->success(dt('Processed @count items from the @name queue in @elapsed sec.', ['@count' => $count, '@name' => $name, '@elapsed' => round($elapsed, 2)]));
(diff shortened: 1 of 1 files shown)


weitzman:
Thanks for the POC. I'm still hesitant. maybe we doa refactor but keep the new command in contrib. Logging will be an issue. Its not clear which queue some log messages and errors are referring to.

DieterHolvoet:
I did some work on improving the logging:
- every new queue where an item is claimed is announced (giving context to any errors that may follow)
- every queue where an item has been claimed gets a summary at the end, the same one as `queue:run`
- at the very end of the command execution, a global summary is shown (amount of items and time elapsed)

CleanShot 2025-01-07 at 16 14 02@2x

To be clear, _'Processed X items'_ does not include failed/skipped items.

DieterHolvoet:
I added the `--daemon` option as well, will work on the `--memory-limit` option now. If not for Drush core, for contrib.

DieterHolvoet:
I added the `--memory-limit` option as well. The MR is all ready now, I just started using the new command in a Supervisor config on a production site as test and it seems to be working so far.

If we decided not to merge this into Drush core, I will probably start a new module with the `queue:run-all` command. I don't think I'll be able to add the changes to the `queue:run` command (the new `--daemon` and `--memory-limit`) in contrib though.

That said, I am still fan of the idea of adding this to Drush core. It doesn't add a lot of code or complexity and it's very useful.

weitzman:
1. I dont especially want to do more memory checking in Drush. We've done that in the migrate commands and it confuses people. See https://github.com/drush-ops/drush/issues/4899#issuecomment-965272931
2. Why do you think the item-limit should span across queues? That seems weird to me.

> Keep the command running indefinitely, or until one of the chosen limits has been reached

If the memory limit is reached, won't folks expect the command to restart itself? Should the above statement only apply to items and time limits?

DieterHolvoet:
> Why do you think the item-limit should span across queues? That seems weird to me.

I don't really see the point of this option in general. In what case would you use it, except to just limit the amount of work that's being done. Would it make more sense to leave it out of `queue:run-all` then?

> If the memory limit is reached, won't folks expect the command to restart itself?

In my use case, software like Supervisor takes care of restarting automatically, but I guess we could add a `--restart` option that restarts Drush in a subprocess?

> Should the above statement only apply to items and time limits?

That wasn't my idea. Why?

DieterHolvoet:
If you're really not sold on the idea of adding the new command and new options, no problem, just let me know and I'll add it to a contrib module. I still wouldn't mind maintaining this in Drush core though. What do you think?

weitzman:
Let's start with this in Contrib. Feel free to refactor our existing commands so that building on them is easier. Also feel free to link to the contrib commands from the description of migrate:run command

DieterHolvoet:
I moved the `queue:run-all` command to a Drupal module: https://www.drupal.org/project/drush_queue_run_all
I discarded the new `queue:run` options for now.

PR #6206, opened 2025-01-09
Typehint variable in batch query to integer

ddaffie:
Updated the query on the batch table to typehint the variable $id to an integer value. This is necessary for working with MongoDB. For MongoDB a condition variable needs to the right type.


ddaffie:
I have talked to @moshe on Drupal Slack about this and he replied with: "looks good. no test needed. simple pr."

Issue #6207, opened 2025-01-10
Add the fix for PR 5869 to the 12.x branch


acha5066:
**Describe the bug**
I'm experiencing the same bug described in #5838 and fixed in #5869. This was fixing a bug introduced on the 12.x branch but the fix was merged only into the 13.x branch. 

**To Reproduce**
N/A

**Expected behavior**
The fix should also be backported to the 12.x branch seeing as it is a bugfix and not a new feature.

**Actual behavior**
N/A

**Workaround**
N/A

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.x
| Drupal version? | 10.3
| PHP version     | 8.2
| OS?             | Linux

**Additional information**
N/A

PR #6208, opened 2025-01-10
 Add the fix for PR 5869 to the 12.x branch

acha5066:
See #6207


weitzman:
Drush 12 is in security fix only state. Please upgrade to 13.

PR #6209, opened 2025-01-12
Allow "drush pm:install" to use composer namespaces in module names (e.g. "drush en drupal/paragraphs")

camya:
This PR makes it more convenient to install/uninstall modules after we have added them with **composer require**.

**Example:**

Install and activate the "Paragraph" module via composer / drush:

```
composer require 'drupal/paragraphs'

drush en drupal/paragraphs
```

The change here is that it's now allowed to use `drush en paragraphs` or `drush en drupal/paragraphs` to install the "Paragraphs" module.

Internally, drush still uses the module name "paragraphs". The namespace part will be ignored.

**Excerpt from the documentation for the normalizeModuleNames method:** In pm:install/pm:unistall it is possible to pass single module names, e.g. "views_ui", or module names with composer namespaces, e.g. "drupal/views_ui". Internally, drush will only process single word module names, e.g. views_ui.

### DIFF
--- a/src/Commands/pm/PmCommands.php
+++ b/src/Commands/pm/PmCommands.php
@@ -83,6 +83,7 @@ final class PmCommands extends DrushCommands
     public function install(array $modules): void
     {
         $modules = StringUtils::csvToArray($modules);
+        $modules = $this->normalizeModuleNames($modules);
         $todo = $this->addInstallDependencies($modules);
         $todo_str = ['!list' => implode(', ', $todo)];
         if ($todo === []) {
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
It seems overly flexible to allow `drush en anything/paragaphs` to be equivalent to `drush en paragraphs`. I might be +1 on allowing `drush en drupal/paragraphs` if we had some way of validating that the paragraphs module was installed from the Composer package.

Note, however, that a Composer package might contain multiple modules. The benefit here is fuzzy, and I'm not sure that the ambiguities that it introduces are worth it.

weitzman:
This is not drush's problem to solve.

PR #6210, opened 2025-01-14
Symlink the testing profile to be treated as a regular one

claudiu-cristea:


tests/unish/UnishTestCase.php
```diff
@@ -671,14 +672,25 @@ protected function installSut($uri = self::INTEGRATION_TEST_ENV, $optionsFromTes
...
+        DeprecationHelper::backwardsCompatibleCall(
+            \Drupal::VERSION,
+            '11.1.0',
+            // Make the 'testing' profile available as a regular profile to
+            // avoid discovery of all testing extensions.
+            // @see https://www.drupal.org/node/3490626
+            fn() => @symlink('tests/testing', "$root/core/profiles/testing"),
+            fn() => null,
+        );
+
```
weitzman: I wa thinking about a symlink too. Seems fine to me. We need to only create the symlink if it does not exist already. Thats needed when running tests locally. I'm fine with not using DeprecationHelper if that makes it any easier.

claudiu-cristea: I've muted with "@" to avoid the error if symlink already exists

Issue #5912, opened 2024-03-15
Allow using rsync from Docker container on the server


back-2-95:
**Is your feature request related to a problem? Please describe.**
I would like to be able to rsync files and sql-syncs from a virtual server which has Drupal running as Docker container.

**Describe the solution you'd like**

```yaml
dev:
  host: somehost
  user: deployment
  uri: https://some-site.tld
  paths:
    drush-script: '/usr/local/bin/drush-script'
    rsync-script: '/usr/local/bin/rsync-script'
```

/usr/local/bin/drush-script: (this works):

```shell
#!/bin/bash

docker exec drupal-container drush --ansi $@
```

/usr/local/bin/rsync-script:

```shell
#!/bin/bash

docker exec drupal-container drush --ansi core:rsync $@
```

**Describe alternatives you've considered**

Bypassing whole Drush and sync files other way.

weitzman:
There was some discussion of this at https://github.com/drush-ops/drush/pull/3835. The transports for docker and kube are available to site aliases. But getting sql:sync to work is not implemented yet, or even designed. You can wire together the pieces though, like sql:sync does.

Also see https://github.com/drush-ops/drush/blob/577edcb72344e7b4789237feed64a8037ac93071/examples/Commands/SyncViaHttpCommands.php. You could write a similar rsync replacement if http is not acceptable - that would let you continue to use sql:sync

PR #6213, opened 2025-01-20
Add remote docker-compose support with docker host (-H) for sql:sync and rsync

christianwiedemann:


weitzman:
For some background on this topic see https://github.com/drush-ops/drush/issues/5912 and its linked issue. 

Unfortunately, too much of the docker details in this PR are bleeding into Drush. The goal is for all of that to be attracted into transports in the [consolidation/site-process](https://github.com/consolidation/site-process) project. I'm aware that this implies that sql:sync and rsync: may never work as that abstraction isn't enough. This is unfortunate, but necessary for manageable maintenance of Drush project.

christianwiedemann:
Hi @weitzman, 
I think it would be great to have rsync and sql:sync support for docker compose and kubernetes. There are a lot of docker compose environments out there and installing an SSH container is not always possible and even yes it is a pain. 
The same for kubernetes. 

So try it with a new idea. :)
1. we can make extend the TransportInterface to handle transport specific adjustments of command and make it easy to implement own transports.  (I found ProcessManager::addTransports which adds transports hard coded. Is there a way right now to register own transports? )

Something like:

```
SshTransport"InDrushPackage"  implements RsyncCommandsAlter, SqlSyncCommandsAlter { 
... 
public function rsyncTransportArgs(DrushCommands $command) {
            $ssh_options = $command->getConfig()->get('ssh.options', '');
            $parameters[] = Escape::shellArg($command->sourceEvaluatedPath->fullyQualifiedPathPreservingTrailingSlash());
            $parameters[] = Escape::shellArg($command->targetEvaluatedPath->fullyQualifiedPath());
           return 'ssh $ssh_options'" . ' ' . implode(' ', array_filter($parameters)';
  }
... 
```
Inside rsync:
 ```
public function rsync($source, $target, array $extra, $options = ['exclude-paths' => self::REQ, 'include-paths' => self::REQ, 'mode' => 'akz']): void
    {
  $transport = $this->getConfig()->getTransport()
  if (transport instanceof RsyncCommandsAlter) {
     $rsync_args = $transport->rsyncTransportArgs($this);
  }
  } 
```

Or we use some kind of events?

weitzman:
It still looks messy to me, but I'll accept it if @greg-1-anderson agrees with the direction. [I pinged him in #drush](https://drupal.slack.com/archives/C62H9CWQM/p1737552101028719) 

Adding own transports is not possible yet. Requested at https://github.com/consolidation/site-process/issues/52

PR #6216, opened 2025-01-23
Make the url logged in field:create a clickable link

DieterHolvoet:
Since we started using clickable links in the `drush en` command, I thought we could use it in `drush field:create` too.


DieterHolvoet:
@weitzman 

> Failed asserting that ' [success] Successfully created field 'field_test3' on unish_article type with bundle 'alpha'
> [success] Further customisation can be done through the edit form.
> [ASCII](length: 162) contains "http://dev/admin/structure/unish_article_types/manage/alpha/fields/unish_article.alpha.field_test3" [ASCII](length: 98).

Looks like those kind of terminal links aren't part of the error output and I can't seem to figure out how to get it. Can we remove that line?

weitzman:
Yes I think we can adjust the expectation or remove it - whatever you think.

PR #6233, opened 2025-02-18
Fix #6232. Make sure redispatchOptions() are used in deploy:hook

### LINKED ISSUE #6232: drush --xdebug [command] hangs forever for commands executing other commands
**Describe the bug**

When executing a Drush command with XDEBUG enabled for debugging the command using `drush --xdebug`, the main process hangs forever waiting for the sub-processes to terminate.

**To Reproduce**
Execute a Drush command which executes other Drush commands, i. e. `drush deploy:hook` which executes the `deploy:hook-status` command and eventually others.

* Create a test deploy hook which does nothing but returning a success message.
* Set a breakpoint on the first line of the deploy hook as well as of the first meaningful line of the Drush script (`vendor/drush/drush/drush.php`), i. e. on the line `$autoloadFile = FALSE;`.
* Enable Xdebug and start listening.
* Execute `drush --xdebug deploy:hook`

**Variant:** Create a deploy hook using Batch API, then the debugger should stop for every batch execution in the deploy hook as long as `$sandbox['#finished'] !== 1`.

**Expected behavior**
Expect the debugger to stop at the following breakpoints: 2 x Drush script (for `deploy:hook` and `deploy:hook-status`), deploy hook.

**Actual behavior**
Debugger only stops once in the Drush script and then hangs forever.

**Workaround**
There is no other way of debugging Drush commands with Xdebug.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 13.3.3.0
| Drupal version? | 10.4.2
| PHP version     | 8.3.4
| OS?             | ddev/ddev-webserver:v1.24.1 on Ubuntu 22.04.5 LTS (WSL2)

**Additional information**
The options the root Drush command is called with (i. e. `--xdebug`) are not passed on to the sub-processes (including Batch API sub-processes). I tried this but it doesn't seem enough to work, maybe because the interactive shell of the sub-processes is disabled?

weitzman:
This makes sure that --xdebug gets passed to subprocesses. I'm going to audit the codebase to see if there are more similar omissions. In the meanwhile, folks can enable xdebug via env variable instead of --xdebug

### DIFF
--- a/src/Commands/core/DeployHookCommands.php
+++ b/src/Commands/core/DeployHookCommands.php
@@ -111,7 +111,7 @@ final class DeployHookCommands extends DrushCommands
             return self::EXIT_SUCCESS;
         }
 
-        $process = $this->processManager()->drush($this->siteAliasManager->getSelf(), self::HOOK_STATUS);
+        $process = $this->processManager()->drush($this->siteAliasManager->getSelf(), self::HOOK_STATUS, [], Drush::redispatchOptions());
         $process->mustRun();
         $this->output()->writeln($process->getOutput());
 


weitzman:
The other instances of this might be deliberate omittions of redispatchOptions so lets merge just this one here.

PR #6234, opened 2025-02-18
fix(php): Fix PHP 8.4 deprecations for nullable types

klausi:
I'm working on upgrading from PHP 8.1 to 8.4. I cannot use Drush 13 yet because that requires PHP 8.2. I want to upgrade as many dependencies as possible before jumping to PHP 8.4, so I need Drush 12 to run on both PHP 8.1 and 8.4.

The changes were created automatically with
```
~/workspace/php-cs-fixer/vendor/bin/php-cs-fixer fix . --rules=nullable_type_declaration_for_default_null_value
```


klausi:
Uploading stable patch file for composer patches

[drush-6234.patch.txt](https://github.com/user-attachments/files/18848207/drush-6234.patch.txt)

klausi:
Thanks! Can you also make a Drush 12 release?

PR #6243, opened 2025-02-28
Allow cache-rebuild to fix more things

alexpott:
If we add this annotation the cache rebuild command becomes even more useful and can recover from more terminal issues.

For example if you change the Drupal core service name cache.backend.database to cache.backend.databas and do a drush cr you'll get an error. On HEAD if you go back and fix this and then run drush cr again it'll still be broken. With this change drush cr is capable of fixing the site.

### DIFF
--- a/src/Commands/core/CacheRebuildCommands.php
+++ b/src/Commands/core/CacheRebuildCommands.php
@@ -15,6 +15,7 @@ use Drush\Commands\DrushCommands;
 use Drush\Drupal\DrushLoggerServiceProvider;
 use Drush\Drupal\Migrate\MigrateRunnerServiceProvider;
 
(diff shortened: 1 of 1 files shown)


alexpott:
Drush cr not rebuilding the container has been observed in https://www.drupal.org/project/entity_usage/issues/3509039 for example.

weitzman:
Nice find. This certainly fixes the scenario described in the OP. Otherwise, the `CacheRebuildsCommands` class doesn't get instantiated until FULL bootstrap. I'm surprised this hasn't come up earlier.

heddn:
I suspect this happens all the time but those in the known just truncated the cache tables in the DB manually. Or at least that's what I did. This will be a great fix.

PR #6246, opened 2025-03-06
Bump laravel/prompts to v0.3.5

jnoordsij:
Bump laravel/prompts to the latest version, to allow installation of this package alongside the latest version of the Laravel/Illuminate components, which require this version.

See also https://github.com/laravel/prompts/releases/tag/v0.3.5

## Notes
- I have not been able to test too many things yet, but by the looks of things this new version does not introduce any relevant breaking changes
- I've opted for a simple bump; if a version union like `^0.1.21 || ^0.3.5` is more appropriate for broader compatibility, let me know!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
     "dflydev/dot-access-data": "^3.0.2",
     "grasmash/yaml-cli": "^3.1",
     "guzzlehttp/guzzle": "^7.0",
-    "laravel/prompts": "^0.1.21",
+    "laravel/prompts": "^0.3.5",
     "league/container": "^4.2",
     "psy/psysh": "~0.12",
     "symfony/event-dispatcher": "^6 || ^7",
--- a/src-symfony-compatibility/v6/Style/DrushStyle.php
+++ b/src-symfony-compatibility/v6/Style/DrushStyle.php
@@ -3,7 +3,6 @@
 namespace Drush\Style;
 
 use Drush\Drush;
-use Illuminate\Support\Collection;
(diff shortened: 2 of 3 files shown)


jnoordsij:
It seems that https://github.com/laravel/prompts/pull/168 now requires adding a dependency on `illuminate/collections` to work with these.

weitzman:
The current constraint allows you to upgrade to any version of Prompts less than 1.0. I dont think this is strictly needed.

jnoordsij:
> The current constraint allows you to upgrade to any version of Prompts less than 1.0. I dont think this is strictly needed.

Actually it does not, as the 0 major is handled differently by Composer:

>  For pre-1.0 versions it also acts with safety in mind and treats `^0.3` as `>=0.3.0 <0.4.0` and `^0.0.3 as >=0.0.3 <0.0.4`.

See https://getcomposer.org/doc/articles/versions.md#caret-version-range

weitzman:
Thanks for teaching me that  0 major is handled differently by Composer.

Lets try not to depend on `laravel/collections`.  I see 3 usages of connect() in src/Commands/ConfiguresPrompts.php. I think those could be adapted by reviewing the model for this file - https://github.com/laravel/framework/blob/12.x/src/Illuminate/Console/Concerns/ConfiguresPrompts.php. I dont think we use laravel/collections outside of this file

weitzman:
Lgtm. Just needs a couple code style fixes. See test failures. Thank you.

jnoordsij:
Should be fixed now!

PR #6249, opened 2025-03-11
Add `--extra` option to `sql:create`, `sql:drop`, and `site:install`

arnested:
With recent versions of MariaDB, we need to add the `--extra=--skip_ssl` option in our development setup. It wasn't possible to do that with some Drush commands.

We discovered this while trying to do a site install using Drush.


tobybellwood (approved):
We've experienced the same issue, and tested this patch successfully.

Thanks for the great work - here's my community +1 on this :heart:

tobybellwood:
For reference, here is our drush.yml snippet with the necessary extra settings:

```
command:
  sql:
    cli:
      options:
        extra: '--disable-ssl'
    query:
      options:
        extra: '--disable-ssl'
    dump:
      options:
        extra: '--disable-ssl'
        extra-dump: '--disable-ssl'
    drop:
      options:
        extra: '--disable-ssl'
    create:
      options:
        extra: '--disable-ssl'
  site:
    install:
      options:
        extra: '--disable-ssl'
 ```

nicksantamaria:
I've successfully tested this patch too. LGTM!

PR #6260, opened 2025-03-26
Introduce a cache:warm command

### LINKED ISSUE #5724: Implement core's cache prewarm API, when it exists
When you clear caches, a certain amount of cache rebuilding happens during the cache clear (route rebuilding etc.), and a certain amount happens on the first new HTTP request to the site.

In https://www.drupal.org/project/drupal/issues/3257725 I'm trying to add a cache prewarm API to core using the PHP 8.1 Fibers feature.

This takes advantage of stampede situations, by moving certain long-running cache building operations that usually happen later in the request earlier, and picking one at random to run if something during bootstrap has called Fiber::suspend(). The idea is that if you get more than one cold cache request come in at the same time, they can build different caches for each other, instead of both trying to build the same ones in sequence.

I think we can go a step further with this in drush.

The patch on https://www.drupal.org/project/drupal/issues/3257725  adds a `cache_prewarmable` service tag - this lets any service define itself as prewarmable, using an interface with a single method - ::preWarm(). There's then a cache_prewarmer service which picks one of these to run and runs it (drush would probably need its own service though).

drush cr could, as well as clearing caches, get and execute them simultaneously in child processes. So for example if there are ten caches that take an average of 100ms to warm, they could be done in 100ms instead of one second.

The main example I've used on the issue so far is the views data cache. In a cold start situation, the views data cache triggers the rebuild of several other caches, this means it takes ~300ms overall. One of the cache builds it triggers is entity_field.manager's entity_field_map, which takes about 100ms.

By prewarming entity_field_map earlier in the request, the views data cache then only takes 200ms itself to build when we get there. If we slice out another nested cache from there to prewarm, we could have three things taking 100ms instead of one thing taking 300ms and run them all at the same time.

This applies to all the usual candidates - theme registry, element info, field and entity type caches, plugin discovery, library info.

Some caches (element info and etc.) rely on the active theme being set. We could either just ignore those, they will need to detect when the active theme isn't set anyway, or we could manually set the active theme to the default theme for the site and prewarm with that in place.

Making a cup of tea while I wait for the page to load.

**Additional context**

weitzman:
Fix #5724

The new command is automatically run during drush deploy. I think almost everyone will want that. Others can make own deploy-like command. I'm not inclined to add this to cache:rebuild since thats a low level command that shouldn't get even more complex IMO.


src/Commands/config/ConfigExportCommands.php
```diff
@@ -142,7 +142,7 @@ public function doExport($options, $destination_dir)
                 $this->logger()->notice($preamble . $preview);
             }
 
-            if (!$this->io()->confirm(dt('The .yml files in your export directory (!target) will be deleted and replaced with the active config.', ['!target' => $destination_dir]))) {
+            if (!$this->io()->confirm(dt('The .yml files in your export directory will be deleted and replaced with the active config.'), hint: dt('Target: !target', ['!target' => $destination_dir]))) {
```
weitzman, line 145: Unrelated change - a long directory was getting truncated

src/Commands/core/CacheWarmCommands.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Pre-warm all caches.
+     */
+    #[CLI\Command(name: self::WARM, aliases: ['warm', 'cache-warm'])]
+    #[CLI\Version(version: '13.5')]
+    public function warm(): void
+    {
+        $this->warmer->preWarmAllCaches();
```
catch56, line 37: We could possibly go a step further, and get the prewarmable services from the container, then start a new process to prewarm each one in parallel. There is not a nice API for this so could maybe be a follow-up.

PR #6269, opened 2025-04-09
Double escape \n in doc string

normanlolx:
Noticed `\Drupal ode\Entity\Node::load(1);` on https://www.drush.org/latest/commands/php_eval/


weitzman:
Nicely spotted. unfortunately, thats going to make the Usage not work if youn copy and paste from the help when viewed in the terminal - `drush help php:eval`. not sure how to proceed.

PR #6279, opened 2025-04-24
Refs: #6278 fix to use the root of the system as the working directory for diff commands.

### LINKED ISSUE #6278: drush config:export / config:import --diff does not work reliably
**Describe the bug**
using a non standard git checkout (such as with git worktree) will result in a non functional .git directory inside the container (this is not a problem in itself).
However what is a bug is that for some unknown reason the drush config --diff options use git to do the diffing (which is fine) and they do it in their own directory (even better) yet the command uses the project root to run from (which does not always have a full git directory available).
The result is that the git command fails and the diff is subsequently empty. (when using ddev as a development environment)

**To Reproduce**
use ddev , do a git worktree checkout of a drupal site, go to the worktree version of the project, change 1 config item and do drush config- --diff, there will be no diff result.

**Expected behavior**
normal diffing behaviour as with non git checkouts and normal git checkouts.

**Actual behavior**
git does not work and yields a silent error. and no diff is shown.

**Workaround**
change the directory the command is run in.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.5.3.0 , but present in the latest release too.
| Drupal version? | 10.4.5, but presetn in the latest release too.
| PHP version     | 8.1.32
| OS?             | Linux (DDEV 1.24.4)

**Additional information**
PR !6279

Sysosmaster-LimoenGroen:
- Fixes the wrong behaviour of running a diff on non-project root files while running git form the project-root, by changing the working directory for this command to be '/'.

### DIFF
--- a/src/Commands/config/ConfigCommands.php
+++ b/src/Commands/config/ConfigCommands.php
@@ -529,7 +529,7 @@ final class ConfigCommands extends DrushCommands implements StdinAwareInterface
             }
         }
         $args = array_merge($prefix, ['-u', $temp_destination_dir, $temp_source_dir]);
-        $process = Drush::process($args);
+        $process = Drush::process($args, "/");
         $process->run();
         return $process->getOutput();
     }


weitzman:
Please provide the commands needed to reproduce the bug. Thanks.

weitzman:
Please xplain whi running this command from the root directory is a fix. That dir does not have a .git subdir either.

Sysosmaster-LimoenGroen:
> Please xplain whi running this command from the root directory is a fix. That dir does not have a .git subdir either.
by NOT being in ANY git repository, the command works reliably and independently of any of the possible configs being used by the users Git. (for example, custom diff command scan also ruin the drush command right now).

as for a easty test plan:
- get a ddev project with drupal that is inside a git repo and has tracked drupal config (sync).
- create an branch `git branch drushgitconfigtest`
- use the git command `git worktree add ../brokendrushcompare drushgitconfigtest`
- go to the newly created worktree: `cd ../brokendrushcompare`
- change some config which is tracked.
- run the `ddev drush config:import --diff ` or `ddev drush config:export --diff` 
- Revel in the discovery there is no output of the diff whatsoever (it silently broke)

ill add it to the issue too, cause I though I had. Sorry for the missing info.

weitzman:
I made a different fix at #6339. Could you please check that it fixes your issue?  I wasn't able to reproduce it even following your steps. I added --no-index option instead of your root dir fix.

I also couldn't push to your branch due to missing perms so I had to open a new PR.

Sysosmaster-LimoenGroen:
@weitzman, unfortunetly git is persistent with trying to read the index files even when there not there.

(as an alternative, you could do the following to test with, this will mock it sufficiently for a test. but it would break the git repo itself

1. got to the git repo root.
2. execute the following command to rename the .git directory: `mv .git .git.stored`
3. create the worktree ,git file with: `echo "gitdir:  /home/sysosmaster/src/git/drupal-web/.git/worktrees/drupal.web-wc" > .git`
4. now typical git commands will not work due to the missing "root" of the git repo (which is outside the container).

To Restore:
1. `mv .git  .git-worktree_test`
2.  `mv .git.stored .git `
3. its now working again... most likely.
)

Currently I see the following behaviour:

`sysosmaster@drupal.web:/var/www/html/htdocs$ git diff --no-index --exit-code --color=always -u /tmp/drush_tmp_1755528324_68a33c8dc4990 /tmp/drush_tmp_1755528325_68a3dc850be76`
Yields: 
`fatal: not a git repository: /home/sysosmaster/src/git/drupal-web/.git/worktrees/drupal.web-wc`

Ergo, the core issue (wrongly assuming the root of the project is a full fledge git repo) is still wrong.

I have found the missing check that would allow edits (sorry for that, all watch for that in the future).

The Git version in the container is: `git version 2.39.5`.

PR #6280, opened 2025-04-27
#6628: Added a command to remove a deploy hook from the list of deployed hooks in the site State

harivansh0:


src/Commands/core/DeployHookCommands.php
```diff
@@ -283,4 +258,201 @@ public function markComplete(): int
...
+            ]);
+            return self::EXIT_SUCCESS;
+        }
+        // Remove the hook from the list.
+        $update_functions = array_filter($deployed_hooks, function ($function) use ($hook_name) {
+            return $function !== $hook_name;
+        });
+
+        // Update the deployed hook list.
+        \Drupal::service('keyvalue')->get('deploy_hook')->set('existing_updates', $update_functions);
```
MurzNN, line 345: We should have this service available as `$this->keyValue`.

harivansh0: Hi @MurzNN 
The DeployHookCommands class does not have a keyValue property. Although the getRegistry() function injects the service, the class extends UpdateRegistry, which defines keyValue as a protected property. Since it's protected and not accessible directly in the subclass, we cannot use it, without modifying the drupal core UpdateRegistry or use refelection.
Or I could using autowire to inject
`        #[Autowire(service: 'keyvalue')]
        protected KeyValueFactoryInterface $keyValueFactory,`

PR #6281, opened 2025-04-28
Use the search input for select lists with lots of options

DieterHolvoet:
Symfony Console has the possibility to filter option lists by default. This functionality was lost when we started using Laravel Prompts. Laravel Prompts does offer this functionality, but under the search input instead of the select input and the multisearch input instead of the multiselect input. I propose using the search inputs in cases where the option lists are (potentially) long, to improve UX.

### DIFF
--- a/src-symfony-compatibility/v6/Style/DrushStyle.php
+++ b/src-symfony-compatibility/v6/Style/DrushStyle.php
@@ -82,8 +82,13 @@ class DrushStyle extends SymfonyStyle
      * @param  array<int|string, string>  $options
      * @param  true|string  $required
      */
-    public function select(string $label, array $options, int|string|null $default = null, int $scroll = 10, ?\Closure $validate = null, string $hint = '', bool|string $required = true): int|string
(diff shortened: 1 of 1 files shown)


weitzman:
IMO, the search UX is more cluttered than the select UX. Lets only use search when we expect 20 or more options. So that rules out cache:clear among others.

DieterHolvoet:
Okay, I reverted for the cache/site install/topic commands. An idea for commands where the amount of options depends on the size of the site: dynamically call either search or select if the amount of options is higher or lower than 20. Like this:
```
$this->io()->{count($choices) > 20 ? 'search' : 'select'}(...)
```
Could be good for eg. the image commands. There can be a lot of image styles depending on the site configuration.
Could we take 10 options as the threshold? Since that's the amount of options that are visible by default, without scrolling.

weitzman:
Could we create a wrapper prompts that automatically switch between select/search and multiselect/multisearch. i kinda think that we should use 20 as the threshold and we experiment with showing 20 by default in select. I think select lists have better UX at lower item counts.

DieterHolvoet:
I simplified the PR by adding extra options to the `select` and `multiselect` methods, automatically enabling search when there are more than 20 options, while keeping the threshold configurable.

Tests are failing because packages.drupal.org is down, I'll run them again later.

weitzman (approved):
LGTM

PR #6287, opened 2025-05-22
Test Drush 8 on php 8.4

greg-1-anderson:


greg-1-anderson:
Tests are failing because a stderr message from mariadb is being mixed in with the program output.

PR #6291, opened 2025-06-10
Move CONTRIBUTING.md physically into docs

phenaproxima:
I am sorry to report that the presence of `docs/contribute/CONTRIBUTING.md`, which is a symlink, wreaks havoc on Drupal sites that are using Package Manager (i.e., Project Browser and Automatic Updates) on Windows. Composer Stager, which is part of Package Manager's plumbing, detects the symlink and throws an error because symlinks are not supported on Windows. It's surprisingly disruptive and we've encountered this specific problem, with this specific symlink, quite a few times in the last several years of developing Automatic Updates and Project Browser.

There are workarounds we could do, but the _cleanest_ fix here is for Drush to remove the symlink, if it's not really needed. I inquired in Slack and @weitzman told me that the symlink could probably be replaced by the actual document it references. So that's what this PR does.


phenaproxima:
To be clear, the problem here is not so much an incompatibility with _Windows_ as with Composer Stager, which completely disallows links on Windows for reasons explained in https://github.com/php-tuf/composer-stager/tree/develop/docs/preconditions#symlinks.

That's why the cleanest solution here is still to remove the symlink if possible, but if that's not acceptable we can find another workaround.

PR #6295, opened 2025-06-11
Fix 'Negated boolean expression is always false'

DieterHolvoet:


DieterHolvoet:
@weitzman this seems to fix it!

PR #6313, opened 2025-07-15
Add tty/interactive options to kubectl docs

froboy:
As per https://github.com/consolidation/site-process/blob/e7fafc40ebfddc1a5ee99ee66e5d186fc1bed4da/src/Transport/KubectlTransport.php#L41-L42, `tty` and `interactive` can be used to force the connection state.

This adds those options to the docs to other devs don't have to dive into the code to find them like I did.


froboy:
Test failures seem unrelated to docs, I think?

PR #6315, opened 2025-07-23
Explain usage of --fields option for user:create command

mlncn:
People are likely to want to set the value for user fields while creating a user, so it is good to be explicit that it is reporting-only by including a `--fields` example.


src/Commands/core/UserCommands.php
```diff
@@ -215,6 +215,7 @@ public function removeRole(string $role, string $names = '', $options = ['uid' =
     #[CLI\DefaultTableFields(fields: self::INF_DEFAULT_FIELDS)]
     #[CLI\FilterDefaultField(field: 'name')]
     #[CLI\Usage(name: "drush user:create newuser --mail='person@example.com' --password='letmein'", description: 'Create a new user account with the name newuser, the email address person@example.com, and the password letmein')]
+    #[CLI\Usage(name: "drush user:create anotheruser --fields=uuid,langcode", description: 'Create a new user account with the name anotheruser and report back the values of the UUID and Language code fields (note it is not possible to set field values)')]
```
weitzman, line 218: Love the motivation, just too many words for one usage :) 

Can you use multiple Usages to make your point? Or add a synopsis. Thats a longer form text that would go in a second paragraph in method description. That will be shown in --help.

Issue #6335, opened 2025-08-07
twig:compile: Surround the twig template loading by try-catch to avoid breaking the script


efpapado:
**Describe the bug**

On the twig:compile command `$this->twig->load($relative);` is called. If the template contains a non-existing twig function, an exception is thrown, and the script is terminated. This prevents other templates (which do not have this problem) from getting compiled. The script can break in a random step of the templates iteration, depending on the names of the modules and themes that are used.

A template that uses a non-existing function is a valid use case. Several base themes contain templates for known and widely used modules, such as commerce in my example. This is not wrong, since the template is not expected to be loaded and used if the module is not in place (and possible custom code is expected to follow naming conventions of the drupal universe). But if the twig:compile command is run, then it will try to compile these templates also, resulting to fail. Using a theme does not require installation of all the possible modules that it covers.

**To Reproduce**

- Install the bootstrap_barrio theme.
- Do NOT install the commerce module.
- Run the drush command.
- The command breaks as soon as the `themes/contrib/bootstrap_barrio/templates/commerce/order/commerce-order-add-list.html.twig` is attempted to be compiled.

**Expected behavior**

An error should be printed, but the iteration should continue with the rest of the available templates.

**Actual behavior**

The script broke, since the excetion thrown by twig is not caught.

**Workaround**

-

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 12.5.3, 13.6.1 (*)
| Drupal version? | 11.x, 10.x
| PHP version     | 8.2, 8.3
| OS?             | Mac, Linux

(*) Expected to happen at least in all versions after this commit : https://github.com/drush-ops/drush/pull/5705

**Additional information**

-

PR #6336, opened 2025-08-07
twig:compile: Surround the template loading with try-catch #6335

efpapado:
Resolves the issue https://github.com/drush-ops/drush/issues/6335


weitzman (approved):
LGTM. Just needs PHPCS fix. We use PSR-12, not Drupal standards.

PR #6354, opened 2025-09-12
Add support for Drupal ContentExportCommand

### LINKED ISSUE #6353: Directly add content:export command from Drupal 11.3
Drupal added a new command, `content:export` to the default content system to aid in recipe development.

In  #5997 , Drush added a wrapper command around Drupal's recipe command for easier development.  Could we do the same thing for `content:export`?

thejimbirch:
Would close #6353

I am not much of a developer, so hopefully I did this right.  I copied what was done in #5997


weitzman:
Merged into 13.x and 14.x - Thanks.

PR #6356, opened 2025-09-14
Resolves #6355: Verify ssl peer for command line should adhere to server config

### LINKED ISSUE #6355: SSL verify peer server and client consistency

Newer versions of the mariadb and mysql clients are turning on SSL by default. This is clear in the latest version of alpine which has switch from mysql to mariadb by default, and turned on SSL connections by default.

This can't (and shouldn't really) be turned off at the `my.cnf` for the server because of the way drush handles `--default-file=` and adding `--extras=` to everything is both prone to error can reduce site security if mishandled.

`PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` can be used to allow self-signed certificates to be used in local development infrastructure; and can be configured in Drupal `settings.php` for _only_ the connection that is used for localhost; without touching the configuration for hosted non-production or production environments.

In a similar way to other PDO options flowing through to the drush cli; I'd like to see this one flow through too.

https://github.com/drush-ops/drush/blob/13.x/src/Sql/SqlMysql.php#L111-L129

The closest match between the PDO required to use self-signed certificates and the command line client argument is `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` to `--disable-ssl-verify-server-cert` and has been tested in our infrastructure.

* We've considered having `--defaults-file` understand how to turn off SSL verify peer when it is built.
* We've considered turning it off at a server level with `my.cnf` but this is both overridden and would turn it off as the containers are deployed through various environments.
* We've considered turning off self-signed certificates but this causes problems in other build environments that connect over a non-local socket which `localhost` and `127.0.0.1` do by default.
* We've considered using a proper certificate authority to sign our certificates. But this causes security concerns because the private key would need to be somewhere semi-public.

**Additional context**

nterbogt:
Passing through the `POD::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` to the `mysql` or `mariadb` CLI; if it's set.


weitzman:
Seems reasonable to me. lets wait a couple days for more input.

nterbogt:
@weitzman Following up on this one.

weitzman (approved):

weitzman:
Merged into 14.x and 13.x. Thank you.

PR #6360, opened 2025-09-30
Get a FormatterManager if needed in FormatterTrait

weitzman:


weitzman:
Abandoned. Avoids many ` Readonly property Drush\Commands\sql\SqlDumpCommand::$formatterManager is assigned outside of the constructor.
         ✏️  Formatters/FormatterTrait.php
 ------ ---------------------------------`

PR #6383, opened 2025-10-04
Enable all options in example.drush.yml and make `sql:dump` work

gitressa:
## examples/example.drush.yml

Wouldn't it be better to streamline options in https://github.com/drush-ops/drush/blob/13.x/examples/example.drush.yml and make them all active? Supposedly, the user knows that commenting out an option is possible.

Also, the `sql:dump` > `structure-tables-key: common` example doesn't work without defining the actual tables under `sql` ("Non-options") so maybe include them as well?

## docs/using-drush-configuration.md

We could also add the `user:login` example on https://www.drush.org/13.x/using-drush-configuration
- https://github.com/drush-ops/drush/blob/13.x/docs/using-drush-configuration.md


weitzman:
I'm not too inclined to uncomment by default. I welcome other content changes

gitressa:
The out-commenting of the line has been restored.

gitressa:
Could the updated MR work?

examples/example.drush.yml
```diff
@@ -29,14 +29,24 @@ command:
...
 # Non-options.
+sql:
+  # List of tables whose *data* is skipped by the 'sql:dump' and 'sql:sync'
+  # commands when the "--structure-tables-key=common" option is provided.
+  structure-tables:
+    common:
+      - cache
+      - 'cache_*'
+      - 'watchdog'
+
```
weitzman:
```suggestion
```

weitzman: This is already documented in docs/using-drush-configuration.md. Lets not duplicate.

weitzman (commented):
Lets target 14.x, and omit one change as Suggested.

weitzman:
Also merged to 14.x

PR #6418, opened 2025-10-29
field-create: Get defaults from existing field

DieterHolvoet:
I changed it so default values are taken from an existing bundle field, which nowadays is the default behaviour for the re-use field form in the UI as well.


DieterHolvoet:
Remaining failures seem unrelated to this PR.

PR #6420, opened 2025-11-03
field-create: Remove unnecessary _none option from allowed formats list

DieterHolvoet:
I think it has become redundant since we have switched to `laravel/prompts`. Either way we have to remove it, since it causes issues when this option is selected, since `_none` is not an actual text format.


weitzman (approved):
Needs 14.x as well.

weitzman:
Merged and cherry picked to 14.x. Thanks.

PR #6431, opened 2025-11-10
Hooks are no more event listeners since Drupal 11.3

claudiu-cristea:


weitzman:
Thanks. Also cherry picked into 14.x

PR #6438, opened 2025-11-17
Install dependencies with theme installation.

### LINKED ISSUE #4248: Potential changes as it will soon be possible for themes to declare dependencies on modules.
After roughly a decade, this issue has been escalated to critical and may actually see completion in the near future.
https://www.drupal.org/project/drupal/issues/474684

In my most recent patch at the time of writing this (#206), I made some changes to `\Drupal\Core\Extension\ModuleHandler::buildModuleDependencies` that prevent uninstalling modules that enabled themes depend on.

fjgarlin:
Fixes https://github.com/drush-ops/drush/issues/4248

Use the second optional parameter defined in core: https://git.drupalcode.org/project/drupal/-/blob/11.x/core/lib/Drupal/Core/Extension/ThemeInstallerInterface.php#L36


weitzman:
For module install, we report what dependencies will be enabled and make user confirm when there are some that were not named on the cli. We dont have to do that for themes, but ideally we would, especially it if wouldnt introduce much new code because we have this for module install already.

fjgarlin:
I gave it a go checking the logic on `PmCommands.php`  and seeing that Drupal also has `ThemeExtensionList.php`. I haven't tested anything yet tho but it'd be great to see if this is what you had in mind.

fjgarlin:
I haven't tested it, but testing instructions are:
- Install a Drupal site via composer with `drush/drush` in it too.
- Bring via composer a theme with module dependencies (eg: https://www.drupal.org/project/gcds/releases/1.0.x-dev): `composer require 'drupal/gcds:1.0.x-dev@dev'`
- Try to install the theme via `drush -y theme:install gcds`

fjgarlin:
Funnily enough, the second parameter to install dependencies on Drupal core does not install module dependencies, it just reports them as uninstalled. With the last commit, it runs the module installer for the found modules.

I followed the above testing steps on a local install and it now works.
```
$ ddev drush theme:install gcds 
The following module(s) and themes(s) will be installed: gcds, twig_tools, twig_tweak

 ┌ Do you want to continue? ────────────────────────────────────┐
 │ Yes                                                          │
 └──────────────────────────────────────────────────────────────┘

 [success] Successfully installed theme: gcds
 
$ ddev drush pm-list --type=Module --status=enabled | grep twig
Other         Twig Tools (twig_tools)                            Enabled   2.1.0    
Other         Twig Tweak (twig_tweak)                            Enabled   3.4.1    
```

src/Commands/pm/ThemeCommands.php
```diff
@@ -55,4 +94,33 @@ public function uninstall(array $themes): void
...
+
+    public function addInstallDependencies($themes): array
+    {
+        $theme_data = $this->getExtensionListTheme()->reset()->getList();
+        $theme_list  = array_combine($themes, $themes);
+        if ($missing_themes = array_diff_key($theme_list, $theme_data)) {
+            // One or more of the given themes doesn't exist.
+            throw new MissingDependencyException(sprintf('Unable to install themes %s due to missing themes %s.', implode(', ', $theme_list), implode(', ', $missing_themes)));
+        }
+        $extension_config = $this->getConfigFactory()->getEditable('core.extension');
```
weitzman, line 115: Is this really the bets way to get the installed module list?

src/Commands/pm/ThemeCommands.php
```diff
@@ -55,4 +94,33 @@ public function uninstall(array $themes): void
         $this->getThemeInstaller()->uninstall($themes);
         $this->logger()->success(dt('Successfully uninstalled theme: !list', ['!list' => implode(', ', $themes)]));
     }
+
+    public function addInstallDependencies($themes): array
```
weitzman, line 107: Needs Doxygen

weitzman (changes requested):
This needs a test. I think we could borrow from tests/functional/PmInUnListInfoTest.php

weitzman:
Did you try using \Drush\Commands\pm\PmTrait to ThemeCommands and using its addInstallDependencies? Maybe it only needs minor tweaks to serve our needs here (at least for dependant modules).

src/Commands/pm/ThemeCommands.php
fjgarlin: Wasn't sure what the rule for this was as `PmCommands.php` (where I copied this from) didn't have it. I can add it.

src/Commands/pm/ThemeCommands.php
fjgarlin: Borrowed this from the sibling class `PmCommands.php`, see https://github.com/drush-ops/drush/blob/13.x/src/Commands/pm/PmCommands.php#L354-L355

src/Commands/pm/ThemeCommands.php
weitzman: I think you didnt have it because your PR is for 13.x. Ideally you target 14.x. I just changed the dfeault branch to 14.x so future PRs are not similarly confused.

fjgarlin: Oh, I didn't know there was a 14.x, I went straight with the default one. Thanks.

fjgarlin:
Will open an MR against 14.x

fjgarlin:
MR against 14.x https://github.com/drush-ops/drush/pull/6439

fjgarlin:
@weitzman as the 14.x cannot be cherry picked to 13.x, should I get this PR fully ready so we can have it as well in 13.x? Asking before I make the effort. I think the code in this PR is (mostly) ok, it's just the test that needs adapting.

Let mek now and if can get it fully ready for 13.x

weitzman:
Yes, I would merge this PR once it is passing.

fjgarlin:
@weitzman - any idea on why the `*-lowest-functional` are failing but the `*-highest-functional` are passing in this case? I see that it's attempting to install the dependencies but it's not doing it for whatever reason. Not sure how to debug.

weitzman:
I dont know offhand. I think its Drupal rejecting the install, not Drush.

weitzman:
To debug, you would try to run this:

I would use xdebug breakpoints to find the issue.

```
ddev start
ddev xdebug enable
composer phphunit -- --filter testEnDisUnList
```

fjgarlin:
The module I was requesting as part of the test required a higher version of Drupal than the one installed during the `lowest` tests. I changed that and it seems happier now.

There is one job failing on an unrelated-to-these-changes test. I cannot re-run it. I think this is ready for your review @weitzman or @greg-1-anderson

fjgarlin:
I triggered a re-run with a change-of-wording commit and all tests are happy now.

PR #6451, opened 2025-11-19
Rector PHPUnit quality

weitzman:


weitzman:
No energy for this one

PR #6456, opened 2025-11-21
Use PHP 8.5 for highest tests. remove check_mergeable job

weitzman:


weitzman:
The remaining failures will hopefully be fixed by https://github.com/wodby/php/issues/212. If not, I will downgrade our Postgres in highest tests.

PR #6461, opened 2025-11-25
field-create: Advertise target-type, target-bundle and allowed-formats options

DieterHolvoet:
As discussed in #5692.


weitzman:
I'm seeing a test failure FYI

DieterHolvoet:
Test failure is fixed!

PR #6463, opened 2025-11-28
Support form option arrays in SiteInstall arguments

istncog:
When using drush si with installation profiles that define complex form elements, some form options are arrays which Drush does not handle correctly.


weitzman:
Changes should start on 14.x and then can backport to 13.x

weitzman:
Needs an Example to document the new capability

PR #6464, opened 2025-12-02
Guard against missing PDO MySQL constants

RobXYZ:
I found commands like sql:drop starting throwing errors after updating from 13.6.2 to 13.7 on my web host.

`Error: Call to a member function command() on null in Drush\Commands\sql\SqlCommands->validate() (line 244 of /drush/drush/src/Commands/sql/SqlCommands.php).`

This seems to be due to undefined PDO::MYSQL_ATTR_* constants. SqlBase::create() swallowed and returned null.

Codex found and wrote this fix so it may not be correct but hopefully points in the right direction.

### DIFF
--- a/src/Sql/SqlMysql.php
+++ b/src/Sql/SqlMysql.php
@@ -110,35 +110,35 @@ EOT;
 
         // Use updated constants
         $attribs = [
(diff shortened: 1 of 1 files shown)


weitzman:
Whats your PHP version?

RobXYZ:
PHP 8.3.27

weitzman:
I'm not sure that its valid for those constants to be missing. Your mPDO would be busted. Anyway, [here is how Drupal core handled this](https://git.drupalcode.org/project/drupal/-/merge_requests/13107/diffs), in case anyone wants to make a different PR. I would merge that approach.

PR #6468, opened 2025-12-03
Convert field commands to Symfony Console commands

DieterHolvoet:


DieterHolvoet:
@weitzman any chance we can keep the `input()`, 'output()` and `io()` methods on command classes? It would make it a lot easier to convert commands. Also, having to create a `DrushStyle` instance every time you want to output something feels overly verbose and like a step backwards. I'm converting `FieldCreateCommands` and I just end up having to pass input and output objects all over the place, adding a lot of noise.

In the simplest form it could be a trait with the before mentioned methods and properties storing the objects, and a `setIo(InputInterface $input, OutputInterface $output)` that should be called at the start of the `execute()` method, but if that last part could be abstracted away, even better.

DieterHolvoet:
@weitzman I just read that custom events are deprecated, seemingly without full alternative. What am I supposed to convert this kind of code to?

```
// Command files may set additional options as desired.
$handlers = $this->getCustomEventHandlers('field-create-set-options');
foreach ($handlers as $handler) {
    $handler($input);
}
```

weitzman:
That would be nice, but we are trying not to extend \Symfony\Component\Console\Command\Command. Once we do that, our commands rely a bit more on Drush-isms and are harder to reuse in other CLIs like a future one by Drupal core. I hear you that it feels like a step backward. Hopefully that step is for a good cause in the end.

weitzman:
You should emit events using event_dispatcher from the container. Then you write listeners to react to those events. See https://github.com/drush-ops/drush/blob/3d4bf9c74dcf94041dbcd82882ead6cdacc30245/src/Commands/sql/sanitize/SanitizeCommand.php#L55-L57. Also there is an example in CacheClearCommand

weitzman:
I have passed around $input->getOptions() in places where I used to pass around $options. If you dont want to pass around a DrushStyle and don't want to keep creating it, you could store it on a property on the command class. You could do that in a trait like you say. Start the trait just for the field commands and then we can consider making it more available.

DieterHolvoet:
These io methods wouldn't have to exist in a base class, we could just provide a trait right? An optional `IoTrait` for big command classes that do a lot of input/output spread across multiple methods. Or I could add this straight to `FieldCreateCommands`, but I feel like it could be useful in more places.

weitzman:
After some more thought, I like the `IoTrait` idea. Feel free to make one that all commands can use. $this->setIO()  should return the DrushStyle object in the fluent style so that the author can immediately start asking questions.

DieterHolvoet:
Conversion is done, tests are passing! Also creating MR's here:
- https://www.drupal.org/project/entity_reference_revisions/issues/3364421
- https://www.drupal.org/project/entity_bundle_scaffold/issues/3563442

src/Commands/field/FieldBaseInfoCommands.php
```diff
@@ -4,75 +4,95 @@
...
-    #[CLI\Usage(name: self::BASE_INFO, description: 'List all base fields and fill in the remaining information through prompts.')]
-    #[CLI\Complete(method_name_or_callable: 'complete')]
-    #[CLI\Version(version: '11.0')]
-    public function info(?string $entityType = null, array $options = [
-        'format' => 'table',
-    ]): RowsOfFields
+    public function execute(InputInterface $input, OutputInterface $output): int
     {
-        $this->input->setArgument('entityType', $entityType ??= $this->askEntityType());
+        $this->setIo($input, $output);
```
weitzman: Lets keep using the pattern for commands that use formatters where execute() is the same 3 lines and the work is done in a doexecute(). For example, `\Drush\Commands\pm\PmListCommand::execute`. This applies elsewhere in this PR.

src/Commands/field/FieldCreateCommands.php
```diff
@@ -469,16 +447,15 @@ protected function createFieldDisplay(string $context): void
...
+        // Event subscribers may customize $values as desired.
+        $event = new FieldCreateEntityDisplayValuesEvent($context, $values, $this->input, $this->output);
+        $this->eventDispatcher->dispatch($event);
+        $values = $event->getValues();
 
         $storage = $this->getEntityDisplay($context);
 
         if (!$storage instanceof EntityDisplayInterface) {
-            $this->logger()->info(
+            $this->io()->info(
```
weitzman: FYI I only changed success messages to io(). The others have remained logger messages. I'm happy to do whatever you prefer.

src/Commands/field/FieldDeleteCommands.php
weitzman, line 1: Commandfile names should be singular, not plural

weitzman (approved):
Thanks so much for this excellent work. I have minor feedback only.

src/Commands/field/FieldCreateCommands.php
DieterHolvoet: Is there a difference between the two? This saves us from injecting a dependency.

weitzman: Yes. They look different, and can have different destinations. A site could concievably provide own logger but is unlikely to bother with io. They do that to send log messages to New Relic, Datadog, etc.

I would think that almost all commands will want to inject the logger. I dont know why we would avoid it. Its a single line in the contructor when using autowire.

DieterHolvoet: I think it's fine here. These are development commands anyway, so external logging doesn't make much sense.

PR #6479, opened 2025-12-23
Sort option lists in field commands

DieterHolvoet:
What the title says. I feel like this was done automatically before the switch to Laravel Prompts but I'm not sure, so I'm changing it in the field commands instead of in `DrushStyle`.


weitzman:
If it was done automatically before, I'd be fine with doing it globally.

DieterHolvoet:
I just checked and I was wrong, it wasn't sorted. Let's stick to this way then.

PR #6481, opened 2025-12-24
Align output styling betwen $io->success() and logger

### LINKED ISSUE #6480: Align the output styling between io() and logger() logs
**Describe the bug**
When logging through `$this->io()->success()` and `$this->logger->success()`, the output formatting is completely different. An example, the first one being the logger and the second one being io.

Image

Image

In addition to styling being different, it looks like [hyperlinks](https://symfony.com/blog/new-in-symfony-4-3-console-hyperlinks) aren't supported when logging through io().

**To Reproduce**
Run `drush field:create`, but change the function being used for logging in `FieldCreateCommand::logResult()`

**Expected behavior**
Have the same styling across functions, because both are used throughout Drush commands. Personally, I'm more a fan of the first styling, it's a lot cleaner and takes less space in the terminal.

**Actual behavior**
The styling is different

**Workaround**
/

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 14.x
| Drupal version? | 11.x
| PHP version     | 8.x
| OS?             | Mac

**Additional information**

DieterHolvoet:
Fixes #6480.

### DIFF
--- a/src-symfony-compatibility/v6/Style/DrushStyle.php
+++ b/src-symfony-compatibility/v6/Style/DrushStyle.php
@@ -2,6 +2,7 @@
 
 namespace Drush\Style;
 
+use Consolidation\Log\LogOutputStyler;
 use Drush\Drush;
 use JetBrains\PhpStorm\Deprecated;
 use Laravel\Prompts\MultiSearchPrompt;
@@ -13,6 +14,9 @@ use Laravel\Prompts\SelectPrompt;
 use Laravel\Prompts\Spinner;
 use Laravel\Prompts\SuggestPrompt;
 use Laravel\Prompts\TextPrompt;
+use Psr\Log\LogLevel;
+use Robo\Log\RoboLogLevel;
+use Robo\Log\RoboLogStyle;
 use Symfony\Component\Console\Question\Question;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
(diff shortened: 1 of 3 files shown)


weitzman:
Thanks for the suggestion ... I'm not bothered by inconsistent styling. There are lots of things about io() methods that are not the same as loggers. We need not pretend that success() is a logging level. Thats a legacy Drush feature that we might as well drop because PSR loggers are never going to add a success level. I'm inclined to leave things as they are.

DieterHolvoet:
Apart from inconsistency I just think the previous styling was a lot cleaner: it takes less screen space, it's a lot less bright and hyperlinks work. Also, it's what people are used to. I don't see why a backend code refactor should result in different (and arguably worse) console output for users.

It's also not a very big change. If at some point those Robo classes aren't used anymore we just replace the constants with the strings they reference. I don't think styling our console output differently than Symfony Console is a bad thing. We're already using Laravel Prompts to get cleaner console output than what Symfony provides by default.

Regarding success messages: I don't see why we should remove those. Like you say, io() and loggers are different things. The fact that there isn't an official success PSR log level doesn't make it less useful in communicating that a command executed successfully.

weitzman:
In the end, we have a difference in taste. I like the new styling. 

I'm not sure what the issue is with links. That sounds like a fixable bug.

DieterHolvoet:
Would you consider making this configurable? A bit like Gin's _Layout density_ option, but for the CLI and only providing two options.

CleanShot 2025-12-31 at 15 38 27

weitzman:
That does not feel like a good solution to me. Lets leave this open and see if we get more +1 on the OP.

PR #6485, opened 2026-01-06
Fix broken hyperlink

DieterHolvoet:
Workaround for an issue discussed in #6481.

### DIFF
--- a/src/Commands/field/FieldCreateCommand.php
+++ b/src/Commands/field/FieldCreateCommand.php
@@ -27,6 +27,7 @@ use Drush\Event\FieldCreateFieldConfigValuesEvent;
 use Drush\Event\FieldCreateFieldStorageConfigValuesEvent;
 use Drush\Event\FieldCreateInputOptionsEvent;
 use Psr\EventDispatcher\EventDispatcherInterface;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\Console\Attribute\AsCommand;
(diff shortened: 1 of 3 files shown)


src/Commands/field/FieldCreateCommand.php
```diff
@@ -495,6 +496,7 @@ protected function getEntityDisplay(string $context): ?EntityDisplayInterface
 
     protected function logResult(FieldConfigInterface $field): void
     {
+        assert($this->logger instanceof SuccessInterface);
         $this->logger->success(
```
weitzman, line 500: Shouldnt we use io when we use success()? Thats what fieldDelete is doing ... Oh, its because we should a hyperlink later. I would rather give up the hyperlink than make success() usage inconsistent. or lets fix io so it can print hyperlinks.

DieterHolvoet: It used logger before, I changed it in #6468. Changing it back now to work around the hyperlink issue. Already spent a lot of time on #6481, don't really feel like diving into that hyperlink issue now. Would be a shame to lose the hyperlink here, I actually use it a lot.

weitzman:
I'm OK with using logger in 13.x, but 14.x can't use logger. We want to support injecting a plain PSR logger and that doesnt have success().

weitzman:
Feel free to emit a regular link (i.e. not `<href>`). Most terminals support clicking on those as well. On Mac Iterm2, you control click them.

Issue #6331, opened 2025-07-30
Allow running `updb` and skipping maintenance mode


mglaman:
**Is your feature request related to a problem? Please describe.**
I do not want `updb` to put my Drupal site into maintenance mode

https://github.com/drush-ops/drush/blob/86f9205f7a8706847aaf8731d5967acb736f1bf9/src/Commands/core/UpdateDBCommands.php#L405C1-L409C10

```
        $original_maint_mode = \Drupal::service('state')->get('system.maintenance_mode');
        if (!$original_maint_mode) {
            \Drupal::service('state')->set('system.maintenance_mode', true);
            $operations[] = ['\Drush\Commands\core\UpdateDBCommands::restoreMaintMode', [false]];
        }
```

**Describe the solution you'd like**
Just like `cache-clear` is an option, `maintenance-mode` should be one allowing `--no-maintenance-mode`

**Describe alternatives you've considered**
I haven't found one, beyond trying to hack in with hooks

**Additional context**
Just trying to have zero downtime deployments

gitressa:
Nice catch @mglaman, I added a note on the https://www.drupal.org/docs/updating-drupal/updating-drupal-core-via-composer doc page:

> 3. Activate [maintenance mode](https://www.drupal.org/node/2827362/) using `drush state:set system.maintenance_mode 1` and then `drush cache:rebuild`. (Note: [updatedb activates maintenance mode](https://github.com/drush-ops/drush/issues/6331).)

weitzman:
I'm curious how folks plan to achieve zero downtime deployments. 

For starters, I think we can agree that only some deployments are safe to run without maint mode. So an org would need to evaluate each deployment and run different commands accordingly. Would this be a human decision or a programmatic decision? Its not easy to decide - the set of update functions, post update, hook_deploy, config import, etc. that get run depends on the state of the target. So local might run dfferent functions from dev which might differ from stage which might differ from prod.

I"m not opposed to the proposal, I'm just skeptical that it will be used.

mglaman:
> So an org would need to evaluate each deployment and run different commands accordingly. 

Exactly this. It'd very high touch.

>  Would this be a human decision or a programmatic decision?

Very human.

We already have this:

```php
  #[CLI\Hook(type: HookManager::PRE_COMMAND_HOOK, target: UpdateDBCommands::UPDATEDB)]
  public function disableUpdbCacheClear(CommandData $commandData): void {
[...]

PR #6486, opened 2026-01-07
Document that `updatedb` automatically sets the site in maintenance mode during the process

gitressa:
## Documentation page

`updatedb`: https://www.drush.org/latest/commands/updatedb/

## Problem

Some users may manually enable and disable maintenance mode via GUI or with Drush, before and after running a database update with `drush updatedb`.

## Solution

`drush updatedb` automatically sets the site in maintenance mode during the process, and we could document this, since many users can then skip the enable/disable maintenance mode steps.

## Extra info

Idea originated from this related issue, raised by @mglaman: 

- https://github.com/drush-ops/drush/issues/6331

### DIFF
--- a/src/Commands/updatedb/UpdateDBCommand.php
+++ b/src/Commands/updatedb/UpdateDBCommand.php
@@ -32,7 +32,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 #[AsCommand(
     name: self::NAME,
-    description: 'Apply any database updates required (as with running update.php).',
+    description: 'Apply any database updates required (as with running update.php). Automatically sets the site in maintenance mode, during the update.',
     aliases: ['updb'],
 )]
 #[CLI\Bootstrap(DrupalBootLevels::NONE)]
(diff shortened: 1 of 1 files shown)


src/Commands/updatedb/UpdateDBCommand.php
```diff
@@ -32,7 +32,7 @@
 
 #[AsCommand(
     name: self::NAME,
-    description: 'Apply any database updates required (as with running update.php).',
+    description: 'Apply any database updates required (as with running update.php). Automatically sets the site in maintenance mode, during the update.',
```
weitzman:
```suggestion
    description: 'Apply any pending database updates. Automatically enables maintenance mode during the update.',
```

gitressa:
Thanks for adding this, I just checked, and it seems to have been the case since at least Drush 9.
https://github.com/drush-ops/drush/blob/9.x/src/Commands/core/UpdateDBCommands.php#L437

Since Drush 13 is still widely used, perhaps it should be added there as well? If it's easier, I can create an MR.
https://www.drush.org/13.x/commands/updatedb/

weitzman:
Yes please. A cherry pick does not apply cleanly.

PR #6491, opened 2026-01-26
Inject optional service

DieterHolvoet:
I added support for optional service injection to AutowireTrait when converting the field commands.


weitzman:
Nice. I'm sure there are more commands I need to update

DieterHolvoet:
I did a quick search in the codebase and I couldn't find any others.

PR #6494, opened 2026-01-30
Backport optional service injection to 13.x

DieterHolvoet:
I updated `ConfigImportCommands` as a test case, as suggested. In the process I discovered that `$configStorageSync` isn't actually optional: the return type of the `getConfigStorageSync()` method isn't nullable. Because of this I made the property non-nullable.


weitzman:
Thank you.

PR #6503, opened 2026-02-02
ssl-verify-server-cert is a boolean and 'empty()' doesn't work with FALSE

nterbogt:
This option is a boolean and behaves differently to the existing ones. Update to handle 'FALSE' being set and update the option value to be a more readable output.


weitzman:
14.x first.

PR #6376, opened 2025-10-03
Ensure the order of database updates shown is consistent with how it would actually run

codebymikey:
Currently the database update is listed in alphabetical order, rather than the actual order it would run.

This is especially the case when the [`hook_update_dependencies`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_update_dependencies/11.x) hook is incorporated.

Still need tests and some minor coding standards fixes, but pushing this up just so I don't lose/forget it.

PR #6520, opened 2026-02-20
Ensure the order of database updates shown is consistent with how it would actually run (14.x)

codebymikey:
14.x version of the #6376 PR.

### DIFF
--- a/src/Commands/updatedb/UpdateDbStatusCommand.php
+++ b/src/Commands/updatedb/UpdateDbStatusCommand.php
@@ -30,7 +30,8 @@ use Symfony\Component\Console\Output\OutputInterface;
     'module' => 'Module',
     'update_id' => 'Update ID',
     'description' => 'Description',
-    'type' => 'Type'
+    'type' => 'Type',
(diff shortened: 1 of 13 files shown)


weitzman:
This looks good to me. At the same time, I'm hesitant to make any changes given that Drush is on its way out in favor of the new CLI in Drupal core. Any chance you are willing to code up an MR for updatedb in core? I would pair up to get that reviewed and RTBC.

codebymikey:
> Any chance you are willing to code up an MR for updatedb in core?

Hmm yeah sure, I can take a crack at it since all the code is already here. What's the best way to do that? Is there an open issue with a list of functionality that needs to be implemented?

> I'm hesitant to make any changes given that Drush is on its way out in favor of the new CLI in Drupal core.

I'm not sure about the timelines until Drupal core CLI will have feature-parity with Drush, but from having a quick look, it seems like there's still going to be a fairly long period where both will be used until `dr` encompasses `drush` feature-wise, so I struggle to see why a fix like this couldn't land until Drush is fully deprecated.

PR #6522, opened 2026-02-20
Fix sql:cli in containerized environments (Docker, ddev)

### LINKED ISSUE #6521: sql:cli fails in containerized environments (Docker, ddev) - false positive stdin detection and missing TTY
When SSH'd into a server and running drush sqlc (directly or via ddev), Drush can't detect the TTY properly; this is evident with ```drush sqlc``` which exits immediately without landing at a prompt.

In my case, I'm ssh'd into a local user and running ddev under that user, so multiple levels of ssh.

douggreen:
Fixes #6521

Two issues prevented sql:cli from working in containerized environments:

1. hasPipedInput() false positive: stream_select() returns false positives in containers because stdin is a pipe/socket rather than a TTY, even when no data is being piped. Fix: check posix_isatty() first, and when not on a TTY, skip the unreliable stream_select() check entirely.

2. Missing TTY allocation: The setTty() call was commented out, so the mysql client had no interactive terminal and would exit immediately. Fix: Enable TTY when Process::isTtySupported() returns true.

### DIFF
--- a/src/Commands/sql/SqlCliCommand.php
+++ b/src/Commands/sql/SqlCliCommand.php
@@ -67,25 +67,31 @@ final class SqlCliCommand extends Command
         }
 
         $process = $this->processManager->shell($sql->connect(), null, $sql->getEnv());
-        // No longer needed?
-        // if (Tty::isTtySupported()) {
-        //     $process->setTty((bool) $this->drushConfig->get('ssh.tty', $input->isInteractive()));
-        // }
+        // Enable TTY for interactive use if supported.
+        if (\Symfony\Component\Process\Process::isTtySupported()) {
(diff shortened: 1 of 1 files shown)


weitzman:
sql:cli works for me in DDEV. Could you try using a clone of this repo, since it has DDEV integration?

````
git clone https://github.com/drush-ops/drush.git
ddev start
ddev ssh
./drush sql:cli
```

If it only fails under some DDEV configs, it would be good to understand that better.

weitzman:
I merged a fix for 2. Its not happenning everywhere but its happenning in my current checkout with 12.x and mariadb so in it goes in #6537. I am not seeing a problem with 1. yet. Leaving this open.

PR #6539, opened 2026-03-23
Fix locale import command documentation typos

RichardDavies:
I corrected some apparent typos and inconsistencies in the documentation. The current documentation also suggests that `--type=not-translated` is a valid option when importing. That doesn't make sense to me and I'm not sure that's actually a valid option... but I'm not 100% sure about that so I didn't change it.

### DIFF
--- a/src/Commands/core/LocaleCommands.php
+++ b/src/Commands/core/LocaleCommands.php
@@ -286,19 +286,19 @@ final class LocaleCommands extends DrushCommands
     }
 
     /**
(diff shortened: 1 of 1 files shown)


weitzman:
If anyone could reroll this for 14.x it would be appreciated. It wont cherry pick cleanly.

RichardDavies:
@weitzman New PR for 14.x https://github.com/drush-ops/drush/pull/6546

PR #6563, opened 2026-05-12
Fix duplicate command discovery.

### LINKED ISSUE #6562: Drush command classes whose namespace is registered with Composer's PSR-4 autoloader are discovered twice.
**Describe the bug**
Drush command classes whose namespace is registered with Composer's PSR-4 autoloader are discovered twice in Drush 13.7.2+: once pre-bootstrap by `ServiceManager::discoverPsr4Commands()` and again post-bootstrap by `ServiceManager::discoverModuleCommands()`. The class is then instantiated twice and any hooks it registers fire twice.
The regression was introduced in #6524 (released in 13.7.2), which widened the PSR-4 search pattern from `/.*Commands\.php$/` to `/.*(Command)s?\.php$/`, so files ending in singular `Command.php` (the convention documented for sanitize plugins and used widely in Drupal contrib) are now matched by the pre-bootstrap PSR-4 scan.

The two discovery results are merged in `Drush\Boot\DrupalBoot8::bootstrapDrupalFull()` with no deduplication:

  ```php
  // src/Boot/DrupalBoot8.php:253
  $bootstrapCommandClasses = array_merge($bootstrapCommandClasses, $commandsInThisModule);
  ```
`ServiceManager::discover()` already does `array_unique` on its own pre-bootstrap merge, but this second merge does not.

**To Reproduce**
1. Create a Drupal module with its own `composer.json` declaring a PSR-4 mapping (the default for contrib modules), e.g. `"Drupal\\my_module\\": "src/"`.
2. Add a sanitize plugin at `src/Drush/Commands/sql/MyModuleSanitizeCommand.php` that implements `SanitizePluginInterface` and registers a `sql-sanitize-confirms` on-event hook plus a `sql:sanitize` post-command hook (the standard pattern).
3. Enable the module and run `drush sql:sanitize`.

**Expected behavior**
The confirmation list shows the plugin's message once, and the `sanitize()` post-command hook runs once.

**Actual behavior**
The message appears twice in the confirmation list, and the `sanitize()` callback runs twice.
Any sanitize plugin whose parent namespace ends up in `vendor/composer/autoload_psr4.php` is affected. Sub-modules whose namespace is only registered by Drupal at runtime are unaffected, because they aren't visible to the pre-bootstrap PSR-4 scan.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 13.7.3 (also reproduces on 13.7.2; 13.7.1 is fine)
| Drupal version? | 11.x / 10.x
| PHP version     | 8.x
| OS?             | Linux

nagyad:
Fixes https://github.com/drush-ops/drush/issues/6562

### DIFF
--- a/src/Boot/DrupalBoot8.php
+++ b/src/Boot/DrupalBoot8.php
@@ -249,7 +249,10 @@ class DrupalBoot8 extends DrupalBoot
             $commandsInThisModule = $this->serviceManager->discoverModuleCommands([$path], "\\Drupal\\" . $moduleId . "\\Drush");
             // TODO: Maybe $bootstrapCommandClasses could use a better name.
             // These are commandhandlers that have static create factory methods.
-            $bootstrapCommandClasses = array_merge($bootstrapCommandClasses, $commandsInThisModule);
+            // array_unique() prevents classes already found pre-bootstrap by
+            // ServiceManager::discoverPsr4Commands() from being instantiated a
+            // second time when their namespace is a Composer PSR-4 prefix.
+            $bootstrapCommandClasses = array_unique(array_merge($bootstrapCommandClasses, $commandsInThisModule));
             // TODO: Support PSR-4 command info alterers, like bootstrapCommandClasses?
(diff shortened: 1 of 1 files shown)


weitzman:
Thank you. Also cherry picked to 13.x.

PR #6568, opened 2026-06-01
Fix Drupal dependencies test

claudiu-cristea:


weitzman:
_that link was for error handlers not exception handler_

Thanks a lot. Can you see if the integration test base class can do something like https://git.drupalcode.org/project/drupal/-/merge_requests/14902/diffs#9b93e918831af61c247dd713cdc77bc8c1f83155_0_50 so we dont have to remember to add an Attribute to each test? Or add that code elsewhere--

Issue #6162, opened 2024-11-14
Ability to execute a config diff without exit status 1 or error messages


jameswilson:
**Is your feature request related to a problem? Please describe.**
Currently, the `drush config:import --no --diff` command outputs a configuration diff and does not execute the import, as expected when the `--no` flag is used. However, the command still exits with status code `1`, indicating a failure. This is problematic when the intent is to retrieve a config diff without performing an import or generating an error in scripts or logs.

Many users, including myself, are frustrated by this behavior because the command, in this context, performs exactly as intended—outputting a diff and abstaining from importing—yet it results in a "failed" exit status. This complicates scripting and logging when only a diff is needed without triggering a script failure.

Current output,

```
$ drush config:import --no --diff
diff --git a/tmp/drush_tmp_1731606322_67363732aed14/system.site.yml b/tmp/drush_tmp_1731606323_6736373347357/system.site.yml
index 0418eb8f19..7d88f504d7 100644
--- a/tmp/drush_tmp_1731606322_67363732aed14/system.site.yml
+++ b/tmp/drush_tmp_1731606323_6736373347357/system.site.yml
@@ -4,7 +4,7 @@ langcode: en
 uuid: XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
 name: 'My Website'
 mail: admin@example.com
-slogan: 'CHANGED ON PROD'
+slogan: ''
 page:
   403: /node/3
   404: /node/4

 !
 ! [WARNING] Import the listed configuration changes?: no.
 !

In ConfigImportCommands.php line 228:

  Cancelled.

Failed to execute command drush config:import --no --diff: exit status 1
```

**Describe the solution you'd like**

1. Introduce a dedicated command `drush config:diff`.

    A `config:diff` command will enable diff-only output without the failed exit code, _WARNING_, or error messages. This new command should allow users to view the configuration differences between the current and new config states without triggering an error status, ensuring compatibility with scripts or logging processes that expect a successful run.  The command can be considered analogous to `git diff` and may in fact pass through status codes from the underlying command used internally to generate the diff.

    The desired behavior is for the diff command to exit with code `0` when it completes successfully, even if no import takes place, reflecting the intention of the `--no` or `--dry-run` (see below) functionality.

2. Support `--reverse` flag.

[...]

PR #6569, opened 2026-06-03
refactor: `drush cex --diff` output to stdout same as `drush cim --diff`

hanoii:
I wanted to use drush cex --diff to do something conditionally on the output.

I am also very used to do its `drush cim  --diff` equivalent.

On https://github.com/drush-ops/drush/issues/6162#issuecomment-4615373283 I metioned that I use

```
drush cim --diff --no 2>/dev/null | grep -v ' !' || true
```

To jsts get the diff. I would have expected drush cex work exactly the same, but I noticed it sends it to stdout instead. I think both should behave the same [and stdout is the right place](https://github.com/drush-ops/drush/blob/bb7505d8260157fd3db168e28928ed156390ee2e/src/Commands/config/ConfigImportCommand.php#L137) for this output.

### DIFF
--- a/src/Commands/config/ConfigExportCommand.php
+++ b/src/Commands/config/ConfigExportCommand.php
@@ -123,7 +123,7 @@ final class ConfigExportCommand extends Command
 
             if ($input->getOption('diff')) {
                 $diff = $this->getDiff($target_storage, $this->configStorageExport, $output);
-                $this->logger->notice($preamble . $diff);
+                $output->writeln($diff);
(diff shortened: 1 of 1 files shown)


weitzman:
Consistency is good. I'm unsure if the diff belongs on stderr or stdout. Can you say more about how you chose stdout?

hanoii:
Mostly so I can separate concerns and be able to redirect/pipe the diff. It's not an error, it's something I am requestignm to be output.

Issue #6566, opened 2026-05-23
Batch callbacks need to support services to match Drupal core


Berdir:
**Describe the bug**
See https://www.drupal.org/node/3570904

Drupal core now supports batch callbacks as services since 11.3, and locale module in 11.4 makes heavy use of that, breaking the batch during module install when using drush.

CallableResolver exits since Drupal 10.2: https://www.drupal.org/node/3368504

**To Reproduce**
What did you do?

**Expected behavior**
Use CallableResolver. 

**Actual behavior**
What happened instead?

**Workaround**
Is there another way to do the desired action?

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 14.x/13.x/12.x
| Drupal version? | 11.3+
| PHP version     | 8.x/7.x
| OS?             | Mac/Linux/Windows

**Additional information**

marcoliverteschke:
Added an attempt at a pull request for this.

rodrigoaguilera:
I just patched my Drupal 11.4.0-beta4 with the code in the PR and I don't get the following error anymore.
```
TypeError: call_user_func_array(): Argument #1 ($callback) must be a valid callback, function "Drupal\locale\LocaleFetch:batchVersionCheck" not found or invalid function name in _drush_batch_worker() (line  
   257 of /var/www/html/vendor/drush/drush/includes/batch.inc).         
```

ptmkenny:
Related issue: https://github.com/drush-ops/drush/issues/5880

Berdir:
The changes have been merged into 13.x and similar changes already existed in 14.x to fix the tests.

However, I noticed that the change in 14.x is incomplete as it didn't remove the is_callable() check, so the finish callback is silently skipped. That needs to be removed.

Additinally, I'd recommend against try/catch on the CallableResolver, core removed that too in drupal.org/node/3539919. This is a developer-oriented error, the callback either exists or not and if it doesn't, then as a developer, I want to see that instead of it just silently skipping that.

I'll try to provide PRs for that in the next days, but if anyone manages to do that first then that would be great and I'll try to review.

marcoliverteschke:
Opened a PR for 14.x that would remove is_callable() and one for 13.x that removes the try-catch in alignment with what Core did, as @Berdir mentioned.

FlorentTorregrosa:
Thanks for this issue and the fix!

weitzman:
This got done.

PR #6571, opened 2026-06-11
Issue #6566: Use CallableResolver to get the callback we want to use in batch

marcoliverteschke:
Seeks to solve #6566 

Not sure if this covers everything…

But these are the problematic places I was able to find.

The fix unblocked some a failing module install I encountered under 11.4.0-beta1.

### DIFF
--- a/includes/batch.inc
+++ b/includes/batch.inc
@@ -27,6 +27,7 @@
 
 use Drupal\Core\Database\Database;
 use Drupal\Core\Database\IntegrityConstraintViolationException;
+use Drupal\Core\Utility\CallableResolver;
 use Drush\Drush;
 
 /**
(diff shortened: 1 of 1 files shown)


weitzman:
Needs someone to review and ideally needs a test. Or maybe the locale commands that are currently failing will be fixed by this and thus no additional test is needed?

weitzman:
Actually this is fixing some test fails so in it goes.

weitzman:
Sorry I had to revert this. I didn't realize that the PR author chose 13.x as target. Changes should go into 14.x first. I tried to cherry pick this into 14.x but it doesnt apply cleanly so I decided to revert.

marcoliverteschke:
@weitzman Sorry, my bad. We're not on Drush 14 over here yet, so I missed looking into that branch first. FWIW, it seems the most recent commit on 14.x (https://github.com/drush-ops/drush/commit/bb7505d8260157fd3db168e28928ed156390ee2e) already includes a similar fix.

weitzman:
Oh I didn't realize that 14.x was fixed already. So your PR *is* correct. I just cherry picked it back onto 13.x

Issue #6572, opened 2026-06-11
Do not treat non-existing post updates as an error that aborts the update process


Berdir:
**Describe the bug**

When a module has a post update function, but is uninstalled by an update function before that post update can run, it causes this:

```
>  [warning] Post update function help_post_update_search_help_dependencies not found.
>  [error]  Update failed: help_post_update_search_help_dependencies 
```

**To Reproduce**
see above.

**Expected behavior**
Only display a warning, not an error.

**Actual behavior**
Warning and error is displayed.

**Workaround**
Is there another way to do the desired action?

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 14.x/13.x/12.x (please be specific, and try latest release)
| Drupal version? | 11.x/10.x/9.x/8.x/7.x
| PHP version     | 8.x/7.x
| OS?             | Mac/Linux/Windows

**Additional information**

Another option would be to recalculate the post updates after regular updates completed.

weitzman:
done

PR #6573, opened 2026-06-12
Do not fail on non-existing post updates

Berdir:
MR for #6572 

I'm not sure if this is really desired, but putting it up in case it's useful, needed a patch for my project.


weitzman:
Seems reasonable to me. Any objection @pfrenssen ?

weitzman:
Merged. i'm guessing this is not suitable for backport.

PR #6574, opened 2026-06-15
Revert "Issue #6566: Use CallableResolver to get the callback we want to use in batch"

weitzman:
Reverts drush-ops/drush#6571


weitzman:
reverting since it needs to go into 14.x first.

PR #6577, opened 2026-06-16
Postgresql schema not used by drush #1523

joejoseph00:
This works and makes sense to me.

### DIFF
--- a/src/Sql/SqlPgsql.php
+++ b/src/Sql/SqlPgsql.php
@@ -6,8 +6,6 @@ namespace Drush\Sql;
 
 use Drush\Drush;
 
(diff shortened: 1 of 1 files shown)


weitzman:
Please describe the bug and the fix.

joejoseph00:
> Please describe the bug and the fix.

Before patching , drush sql-drop only works if your schema/database is called 'public' . Who does this?  After patching, drush sql-drop functions as expected.


Install postgresql
Create schéma / database called mydrupal
Create user with privileges for mydrupal
Install drupal 11.3.12 on the mydrupal schema
Install drush 

Run drush sql-drop
Nothing happens because drush thinks your schema is called 'public'

Solution: apply patch 

Now run drush sql-drop

All tables are dropped as expected

weitzman:
Thanks. Needs thorough review by a postgres user.

PR #6578, opened 2026-06-16
fix(updatedb): support RequirementSeverity enum in updateCheckRequirements

### LINKED ISSUE #6528: UpdateDBCommands::updateCheckRequirements doesn't support new requirements enum
**Describe the bug**

Running drush updb on 11.2+ shows a warning about JSON database support. However, that's not actually meant to be a warning, just an OK that is supposed to be ignored.

https://www.drupal.org/project/drupal/issues/3532243

While it's uncommon from core to return OK requirements for this phase, it's supported and the code is there to handle that. But it doesn't support the new enums, which neither equals to a warning nor an error, so as a result, it is shown as a warning which is the fallback.

The code will need to be adjusted to support both constants and enums, probably a helper function(s) somewhere, like isErrorSeverity() and so on?

**To Reproduce**

Run drush updb

**Expected behavior**

No requirements output.

**Actual behavior**

```
 [warning] Drupal requires databases that support JSON storage.  (Currently using Database support for JSON Available)
```

**Workaround**

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 13.7.1.0
| Drupal version? | 11.3
| PHP version     | 8.4
| OS?             | Linux

**Additional information**

tobiasbaehr:
Fixes #6528

Drupal 11.2+ replaced integer-based requirement severity constants with a backed enum (`RequirementSeverity`). The inner loop in `updateCheckRequirements()` compared `$requirement['severity']` directly against integers `0` and `2`, causing enum values to never match `REQUIREMENT_OK` and always fall through as `[warning]`.

Adds `drush_drupal_requirement_severity_int()` to `legacy.inc` — mirroring the existing `drush_drupal_requirements_severity()` pattern — and uses it for both comparisons in the inner loop.

### DIFF
--- a/includes/legacy.inc
+++ b/includes/legacy.inc
@@ -21,6 +21,19 @@ function drush_drupal_requirements_severity(&$requirements) {
     }
 }
 
+/**
+ * Normalizes a single requirement severity to an integer.
+ *
+ * Drupal 11.2+ uses RequirementSeverity enums; older versions use integer
+ * constants (REQUIREMENT_OK=0, REQUIREMENT_WARNING=1, REQUIREMENT_ERROR=2).
+ */
+function drush_drupal_requirement_severity_int(mixed $severity): int {
+    if ($severity instanceof RequirementSeverityAlias) {
+        return $severity->value;
+    }
+    return (int) $severity;
+}
+
 function drush_locale_translation_default_update_options() {
     if (function_exists('_locale_translation_default_update_options')) {
         return _locale_translation_default_update_options();
--- a/src/Commands/updatedb/UpdateDBCommand.php
+++ b/src/Commands/updatedb/UpdateDBCommand.php
@@ -129,18 +129,17 @@ final class UpdateDBCommand extends Command
         $severity = drush_drupal_requirements_severity($requirements);
 
         // If there are issues, report them.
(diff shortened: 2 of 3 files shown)


weitzman:
the drush 14 composer.json says that it supports 11.3+. So can we assume the enum is always there? Maybe there is more simplifying we can do here?

weitzman:
If you merge in latest changes you will get rid of the test failures. I dont have perms to do that in your repo.

tobiasbaehr:
We should support the old severity constants until drupal 12.

weitzman:
OK. Please update this branch or give me access as PR reviewer. I'd like to see this green before merge.

tobiasbaehr:
I updated the MR already, but I squashed the again (to remove the ugly merge commit)  and "Allow edits by maintainers" is enabled.

Ok fix now also the phpstan issue. I have no idea why phpstan is now unhappy.

weitzman:
Merged. Does 13.x need this change as well? The commit wont cherry pick cleanly so a new PR would be needed.

tobiasbaehr:
Done https://github.com/drush-ops/drush/pull/6586. Drush 13.x + Drupal 11 is main reason why I created the fix. :D

PR #6581, opened 2026-06-30
Additional info about hook locations and triggers

obriat:


weitzman:
then intent is to link to the drupal docs, not duplicate them

obriat:
There's no central doc on drupal.org, I thought this page could be the ideal spot to details all the steps of drush deploy, my bad

PR #6589, opened 2026-07-05
[13.x] Update dependencies in Dependency commands

claudiu-cristea:


claudiu-cristea:
* Fixed the Drupal dependencies test expectancy
* PHP 8.5 support

PR #6593, opened 2026-07-08
Fix locale:update command on 11.4

alexpott:
Drush 13 fix for locale command on 11.4 Drupal core. See https://www.drupal.org/node/3591660


weitzman:
Released https://github.com/drush-ops/drush/releases/tag/13.7.6

PR #6598, opened 2026-07-08
Relative PSR4 for phpstan-bootstrap.php

### LINKED ISSUE #6596: phpstan-bootstrap.php should add the relative directory for consuming project
**Describe the bug**
We have a project using drush which has it's own drush commands. We get errors from `DrushStyle` etc.

We tried including `vendor/drush/drush/phpstan-bootstrap.php` in the `bootstrapFiles`, but that doesn't resolve the issue.

The reason is that `phpstan-bootstrap.php` adds `$loader->addPsr4('Drush\\', 'src-symfony-compatibility/v6');` (or v7 in later versions). However, this runs relative to the project, meaning it only work for drush internally.

Would be better if it used `__DIR__` or similar to ensure the correct path for the directory.

**To Reproduce**
1. Add drush commands
2. Run `phpstan` and get errors, e.g. `$this->io()->text(...)`
3. Add `vendor/drush/drush/phpstan-bootstrap.php` to `bootstrapFiles`
4. Run `phpstan` and _still_ get errors, e.g. `$this->io()->text(...)`

**Expected behavior**
No errors after including the bootstrap.

**Actual behavior**
Still get errors because the class loader gets `Drush\` => `<project root>/src-symfony-compatibility/v6`.

**Workaround**
Use include files instead of bootstrap, but then that needs to be maintained in line with Drush.

### System Configuration
| Q               | A
| --------------- | ---
| Drush version?  | 13.6.2.0
| Drupal version? | 11.x
| PHP version     | 8.3
| OS?             | Linux (DDEV)

andrewbelcher:
Fixes #6596

### DIFF
--- a/phpstan-bootstrap.php
+++ b/phpstan-bootstrap.php
@@ -4,5 +4,5 @@ use Composer\Autoload\ClassLoader;
 
 // Deal with dynamic autoloader that we use for symfony cross version compat.
 $loader = new ClassLoader();
-$loader->addPsr4('Drush\\', 'src-symfony-compatibility/v6');
+$loader->addPsr4('Drush\\', __DIR__ . '/src-symfony-compatibility/v6');
 $loader->register();


weitzman:
Released https://github.com/drush-ops/drush/releases/tag/13.7.6

PR #6603, opened 2026-07-27
Fix return type hint for dt() in output.inc

mbomb007:
Intelephense detects the return type of `dt()` as "The" without the fix.

### DIFF
--- a/includes/output.inc
+++ b/includes/output.inc
@@ -19,7 +19,7 @@ use Drush\Utils\StringUtils;
  * @return string
  *   The processed string.
  */
-function dt($string, $args = []) {
+function dt($string, $args = []): string {
   return StringUtils::interpolate($string, $args);
 }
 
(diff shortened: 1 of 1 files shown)


mbomb007:
This should prevent all the annoying complaints about `dt()` by commonly-used code sniffers and intelephense.

includes/output.inc
```diff
@@ -16,7 +16,7 @@ use Drush\Utils\StringUtils;
  * @param array
  *  An associative array of replacement items.
  *
- * @return
+ * @return string
```
weitzman, line 19: Instead of this, add a real return type to the function. It always returns a string.

mbomb007: Okay. I didn't know if it was okay to change the function signature.

mbomb007: I added the return type to both functions.

weitzman:
And cherry picked onto 13.x

PR #6604, opened 2026-08-17
Document dt() return type

mfb:
Documenting the dt() return type should resolve phpstan errors that callers hit when passing dt() output to a string param.


weitzman:
Dupe of #6603. Would that work for you?

mfb:
Oops - was so trivial I neglected to check

PR #6607, opened 2026-09-03
Wrong class is created

knegyesi:
1. Have an abstract base class which uses autowiretrait
2. Have a command extend this class but doesn't use autowiretrait
3. kaboom

Workaround: add `use AutowireTrait` to children classes.


weitzman:
Fixed in #6476

Issue #6595, opened 2026-07-08
`migrate:import --delete` can fail silently and result in data loss when using container built outside drush's bootstrap


marcusmcdowell:
**Describe the bug**
When Drupal's container is built outside of Drush's bootstrap and you later execute a `migrate:import` with the `--delete` option, all items are rolled back.

**To Reproduce**

1. Install drupal `ddev drush site-install standard`
2. Install the migrate module (the hook is only registered if the module is enabled) `ddev drush install migrate`
3. Check that the hook was registered successfully `ddev drush eval "var_dump(\Drupal::moduleHandler()->hasImplementations('migrate_prepare_row', ['system']));"`
4. Invalidate the container `ddev drush cc container`
5. Curl the front page so container is rebuilt without drush `ddev exec curl -s -o /dev/null http://localhost`
6. This time we can see that the `migrate_prepare_row` is not registered with `ddev drush eval "var_dump(\Drupal::moduleHandler()->hasImplementations('migrate_prepare_row', ['system']));"`

If you have a migration with `track_changes` enabled and you run the import after step 6 with the `--delete` option, all items will be rolled back.

<details>
<summary>migration_example.yml</summary>

_You need to create a `page` content type before using this example._

```
id: migration_example
label: "Example migration"
source:
  track_changes: true
  plugin: embedded_data
  data_rows:
    - id: 1
      title: "Node 1"
    - id: 2
      title: "Node 2"
  ids:
    id:
      type: integer
process:
  title: title
destination:
  default_bundle: page
  plugin: entity:node
```

</details>

**Expected behavior**
`migrate:import` should either be terminated before being executed to prevent data loss. Or better yet be able to make sure the hook is registered before proceeding. 

**Actual behavior**
Running `migrate:import` with the `--delete` options rolls back all migrated items.

**Workaround**
We're running migrations on cron on our site and we've added a validator hook which terminates the import if the `migrate_prepare_row` hook isn't registered.

```php
  #[CLI\Hook(type: HookManager::ARGUMENT_VALIDATOR, target: 'migrate:import')]
  public function validateMigrateImport(): ?CommandError {
    if ($this->moduleHandler->hasImplementations('migrate_prepare_row', ['system'])) {
      return NULL;
    }

[...]

PR #6611, opened 2026-09-04
migrate:import --delete can fail silently and result in data loss 

weitzman:
... when using container built outside drush's bootstrap

See #6595


src/Drupal/Migrate/MigratePrepareRowModuleHandler.php
```diff
@@ -0,0 +1,170 @@
...
+
+    public function hasImplementations(string $hook, $modules = null): bool
+    {
+        return $this->decorated->hasImplementations($hook, $modules);
+    }
+
+    public function invokeAllWith(string $hook, callable $callback): void
+    {
+        $this->decorated->invokeAllWith($hook, $callback);
+    }
```
kristiaanvandeneynde, line 134: If core switches SourcepluginBase::prepareRow() to use invokeAllWith() at some point, this bug will rear its ugly head again.

Maybe it's safer to also apply the guard here?

src/Drupal/Migrate/MigrateExecutable.php
```diff
@@ -162,6 +163,43 @@ public function __construct(MigrationInterface $migration, MigrateMessageInterfa
...
+     * dispatches MigrateEvents::DRUSH_MIGRATE_PREPARE_ROW for every scanned
+     * source row, including rows skipped by track_changes/high-water. The
+     * interception is per-instance and in-memory; unlike the former
+     * compile-time hook registration it cannot leak into the hook lists that
+     * Drupal >= 11.3 persists in shared keyvalue storage.
+     *
+     * @return bool
+     *   Whether the interception could be attached.
+     */
+    private function interceptPrepareRow(MigrationInterface $migration): bool
```
kristiaanvandeneynde, line 189: This method returns true when it successfully replaced the property, but that can resurface this very same bug later down the line (see comment below on MigratePrepareRowModuleHandler.php)

kristiaanvandeneynde (commented):
This PR looks good already, but it leaves room for the same silent data loss to occur in the future. The issue where this behavior was reported had a suggestion:

> makes MigrateExecutable::handleMissingSourceRows() refuse to act when zero source rows were observed but the source is not empty?

That seems like a good way to make sure the bug doesn't surface again when Drupal changes the internals of SourcePluginBase, more changes to hooks are made in core, or other unforeseen changes alter the behavior so that the fix here no longer works.

weitzman:
> makes MigrateExecutable::handleMissingSourceRows() refuse to act when zero source rows were observed but the source is not empty

Good idea — done in 78e465100. `handleMissingSourceRows()` now throws when no source rows were observed but the source count is > 0, so any future breakage of the interception becomes a hard error instead of a mass rollback. Empty and uncountable sources keep the current behavior.

I'm skipping the `invokeAllWith()` interception for now. Core doesn't route `migrate_prepare_row` through it today, so it would be untestable speculation, and intercepting both risks double-dispatching if core ever called both paths. Same reasoning for the `interceptPrepareRow()` return value: no compile-time check can prove the event will actually fire, which is why the new guard verifies the runtime outcome (rows observed) at the only point where the damage could happen.

weitzman:
Merged, and backported to 13.x. 

I suggest these Migrate commands get removed from Drush core in 14.x. Follow https://github.com/drush-ops/drush/issues/6613 for updates.

PR #6619, opened 2026-09-11
Add deployment identifier to status output

### LINKED ISSUE #6554: Add deployment identifier to `drush status`

I'd appreciate being able to see the deployment identifier in the `drush status` command. This way I can know what Drupal/Drush has fully evaluated the identifier to be.

- `drush status` -> See the value in table
- `drush status` with --fields=deployment-identifier & format=json = `{ "deployment-identifier": "ABC123" }`

Eval, custom commands

novia713:
Fixes #6554.

Adds the evaluated Drupal deployment identifier to the `core:status` property list.

The field is not included in the default table output. It remains available through structured formats or when explicitly requested. For example:

```shell
drush status \
  --fields=deployment-identifier \
  --format=json
```

### DIFF
--- a/src/Commands/core/StatusCommand.php
+++ b/src/Commands/core/StatusCommand.php
@@ -72,9 +72,10 @@ use Symfony\Component\Filesystem\Path;
     // config-sync is deprecated. Use 'config' instead.
     'config-sync' => 'Drupal config',
     'config' => 'Drupal config',
+    'deployment-identifier' => 'Deployment identifier',
     '%paths' => 'Other paths'
 ])]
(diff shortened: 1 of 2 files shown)


weitzman:
Thanks. I'm a little hesitant to add it as a default field. I dont think its needed very often. If omitted from there, one would ask for it directly as you showed or --format=json shows all fields.

novia713:
Thanks for the feedback. I added it to the default fields because the original feature request explicitly asks for the value to be shown in the standard drush status table.

If that part of the request is no longer considered desirable, would you prefer me to remove it from the default fields and keep it available only when explicitly requested or through structured output? Perhaps the issue description could also be updated to reflect that narrower scope.

weitzman:
As mentioned, please remove from default fields. Also omit the dedicated test, thats AI slop IMO. If anything you could add it to any existing test for StatusCommand.

novia713:
Thanks. I removed deployment-identifier from the default fields and removed the dedicated functional test.

The field remains available when explicitly requested or through structured output. I also updated the PR description accordingly.

